PR #33, opened 2015-05-29
Make sentry a core feature but optionnal

noirbizarre:
The sentry extension is quite small.

Sentry gain to be tighly integrated into udata, event if it's optionnal.


davidbgk:
r+

Issue #23, opened 2015-05-19
Add CSV export for issues and discussions related to organizations


davidbgk:
As an organization, I want to download a CSV of issues and discussions related to my datasets in order to participate/answer.

davidbgk:
Fixed in https://github.com/etalab/udata/pull/35

PR #35, opened 2015-06-01
Add CSV export for issues and discussions, refs #23

davidbgk:


udata/core/organization/views.py
```diff
@@ -291,6 +296,26 @@ def datasets_csv(org):
     return csv.stream(adapter, '{0}-datasets'.format(org.slug))
 
 
+@blueprint.route('/<org:org>/issues.csv')
+def issues_csv(org):
+    datasets = Dataset.objects.filter(organization=str(org.id))
+    issues = [DatasetIssue.objects.filter(subject=dataset)
```
noirbizarre: I think ``DatasetIssue.objects(subject__in=datasets)`` should be more efficient

udata/core/organization/views.py
```diff
@@ -291,6 +296,26 @@ def datasets_csv(org):
...
+              for dataset in datasets]
+    # Turns a list of lists into a flat list.
+    adapter = IssuesOrDiscussionCsvAdapter(itertools.chain(*issues))
+    return csv.stream(adapter, '{0}-issues'.format(org.slug))
+
+
+@blueprint.route('/<org:org>/discussions.csv')
+def discussions_csv(org):
+    datasets = Dataset.objects.filter(organization=str(org.id))
+    discussions = [DatasetDiscussion.objects.filter(subject=dataset)
```
noirbizarre: I think ``DatasetDiscussion.objects(subject__in=datasets)`` should be more efficient

udata/core/organization/views.py
davidbgk: Indeed! Thanks.

davidbgk: After trying a few things, it appears that the planned `ReferenceField` refactoring is necessary to get rid of the manual list comprehension. I'll keep it as-is for now.

PR #39, opened 2015-06-01
Use editorial selection in homepage for dataset and reuse

noirbizarre:
No more automatic selection in homepage, it now use the manual editorial selection.

The default home template alos use real data now.

### DIFF
--- a/udata/templates/home.html
+++ b/udata/templates/home.html
@@ -23,12 +23,13 @@
 
     <!-- Wrapper for slides -->
     <div class="carousel-inner">
+        {% if last_post %}
         <div class="item active">
             <div class="container">
 
                 <div class="row">
                     <div class="col-lg-12">
-                        <h1 class="page-header">Titre de l'actu
+                        <h1 class="page-header">{{last_post.name}}
                             <small>Actualité du moment</small>
                         </h1>
                     </div>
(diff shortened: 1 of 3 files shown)


udata/templates/home.html
```diff
@@ -36,81 +37,77 @@ <h1 class="page-header">Titre de l'actu
...
-                                <a href="#">
-                                    <img class="img-responsive portfolio-item" src="http://placehold.it/300x200">
-                                </a>
-                            </div>
-
-                            <div class="col-sm-3 col-xs-6">
-                                <a href="#">
-                                    <img class="img-responsive portfolio-item" src="http://placehold.it/300x200">
-                                </a>
+                            {% for dataset in reuse.datasets[:4] %}
```
davidbgk: I think that it's more "secure" to put limits in the view but YMMV.

davidbgk:
Do we remove the related code in udata-gouvfr in the meantime?

noirbizarre:
I already removed some parts (etalab/udata-gouvfr#16).

More will require a theme refactoring, which, I think, will be an whole story by itself.

udata/templates/home.html
noirbizarre: Yep, I was wondering which one should I choose between
- no limits in view and slicing in template => allow theme to limit itself the amount of required datasets/reuses
- limits in view => theme is limited by the view itself

davidbgk: Oh good point. Let's discuss that tomorrow.

noirbizarre: Good to me.

davidbgk:
r+

PR #46, opened 2015-06-04
Sitemap refactoring

noirbizarre:
A little refactoring including:

- Allow each module to register its own URLs to the sitemap (by making the sitemap instance global)
- Fix cache handling during tests
- Reuse the application-wide cache
- Only use https in URLs when ``USE_SSL`` setting is set
- Avoid reaching maximum recursion depth during tests
- Add the posts to the sitemap


davidbgk:
Great! r+

PR #48, opened 2015-06-05
Add URL checking on submission and pre-fill of metadata, fix #47

### LINKED ISSUE #47: Add URL checking on submission and pre-fill of metadata
As a producer, I want to get a visual feedback on URL submission to avoid submitting 4XX or 5XX pages.

Part of the integration with https://github.com/davidbgk/croquemort

davidbgk:
As a producer, I want to get a visual feedback on URL submission to
avoid submitting 4XX or 5XX pages.

Part of the integration with https://github.com/davidbgk/croquemort

### DIFF
--- a/udata/static/less/forms.less
+++ b/udata/static/less/forms.less
@@ -210,6 +210,19 @@ form {
     }
 }
 
+.form-upload-fields {
+    .success {
+        background-color: lighten(#3f903f, 40%) !important;
+    }
+
+    .warning {
+        background-color: lighten(#f0ad4e, 20%) !important;
+    }
+
+    .danger {
+        background-color: lighten(#d9534f, 20%) !important;
+    }
+}
 
 .image-picker {
     color: lighten(@gray, 5%);
--- a/udata/static/less/forms.less
+++ b/udata/static/less/forms.less
@@ -210,20 +210,6 @@ form {
     }
 }
 
-.form-upload-fields {
-    .success {
-        background-color: lighten(#3f903f, 40%) !important;
-    }
-
-    .warning {
-        background-color: lighten(#f0ad4e, 20%) !important;
-    }
-
-    .danger {
-        background-color: lighten(#d9534f, 20%) !important;
-    }
-}
-
 .image-picker {
     color: lighten(@gray, 5%);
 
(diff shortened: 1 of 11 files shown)


davidbgk:
Remaining questions:

* do we block the submission if the status code returned is > 400?
* do we turn into settings vars like the number of attempts or the introduced delay in fetching?
* do we add more visual feedback or help?
* do we try to pre-fill the checksum too?

capture d ecran 2015-06-04 a 23 24 38

davidbgk:
Note: for the moment tests are skipped on Travis because of the complexity of the setup/mock. Thoughts?

noirbizarre:
About questions:
- **do we block the submission if the status code returned is > 400?** I think we should, we already had a lot of bad submissions.
- **do we turn into settings vars like the number of attempts or the introduced delay in fetching?** Good idea, it will allows tuning without redeploying just by changing the configuration.
- **do we add more visual feedback or help?** Let's keep it integrated and uniform with other existing visual feedback (jquery-validation: ``udata/static/js/form/common.js``). Maybe we could later add:
  - details about the error, what does it means...
  - some icon feedback: http://getbootstrap.com/css/#with-optional-icons
- **do we try to pre-fill the checksum too?** Could be great. Let's iterate and add it later.

udata/static/less/forms.less
```diff
@@ -210,6 +210,19 @@ form {
     }
 }
 
+.form-upload-fields {
```
noirbizarre: I think it's already globally defined.
See ``udata/static/js/form/common.js`` and http://getbootstrap.com/css/#forms-control-validation

We should keep it uniform by reusing the same classes and addind the background feedback globally.

noirbizarre:
r+wc

udata/tests/api/test_url_api.py
```diff
@@ -0,0 +1,41 @@
...
+from . import APITestCase
+
+
+class CheckUrlAPITest(APITestCase):
+
+    def setUp(self):
+        # If you set up a Croquemort URL, don't forget to clean up the
+        # Redis database after each test. Otherwise the test against the
+        # delayed URL will not pass (existing entry already fetched).
+        self.CROQUEMORT_URL = self.create_app().config.get('CROQUEMORT_URL')
```
noirbizarre: I think it should be written:
```python
def serUp(self):
    super(CheckUrlAPITest, self).setUp()
    self.CROQUEMORT_URL = self.app.config.get('CROQUEMORT_URL')
    if self.CROQUEMORT_URL is None:
        raise unittest.SkipTest
```
It will avoid instanciating 2 or more apps and ensure parents setups are executed.

PR #55, opened 2015-06-09
Add badges for organizations and datasets, fix #54

### LINKED ISSUE #54: Add badges for organizations and datasets
As an administrator, I want to be able to set/unset badges to organizations and datasets in order to highlight trusted sources.

davidbgk:


udata/core/badges/models.py
```diff
@@ -0,0 +1,29 @@
...
+class Badge(db.EmbeddedDocument):
+    kind = db.StringField(choices=[], required=True)
+    created = db.DateTimeField(default=datetime.now, required=True)
+    created_by = db.ReferenceField('User')
+    removed = db.DateTimeField()
+    removed_by = db.ReferenceField('User')
+
+    meta = {
+        'allow_inheritance': True,
+        'ordering': ['created'],
```
noirbizarre: Maybe better with ``-created`` (last created first)

udata/core/dataset/api.py
```diff
@@ -122,6 +126,39 @@ def delete(self, dataset):
         return dataset
 
 
+@ns.route('/<dataset:dataset>/badges/', endpoint='dataset_badges')
+class DatasetBadgesAPI(API):
+    @api.secure
```
noirbizarre: Not needed because of following ``@api.secure(admin_permission)``

udata/core/dataset/api.py
```diff
@@ -122,6 +126,39 @@ def delete(self, dataset):
...
+        badge = DatasetBadge(created=datetime.now(),
+                             created_by=current_user.id)
+        form.populate_obj(badge)
+        dataset.add_badge(badge)
+        return badge, 201
+
+
+@ns.route('/<dataset:dataset>/badges/<badge_kind>/', endpoint='dataset_badge')
+class DatasetBadgeAPI(API):
+    @api.secure
```
noirbizarre: Not needed because of following ``@api.secure(admin_permission)``

udata/core/organization/api.py
```diff
@@ -101,6 +107,39 @@ def delete(self, org):
         return '', 204
 
 
+@ns.route('/<org:org>/badges/', endpoint='organization_badges')
+class OrganizationBadgesAPI(API):
+    @api.secure
```
noirbizarre: Not needed because of following ``@api.secure(admin_permission)``

udata/core/organization/api.py
```diff
@@ -101,6 +107,39 @@ def delete(self, org):
...
+        badge = OrganizationBadge(created=datetime.now(),
+                                  created_by=current_user.id)
+        form.populate_obj(badge)
+        org.add_badge(badge)
+        return badge, 201
+
+
+@ns.route('/<org:org>/badges/<badge_kind>/', endpoint='organization_badge')
+class OrganizationBadgeAPI(API):
+    @api.secure
```
noirbizarre: Not needed because of following ``@api.secure(admin_permission)``

udata/core/site/views.py
```diff
@@ -169,13 +170,6 @@ def get_context(self):
                     'endpoint': 'organizations.list',
                 },
                 {
```
noirbizarre: Don't we have an equivalent with badges ?

udata/core/badges/metrics.py
```diff
@@ -0,0 +1,25 @@
...
+
+from .signals import on_badge_added, on_badge_removed
+
+
+class BadgesMetric(Metric):
+    name = 'badges'
+    display_name = _('Badges')
+
+    def get_value(self):
+        return (self.model_badge.objects(subject=self.target, removed=None)
```
noirbizarre: I think this is not working. It should be ``self.model.objects`` if I'm right

udata/core/site/views.py
davidbgk: I'm not sure that a metric for that is useful?

udata/core/badges/metrics.py
davidbgk: Actually, `self.model` is the subject, see https://github.com/etalab/udata/pull/55/files#diff-8b1584d4d7d89f813bfbf43533ce51dfR50 for instance and `model_badge` is the inherited model from `Badge`.

udata/core/badges/models.py
davidbgk: Given the display on the page, I think it's relevant to keep badges ordered by creation date. This way there will be "classic" ones first (public service, certified) and then additional and maybe more pertinent ones easier to scan for the visitor. I'm not sure about that though so maybe we should try and see once we've got more data?

udata/core/dataset/api.py
davidbgk: Thanks, copy-pasta evilness.

udata/core/badges/metrics.py
noirbizarre: ok

udata/core/badges/models.py
noirbizarre: Fine to me, seems more logical with the explanation :)

udata/core/site/views.py
noirbizarre: Not very much, but it's something that is used by the team to follow a little bit our "lobbying" progression.
Keep it like, we will see if someone ask for this.

noirbizarre:
r+wc

udata/core/site/views.py
davidbgk: Given that we set those badges it's a bit biased ;)

noirbizarre: It's only for keeping track of public services count ;)

PR #58, opened 2015-06-16
Merge reindex and unindex into a single task performing both

noirbizarre:
This PR merge reindex and unindex into a single one task, mostly allowing lazy action resolving (avoid unindexing when object has change and should be reindexed).


davidbgk:
r+

PR #59, opened 2015-06-16
Added organization datasets and reuses APIs

noirbizarre:
This pull-request add organisation specific APIs (same as user) for listing datasets and reuses.
This mainly allow:
- listing without using search
- display of private datasets and reuses into backoffice (and any other API based service).


davidbgk:
r+

PR #61, opened 2015-06-18
Add Piwik statistics to the API, fix #60

### LINKED ISSUE #60: Add Piwik statistics to the API
As an administrator, I want to log API requests in order to identify repetitive paths and improve the developer experience.

davidbgk:


udata/api/__init__.py
```diff
@@ -149,6 +173,38 @@ def set_api_language():
...
+
+@apiv1.after_request
+def collect_stats(response):
+    action_name = extract_name_from_path(request.full_path)
+    params = {
+        'action_name': urllib.quote(action_name),
+        'user_ip': request.remote_addr
+    }
+    if current_user.is_authenticated():
+        params['_id'] = str(hash(current_user.email)).strip('-')
```
noirbizarre: email as ID or user ID ?

udata/core/reuse/views.py
```diff
@@ -125,6 +134,14 @@ def initialize_form(self, form):
     def on_form_valid(self, form):
         response = super(ReuseCreateView, self).on_form_valid(form)
         notify_new_reuse.delay(self.object)
+        params = {
```
noirbizarre: Maybe we should encapsulate the action sending because it's always the same f(name, signal).
At least user extraction, remote adresse... will be coded once for all.

udata/tasks.py
```diff
@@ -53,6 +53,18 @@ def get_logger(name):
     return logger
 
 
+def connect(signal):
+    def wrapper(func):
+        t = task(func)
+
+        def call_task(issue, **kwargs):
```
noirbizarre: Not issue specific anymore. The parameter should be something more generic like ``subject``, ``item``...

davidbgk: Good catch, thanks!

udata/api/__init__.py
davidbgk: Good question. The length of the `current_user.id` is 24 chars and Piwik allows only 16 chars, is it safe to truncate it?

noirbizarre: I don't think so. i've seen many ObjectIDs differing only by the last character.
But maybe we can simply use:
```python
param['_id'] = str(hash(current_user.id)).strip('-')
```

By the way, we need to ensure the same UID is used client-side.

noirbizarre: Maybe the solution is to use ``uid`` instead of ``_id``: http://piwik.org/docs/user-id/

But this is only for logged in users (which is fine beacause we only track logged actions serverside)

Piwik can set a cookie ID: http://piwik.org/faq/how-to/faq_175/

I can see a ``PIWIK_SESSID`` in cookies but I don't know if it's reusable.

davidbgk: I used the `_id` because that's what is recommended in the [tracking API documentation](http://developer.piwik.org/api-reference/tracking-api) (unfortunately there is no explicit comparison with the `uid` mentioned in that same page and both description are a bit cryptic to understand which one is pertinent).

udata/core/reuse/views.py
davidbgk: Agreed.

udata/core/metrics/utils.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
```
noirbizarre: Is it the better place ?
Maybe be we should introduce a tracking module in order to group tracking related helpers ?

noirbizarre:
r+wc

PR #62, opened 2015-06-19
Added missing placeholder on mail reuse card

noirbizarre:
Fix missing reuse placeholder on mails.

### DIFF
--- a/udata/templates/reuse/mail_card.html
+++ b/udata/templates/reuse/mail_card.html
@@ -4,7 +4,7 @@
     <table height="94" cellpadding="0" cellspacing="0" style="margin: 0;padding: 0;width: 100%;border: 1px solid #eee;">
         <tr style="margin: 0;padding: 0;">
             <td width="98" style="margin: 0;padding: 0;">
-                <img src="{{ reuse.image(94, external=True) }}" style="margin: 0;padding: 0;width: 94px;height: 94px" width="94" height="94">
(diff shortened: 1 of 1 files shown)


udata/templates/reuse/mail_card.html
```diff
@@ -4,7 +4,7 @@
     <table height="94" cellpadding="0" cellspacing="0" style="margin: 0;padding: 0;width: 100%;border: 1px solid #eee;">
         <tr style="margin: 0;padding: 0;">
             <td width="98" style="margin: 0;padding: 0;">
-                <img src="{{ reuse.image(94, external=True) }}" style="margin: 0;padding: 0;width: 94px;height: 94px" width="94" height="94">
+                <img src="{{ reuse.image(94, external=True)|placeholder('reuse', True) }}" style="margin: 0;padding: 0;width: 94px;height: 94px" width="94" height="94">
```
davidbgk: I wonder if the inline style is still required now that we have explicit `width` and `height` attributes (not directly related but while you're at it :))

davidbgk:
r+wc

udata/templates/reuse/mail_card.html
noirbizarre: Sadly, it depends on the mail reader, some of them only interpret the width/height, others only interprets style, so both are required :/
Quite difficult to test, I didn't take the risk to break half of the clients.

Issue #69, opened 2015-06-24
Add tracking on the client side


davidbgk:
As an administrator, I want to track dedicated goals in order to make strategic decisions

PR #75, opened 2015-06-29
Add the share tracking goal, refs #69

davidbgk:
+ fix a bug related to FF+OSX/Safari+iOS

### DIFF
--- a/udata/commands/__init__.py
+++ b/udata/commands/__init__.py
@@ -30,8 +30,10 @@ def register_commands(manager):
     manager.add_command('clean', Clean())
     manager.add_command('urls', ShowUrls())
 
+    settings = os.environ.get('UDATA_SETTINGS', join(os.getcwd(), 'udata.cfg'))
     manager.add_command('serve', Server(port=6666, use_debugger=True,
-                                        use_reloader=True, threaded=True))
+                                        use_reloader=True, threaded=True,
+                                        extra_files=[settings]))
 
     # Load all commands submodule
     for filename in iglob(join(dirname(__file__), '[!_]*.py')):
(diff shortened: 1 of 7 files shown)


udata/static/js/widgets/share-btn.js
```diff
@@ -16,11 +17,17 @@ define([
...
             placement: 'top',
             container: 'body',
-            trigger: 'focus',
+            // trigger: 'focus', // Doesn't work on OSX+FF/iOS+Safari
             content: tpl({
                 url: $this.data('share-url'),
                 title: $this.data('share-title')
             })
+        }).on('shown.bs.popover', function() {
+            $(document.body).find('.share-click').each(function() {
```
noirbizarre: I think you can switch $(document.body) to $(this).

davidbgk: Unfortunately, I can't because `this` refers to the clicked element (`btn-share`) and links are appended at the end of the DOM, just before the `</body>`

udata/commands/__init__.py
```diff
@@ -30,9 +30,8 @@ def register_commands(manager):
     manager.add_command('clean', Clean())
     manager.add_command('urls', ShowUrls())
 
-    custom_settings = join(os.getcwd(), 'udata.cfg')
-    manager.add_command('serve', Server(port=6666, use_debugger=True, use_reloader=True))
-    # manager.add_command('serve', Server(port=6666, use_debugger=True, use_reloader=True, extra_files=[custom_settings]))
```
noirbizarre: I kept that until mitsuhiko/werkzeug#662 is fixed.
It seems to be done, maybe we should reintroduce the extra file.

udata/static/js/widgets/share-btn.js
noirbizarre: OK

noirbizarre:
r+wc

udata/commands/__init__.py
davidbgk: Is that better?

noirbizarre: Great ! :)

PR #81, opened 2015-06-30
Prefill suggest tag textarea, fix #80

### LINKED ISSUE #80: Bug: prefill the suggest tag textarea
When I click on the suggest tag button, the textarea is not always prefilled.

davidbgk:


noirbizarre:
r+

PR #83, opened 2015-06-30
Switch to CommonMark on the backend, fix #82

### LINKED ISSUE #82: Switch to CommonMark
As a writer, I want a similar Markdown rendering when I preview and display the same content to avoid any confusion. The solution is to use a consistent Markdown implementation: http://commonmark.org/

davidbgk:
Better sanitization too.


noirbizarre:
r+

Issue #93, opened 2015-07-08
Be kind with FTP checks and croquemort


davidbgk:
ATM, FTP servers are displayed as red resources (server not found), let's find a better way to handle that special case.

davidbgk:
Fixed in https://github.com/etalab/udata/commit/500b4638098f84e76ba917d90f746d30662b7d98

PR #94, opened 2015-07-08
Handling errors with FTP servers, refs #93

davidbgk:
Before:

capture d ecran 2015-07-08 a 18 01 26

After:

capture d ecran 2015-07-08 a 18 01 12


noirbizarre:
r+ (Good hack :))

PR #96, opened 2015-07-09
Display badges facets for organizations, fix #95

### LINKED ISSUE #95: Add facets for badges

davidbgk:


noirbizarre:
r+

udata/core/dataset/api.py
```diff
@@ -146,6 +146,9 @@ def post(self, dataset):
         badge = DatasetBadge(created=datetime.now(),
                              created_by=current_user.id)
         form.populate_obj(badge)
+        for existing_badge in dataset.badges:
+            if existing_badge.kind == badge.kind:
+                return badge
```
noirbizarre: It should be existing_badge as return value as its the one which is actually stored

noirbizarre:
r+ bis :)

Issue #99, opened 2015-07-10
Add moderation for discussions


davidbgk:
As an administrator, I want to be able to delete a discussion in order to avoid trolls and haters.

PR #100, opened 2015-07-10
Add the ability to delete a discussion, refs #99

davidbgk:


noirbizarre:
r+
I think we maybe need to be able to remove a single post from a discussion in a close futur but right now, it's fine to me

PR #102, opened 2015-07-10
Properly plug Sentry into Celery

noirbizarre:
This pull request apply recommandations from http://raven.readthedocs.org/en/latest/integrations/celery.html to ensure Sentry and Celery plays well togother with logging.


davidbgk:
r+

PR #107, opened 2015-07-17
Update mongoengine to 0.10.0 and pymongo to 3.0.3

davidbgk:


requirements/install.pip
```diff
@@ -22,10 +22,11 @@ Flask-Themes2==0.1.4
 Flask-Uploads==0.1.3
 Flask-WTF==0.11.0
 Flask==0.10.1
-mongoengine==0.9.0
+mongoengine==0.10.0
 pillow
 py-bcrypt==0.4
 pyliblzma==0.5.3
+pymongo==3.0.3
```
noirbizarre: Can be removed (transitive with mongoengine).
Or do we force it to pymongo>3.0

udata/core/followers/api.py
```diff
@@ -46,7 +46,12 @@ def get(self, id):
     @api.doc(id='follow', description=NOTE)
     def post(self, id):
         '''Follow an object given its ID'''
-        follow, created = self.model.objects.get_or_create(follower=current_user.id, following=id, until=None)
+        try:
+            follow = self.model.objects.get(follower=current_user.id, following=id, until=None)
+            created = False
+        except self.model.DoesNotExist:
+            follow = self.model.objects.create(follower=current_user.id, following=id, until=None)
+            created = True
```
noirbizarre: Do we make it a reusable get_or_create here:
https://github.com/etalab/udata/blob/master/udata/models/__init__.py#L84-L91

udata/core/site/views.py
```diff
@@ -26,10 +26,14 @@
...
-            'keywords': current_app.config.get('SITE_KEYWORDS', []),
-        })
+        try:
+            g.site = Site.objects.get(id=site_id)
+        except Site.DoesNotExist:
+            g.site = Site.objects.create(
+                id = site_id,
+                title = current_app.config.get('SITE_TITLE'),
+                keywords = current_app.config.get('SITE_KEYWORDS', []),
+            )
```
noirbizarre: Same as above

udata/models/slug_fields.py
```diff
@@ -107,10 +107,16 @@ def populate_slug(instance, field):
...
+        try:
+            slug_follower = SlugFollow.objects.get(
+                namespace=instance.__class__.__name__,
+                old_slug=old_slug
+            )
+        except SlugFollow.DoesNotExist:
+            slug_follower = SlugFollow.objects.create(
+                namespace=instance.__class__.__name__,
+                old_slug=old_slug
+            )
```
noirbizarre: Same as above

udata/tests/factories.py
```diff
@@ -128,7 +128,10 @@ class Meta:
 class AdminFactory(UserFactory):
     @factory.lazy_attribute
     def roles(self):
-        admin_role, _ = models.Role.objects.get_or_create(name='admin')
+        try:
+            admin_role = models.Role.objects.get(name='admin')
+        except models.Role.DoesNotExist:
+            admin_role = models.Role.objects.create(name='admin')
```
noirbizarre: Same as above

noirbizarre:
r+wc

udata/core/followers/api.py
davidbgk: I'm a bit afraid of adding a method they removed upstream because of the [impossibility to do that reliably with Mongo](https://github.com/MongoEngine/mongoengine/issues/300#issuecomment-111854695).

We might want to document why we think that we can do that safely for each of our case though.

udata/models/__init__.py
```diff
@@ -90,6 +90,33 @@ def bulk_list(self, ids):
...
+
+        Returns a tuple of ``(object, created)``, where ``object`` is
+        the retrieved or created object and ``created`` is a boolean
+        specifying whether a new object was created.
+
+        Taken back from:
+
+        https://github.com/MongoEngine/mongoengine/pull/1029/files#diff-05c70acbd0634d6d05e4a6e3a9b7d66b
+        """
+        defaults = query.get('defaults', {})
```
noirbizarre: query.pop('defaults', {}) ?

davidbgk: Good catch, I should review my copy-pasta…

PR #111, opened 2015-07-20
Add the --immediate option on travis

davidbgk:
Print errors and failures as they happen, as well as at the end.


noirbizarre:
r+

PR #112, opened 2015-07-20
Splitting a translation to avoid newlines

davidbgk:


noirbizarre:
Need to update the translations files, by either removing the references or updating them:
https://github.com/etalab/udata/blob/master/js/locales/udata.en.json#L59
https://github.com/etalab/udata/blob/master/js/locales/udata.es.json#L59
https://github.com/etalab/udata/blob/master/js/locales/udata.fr.json#L59

noirbizarre:
But R+ ;)

davidbgk:
@noirbizarre isn't it safer to do that via transifex?

noirbizarre:
Yes, but right now the extraction script only extract :)
So we will have a dead key. But maybe the right solution is a check script

PR #113, opened 2015-07-20
Fix js translatable string extraction

noirbizarre:
This PR fixes JS string extraction, move the function from udata commands to an invoke task (and cleanup invoke tasks)


davidbgk:
Flake8 and JSHint, yay!

noirbizarre:
Yes, we will need to fix lints soon, there are a lot of them ! :open_mouth:

davidbgk:
I can take care of that during a train commute ;)

noirbizarre:
Good :)
After that, maybe we need to integrate the QA check into the Travis build.
I don't know if it's a little bit violent...

davidbgk:
+1 for adding it to the Travis build, or maybe a pre-commit hook?

r+

Issue #115, opened 2015-07-21
Allow manual setting of published date for resources


davidbgk:
As a producer, I want to be able to control the published date for my resources in order to be pertinent for reusers.

PR #116, opened 2015-07-21
Add the ability to edit published date via the API, refs #115

davidbgk:
First step prior to hack the admin part.

### DIFF
--- a/udata/forms/fields.py
+++ b/udata/forms/fields.py
@@ -2,7 +2,7 @@
 from __future__ import unicode_literals
 
 import re
-
+from datetime import datetime
 from dateutil.parser import parse
 
 from flask import url_for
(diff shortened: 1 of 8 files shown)


udata/forms/fields.py
```diff
@@ -76,6 +76,14 @@ class IntegerField(FieldHelper, fields.IntegerField):
     pass
 
 
+class DateTimeField(Field):
+
+    def process_formdata(self, valuelist):
+        if valuelist:
+            date_time = valuelist[0].split('.')[0]  # Remove milliseconds.
+            self.data = datetime.strptime(date_time, "%Y-%m-%dT%H:%M:%S")
```
noirbizarre: OK but I think giving it to dateutil.parser.parse will avoid splitting on dot.

noirbizarre:
r+wc

PR #118, opened 2015-07-23
Emit signals on Dataset.add_resource

noirbizarre:
This pull request fix some missing signals emission on `Dataset.add_resource` and `Dataset.add_community_resource`.
By extension, it fixes reindexation on these operations.

### DIFF
--- a/udata/core/dataset/models.py
+++ b/udata/core/dataset/models.py
@@ -5,7 +5,7 @@ import datetime
 
 from collections import OrderedDict
 
-from blinker import Signal
+from blinker import signal
 from flask import url_for
 from mongoengine.signals import pre_save, post_save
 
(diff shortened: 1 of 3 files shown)


udata/core/dataset/models.py
```diff
@@ -5,7 +5,7 @@
 
 from collections import OrderedDict
 
-from blinker import Signal
+from blinker import signal
```
davidbgk: Wow, they changed the API without any breakage/warning?

davidbgk:
r+

udata/core/dataset/models.py
noirbizarre: No, to ease the debugging, I switch to named signals ;)

Issue #122, opened 2015-07-27
Put discussions and issues in context within the admin


davidbgk:
As an administrator, I want to see which dataset or reuse the discussions and issues are related to.

PR #123, opened 2015-07-27
Expose subject's class through the API, refs #122

davidbgk:


noirbizarre:
r+
Glad you're using ClassName

PR #129, opened 2015-07-28
Prefill the correct resource subform, fix #128

### LINKED ISSUE #128: Prefill the right resource subform on edit

davidbgk:


noirbizarre:
r+

PR #133, opened 2015-07-29
Improve the static collect command

noirbizarre:
This PR improves and fixes the static collect command:
- ensure every file is copied into the right folder:
  - app static
  - blueprint static
  - theme static and STATIC_URLS
- uses the command logger instead of raw `print` statements
- stop being verbose for each copied file.

### DIFF
--- a/udata/commands/static.py
+++ b/udata/commands/static.py
@@ -24,29 +24,45 @@ log = logging.getLogger(__name__)
 def collect(path, input):
     '''Collect static files'''
     if exists(path):
-        print('"{0}" directory already exists and will be erased'.format(path))
+        msg = '"%s" directory already exists and will be erased'
+        log.warning(msg, path)
         if input and not prompt_bool('Are you sure'):
             exit(-1)
 
-        print('Deleting static directory {0}'.format(path))
+        log.info('Deleting static directory "%s"', path)
         shutil.rmtree(path)
 
-    print('Copying assets into "{0}"'.format(path))
-    shutil.copytree(current_app.static_folder, path)
+    prefix = current_app.static_url_path or current_app.static_folder
+    if prefix.startswith('/'):
+        prefix = prefix[1:]
+    destination = join(path, prefix)
+    log.info('Copying application assets into "%s"', destination)
+    shutil.copytree(current_app.static_folder, destination)
+
+    for blueprint in current_app.blueprints.values():
+        if blueprint.has_static_folder:
+            prefix = current_app.static_prefixes.get(blueprint.name)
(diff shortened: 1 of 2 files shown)


udata/commands/static.py
```diff
@@ -24,29 +24,45 @@
...
 
-        print('Deleting static directory {0}'.format(path))
+        log.info('Deleting static directory "%s"', path)
         shutil.rmtree(path)
 
-    print('Copying assets into "{0}"'.format(path))
-    shutil.copytree(current_app.static_folder, path)
+    prefix = current_app.static_url_path or current_app.static_folder
+    if prefix.startswith('/'):
+        prefix = prefix[1:]
```
davidbgk: Can't we fix that at the source?

udata/commands/static.py
```diff
@@ -24,29 +24,45 @@
...
+    log.info('Copying application assets into "%s"', destination)
+    shutil.copytree(current_app.static_folder, destination)
+
+    for blueprint in current_app.blueprints.values():
+        if blueprint.has_static_folder:
+            prefix = current_app.static_prefixes.get(blueprint.name)
+            prefix = prefix or blueprint.url_prefix or ''
+            prefix += blueprint.static_url_path or ''
+            if prefix.startswith('/'):
+                prefix = prefix[1:]
```
davidbgk: Same here

udata/commands/static.py
```diff
@@ -24,29 +24,45 @@
 def collect(path, input):
     '''Collect static files'''
     if exists(path):
-        print('"{0}" directory already exists and will be erased'.format(path))
+        msg = '"%s" directory already exists and will be erased'
+        log.warning(msg, path)
```
davidbgk: Just curious, what are the pros of `msg, path` (vs. `msg % path` or `msg.format(path)`)?

davidbgk:
r+wc

udata/commands/static.py
noirbizarre: Not really: the `static_url_path` must start by a '/'

udata/commands/static.py
noirbizarre: Same here :)

udata/commands/static.py
noirbizarre: Personnally, I prefer the `msg.format()` template syntax.
I that case, `logging` only support string interpolation with `%` syntax. The interpolation is lazy: no logger active, no interpolation done.
When you log a lot of line in debug, it avoid evaluating it if the logging level is `warning`.

I don't think this is critical here, but I'm used to use this syntax with logging when possible to prevent hidden performance loss.

PR #135, opened 2015-07-31
Cache improvements

noirbizarre:
This pull request reduce the amount of db hits by caching, mostly some jinja templates.
Cache is set to 5 minutes everywhere.

Spatial granularities are cached forever.

### DIFF
--- a/udata/templates/dataset/card.html
+++ b/udata/templates/dataset/card.html
@@ -1,3 +1,4 @@
+{% cache 60*5, 'dataset-card', dataset.id|string, g.lang_code %}
 {% set dataset_url = url_for('datasets.show', dataset=dataset) %}
 <div class="card dataset-card">
     {% if dataset.organization %}
@@ -105,3 +106,4 @@
 
     </footer>
 </div>
+{% endcache %}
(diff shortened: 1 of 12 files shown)


udata/core/site/views.py
```diff
@@ -26,10 +26,15 @@
 def get_current_site():
     if getattr(g, 'site', None) is None:
         site_id = current_app.config['SITE_ID']
-        g.site, _ = Site.objects.get_or_create(id=site_id, defaults={
-            'title': current_app.config.get('SITE_TITLE'),
-            'keywords': current_app.config.get('SITE_KEYWORDS', []),
-        })
+        try:
+            g.site = Site.objects.no_dereference().get(id=site_id)
```
davidbgk: Oh, what was the issue here? Is it worth fixing it in our own `get_or_create`?

udata/templates/dataset/card.html
```diff
@@ -1,3 +1,4 @@
+{% cache 60*5, 'dataset-card', dataset.id|string, g.lang_code %}
```
davidbgk: We need a setting for that value.

udata/core/site/views.py
noirbizarre: Sorry, shouldn't have been in the PR, I will remove it now.

udata/templates/dataset/card.html
noirbizarre: Agreed, but how do we manager that: same cache for all snippets ?

davidbgk: At first, yes. It's already the case anyway ;) We can be more modular later.
Another option is dict + default value.

noirbizarre: I will go for the simpler as a start: TEMPLATE_CACHE_DURATION setting

davidbgk:
r+

PR #140, opened 2015-08-18
Add the frequency date, fix #139

### LINKED ISSUE #139: Add a frequency reminder date for datasets
As a producer, I want to be notified if I forget to update a dataset.

davidbgk:
Note that the front form is not updated and will not given that it is removed in https://github.com/etalab/udata/pull/136

### DIFF
--- a/udata/core/dataset/api_fields.py
+++ b/udata/core/dataset/api_fields.py
@@ -106,6 +106,9 @@ dataset_fields = api.model('Dataset', {
     'frequency': fields.String(
         description='The update frequency', required=True,
         enum=UPDATE_FREQUENCIES.keys(), default=DEFAULT_FREQUENCY),
+    'frequency_date': fields.ISODateTime(
+        description=('Next expected update date, you will be notified '
+                     'once that date is reached.')),
     'extras': fields.Raw(description='Extras attributes as key-value pairs'),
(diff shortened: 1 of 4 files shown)


udata/core/dataset/forms.py
```diff
@@ -33,6 +33,7 @@ class DatasetForm(ModelForm):
         choices=UPDATE_FREQUENCIES.items(), default=DEFAULT_FREQUENCY,
         validators=[validators.optional()],
         description=_('The frequency at which data are updated.'))
+    frequency_date = fields.DateTimeField(_('Reminder frequency date'))
```
noirbizarre: Reminder or Expected ?

udata/core/dataset/api_fields.py
```diff
@@ -106,6 +106,9 @@
     'frequency': fields.String(
         description='The update frequency', required=True,
         enum=UPDATE_FREQUENCIES.keys(), default=DEFAULT_FREQUENCY),
+    'frequency_date': fields.ISODateTime(
+        description=('Future date of update, you will be notified '
```
noirbizarre: Maybe 'Next expected update date' instead of 'Futur date of update'. What do you think about ?

noirbizarre:
r+wc

PR #145, opened 2015-08-21
Integrate pubsub from udata-admin

davidbgk:


js/dataset/display.js
```diff
@@ -96,7 +96,7 @@ define([
                     } else {
                         eventName = 'RESOURCE_REDIRECT';
                     }
-                    pubsub.publish(eventName);
+                    pubsub.pubsub.publish(eventName);
```
noirbizarre: Maybe this is worth migrating this file to ES6 and being able to write:
```javascript
import pubsub from 'pubsub';
```

js/widgets/share-btn.js
```diff
@@ -25,7 +25,7 @@ define([
         }).on('shown.bs.popover', function() {
             $(document.body).find('.share-click').each(function() {
                 $(this).on('click', function(e) {
-                    pubsub.publish('SHARE');
+                    pubsub.pubsub.publish('SHARE');
```
noirbizarre: Same as above

PR #147, opened 2015-08-31
Suggest fixes

noirbizarre:
This pull-request fix bad deduplications and allow suggest by ID on users and organizations suggesters.
It also fill the empty acronym field in org suggest.

This pull-request needs reindexing users and organizations.


davidbgk:
r+

PR #156, opened 2015-09-11
Timeline adjustments

noirbizarre:
Little adjustments and fixes on timeline

### DIFF
--- a/js/components/activities/timeline.vue
+++ b/js/components/activities/timeline.vue
@@ -8,14 +8,13 @@
                 <h5>
                     <span v-if="activity.aggregatedFollowing">
                         <span v-repeat="actor in activity.aggregaterActors">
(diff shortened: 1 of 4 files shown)


js/components/activities/timeline.vue
```diff
@@ -39,6 +39,10 @@ import moment from 'moment';
 import ActivityPage from 'models/activities';
 import URLs from 'urls';
 
+function actor(activity) {
```
davidbgk: Oh nice trick :)

davidbgk:
r+wc

js/components/activities/timeline.vue
noirbizarre: ;)
Which `v-if`, these one ? https://github.com/etalab/udata/blob/timeline-adjustments/js/components/activities/timeline.vue#L17-L18

davidbgk: Yes, and now I realize that the label is not the same… nevermind.

PR #157, opened 2015-09-11
Adding integration tests, fix #153

### LINKED ISSUE #153: Add integration tests for the frontend
As a developer, I want to be sure that core workflows are working before pushing to production.

davidbgk:
Add tests for creating datasets, orgs and reuses using watai and selenium.

You have to install and launch the selenium driver as explained [in the watai documentation](https://github.com/MattiSG/Watai#installing).

There are probably rooms for improvements given that I don't know either watai or selenium. @MattiSG your review would be really appreciated :)

Known issues:

* [x] credentials should be part of the configuration
* [x] using `setTimeout` to have the time to manually upload an image is the wrong way to do it and leads to error from watai
* [x] I should probably split a few files to be more atomic and reusable

The command to run tests:
```
$ selenium-server -p 4444
$ watai specs/integration/search/ --config '{"your":"options"}'
```


MattiSG:
I'm in da place  ;)

package.json
```diff
@@ -51,6 +51,7 @@
     "sinon-chai": "^2.8.0",
     "style-loader": "^0.12.3",
     "vue-loader": "3.0.3",
+    "watai": "0.6.2",
```
MattiSG: Watai follows SemVer, you can safely use `~0.6.2`.

specs/integration/search/2 - LoginFeature.js
```diff
@@ -0,0 +1,10 @@
+description: "Login with credentials.",
+
+scenario: [
+    LoginWidget.home(),
+    LoginWidget.open(),
```
MattiSG: Could be renamed for better readability. As an external reader, I understand `home`, but not `open`.

specs/integration/search/3 - CreateDatasetFeature.js
```diff
@@ -0,0 +1,20 @@
+description: "Create a dataset.",
+
+scenario: [
+    CreateDatasetWidget.contribute(),
+    CreateDatasetWidget.publishDataset(),
+    CreateDatasetWidget.next(),
+    CreateDatasetWidget.fill(datasetName, datasetDescription),
```
MattiSG: > What you're doing here is already excellent, it's very clear and tidy, it is a great initial investment. When you add tests on a per-feature basis though, it might be worth adding a few intermediary assertions to catch errors in the navigation path  :)

specs/integration/search/AdminWidget.js
```diff
@@ -0,0 +1 @@
+homeLink: ".main-header a.logo"
```
MattiSG: > Tip: make a habit of adding trailing commas. Widgets have a tendency to grow once they exist, and it will ease interventions on tests later on  :)

specs/integration/search/CreateDatasetData.js
```diff
@@ -0,0 +1,2 @@
+datasetName = "Test dataset " + Math.random().toString(36).substring(7)
```
MattiSG: > Nice, I didn't know this trick, thanks for the discovery  :)

specs/integration/search/CreateDatasetData.js
```diff
@@ -0,0 +1,2 @@
...
+datasetDescription = "Test dataset description"
```
MattiSG: Rather use `datasetName + " description"`, it can help in cleaning the database if something goes awry in tests.

specs/integration/search/CreateDatasetWidget.js
```diff
@@ -0,0 +1,12 @@
+contributeLink: 'a[title="Contribute!"]',
```
MattiSG: > You could also use `{ a: "Contribuer" }` as a [location method](https://github.com/MattiSG/Watai/wiki/Widgets#alternative-location-methods) ([example](https://github.com/sgmap/mes-aides-ui/blob/rewrite/test/integration/HomepageWidget.js#L3)).

Use what's the most likely to be stable: either the title attribute or part of the text content.

specs/integration/search/CreateDatasetWidget.js
```diff
@@ -0,0 +1,12 @@
...
+publishDatasetLink: ".actions-list a:first-child",
+nextLink: "button.pointer:last-child",
+titleField: "input#title",
+descriptionField: "textarea#description",
+setFrequencyOption: 'select#frequency option[value="punctual"]',
```
MattiSG: I guess you've encountered this issue where Selenium complains it wants `"` and not `'` for a valid CSS selector  ;)  That's why I'd advise using only single quotes throughout integration tests, to form a habit and avoid switching delimiters on a case-by-case basis.

specs/integration/search/CreateOrgData.js
```diff
@@ -0,0 +1,2 @@
+orgName = "Test org " + Math.random().toString(36).substring(7)
+orgDescription = "Test org description"
```
MattiSG: Same as above: I'd recommend using `orgName + " description"`.

specs/integration/search/CreateResourceWidget.js
```diff
@@ -0,0 +1,10 @@
+titleField: "input#title",
+openFileUploaderLink: '.resource-upload-dropzone input[type="file"]',
+
+fill: function fill(name) {
+    return  this.setTitleField(name)();
```
MattiSG: Why wrap this action and not simply use `setTitle` in the feature?

specs/integration/search/CreateResourceWidget.js
```diff
@@ -0,0 +1,10 @@
...
+},
+
+waitForUpload: function waitForUpload() {
+    setTimeout(function() {}, 7000); // TODO: fragile but improve your APM ;)
```
MattiSG: This can't work, as it doesn't return a promise. What's probably happening is that you add a step that Watai retries for its own timeout.

Watai tries really hard to avoid any hard timeouts, which are the plague of integration testing, as they're unreliable and decrease performance. You should always have a way to work around this by using UI elements, with the side-effect that you can trace race conditions your users could encounter.

Here, for example, the best solution would be to replace this step with an intermediate assertion in the feature that would assert the absence of a loading indicator (`{ uploadSpinner: false }`, for example).

> If necessary, you could [adjust the timeout](https://github.com/MattiSG/Watai/wiki/Features#options) for this specific step.

If you don't have such a loading indicator, well, it might be worth considering adding it to the UI  :)  Making code testable usually improves its quality, the same usually goes for UI testing  :)

- - - - -

If really you can't find something your users have to understand when the step is done, I'd advise you to use a [custom function](https://github.com/MattiSG/Watai/wiki/Features#custom-function) straight in the feature (or in a data helper file) to wait for that time.

specs/integration/search/CreateReuseWidget.js
```diff
@@ -0,0 +1,18 @@
+contributeLink: 'a[title="Contribute!"]',
+publishReuseLink: ".actions-list a:first-child + a",
+nextLink: "button.pointer:last-child",
+titleField: "input#title",
+urlField: "input#url",
+descriptionField: "textarea#description",
+setTypeOption: 'select#type option[value="post"]',
+openFileUploaderLink: '.image-picker input[type="file"]',
+
+fill: function fill(name, url, description) {
```
MattiSG: Why not get rid of this wrapper and simply call `setTitle`, `setUrl` and `setDescription` in the feature?

MattiSG:
Done reviewing, I didn't repeat some points everywhere. Global feedback:

1. This is great.
2. Consider intermediate assertions as a way to both handle timing issues and checking your UI gives proper timing feedback to users.
3. Considering the current syntax ugliness, avoid custom widget actions when it's only about chaining magic actions.

specs/integration/search/CreateReuseWidget.js
davidbgk: I hesitate between the two options as described [in the documentation](https://github.com/MattiSG/Watai/wiki/Features#example), I'm wondering if I left the responsibility to fill the whole form(s) to widgets or features.

specs/integration/search/CreateResourceWidget.js
davidbgk: Oh it explains everything, thanks for the suggestions, I'll go with the assertion on upload completed.

specs/integration/search/CreateReuseWidget.js
MattiSG: Mmh… Re-reading the doc, I might need to rethink either this piece of advice or the doc itself  ;)

MattiSG: Actually, I think the doc is right. If you need to reuse this widget in some other feature, you'll probably want to call an action.

davidbgk:
All tests are now passing (and documented!).

MattiSG:
Oh. After squashing, reviewing post-review changes only is impossible, making a second review more cumbersome  :(

specs/integration/1 - SearchFeature.js
```diff
@@ -0,0 +1,33 @@
+/*
+As an anonymous user, I want to perform various search queries from the homepage
+in order to find datasets, orgs, reuses and users.
```
MattiSG: Out of curiosity, why not store this BDD-style description as a comment rather than as the `description` itself?

MattiSG:
GTM, great job!

specs/integration/1 - SearchFeature.js
davidbgk: Good idea, I changed that :)

specs/integration/config.js
```diff
@@ -0,0 +1,13 @@
+/* You can overwrite this config using the --config command-line option.
+See: https://github.com/MattiSG/Watai/wiki/Configuration
+E.g.: $ watai specs/integration/  --config '{"baseURL":"http://example.org"}'
+
+Note that you will have to set at least these config options
+in order to perform authenticated operations:
+
+$ --config '{"email":"david@larlet.fr","password":"testpassword","username":"David Larlet"}'
```
MattiSG: Why not store this in the config directly, rather than as a comment? Furthermore, that would make tests runnable without a CLI switch.

davidbgk: Because each env will have its own config and I want it to let it be overwritten by the `~/.watai/config` file and that config file has a higher priority. It's a workaround to be able to use personal config or CLI. Is there a better option?

davidbgk:
@noirbizarre I think that one can still be merged on master (vs. dev).

specs/integration/config.js
MattiSG: Mmh… Not an exact answer to the specific problem you bring up, but you could be conceptually interested in what I do [here](https://github.com/sgmap/cout-embauche/blob/gh-pages/circle.yml#L20-L21).

PR #158, opened 2015-09-11
Fix apidoc

noirbizarre:
This fix API documentation display and add the missing API key management to the admin.


udata/core/user/api.py
```diff
@@ -175,6 +176,28 @@ def get(self):
...
 
+@me.route('/apikey', endpoint='my_apikey')
+class ApiKeyAPI(API):
+    @api.secure
+    @api.doc('generate_apikey')
+    @api.marshal_with(apikey_fields)
+    @api.response(201, 'API Key generated')
+    def post(self):
+        '''(Re)Generate my API Key'''
+        current_user.generate_api_key()
```
davidbgk: How long does it take? Is that a potential source of DoS? Should we threshold that API?

davidbgk:
r+wc

udata/core/user/api.py
noirbizarre: It's quite rapid, but you're right, there are some endpoint we should protect from DoS with some throttling.

PR #160, opened 2015-09-14
Fix transfer

noirbizarre:
This PR fix transfer request and response


davidbgk:
That would be nice to add an integration test for that ;)

PR #161, opened 2015-09-15
Force assets update on new version

noirbizarre:
This pull request make use of a cache bust parameter on static assets to force update on new delivery

### DIFF
--- a/udata/frontend/helpers.py
+++ b/udata/frontend/helpers.py
@@ -3,8 +3,10 @@ from __future__ import unicode_literals
 
 import calendar
 import logging
+import pkg_resources
 
 from datetime import date
+from time import time
 from urlparse import urlsplit, urlunsplit
 
 from flask import url_for, request, current_app, json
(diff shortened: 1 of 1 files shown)


udata/frontend/helpers.py
```diff
@@ -23,7 +25,11 @@
 
 @front.app_template_global(name='static')
 def static_global(filename):
-    return url_for('static', filename=filename)
+    if current_app.config['DEBUG']:
+        burst = time()
+    else:
+        burst = pkg_resources.get_distribution('udata').version
+    return url_for('static', filename=filename, _=burst)
```
davidbgk: I'm curious about `_=burst`, isn't it `_anchor=burst`? That parameter doesn't look to exist in the [Flask's documentation](http://flask.pocoo.org/docs/0.10/api/#flask.url_for).

noirbizarre: I took the same parameter as webpack, it makes short urls.
It will produce: ``http://xxx.com/static/filename?_=bust``.
It's not a flask parameter, flask adds the extra parameters (not registered in the url endpoint) as querystring

davidbgk:
r+

PR #171, opened 2015-09-18
Make check URL more clever, fix #170

### LINKED ISSUE #170: Make check URL more clever
Do not check FTP resources and `*.data.gouv.fr` domains.

davidbgk:
Do not check FTP resources and *.data.gouv.fr domains. And do not display an error when the croquemort server is down.

### DIFF
--- a/js/dataset/display.js
+++ b/js/dataset/display.js
@@ -22,7 +22,10 @@ import 'widgets/share-btn';
 let user_reuses;
 
 function startsWith(data, input) {
-    return (data.substring(0, input.length) === input);
+    return data.substring(0, input.length) === input;
+}
+function endsWith(data, input) {
+    return data.indexOf(input, data.length - input.length) !== -1;
 }
 
 function addTooltip($element, content) {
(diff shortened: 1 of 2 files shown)


js/dataset/display.js
```diff
@@ -52,24 +55,29 @@ function prepare_resources() {
...
+                            $self.addClass('format-label-success');
+                        } else if (data.status == '404') {
+                            $self.addClass('format-label-warning');
+                            addTooltip($self, i18n._('The resource cannot be found.'));
+                        }
+                    }).fail(function() {
+                        // The API check returns a 503 if the croquemort server itself is unreachable
+                        if (data.status !== '503') {
+                            $self.addClass('format-label-danger');
+                            addTooltip($self, i18n._('The server cannot be found.'));
```
noirbizarre: Maybe we can be more explicit about the fact it's croqumort, not the remote server which is unavailable.

udata/core/dataset/api.py
```diff
@@ -362,8 +362,11 @@ def get(self):
         if CROQUEMORT is None:
             return {'error': 'Check server not configured.'}, 500
         check_url = '{url}/check/one'.format(url=CROQUEMORT['url'])
-        response = requests.post(check_url,
-                                 data=json.dumps({'url': args['url']}))
+        try:
+            response = requests.post(check_url,
+                                     data=json.dumps({'url': args['url']}))
```
noirbizarre: I think we can be a little more defensive with a timeout

noirbizarre:
r+wc

js/dataset/display.js
davidbgk: Actually, it only displays the error when the remote server is unavailable, in case of croquemort nothing is displayed at all to the user and the resources remains with the default color/style.

udata/core/dataset/api.py
davidbgk: You're right, I updated the code for that.

PR #177, opened 2015-09-23
Add the ability to deal with community resources within the admin

davidbgk:


js/admin.vue
```diff
@@ -273,11 +274,11 @@ export default {
                     this.$nextTick(cb);
                 }
             } else {
-                require(['./views/' + name + '.vue'], function(options) {
-                    self.$options.components[name] = Vue.extend(options);
-                    self.view = name;
+                require(['./views/' + name + '.vue'], (options) => {
+                    this.$options.components[name] = Vue.extend(options);
```
noirbizarre: Are you sure about this ? Last time I checked, ``require`` override the ``this`` scope, even on arrow functions.

udata/core/dataset/api.py
```diff
@@ -30,15 +32,36 @@
...
 )
 from .permissions import DatasetEditPermission
-from .forms import ResourceForm, DatasetForm
+from .forms import ResourceForm, DatasetForm, CommunityResourceForm
 from .search import DatasetSearch
 
 ns = api.namespace('datasets', 'Dataset related operations')
 search_parser = api.search_parser(DatasetSearch)
-
+parser = api.parser()
```
noirbizarre: As there is multiple parsers in this file, maybe it can have a more explicit name like ``community_resources_parser``

udata/core/dataset/api_fields.py
```diff
@@ -76,6 +76,54 @@
...
+    'title': fields.String(description='The dataset title', readonly=True),
+    'uri': fields.UrlFor(
+        'api.dataset', lambda d: {'dataset': d},
+        description='The API URI for this dataset', readonly=True),
+    'page': fields.UrlFor(
+        'datasets.show', lambda d: {'dataset': d},
+        description='The web page URL for this dataset', readonly=True),
+})
+
+community_resource_fields = api.model('CommunityResource', {
```
noirbizarre: Inheriting from Resource could lead to a smaller specs, no ?

udata/core/dataset/forms.py
```diff
@@ -103,3 +103,33 @@ class ResourceForm(ModelForm):
     published = fields.DateTimeField(
         _('Publication date'),
         description=_('The publication date of the resource'))
+
+
+class CommunityResourceForm(ModelForm):
```
noirbizarre: Could be inherited from ResourceForm, no ?

udata/core/dataset/models.py
```diff
@@ -160,6 +155,14 @@ def is_available(self):
         return self.check_availability(group=None)
 
 
+class Resource(ResourceMixin, WithMetrics, db.EmbeddedDocument):
+    id = db.AutoUUIDField()
+    owner = db.ReferenceField('User')
```
noirbizarre: Does the resource still have an owner ? I think this one was only for community resources.

udata/routing.py
```diff
@@ -59,7 +60,10 @@ class ModelConverter(BaseConverter):
     model = None
 
     def to_python(self, value):
-        obj = self.model.objects(slug=value).first()
+        try:
+            obj = self.model.objects(slug=value).first()
+        except InvalidQueryError:  # If the model doesn't have a slug.
```
noirbizarre: Can't it be avoided with an `` if hasattr(self.model, 'slug')`` ?

js/components/communityresource/list.vue
```diff
@@ -0,0 +1,81 @@
...
+</template>
+
+
+<script>
+import Vue from 'vue';
+import CommunityResource from 'models/communityresource';
+
+export default {
+    name: 'community-widget',
+    props: ['communities', 'withoutDataset'],
```
noirbizarre: WHy not using resource(s) everywhere instead of communit(y|ies) ? It seem a little weird to me :/

davidbgk: Well, to avoid confusion between the two because the behavior is not the same and will probably diverge across time.

udata/routing.py
davidbgk: It's a matter of [EAFP](https://docs.python.org/2/glossary.html#term-eafp) vs. [LBYL](https://docs.python.org/2/glossary.html#term-lbyl). In that particular case, the fact that a slug is not present is the exception. BUT, we lost the specificity of catching only the issue if the query is invalid because of the missing slug… no I'm in doubt :)

js/admin.vue
davidbgk: I changed it and nothing broke, how can I be sure that it still works?

udata/routing.py
noirbizarre: lol, I didn't kown these Python documentation links ! :)
You got the point, keep it like that :)

noirbizarre:
r+wc

PR #178, opened 2015-09-25
Refactor notifications

noirbizarre:
This pull request extract notifications into its own module and allow to split declarations in their respective modules


davidbgk:
r+

PR #179, opened 2015-09-28
Hide deleted reuse on dataset page

noirbizarre:
This pull request ensure that private reuse, deleted and invalid reuse are neither dispayed on dataset page nor counted.


davidbgk:
r+

PR #184, opened 2015-10-01
Fix API references resolutions

noirbizarre:
Ensure that references (``$ref``) is the same everywhere and is compatible with last swagger-client version.


davidbgk:
Nice factorization, r+!

noirbizarre:
Thank ;)

PR #186, opened 2015-10-02
Fix user suggestion within the admin, fix #185

### LINKED ISSUE #185: Fix the user suggestion within the admin
As an administrator, I want to be able to preview users that I can add as a member to an organization.

davidbgk:


js/components/organization/members.vue
```diff
@@ -72,7 +72,7 @@
              v-repeat="member:org.members">
             <a class="pointer"
                  v-on="click: member_click(member)">
-                <img class="img-circle" alt="{{ _('User Image') }}"
+                <img class="img-circle" alt="{{ _('User Image') }}" width="60" height="60"
```
davidbgk: That modification is not part of the bug but it annoys me for months :)

noirbizarre: Be carefull, right now it scales to the display.
Have you tried with different display sizes ?

udata/core/user/api.py
```diff
@@ -169,8 +169,7 @@ def get(self):
         return [
             {
                 'id': opt['text'],
-                'first_name': opt['payload']['first_name'],
-                'last_name': opt['payload']['last_name'],
+                'fullname': opt['payload']['fullname'],
```
noirbizarre: It's intended to be the same fields as user or user reference.
It also match the fields from ``user/search.py``.

I think this is the ``js/components/form/user-completer.vue`` which need to be fixed.

js/components/organization/members.vue
davidbgk: Indeed. I tried to fix that on the CSS side, not ideal but better than fixed.

davidbgk:
@noirbizarre r?

noirbizarre:
r+

PR #189, opened 2015-10-02
Sidebar improvements

noirbizarre:
This PR update Vue and AdminLTE and alos display the org acronym in the sidebar when possible.


davidbgk:
r+

PR #192, opened 2015-10-05
Ensure sitemap cache key is a prefixed string

noirbizarre:
This pull-request nesure that sitemap cache keys are uniques (prefixed by sitemap-page-) and are strings (need by flask-cache).


davidbgk:
r+

PR #193, opened 2015-10-06
Fix the membership feature within the admin

davidbgk:


js/components/organization/members.vue
```diff
@@ -182,11 +191,15 @@ module.exports = {
...
-            // this.org.refuse_membership(request, comment, function(member) {
-            //     console.log('accepted', member);
-            // })
-            console.log('refuse', request);
+            request.$set('refused', true);
+        },
+        confirm_refusal: function(request, index) {
+            let comment = this.$children[index].$$[request.id].value;
+            this.org.refuse_membership(request, comment, (response) => {
+                console.log('refused', response);
```
noirbizarre: Need to be removed ;)

noirbizarre:
r+wc

js/components/organization/members.vue
davidbgk: Note to self: deal with refresh of the list in `accept_request` too.

PR #197, opened 2015-10-13
Dataset api improvements

noirbizarre:
This pull-request add some improvements to the Dataset API:

- handle bulk resources create and update
- handle geom field

It also consolidate the whole spatial module.

PR #199, opened 2015-10-14
Dataset api improvements rebased

noirbizarre:
This pull-request add some improvements to the Dataset API:

- handle bulk resources create and update
- handle geom field

It also consolidate the whole spatial module.

Replace #197


udata/core/dataset/api.py
```diff
@@ -233,18 +235,19 @@ def post(self, dataset):
...
     def put(self, dataset):
         '''Reorder resources'''
         DatasetEditPermission(dataset).test()
-        new_resources = []
-        for rid in request.json:
-            resource = get_by(dataset.resources, 'id', UUID(rid))
-            new_resources.append(resource)
-        dataset.resources = new_resources
-        dataset.save()
+        data = {'resources': request.json}
```
davidbgk: I think we should not have to do this at the server side but directly in the client/admin. Thoughts?

udata/core/dataset/forms.py
```diff
@@ -73,14 +42,8 @@ def __init__(self, label=None, validators=None, **kwargs):
...
 
-    def populate_obj(self, obj, name):
-        self._obj = self._obj or Checksum()
-        super(ChecksumField, self).populate_obj(obj, name)
-
-
-class ResourceForm(ModelForm):
-    model_class = Resource
 
+class BaseResourceForm(ModelForm):
```
davidbgk: I thought I already did this, weird…

udata/core/spatial/forms.py
```diff
@@ -0,0 +1,70 @@
...
+class GeomField(StringField):
+    def process_formdata(self, valuelist):
+        if valuelist:
+            value = valuelist[0]
+            try:
+                if isinstance(value, basestring):
+                    self.data = geojson.loads(value)
+                else:
+                    self.data = geojson.GeoJSON.to_instance(value)
+            except:
```
davidbgk: Can't we target more specific exceptions?

davidbgk:
r+wc

udata/core/dataset/api.py
noirbizarre: You suggest to replace the list by an object into the expected payload ? Is it not redundant with the path `/datasets/<id>/resources/` ?

davidbgk: Indeed, it's actually not really RESTful to have a method for reordering ;)

udata/core/dataset/forms.py
noirbizarre: You had:
```python
class ResourceForm(ModelForm):
    pass

class CommunityResourceForm(ResourceForm):
    pass
```
I had to refactor it into:
```python
class BaseResourceForm(ModelForm):
    pass

class ResourceForm(BaseResourceForm):
    pass

class CommunityResourceForm(BaseResourceForm):
    pass
```

davidbgk: Yup, I'm just surprised I didn't already did this that way :) We discussed this but I probably missed it before merging.

udata/core/spatial/forms.py
noirbizarre: I can but:
- geojson raise different exceptions that do not have a common ancestor
- it will raise a 500 instead of failing the data parsing with a validation error

udata/core/dataset/api.py
noirbizarre: WIth this PR, it's not only reordering now but buk update, which includes create, update and reordering.

PR #201, opened 2015-10-14
Esdoc/NVM

noirbizarre:
This PR add esdoc and nvm support


.nvmrc
```diff
@@ -0,0 +1 @@
+v4.2.1
```
davidbgk: Warning, some lib (at least `swifter`) are not compiling on the latest OSX with that version of nodejs. Still investigating.

noirbizarre: OK, I just rebased and rebuilt https://github.com/noirbizarre/sifter.js/tree/nested-properties-built until brianreavis/sifter.js#27 is merged. It should work on Mac too.

davidbgk: @vinyll can you try?

vinyll:
`npm install`ed successfully! :)

noirbizarre:
Great ! :)

PR #204, opened 2015-10-15
Migrations

noirbizarre:
This pull-request provide commands to handle database migrations.
It also add the initial migrations to the init command.


udata/commands/db.py
```diff
@@ -0,0 +1,142 @@
...
+
+m = submanager(
+    'db',
+    help='Databse related operations',
+    description='Handle all databse related operations and maintenance'
+)
+
+# A migration script wrapper recording the stdout lines
+SCRIPT_WRAPPER = '''
+function(plugin, filename, script) {{
```
davidbgk: JavaScript within Python, what could possibly go wrong here? ;)

udata/commands/db.py
```diff
@@ -0,0 +1,142 @@
...
+        filename: filename,
+        date: ISODate(),
+        script: script,
+        output: 'Recorded only'
+    }});
+}}
+'''
+
+# Date format used to for display
+DATE_FORMAT = '%Y-%m-%d %H:%M'
```
davidbgk: Why not [isoformat](https://docs.python.org/2/library/datetime.html#datetime.datetime.isoformat)?

noirbizarre: It's for command line display, so I wanted something more readable (space instead of T, no millis...)

udata/commands/db.py
noirbizarre: :)
It's the only way I found to keep existing migration, allow to keep writing JS migrations and to not depends on mongo client.

davidbgk:
Can we have a command to `unrecord` a migration? Use case: switching from a branch with a migration already done to another one without it would require the actual:
    mongorestore -h 127.0.0.1 -p 27017 -d udata ../backup/dataset.bson --drop

Plus, removing the record manually. Might be interesting to run a command instead, thoughts?

udata/commands/db.py
davidbgk: Externalizing it would help on syntax check and so on but add extra complexity, not sure it's better.

noirbizarre:
Yes I can do it easily.

It won't perform a reverse migration but simply remove the record allowing to replay it.

BTW, existing migrations are aleady skipped.

udata/commands/db.py
davidbgk: Alright.

noirbizarre:
Done ;)

davidbgk:
Great! r+

PR #206, opened 2015-10-16
Fix image picker

noirbizarre:
This pull-request fix mutualize image upload handling, fix image picker and ModelList.


udata/core/storages/api.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def parse_uploaded_image(field):
+    '''Parse an uploaded image and save into a db.ImageField()'''
+    args = image_parser.parse_args()
+
+    image = args['file']
+    bbox = args.get('bbox', None)
+    if bbox:
+        bbox = [int(float(c)) for c in bbox.split(',')]
```
davidbgk: `int` is not enough?

udata/forms/fields.py
```diff
@@ -370,15 +370,11 @@ def _value(self):
     def process_formdata(self, valuelist):
         if not valuelist:
             return []
-        if len(valuelist) > 1:
-            oids = [clean_oid(id, self.model) for id in valuelist]
-        elif isinstance(valuelist[0], basestring):
-            oids = [
-                clean_oid(id, self.model)
-                for id in valuelist[0].split(',') if id]
+        if len(valuelist) == 1 and isinstance(valuelist[0], basestring):
```
davidbgk: `if valuelist` looks to be enough or I miss a case?

davidbgk:
r+

udata/core/storages/api.py
noirbizarre: float to cast from string to float, int to round to the pixel.

PR #208, opened 2015-10-17
Fix sitemap

noirbizarre:
This pull-request should fix sitemap processing on some servers by greatly reducing the MongoDB payload.
It also add some missing URLs and adjust some priorities.


davidbgk:
Tests need to be updated accordingly but that's a great improvement for perfs! r+

PR #209, opened 2015-10-19
Added alternative Open License labelling handling

noirbizarre:
This is a temporary fix. It will be handled by mappings next (as both ODS and uData have dynamic licenses lists)

### DIFF
--- a/udata/harvest/backends/ods.py
+++ b/udata/harvest/backends/ods.py
@@ -22,6 +22,7 @@ class OdsHarvester(BaseBackend):
     LICENSES = {
         "Open Database License (ODbL)": "odc-odbl",
         "Licence Ouverte (Etalab)": "fr-lo",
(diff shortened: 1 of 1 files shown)


udata/harvest/backends/ods.py
```diff
@@ -22,6 +22,7 @@ class OdsHarvester(BaseBackend):
     LICENSES = {
         "Open Database License (ODbL)": "odc-odbl",
         "Licence Ouverte (Etalab)": "fr-lo",
+        "Licence ouverte / Open Licence": "fr-lo",
```
davidbgk: `Ouverte`? I don't know the official naming but on of these two lines needs to be adapted.

davidbgk:
r+wc

udata/harvest/backends/ods.py
noirbizarre: I just took the wording from their API :/

PR #210, opened 2015-10-19
Fix badges duplication

noirbizarre:
This pull request consolidate badges feature and fix badges duplication.

It also include a temporary fix until MongoEngine/mongoengine#1131 is released.

Also improve db migrations needed during development.


udata/migrations/2015-10-19-deduplicate-badges.js
```diff
@@ -0,0 +1,42 @@
...
+        }
+    });
+
+    if (doc.badges.length !== length) {
+        collection.save(doc);
+        return true;
+    }
+}
+
+process_collection(db.dataset);
```
davidbgk: Don't you think that we should remove `removed` and `removed_by` keys too?

udata/commands/db.py
```diff
@@ -134,18 +147,21 @@ def status():
 
 @m.option('-r', '--record', action='store_true',
           help='Only records the migrations')
-def migrate(record):
+@m.option('-d', '--dry-run', action='store_true', dest='dryrun',
```
davidbgk: yay!

udata/core/badges/api.py
```diff
@@ -3,8 +3,43 @@
...
+
+
 badge_fields = api.model('Badge', {
     'kind': fields.String(description=('Kind of badge (certified, etc), '
                                        'specific to each model'),
                           required=True),
 })
+
+
+def add_badge_api(obj):
```
davidbgk: I'm not fond of that naming. Proposition with pseudo-code:

* `badges/api.py`: `add` and `remove`
* `badges/__init__.py`: `import add, remove`
* `dataset/anywhere.py`: `from core import badges` then in code `badges.add()`

Thoughts?

udata/core/badges/models.py
```diff
@@ -14,15 +17,9 @@
...
 class Badge(db.EmbeddedDocument):
-    kind = db.StringField(choices=[], required=True)
+    kind = db.StringField(required=True)
     created = db.DateTimeField(default=datetime.now, required=True)
     created_by = db.ReferenceField('User')
-    removed = db.DateTimeField()
-    removed_by = db.ReferenceField('User')
-
-    meta = {
-        'ordering': ['created'],
```
davidbgk: A default ordering can be useful to memorize the consistent place of the badge in a list for mass addition for instance. To be confirmed by admins :)

davidbgk:
I'm nitpicking a bit but that's a great factorization, I just want to be sure that we have a clean API if we decide to generalize that pattern :)

udata/migrations/2015-10-19-deduplicate-badges.js
noirbizarre: I checked first: they were not used, no document has it so it's ot necessary to remove it. ;)

udata/commands/db.py
noirbizarre: :)
This is a first attempt, a poor man dry run: It only print the request instead of executing it. But it's a start.

udata/core/badges/api.py
noirbizarre: I like the factory pattern but I will do like this, to be explicit about the fact it's meant for the API
```python
#badges/api.py
def add():
    pass

def remove():
    pass

#somewhere/somemodule.py
# from udata.features import badges (futur)
from udata.core import badges
badges.api.add()
```
Or other alternative: 
```python
from udata.core.badges import api as badge_api
badges_api.add()
```

I have to check if when importing badges in other modules not needing API there is no side effects.

udata/core/badges/models.py
noirbizarre: Default ordering is only used for document on find without sort.
EmbeddedDocumentList preserve order set on save.

noirbizarre:
Agreed. Each time, I try to improve the API and I think this is the right moment to think about naming, packaging, conventions... so you're not nitpicking, you're right ;)

udata/core/badges/api.py
davidbgk: I prefer the first one

PR #211, opened 2015-10-20
Expect an utf8-compatible terminal by default

noirbizarre:
This pull-request set the default output to utf-8 and fiix terminal output and encoding error when detection fail (ie. Jenkins)


davidbgk:
r+ hopefully it doesn't affect other parts on import

noirbizarre:
I already used this solution and the only part affected was stdout encoding detection. Let's hope this is the same here :pray:

PR #214, opened 2015-10-21
Fix admin page metadatas

noirbizarre:
This pull-request fix org page title not updated issue by fixing an issue with local sorting with Sifter.

It also ensure page metadata get properly reinitialized and upgrade some dependencies.


davidbgk:
r+

PR #215, opened 2015-10-23
LESS @input-color set to inherit. Fix #368

vinyll:


vinyll:
Cancelled for https://github.com/etalab/udata/pull/217

Issue #216, opened 2015-10-23
input colors are too light to be readable


vinyll:
See the _titre_ and _Commentaire_ field on the picture
<img width="856" alt="capture_d ecran_2015-10-16_a_20 54 36" src="https://cloud.githubusercontent.com/assets/145172/10692870/934d2c8a-7994-11e5-81aa-3087f3cce915.png">

They should be darkened maybe to match the regular text.

PR #217, opened 2015-10-23
Fixing the unreadable input colors

vinyll:
See bug #216


noirbizarre:
I think you need to cherry-pick the commit

vinyll:
replaced by #218

PR #218, opened 2015-10-23
LESS @input-color set to inherit. Fix #216

### LINKED ISSUE #216: input colors are too light to be readable

vinyll:
fixes #216, replaces #217


noirbizarre:
r+

PR #219, opened 2015-10-23
Fix vue loader peer dependencies

noirbizarre:
This PR switch from webpack html-loader to vue-html-loader ensuring compatibility with vue.js.
It also fixes some invoke tasks ensuring local binaries are used instead of global ones.


davidbgk:
r+, I didn't know about the `nrun`, nice!

noirbizarre:
Yes, I was tired of ``node_modules/.bin/whatever`` :)

PR #220, opened 2015-10-26
Fix resources forms

noirbizarre:
This pull request fixes resources forms validations, serializations and ergonomy.


davidbgk:
r+discussion about ergonomic on the dedicated Trello card.

PR #221, opened 2015-10-26
Api model resolution

noirbizarre:
This pull-request provides utils to perform model class resolution and/or checking client or server-side.
This fixes the transfer API and UI.

Note: Discussions and Issues can be simplified but this out of the scope of this pull-request:
- inheritance can be removed
- can use the is filter once refactored


davidbgk:
r+

PR #222, opened 2015-10-27
Ensure zone not found does not raise Exception in labelizer

noirbizarre:
This pull-request prevent exception on empty zone identifier un search


davidbgk:
r+is that possible to have more than one object?

noirbizarre:
No it's not. It's used on each querystring value. So if there is more than one value, it will be called multiple times.

Issue #223, opened 2015-10-27
Deal with server-side errors on admin updates


davidbgk:
As a logged in user, I want to be informed of errors raised by the backend/API when I update objects through the administration interface.

PR #224, opened 2015-10-27
Deal with server-side errors on admin, refs #223

davidbgk:
capture d ecran 2015-10-27 a 11 38 33

### DIFF
--- a/js/components/form/base-form.js
+++ b/js/components/form/base-form.js
@@ -197,13 +197,13 @@ export default {
                 try {
                     data = JSON.parse(response.data);
                 } catch (e) {
-                    console.log('Parsing error:', e);
+                    log.warn('Parsing error:', e);
                     return;
                 }
                 if ('errors' in data) {
                     this.fill_errors(data.errors);
                 } else {
-                    console.log('Unknown error:', e);
+                    log.warn('Unknown error:', e);
                 }
             }
         },
@@ -211,7 +211,7 @@ export default {
             [...this.$form.querySelectorAll('input,textarea,select')].forEach((element) => {
                 if (element.name in errors) {
                     let name = element.name;
-                    let error = errors[element.name][0];
+                    let error = errors[name][0];
                     let errorElement = `<label for="${name}" class="help-block" id="${name}-error">${error}</label>`;
                     $(element).closest('.form-group,.field-wrapper').removeClass('has-success').addClass('has-error').append(errorElement);
                 }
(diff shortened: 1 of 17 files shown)


js/components/form/base-form.js
```diff
@@ -190,6 +190,32 @@ export default {
...
             return out;
-        }
+        },
+        on_error: function (response) {
+            if ('data' in response) {
+                let data = {};
+                try {
+                    data = JSON.parse(response.data);
+                } catch (e) {
+                    console.log('Parsing error:', e);
```
noirbizarre: Use logger instead of console

js/components/form/base-form.js
```diff
@@ -190,6 +190,32 @@ export default {
...
+                    return;
+                }
+                if ('errors' in data) {
+                    this.fill_errors(data.errors);
+                } else {
+                    console.log('Unknown error:', e);
```
noirbizarre: same as above

js/components/form/base-form.js
```diff
@@ -190,6 +190,32 @@ export default {
...
+                }
+            }
+        },
+        fill_errors: function(errors) {
+            [...this.$form.querySelectorAll('input,textarea,select')].forEach((element) => {
+                if (element.name in errors) {
+                    let name = element.name;
+                    let error = errors[element.name][0];
```
noirbizarre: `errors[name]` ?

js/models/base.js
```diff
@@ -94,15 +94,16 @@ export class Base {
...
-     * @param  {Function} callback The callback function to call on success.
+     * @param  {Function} on_success The callback function to call on success.
+     * @param  {Function} on_error The callback function to call on error.
      */
-    $api(endpoint, data, callback) {
-        var parts = endpoint.split('.'),
-            namespace = parts[0],
-            method = parts[1],
-            operation = API[namespace][method];
+    $api(endpoint, data, on_success, on_error=()=>{}) {
```
noirbizarre: I think it will introduce multiple binding and then memory leak.
May be better to have ``$api(endpoint, data, on_success, on_error)``
and then ``(on_error || () =>{}).bind(this)``

noirbizarre:
r+wc

js/models/base.js
davidbgk: I don't think that's the case, contrary to Python default parameters are evaluated at call time: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#Evaluated_at_call_time

noirbizarre: Oh, good to know !

PR #226, opened 2015-10-27
Add discussions to timelines, fix #225

### LINKED ISSUE #225: Add discussions to the timeline of orgs, users and dashboard
Let's make it more social!

davidbgk:
capture d ecran 2015-10-27 a 14 17 19

### DIFF
--- a/udata/core/user/activities.py
+++ b/udata/core/user/activities.py
@@ -7,7 +7,9 @@ from udata.i18n import lazy_gettext as _
 from udata.models import db, User, Dataset, Reuse, Organization, Activity
 from udata.core.activity.tasks import write_activity
 from udata.core.followers.signals import on_follow
-
+from udata.core.discussions.signals import (
+    on_new_discussion, on_new_discussion_comment
+)
 from udata.core.dataset.activities import DatasetRelatedActivity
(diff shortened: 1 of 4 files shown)


udata/core/user/activities.py
```diff
@@ -70,3 +87,17 @@ def write_activity_on_follow(follow, **kwargs):
             write_activity.delay(
                 UserFollowedUser, current_user._get_current_object(),
                 follow.following)
+
+
+@on_new_discussion.connect
+@on_new_discussion_comment.connect
+def write_activity_on_discuss(discussion, **kwargs):
+    if current_user.is_authenticated:
```
noirbizarre: I think this is `is_authenticated()` while using Flask-Login 0.2.x. The property is available starting Flask-Login 0.3.x (we need to migrate).
So, it's a false positive as it's always true.

noirbizarre:
r+wc

udata/core/user/activities.py
davidbgk: Wow, that's scary. I updated that everywhere else and given the number of places I'm surprised we didn't hit any bug related to that!

PR #239, opened 2015-11-04
Improve search relevance, fix #238

### LINKED ISSUE #238: Improve search pertinence
As a visitor I want to find popular and relevant datasets first.

davidbgk:
Not ideal for now because it requires an update of https://github.com/etalab/geozones to load all postal codes but better than nothing.

### DIFF
--- a/udata/core/dataset/search.py
+++ b/udata/core/dataset/search.py
@@ -249,8 +249,7 @@ class DatasetSearch(ModelSearchAdapter):
                 geozones.append({
                     'id': zone.id,
                     'name': zone.name,
-                    'keys': [key for key in zone.keys
-                             if not key.startswith('-')]  # Avoid -99.
+                    'keys': zone.keys_values
                 })
                 parents |= set(zone.parents)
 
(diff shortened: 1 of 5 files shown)


udata/core/dataset/search.py
```diff
@@ -246,7 +249,8 @@ def serialize(cls, dataset):
                 geozones.append({
                     'id': zone.id,
                     'name': zone.name,
-                    'code': zone.code
+                    'keys': [key for key in zone.keys
```
noirbizarre: I think there is now two path for this with etalab/geozones#4:
- zone.keys is a string
- zone.keys is an array of string

noirbizarre:
I think spatial API, models,commands and/or search may need update to support keys as array.

udata/core/spatial/search.py
```diff
@@ -28,7 +28,9 @@ def serialize(cls, zone):
         return {
             'zone_suggest': {
                 'input': list(set(
-                    [zone.name, zone.code] + zone.keys.values())),
+                    [zone.name, zone.code]
+                    + [key for key in zone.keys.values()]
```
noirbizarre: Is it working for postal codes (when key is a list) ?

noirbizarre:
r+

davidbgk:
@noirbizarre did you try to run tests for that branch on your local instance?

noirbizarre:
no, I will

PR #241, opened 2015-11-05
Fix the generic error handling on updates, fix #240

### LINKED ISSUE #240: Fix the generic error handling on dataset/reuse/organization updates

davidbgk:
And some refactoring

### DIFF
--- a/udata/api/errors.py
+++ b/udata/api/errors.py
@@ -1,0 +1,6 @@
+from udata.i18n import gettext as _
+
+
+VALIDATION_ERROR = _('Validation error: your data cannot be updated for '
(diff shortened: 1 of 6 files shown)


udata/api/errors.py
```diff
@@ -0,0 +1,6 @@
+from udata.i18n import gettext as _
```
noirbizarre: You should be using lazy_gettext in order to have the language resolved on query time instead of declaration time.

noirbizarre:
r+wc

udata/api/errors.py
davidbgk: We tried, quite hard, and it doesn't work at that level… :(

PR #245, opened 2015-11-07
Harvest attach

noirbizarre:
This pull-request add the `udata harvest attach` command allowing to link existing datasets to their remote ID from a CSV before harvest.

It also fixes the ODS backend HTML to Markdown conversion

### DIFF
--- a/udata/harvest/actions.py
+++ b/udata/harvest/actions.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
-import csv
 import logging
+import unicodecsv as csv
 
 from collections import namedtuple
 from datetime import datetime
--- a/udata/harvest/actions.py
+++ b/udata/harvest/actions.py
@@ -1,14 +1,16 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+import csv
 import logging
 
 from datetime import datetime
 
+from bson import ObjectId
 from flask import current_app
 
 from udata.auth import current_user
-from udata.models import User, Organization, PeriodicTask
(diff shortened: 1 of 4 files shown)


udata/harvest/actions.py
```diff
@@ -162,3 +164,43 @@ def unschedule(ident):
...
+    The expected csv file format is the following:
+
+    - a column with header "local" and the local IDs or slugs
+    - a column with header "remote" and the remote IDs
+
+    The delimiter should be ";". columns order
+    and extras columns does not matter
+    '''
+    count = 0
+    with open(filename) as csvfile:
```
davidbgk: No potential encoding issue?

udata/harvest/actions.py
```diff
@@ -162,3 +164,43 @@ def unschedule(ident):
...
+        reader = csv.DictReader(csvfile,
+                                delimiter=b';',
+                                quotechar=b'"')
+        for row in reader:
+            try:
+                dataset = Dataset.objects.get(id=ObjectId(row['local']))
+            except:  # noqa  (Never stop on failure)
```
davidbgk: Might be interesting to have a counter for failures too? (I don't know if it happens often)

davidbgk:
r+wc

udata/harvest/actions.py
noirbizarre: It's only IDs, no unicode but your right, maybe I should use unicodecsv here too.

udata/harvest/actions.py
noirbizarre: Errors are logged into the shell but I can also return a tuple `(success, errors)`

PR #248, opened 2015-11-17
removed all references of a supplier. fix #247

### LINKED ISSUE #247: remove any reference of 'supplier'
The code has references to a _supplier_ property for models such as _Dataset_.
It should be removed from the code as well as other references to it.

vinyll:
Let's first see how Jenkins react.
The API was affected removing the field _supplier_. Shouldn't we increment the API version?


davidbgk:
I relaunched the Travis build, not sure the error was related to the modification.
About the API, we don't have such policy yet but that might be a good idea, @noirbizarre?

davidbgk:
r+

.gitignore
```diff
@@ -38,6 +38,7 @@ js-built
 /node_modules/
 /udata/static/
 /jsdoc/
+dump.rdb
```
noirbizarre: *.rdb ?

davidbgk: redis, if launched in the same folder.

udata/migrations/2015-11-16-remove-supplier.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Remove supplier property from datasets
+ */
+function process_collection(collection) {
+    collection.update({supplier: {$exists: true}},
+                                 {$unset: {'supplier': 1}}, false, true);
```
noirbizarre: It should be indented with `{supplier` instead of `{$exists`

.gitignore
vinyll: The demand is to wildcard rdb extensions, right?

udata/migrations/2015-11-16-remove-supplier.js
```diff
@@ -0,0 +1,11 @@
...
+    var count = db.runCommand({getLastError : 1}).n;
```
noirbizarre: You can also use the update return value which contain `nModified`. The current pattern will only work if there was no other operation in the interval.

.gitignore
noirbizarre: Exactly ;)

noirbizarre:
r+wc

PR #249, opened 2015-11-17
Added theme variants support

noirbizarre:
This pull-request add theme variants support allowing to change theme details with a simple configuration entry (`THEME_VARIANT`)

### DIFF
--- a/udata/templates/raw.html
+++ b/udata/templates/raw.html
@@ -58,7 +58,7 @@
 
   </head>
 
-  <body class="{{body_class}}">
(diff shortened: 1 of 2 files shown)


davidbgk:
Maybe we should add/recommend some kind of namespacing convention for themes' CSS. r+

noirbizarre:
Yep, I made it quick but we need to cleanup the less files, it will be the right time to provide conventions.

This PR introduce it `.theme-default`/`.theme-{variant}` namespace.

PR #252, opened 2015-11-18
Make the admin search actually working, fix #251

### LINKED ISSUE #251: Make the admin search actually working

davidbgk:
capture d ecran 2015-11-18 a 15 29 32

### DIFF
--- a/js/components/communityresource/list.vue
+++ b/js/components/communityresource/list.vue
@@ -18,16 +18,22 @@ export default {
     props: {
         communities: {
             type: Object,
-            default: function() {
+            default() {
                 return new CommunityResources();
             }
         },
-        withoutDataset: Boolean
+        withoutDataset: Boolean,
+        title: {
+            type: String,
+            default() {
+                return this._('Community resources');
+            }
+        }
     },
     components: {
          datatable: require('components/datatable/widget.vue')
     },
-    data: function() {
+    data() {
         let fields = [{
             label: this._('Title'),
             key: 'title',
@@ -47,13 +53,12 @@ export default {
             });
         }
         return {
-            title: this._('Community resources'),
             fields: fields,
             community: new CommunityResource(),
         };
     },
     events: {
-        'datatable:item:click': function(community) {
+        'datatable:item:click'(community) {
             this.$go({name: 'dataset-community-resource', params: {
                 oid: community.dataset.id, rid: community.id
             }});
--- a/js/components/dataset/list.vue
+++ b/js/components/dataset/list.vue
@@ -13,9 +13,8 @@ export default {
     components: {
         datatable: require('components/datatable/widget.vue')
     },
-    data: function() {
+    data() {
         return {
-            title: this._('Datasets'),
             fields: [{
                 label: this._('Title'),
                 key: 'title',
@@ -72,7 +71,7 @@ export default {
         };
     },
     events: {
-        'datatable:item:click': function(dataset) {
+        'datatable:item:click'(dataset) {
             this.$go('/dataset/' + dataset.id + '/');
         }
     },
@@ -80,8 +79,14 @@ export default {
         datasets: null,
         downloads: {
             type: Array,
-            default: function() {
+            default() {
                 return [];
+            }
+        },
+        title: {
+            type: String,
+            default() {
+                return this._('Datasets');
             }
         }
     }
--- a/js/components/sidebar.vue
+++ b/js/components/sidebar.vue
@@ -3,10 +3,10 @@
     <section class="sidebar">
         <!-- search form -->
         <div class="sidebar-form">
-            <form method="get">
+            <form method="get" @submit.prevent="onSearch">
                 <div class="input-group">
                     <input type="text" name="q" class="form-control"
-                        :placeholder="search_label" />
+                        :placeholder="search_label" v-el:terms />
                     <span class="input-group-btn">
                         <button type="submit" name="search" id="search-btn" class="btn btn-flat">
                             <i class="fa fa-search"></i>
@@ -67,11 +67,11 @@ export default {
         };
     },
     computed: {
-        menu: function() {
+        menu() {
             var menu = MENU.concat(this.organizations_menus);
             return this.$root.me.has_role('admin') ? menu.concat(bottom_menu) : menu;
         },
-        organizations_menus: function() {
+        organizations_menus() {
             if (!this.$root.me.organizations) {
                 return [];
             }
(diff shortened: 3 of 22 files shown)


js/views/search.vue
```diff
@@ -0,0 +1,93 @@
...
+    </div>
+    <div class="row">
+        <issues class="col-xs-12" :issues="issues" :title="issuesTitle">
+        </issues>
+    </div>
+    <div class="row">
+        <discussions class="col-xs-12" :discussions="discussions" :title="discussionsTitle">
+        </discussions>
+    </div>
+</template>
```
noirbizarre: Missing closing `</layout>`

js/components/sidebar.vue
```diff
@@ -83,6 +83,15 @@ export default {
                 };
             });
         }
+    },
+    methods: {
+        onSearch() {
+            const terms = this.$els.terms.value;
+            if (terms && terms.length > 2) {
+                this.$go({name: 'search', query: { terms: terms }});
+                this.$root.me.$emit('updated');
```
noirbizarre: Why ?

js/components/communityresource/list.vue
```diff
@@ -22,7 +22,13 @@ export default {
                 return new CommunityResources();
             }
         },
-        withoutDataset: Boolean
+        withoutDataset: Boolean,
+        title: {
+            type: String,
+            default: function() {
```
vinyll: Some code below is converted to ES6 syntax. Why not `default() {…}` here and in the rest?

js/models/base.js
```diff
@@ -515,4 +515,16 @@ export class PageList extends List {
         this.populate();
         return this;
     }
+
+    /**
+     * Perform a server-side search
+     * @param  {String} query The query string to perform the search on.
+     * @return {Object} Return itself allowing to chain methods.
+     */
+    search_server(query) {
```
noirbizarre: I prefer:
```javascript
on_fetched(data) {
    super.on_fetched(data);
    this.page = 1;  // Clear the paging
}
```
And then simply call `fetch({q: 'my seach'})`.
Like that it does not create a special path.

js/views/search.vue
```diff
@@ -0,0 +1,93 @@
...
+    },
+    detached() {
+        this._handler.remove();
+    },
+    methods: {
+        update() {
+            if (this.$root.me.id) {
+                const terms = this.$route.query.terms;
+                this.datasetsTitle = this._(`Datasets related to your data about "${terms}"`);
+                this.datasets.search_server(terms);
```
noirbizarre: given the comment on `server_search`, it should/could/will be:
```javascript
this.datasets.fetch({q: terms});
```
And so on for the others.

js/components/sidebar.vue
davidbgk: That's the only way I found to trigger an event which reset the search :/

udata/core/dataset/models.py
```diff
@@ -83,7 +83,15 @@ def __unicode__(self):
         return self.title
 
 
-class DatasetQuerySet(db.BaseQuerySet):
+class OwnedByQuerySet(db.BaseQuerySet):
```
noirbizarre: As the same pattern is used by `Reuse`, maybe it should be extracted as a mixin into `udata/models/__init__.py` so we could write something like:
```python
class OwnedByQuerySet(db.BaseQuerySet, db.OwnedQuerySetMixin):
    pass

class DatasetQuerySet(db.BaseQuerySet, db.OwnedQuerySetMixin):
    def visible(self):
        # ...
```

js/models/base.js
davidbgk: By `on_fetched` you mean `fetch`?

js/views/search.vue
davidbgk: Well, it does work as-is. But know I wonder why ;)

udata/core/user/api.py
```diff
@@ -30,6 +34,10 @@
 ns = api.namespace('users', 'User related operations')
 me = api.namespace('me', 'Connected user related operations')
 search_parser = api.search_parser(UserSearch)
+filter_parser = api.parser()
+filter_parser.add_argument(
+    'q', type=unicode, help='The string to filter items',
```
noirbizarre: Given the `from __future__ import unicode_literals` it should be `str` instead of `unicode`

js/components/dataset/list.vue
```diff
@@ -80,9 +79,15 @@ export default {
         datasets: null,
         downloads: {
             type: Array,
-            default: function() {
+            default() {
```
vinyll: A step further would be

    default: () => [];

:smile:

js/models/base.js
noirbizarre: no, I clear paging on data received avoiding going back to first page before new data is present.

js/components/sidebar.vue
noirbizarre: By reset, do you mean "empty the search field" or "trigger another server search" ?

js/components/dataset/list.vue
davidbgk: Sadly, you can't do that because it's already some kind of magic trickery (note that it's `default()`, not `default: ()`)

js/components/sidebar.vue
davidbgk: Trigger another event from the sidebar which refresh the view, I just want to update results when the search terms are modified in the sidebar.

udata/core/user/api.py
davidbgk: Alright, I'll change that everywhere.

js/components/sidebar.vue
noirbizarre: OK, so I think you can remote this emit and just update on route change in search.vue.
I will comment in search.vue

js/views/search.vue
```diff
@@ -0,0 +1,94 @@
...
+    attached() {
+        this.update();
+        this._handler = this.$root.me.$on('updated', this.update.bind(this));
+    },
+    detached() {
+        this._handler.remove();
+    },
+    methods: {
+        update() {
+            if (this.$root.me.id) {
```
noirbizarre: You can remove this line (User id is not needed as it's resolved server-side).
But you can trigger search update on route change:
```javascript
route: {
        data() {
            this.update();
        }
    },
```
Or, given it's updated only on route change, you can directly move the update code into route.data()

noirbizarre: And then, `attached` and `detached` are not needed anymore too

noirbizarre:
r+wc

PR #255, opened 2015-11-18
replacing marked.js with commonmark. fix #253

### LINKED ISSUE #253: markdown server and client have different spec
We should use http://commonmark.org/ implementation on both Py and JS to guarantee a similar behavior.

vinyll:
fixing #253 
- Replaced marked with [markdown-it](https://github.com/markdown-it/markdown-it) which is commonmark compliant
- created a preconfigured and easy to call module (helpers/commonmark.js)

### DIFF
--- a/js/components/form/markdown-editor.vue
+++ b/js/components/form/markdown-editor.vue
@@ -52,10 +52,10 @@
 import $ from 'jquery';
 import Vue from 'vue';
 import {FieldComponentMixin} from 'components/form/base-field';
+import commonmark from 'helpers/commonmark';
 
 const EXCERPT_TOKEN = '<!--- excerpt -->';
 
-window.marked = require('marked');
 require('bootstrap-markdown/js/bootstrap-markdown');
 
 $.fn.markdown.messages[Vue.lang] = {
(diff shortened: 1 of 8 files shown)


js/components/form/markdown-editor.vue
```diff
@@ -52,10 +52,10 @@
 import $ from 'jquery';
 import Vue from 'vue';
 import {FieldComponentMixin} from 'components/form/base-field';
+import commonmark from 'helpers/commonmark';
 
 const EXCERPT_TOKEN = '<!--- excerpt -->';
 
-window.marked = require('marked');
```
vinyll: this was used for bootstrap-markdown (which uses a 'marked' global object)

js/components/form/markdown-editor.vue
```diff
@@ -89,6 +89,7 @@ export default {
             savable: false,
             resize: 'both',
             iconlibrary: 'fa',
+            onPreview: (e) => commonmark(e.getContent()),
```
vinyll: this substitutes the absence of `window.marked`.

noirbizarre:
r+

PR #257, opened 2015-11-19
updated search tabs alignement for result count, IE bug. fixed #256

### LINKED ISSUE #256: IE display alignement broken for search tabs
IE9, IE10, IE10 and maybe others.
example: https://www.data.gouv.fr/fr/search/?q=cerema
cejbajed

The number should be beside the title, aligned.

vinyll:


davidbgk:
r+

PR #265, opened 2015-11-20
fixing API for Issues whith the false parameter. fix #264

### LINKED ISSUE #264: In the API boolean parameters cannot be "false"
When a parameter is a boolean in the API it never has the value _false_ but always _true_ and can also be _none_ if no default is set.
_Issues_ API is concerned and possibly metrics also.

vinyll:
fix #264

### DIFF
--- a/udata/core/issues/api.py
+++ b/udata/core/issues/api.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 from datetime import datetime
 
 from flask.ext.security import current_user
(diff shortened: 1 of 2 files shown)


davidbgk:
Tests are failing, we probably need to be more explicit both in test and real implementation.

davidbgk:
Same test update as in https://github.com/etalab/udata/pull/263 is probably needed.

noirbizarre:
r+

PR #267, opened 2015-11-24
Less confusing displayed dates for data and metadata, fix #266

### LINKED ISSUE #266: Less confusing displayed dates for data and metadata

davidbgk:
+ retrieve quality field (full dataset) for search results

capture d ecran 2015-11-24 a 09 39 54

### DIFF
--- a/udata/core/dataset/api_fields.py
+++ b/udata/core/dataset/api_fields.py
@@ -160,6 +160,8 @@ dataset_page_fields = api.model('DatasetPage', fields.pager(dataset_fields))
 
 dataset_full_fields = api.inherit('DatasetFull', dataset_fields, {
     'quality': fields.Raw(description='The dataset quality', readonly=True),
+    'last_update': fields.ISODateTime(
+        description='The resources last modification date', required=True),
 })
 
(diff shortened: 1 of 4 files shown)


udata/core/dataset/api_fields.py
```diff
@@ -160,6 +160,8 @@
 
 dataset_full_fields = api.inherit('DatasetFull', dataset_fields, {
     'quality': fields.Raw(description='The dataset quality', readonly=True),
+    'last_update': fields.ISODateTime(
```
noirbizarre: May be interesting to always have this information ?

noirbizarre:
r+wc

udata/core/dataset/api_fields.py
davidbgk: I hesitated but given that `last_update` is a computation which iterate on each resource, I eventually chose to put it in the `full` version. Waiting for fields-on-demand to fix that once and for all ;-).

noirbizarre: OK for me. If we need it before fields on demand we'll move it.

PR #269, opened 2015-11-24
Complete csvs

noirbizarre:
This PR add some missing fields into datasets.csv.

It also provide a reusable helper to fetch nested attributes/keys on any object


udata/utils.py
```diff
@@ -175,3 +175,25 @@ def to_bool(value):
...
+    Get an attribute or a key recursively.
+
+    :param obj: The object to fetch attribute or key on
+    :type obj: object|dict
+    :param key: Either a string in dotted-notation ar an array of string
+    :type key: string|list|tuple
+    '''
+    if not obj or not key:
+        return
+    parts = key.split('.') if isinstance(key, basestring) else key
```
davidbgk: `key.split('.', 1)` is probably more interesting here.

davidbgk: Well, not that much actually.

udata/utils.py
```diff
@@ -175,3 +175,25 @@ def to_bool(value):
...
+    key = parts[0]
```
davidbgk: Using `parts.pop(0)` makes it possible to ease the end of the test.

davidbgk:
r+wc

udata/utils.py
noirbizarre: Yep, I'm splitting only once

udata/utils.py
noirbizarre: OK, I will try that

PR #271, opened 2015-11-26
Ensure CommunityResources have an UUID as primary key

noirbizarre:
This PR ensure community resources have an UUID as primary key and routing on UUID is correct.

### DIFF
--- a/udata/migrations/2015-11-23-set-uuid-as-community-resources-id.js
+++ b/udata/migrations/2015-11-23-set-uuid-as-community-resources-id.js
@@ -1,0 +1,29 @@
+/*
+ * Ensure all community resources have an UUID string as ID
+ */
+
+/**
+ * Generate a rfc4122 version 4 UUID
+ */
+function uuid() {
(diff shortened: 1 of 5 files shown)


udata/migrations/2015-11-23-set-uuid-as-community-resources-id.js
```diff
@@ -0,0 +1,29 @@
+/*
+ * Ensure all community resources have an UUID string as ID
+ */
+
+/**
+ * Generate a rfc4122 version 4 UUID
+ */
+function uuid() {
```
davidbgk: `python -c "import uuid; print(uuid.uuid4())"` :trollface:

davidbgk:
r+

udata/migrations/2015-11-23-set-uuid-as-community-resources-id.js
noirbizarre: This migrations drived me crazy between the lack of nantive uuid support and `replace()` that only replace the first occurence !
I want a python shell for Mongo !!! :trollface:

PR #278, opened 2015-12-04
Validate URLs within the admin, fix #277

### LINKED ISSUE #277: Validate URLs within the admin
As an administrator, I want to be notified when I submit an incorrect URL in order to avoid confusion for users trying to download a resource.

davidbgk:


vinyll:
I'm not familiar enough with _Vue_ to review the concept but it all makes sense to me.

js/components/form/base-form.js
```diff
@@ -136,7 +136,7 @@ export default {
                 }
             },
             errorPlacement: function(error, element) {
-                $(element).closest('.form-group,.field-wrapper').append(error);
+                $(element).closest('.form-group,.field-wrapper').children('div').append(error);
```
noirbizarre: Are you sure ? I think the extra `<div/>` breaks the bootstrap style: http://getbootstrap.com/css/#forms-control-validation

js/components/form/url-field.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+<div>
+    <input type="text" class="form-control" v-el:url-field
```
noirbizarre: Using html 5 `type="url"` should automatically add the validation rule

js/components/form/url-field.vue
```diff
@@ -0,0 +1,29 @@
...
+import {FieldComponentMixin} from 'components/form/base-field';
+
+export default {
+    name: 'url-field',
+    replace: true,
+    mixins: [FieldComponentMixin],
+    events: {
+        'form:ready'() {
+            // Perform all validations on end field because performing on start field unhighlight.
+            $(this.$els.urlField).rules('add', {
```
noirbizarre: not needed if using `type=url"`

noirbizarre:
r+wc

js/components/form/base-form.js
davidbgk: Actually, I'm not creating a div but looking for it? Visually it looks better.

js/components/form/url-field.vue
davidbgk: Oh didn't think about that possibility. Clever.

js/components/form/base-form.js
noirbizarre: OK, does it work with both horizontal and vertical form ? If yes, it's fine to me.

davidbgk: I confirm that it renders correctly:

capture d ecran 2015-12-09 a 16 35 40

capture d ecran 2015-12-09 a 16 35 47

js/components/form/url-field.vue
davidbgk: It works!

PR #283, opened 2015-12-08
displaying a closed discussion and disabled reply

vinyll:


udata/templates/discussion/item.html
```diff
@@ -37,6 +38,7 @@ <h4 class="list-group-item-heading">
         {{ _('Add a comment') }}
     </h4>
 </a>
+{% endif %}
```
davidbgk: IIRC, you might even close the `if` after the next `div`?

davidbgk:
r+wc

udata/templates/discussion/item.html
vinyll: I may misunderstand here but commenting the next `<div>` would remove the link, user avatar and new discussion form.
We only want to remove the _Add a comment_ as the discussion is closed yet the threat is still open.

davidbgk: My bad, you're right!

PR #285, opened 2015-12-08
Introduce territories, fix #284

### LINKED ISSUE #284: Introduce territories
As a citizen, I want to retrieve available informations about my city in order to produce value locally.

davidbgk:
That PR is not ready for a review, it's only a way to document a POC for now.

As a citizen, I type my city name within the search field and now a new tab appears:

capture d ecran 2015-12-08 a 22 39 01

The tab appears also on INSEE code and postal codes:

capture d ecran 2015-12-08 a 22 40 19

Once you click that tab, a dynamic page is displayed with a couple of auto-generated links to related data, each card redirecting to external pages/data:

capture d ecran 2015-12-11 a 15 34 48

New iteration with typeahead + inclusion of related existing datasets within the displayed list, still completely auto-generated:

territoires


udata/core/search/views.py
```diff
@@ -35,5 +67,7 @@ def render_search():
         ]
         results_labels = ['datasets', 'reuses', 'organizations', 'users']
     results = search.multiquery(*search_queries)
-    return theme.render('search.html',
-                        **dict(zip(results_labels, results)))
+    context = dict(zip(results_labels, results))
+    if current_app.config['ACTIVATE_TERRITORIES']:
```
vinyll: `current_app.get('ACTIVATE_TERRITORIES')` in case the key would not be defined?

davidbgk:
capture d ecran 2015-12-17 a 15 31 19

davidbgk:
capture d ecran 2015-12-17 a 15 33 33

davidbgk:
capture d ecran 2015-12-17 a 15 35 22

davidbgk:
marseille

davidbgk:
capture d ecran 2015-12-17 a 15 49 40

davidbgk:
The territories’ datasets are now displayed as embeds, dog-fooding our own stuff, so I improved a bit the rendering using emojis as a test. Maybe foundations for a new display of datasets’ cards?

Integration within an external page:

capture d ecran 2015-12-22 a 11 57 07

Integration on the territory page:

capture d ecran 2015-12-22 a 11 57 38

udata/api/__init__.py
```diff
@@ -304,6 +305,22 @@ def get(self):
...
+                and current_app.config.get('ACTIVATE_TERRITORIES')):
+            from udata.models import TERRITORY_DATASETS
+            try:
+                country, town, code, kind = item_id.split('-')
+            except ValueError:
+                return api.abort(400, 'Invalid URL.')
+            try:
+                geozone = GeoZone.objects.get(code=code)
+            except GeoZone.DoesNotExist:
+                return api.abort(400, 'Incorrect ID.')
```
noirbizarre: More explicit with: "Unknown territory/zone identifier". (choose between:))

udata/api/__init__.py
```diff
@@ -304,6 +305,22 @@ def get(self):
...
+            if kind in TERRITORY_DATASETS:
+                item = TERRITORY_DATASETS[kind](geozone)
+            else:
+                return api.abort(400, 'Incorrect kind of territory.')
```
noirbizarre: s/Incorrect/Unknown/

udata/core/spatial/models.py
```diff
@@ -53,6 +58,24 @@ def __unicode__(self):
     def __html__(self):
         return gettext(self.name) + ' <i>(' + self.code + ')</i>'
 
+    def logo_url(self, external=False):
+        path = ((self.logo.filename
+                 and self.logo.fs.exists(self.logo.filename)
+                 and self.logo())
+                or '')
+        if path:
+            # Any attempt to put the `external=True` parameter within
```
noirbizarre: My bad, I need to publish the repository.
You can have an external URL with `self.logo.best_url(external=True)` or with `self.logo.fs.url(self.logo.filename, external=True)`

udata/core/spatial/models.py
```diff
@@ -53,6 +58,24 @@ def __unicode__(self):
     def __html__(self):
         return gettext(self.name) + ' <i>(' + self.code + ')</i>'
 
+    def logo_url(self, external=False):
+        path = ((self.logo.filename
```
noirbizarre: I think you can write: `self.logo.filename if self.logo else ''`

udata/core/spatial/models.py
```diff
@@ -64,6 +87,17 @@ def keys_values(self):
                 keys_values.append(value)
         return keys_values
 
+    @property
+    def url(self):
+        return url_for('gouvfr.territory', territory=self)
+
+    @property
+    def external_url(self):
+        return '{scheme}://{domain}{path}'.format(
```
noirbizarre: `return url_for('gouvfr.territory', territory=self, _external=True)`
The domain and scheme will be automatically set.

udata/frontend/helpers.py
```diff
@@ -28,7 +28,10 @@ def static_global(filename, _burst=True, **kwargs):
     if current_app.config['DEBUG']:
         burst = time()
     else:
-        burst = pkg_resources.get_distribution('udata').version
+        try:
+            burst = pkg_resources.get_distribution('udata').version
+        except pkg_resources.DistributionNotFound:
+            burst = time()  # Useful for gouvfr tests?
```
noirbizarre: It depends on how you set your development environment.
If you install udata with `pip install -e` or `python setup.py`, distribution will be set.
The main problem is how to use a different cache burst for theme assets.

udata/models/__init__.py
```diff
@@ -180,8 +180,10 @@ def init_app(app):
...
         try:
-            __import__(name)
-        except ImportError:
-            pass
-        except Exception as e:
+            module_obj = importlib.import_module(name)
+            for item in getattr(module_obj, '__all__'):
+                globals()[item] = getattr(module_obj, item)
+        except ImportError as e:
             log.error('Error importing %s: %s', name, e)
```
noirbizarre: Maybe in debug only because not all modules exports models and in this case the ImportError is expected.

udata/models/__init__.py
```diff
@@ -180,8 +180,10 @@ def init_app(app):
...
         name = 'udata.ext.{0}.models'.format(plugin)
         try:
-            __import__(name)
-        except ImportError:
-            pass
-        except Exception as e:
+            module_obj = importlib.import_module(name)
+            for item in getattr(module_obj, '__all__'):
+                globals()[item] = getattr(module_obj, item)
+        except ImportError as e:
```
noirbizarre: Maybe in debug only because not all modules exports models and in this case the ImportError is expected and we dont want to spam the logs.

udata/models/__init__.py
```diff
@@ -180,8 +180,10 @@ def init_app(app):
     for plugin in app.config['PLUGINS']:
         name = 'udata.ext.{0}.models'.format(plugin)
         try:
-            __import__(name)
-        except ImportError:
-            pass
-        except Exception as e:
+            module_obj = importlib.import_module(name)
+            for item in getattr(module_obj, '__all__'):
+                globals()[item] = getattr(module_obj, item)
```
noirbizarre: Do we really want to make extensions models globals ?

udata/settings.py
```diff
@@ -128,7 +130,7 @@ class Testing(object):
     DEBUG_TOOLBAR = False
     SERVER_NAME = 'localhost'
     DEFAULT_LANGUAGE = 'en'
-
+    ACTIVATE_TERRITORIES = True
```
noirbizarre: Due to the overhead and to the already long running test, we might activate territories only on necessary test cases

udata/templates/embed-dataset.html
```diff
@@ -1,81 +1,149 @@
 <style scoped>
+    {# Following http://rscss.io/ conventions. #}
```
noirbizarre: :+1:

udata/templates/embed-dataset.html
```diff
@@ -1,81 +1,149 @@
...
+        </p>
+        <p class="description">
+            <a href="{{ item.external_url or item.url }}" title="{{ item.title }}">
+                {{ item.description | mdstrip(110) }}
+            </a>
+        </p>
+    </div>
+    <footer class="meta">
+        <p>
+            <a href="">🔃 {{ _('Embed') }}</a>
```
noirbizarre: I love unicode !!

udata/core/spatial/models.py
```diff
@@ -64,6 +87,17 @@ def keys_values(self):
                 keys_values.append(value)
         return keys_values
 
+    @property
+    def url(self):
+        return url_for('gouvfr.territory', territory=self)
```
noirbizarre: udata can't reference gouvfr

udata/core/spatial/models.py
```diff
@@ -53,6 +58,24 @@ def __unicode__(self):
     def __html__(self):
         return gettext(self.name) + ' <i>(' + self.code + ')</i>'
```
vinyll: `return '%s <i>(%s)</i>' % (gettext(self.name), self.code)`

udata/frontend/markdown.py
```diff
@@ -40,7 +40,7 @@ def __call__(self, stream):
             styles=current_app.config['MD_ALLOWED_STYLES'],
             strip_comments=False)
         # Turn markdown to HTML.
-        ast = self.parser.parse(stream)
+        ast = self.parser().parse(stream)
```
vinyll: Should we not keep this as a property?
If not maybe changing the line below from `renderer` to `renderer()` would make it uniform.

udata/templates/embed-dataset.html
```diff
@@ -1,81 +1,149 @@
...
-        max-width: {{ width }}px;
-        max-height: {{ height }}px;
         min-width: 320px;
-        min-height: 90px;
-        padding: 5px;
+        min-height: 160px;
         clear: both;
         margin: 1em;
         background: #FFF;
+        display: flex;
```
vinyll: IE11 has knows issues when using flexbox with min-height. Not sure if they'd appear here.
http://caniuse.com/#search=flex

udata/templates/embed-dataset.html
```diff
@@ -1,81 +1,149 @@
...
+        </p>
+    </div>
+    <footer class="meta">
+        <p>
+            <a href="">🔃 {{ _('Embed') }}</a>
+        </p>
+        {% if item.temporal_coverage %}
+        <p>
+            {% if item.territory %}
+                📅 {{ item.temporal_coverage.start }}
```
vinyll: are

udata/templates/embed-dataset.html
```diff
@@ -1,81 +1,149 @@
...
+                {% if item.temporal_coverage.end %}
+                    {{ _('to')}} {{ item.temporal_coverage.end }}
+                {% endif %}
+            {% else %}
+                📅 {{ item.temporal_coverage|daterange }}
```
vinyll: you

udata/templates/embed-dataset.html
```diff
@@ -1,81 +1,149 @@
...
+            {% endif %}
+        </p>
+        {% endif %}
+        {% if item.spatial and item.spatial.granularity and item.spatial.granularity_label != 'Other' %}
+        <p>📍
```
vinyll: serious?

udata/templates/embed-dataset.html
```diff
@@ -1,81 +1,149 @@
...
+        </p>
+        {% elif item.territory %}
+        <p>
+            <a href="{{ item.territory.external_url }}">
+                📍 {{ item.territory.name }}
+            </a>
+        </p>
+        {% endif %}
+        {% if item.frequency and item.frequency_label != 'Unknown' %}
+        <p>📊
```
vinyll: <3

udata/tests/api/test_oembed_api.py
```diff
@@ -31,7 +31,7 @@ def test_oembed_dataset_api_get(self):
         self.assertIn(dataset.title, data['html'])
         self.assertIn(dataset.external_url, data['html'])
         self.assertIn('placeholders/default.png', data['html'])
-        self.assertIn(dataset.description, data['html'])
+        self.assertIn(dataset.description[:90], data['html'])
```
vinyll: comment here to explain why limiting to first 90

udata/core/spatial/models.py
davidbgk: `self.logo.fs.url(self.logo.filename, external=external)` does return an external URL but `self.logo.best_url(external=external)` doesn't, maybe a bug?

udata/core/spatial/models.py
davidbgk: It doesn't work because my use-case is a bit special: sometimes the path is filed within the DB but the file cannot be download so I need to check the existence directly against the file system…

udata/frontend/markdown.py
davidbgk: It looks like we can keep a single instance of the `renderer` so I would keep it for performance matters.
The new instance of the parser is required in order to deal with thread-safe issues we have when multiple commonmark conversions are issued at the same time. Maybe we should declare the bug upstream but it's hard to reproduce/provide a test for this.

udata/models/__init__.py
davidbgk: That silent error cost me one afternoon, I really think we should keep it, logs are cheap :)

udata/settings.py
davidbgk: Alright.

udata/templates/embed-dataset.html
davidbgk: Well, about that unicode thing. It was mostly a test and I kept it as a demo but I'm not sure we should keep those.

The issue is to have some pictos without requiring to download those from our server. My query budget to render an embed is as low as possible so no font awesome or similar. Maybe base64 embedded icons?

udata/core/spatial/models.py
davidbgk: Not really part of the PR :p

udata/models/__init__.py
davidbgk: That's the only way I found to be able to import models from gouvfr via `udata.models` but I'm open to better solutions because it's ugly. To test by yourself: retrieve both udata and udata-gouvfr branches + generate and import new geozones, mmh I should write that somewhere else.

udata/core/spatial/models.py
```diff
@@ -51,7 +56,15 @@ def __unicode__(self):
     __str__ = __unicode__
 
     def __html__(self):
-        return gettext(self.name) + ' <i>(' + self.code + ')</i>'
+        return '{name} <i>({code})</i>'.format(
+            name=gettext(self.name), code=self.code)
+
+    def logo_url(self, external=False):
```
davidbgk: @noirbizarre what do you think of the new version?

udata/core/spatial/models.py
davidbgk: That's where it's starting to be tricky. These properties are only called when the gouvfr extension is activated…

I'm open to better suggestions :/

vinyll: Can't we think of something like `return url_for('%s.territory' % theme, territory=self)` where theme would come from the config for example?

davidbgk: or maybe the whole `spatial` app should move to gouvfr?

js/widgets/site-search.js
```diff
@@ -6,9 +6,9 @@ define([
     'dataset/typeahead',
     'reuse/typeahead',
     'organization/typeahead',
-    // 'user/typeahead',
+    'territory/typeahead',
```
davidbgk: Here we're hitting the server for territories’ suggestions even if the feature is deactivated.

less/site.less
```diff
@@ -9,6 +9,7 @@
 @import "udata/oauth";
 @import "udata/activity";
 @import "udata/resource";
+@import "udata/territories";
```
davidbgk: Here we're loading the CSS for territories even if the feature is deactivated.

udata/api/__init__.py
```diff
@@ -281,49 +278,6 @@ class API(Resource):  # Avoid name collision as resource is a core model
     pass
 
 
-oembed_parser = api.parser()
```
davidbgk: Splitting to its own `features` app.

js/widgets/site-search.js
noirbizarre: We may expose the feature availability in the raw.html page header like:
```html+jinja
<meta name="territory-enabled" content="{{ config.ACTIVATE_TERRITORIES }}">
```
and use it in config.js
```javascript
export default {
    territory_enabled: $('meta[name=territory-enabled]').attr('content').toLowerCase() === 'true'
};
```

less/site.less
noirbizarre: Given the size of the added style, it's not bad

udata/core/spatial/models.py
noirbizarre: Fine to me.
I may ad this to flask-fs before publication.

I'm wondering: do we use a placeholder for logo ?

udata/features/territories/api.py
```diff
@@ -0,0 +1,31 @@
...
+    def get(self):
+        args = suggest_parser.parse_args()
+        territory = check_for_territory(args['q'].decode('utf-8'))
+        if territory:
+            return [
+                {
+                    'id': territory.id,
+                    'title': territory.name,
+                    'image_url': territory.logo_url() or '',
+                    'page': url_for('territories.territory',
```
noirbizarre: `territory.external_url` ? :)

udata/features/territories/api.py
```diff
@@ -0,0 +1,31 @@
...
+    @api.doc(id='suggest_datasets', parser=suggest_parser)
+    def get(self):
+        args = suggest_parser.parse_args()
+        territory = check_for_territory(args['q'].decode('utf-8'))
+        if territory:
+            return [
+                {
+                    'id': territory.id,
+                    'title': territory.name,
+                    'image_url': territory.logo_url() or '',
```
noirbizarre: `territory.logo_url(external=True)` ?

udata/features/territories/commands.py
```diff
@@ -0,0 +1,68 @@
...
+    'territories',
+    help='Territories specifics operations',
+    description='Handle all territories related operations and maintenance'
+)
+
+
+@m.command
+def fetch_territories_logos():
+    """Retrieves images of flags or blazons from wikimedia."""
+    log.info('As of January 2016, the download is about 1GB. Time to relax.')
```
noirbizarre: :smile:

udata/features/territories/api.py
davidbgk: Why external? This is the API in use for the typeahead in the local search box.

udata/core/spatial/models.py
davidbgk: I use the default placeholder (glasses) but a dedicated one would be better. Any suggestion?

js/widgets/site-search.js
davidbgk: Alright, I'll try that.

udata/features/territories/models.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @property
+    def server(self):
+        return '{scheme}://{domain}'.format(
+            scheme=current_app.config['USE_SSL'] and 'https' or 'http',
+            domain=current_app.config['SERVER_NAME'])
+
+    @property
+    def api_url(self):
+        return '{server}/api/1/territories/{path}/'.format(
```
noirbizarre: `url_for` instead ?

davidbgk: There is no dedicated endpoint for that URL, maybe I should add one just to have the `url_for`?

udata/models/__init__.py
noirbizarre: OK, but I in tis case it should be leveled `warning` instead of `error`.

davidbgk: Fair enough.

udata/core/spatial/models.py
noirbizarre: A blasoon with question mark ?

udata/features/territories/api.py
noirbizarre: All URLs in the API are absolutes (or at least, they should be) allowing external usages

udata/features/territories/models.py
noirbizarre: Yes, I think it's cleaner. Relying on flask mecanism allow more flexibility

js/widgets/site-search.js
davidbgk: Done.

davidbgk:
Test with glyphicons and default blazon:

capture d ecran 2016-01-14 a 17 39 26

capture d ecran 2016-01-14 a 17 39 41

js/config.js
```diff
@@ -8,5 +8,6 @@ export default {
     lang: $('html').attr('lang') || 'en',
     debug: DEBUG,
     collapse_width: 992,
-    theme_static: $('link[rel=theme-static]').attr('href')
+    theme_static: $('link[rel=theme-static]').attr('href'),
+    territory_enabled: $('meta[name=territory-enabled]').attr('content') ? $('meta[name=territory-enabled]').attr('content').toLowerCase() === 'true' : false
```
noirbizarre: Will need to be migrated to vanilla after merge from master to dev

js/widgets.js
```diff
@@ -1,27 +1,97 @@
...
+    if (response.status >= 200 && response.status < 300) {
+        return response;
+    } else {
+        const error = new Error(response.statusText);
+        error.response = response;
+        throw error;
+    }
 }
+function zip(arrays) {
+    // Equivalent to `zip()` in Python.
```
noirbizarre: Missing the batteries included ? ;)

davidbgk: :trollface:

js/widgets.js
```diff
@@ -1,27 +1,97 @@
...
+        if (el.hasAttribute('data-udata-territory-id')) {
+            const territoryId = el.getAttribute('data-udata-territory-id');
+            return `territory-${territoryId}`;
+        } else if (el.hasAttribute('data-udata-dataset-id')) {
+            const datasetId = el.getAttribute('data-udata-dataset-id');
+            return `dataset-${datasetId}`;
+        }
+    });
+    const url = `${baseURL}/api/1/oembeds/?references=${references}`;
     fetch(url)
```
noirbizarre: OMG this is so good. Well translated by babel ?

js/widgets/site-search.js
```diff
@@ -1,34 +1,36 @@
...
+        highlight: true,
+        classNames: {
+            menu: 'tt-dropdown-menu'
+        }
+    };
 
     // Expandable main search bar
     $('#main-search')
-        .focusin(function() {
+        .focusin(() => {
```
noirbizarre: Can't do this in this case: jQuery set the this/scope to the select object. `$(this)` is incompatible with arrow functions.

js/widgets/site-search.js
```diff
@@ -1,34 +1,36 @@
...
             $('.collapse-on-search').addClass('hide');
             $(this).closest('.navbar-nav')
                 .removeClass(SEARCH_UNFOCUS_CLASS)
                 .addClass(SEARCH_FOCUS_CLASS);
         })
-        .focusout(function() {
+        .focusout(() => {
```
noirbizarre: Same as above

js/widgets.js
davidbgk: Babel is keeping the native but quite new `fetch` as-is. We might want to use the [polyfill by github](https://github.com/github/fetch) but it adds some weight. Or find a dedicated plugin for babel?

udata/features/oembed/api.py
```diff
@@ -0,0 +1,80 @@
...
+            result.append({
+                'type': 'rich',
+                'version': '1.0',
+                'html': html,
+                'width': width,
+                'height': height,
+                'maxwidth': maxwidth,
+                'maxheight': maxheight,
+            })
+        return output_json(result, 200)
```
noirbizarre: Might be more coherent with other APIs with a `@api.marshal_with` decorator and simply:
```python
return result
```

udata/features/territories/views.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    args = multi_to_dict(request.args)
+    if 'territory_attr' not in args or 'csv_column' not in args:
+        return abort(404)
+    for resource in dataset.resources:
+        if resource.id == resource_id:
+            break
+
+    resource_path = references.path(resource.url.split('/')[-1])
+    match = getattr(territory, args['territory_attr']).encode('utf-8')
```
noirbizarre: Will raise 500 if `args['territory_attr']` is not found in territory.

udata/features/territories/views.py
```diff
@@ -0,0 +1,74 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+import csv
```
noirbizarre: You might not have to deal with encoding issues if using `unicodecsv`, already a dependency.

udata/templates/raw.html
```diff
@@ -40,6 +40,7 @@
         data-roles="{{ current_user.roles|join(',') }}"
     />
     {% endif %}
+    <meta name="territory-enabled" content="{{ config.ACTIVATE_TERRITORIES }}">
```
noirbizarre: Will have to move into `metadata.html` after merge from master in dev.

udata/templates/territories/territory.html
```diff
@@ -0,0 +1,69 @@
...
+            {{ _('Are you in charge of that city? You can ask for a conversion of that territory into an organization in order to add new datasets, control publications and interact with the community.') }}
+        </p>
+        <button type="button" rel="tooltip" class="btn btn-primary btn-block btn-sm icon-left membership"
+            data-placement="left"
+            title="{{ _('I belong to this city') }}">
+            <span class="glyphicon glyphicon-user"></span>
+            I’m legitimate<br/>
+            to rule this city
+        </button>
+        -->
```
noirbizarre: Why keep it as comment ?

js/widgets.js
noirbizarre: We might need, compatibility is too short: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#Browser_compatibility

davidbgk: Updated with the polyfill.

js/widgets/site-search.js
davidbgk: Oh. Sad.

udata/features/territories/views.py
davidbgk: Except that we should return a 404 a couple of lines above in that case ;)

udata/features/oembed/api.py
davidbgk: AFAIK, `marshal_with` needs a list of fields but we just want to return the JSON here not based on any model. Am I missing something?

js/config.js
```diff
@@ -114,6 +114,12 @@ if (sentryEl) {
  */
 export const notify_in = _meta('notify-in');
 
+/**
+ * Whether territories are enabled or not.
+ */
+export const is_territory_enabled = JSON.parse(_meta('territory-enabled'));;
```
noirbizarre: ;)

js/territory/typeahead.js
```diff
@@ -0,0 +1,43 @@
...
+    var engine = new Bloodhound({
+            queryTokenizer: Bloodhound.tokenizers.whitespace,
+            datumTokenizer: function(d) {
+                return Bloodhound.tokenizers.whitespace(d.name);
+            },
+            identify: function(d) {
+                return d.id;
+            },
+            remote: {
+                url: API.build_url('/territory/suggest/') + '?q=%QUERY&size=',
```
noirbizarre: I think the size parameter value is missing.

js/territory/typeahead.js
```diff
@@ -0,0 +1,43 @@
...
+                wildcard: '%QUERY',
+                // Keep until model is uniformised
+                transform: function(response) {
+                    return $.map(response, function(row, idx) {
```
noirbizarre: If response is a list, you can write `response.map(function(row) {})`

js/widgets.js
```diff
@@ -1,27 +1,97 @@
...
+    }
 }
+function zip(arrays) {
+    // Equivalent to `zip()` in Python.
+    return arrays[0].map((_, i) => {
+        return arrays.map((array) => array[i]);
+    });
+}
+function buildIntegrationFragment(reference) {
+    const [kind, ...ids] = reference.split('-');
```
noirbizarre: wow, nice !

js/territory/typeahead.js
davidbgk: Actually, I'm only returning one result anyway so maybe I should just remove the parameter? I initially let it to be consistent with other calls.

udata/features/oembed/api.py
noirbizarre: `marshal_with` also document. But it's not essential here, you are right.

udata/features/territories/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+            return GeoZone.objects.get(
+                name__istartswith=query, level='fr/town')
+        except GeoZone.MultipleObjectsReturned:
+            try:
+                return GeoZone.objects.get(name__iexact=query, level='fr/town')
+            except GeoZone.MultipleObjectsReturned:
+                # Finally, we fall back on the most popular one.
+                # Warning: it means that we're dropping some territories
+                # here, e.g.: Vitrolles.
+                return (GeoZone.objects
```
noirbizarre: If you want to avoid multiline and parenthesis, you can write:
```python
qs = GeoZone.objects.filter(name__iexact=query, level='fr/town')
qs = qs.order_by('-population', '-area')
return qs.first()
```

udata/features/territories/commands.py
```diff
@@ -0,0 +1,68 @@
...
+    # A bit tricky but otherwise you cannot guess the final file URL.
+    DBPEDIA_MEDIA_URL = 'http://commons.wikimedia.org/wiki/Special:FilePath/'
+    LOGOS_FOLDER_PATH = logos.root
+    if not os.path.exists(LOGOS_FOLDER_PATH):
+        os.makedirs(LOGOS_FOLDER_PATH)
+
+    for geozone in (GeoZone.objects.filter(level='fr/town')
```
noirbizarre: Same here:
```python
qs = GeoZone.objects.filter(level='fr/town')
for geozone in qs.only('flag', 'blazon'):
```

udata/features/territories/views.py
noirbizarre: Only if absent from the query, not from territory. If someone query `?territory_attr=wtf` it will raise a 500. It should raise a 400.

davidbgk: Fair enough :)

udata/tests/api/test_oembed_api.py
```diff
@@ -9,18 +9,18 @@
...
-    def test_oembed_dataset_api_get(self):
+    def test_oembeds_dataset_api_get(self):
         '''It should fetch a dataset in the oembed format.'''
         with self.autoindex():
             dataset = DatasetFactory()
 
-        dataset_url = url_for(
-            'api.dataset', dataset=dataset.id, _external=True)
-        response = self.get(url_for('api.oembed', **{'url': dataset_url}))
+        response = self.get(
```
noirbizarre: Maybe split on 2 lines '`url=...` + `self.get(url)`

udata/tests/api/test_oembed_api.py
```diff
@@ -34,66 +34,52 @@ def test_oembed_dataset_api_get(self):
...
+    def test_oembeds_dataset_api_get_with_organization(self):
         '''It should fetch a dataset in the oembed format with org.'''
         with self.autoindex():
             organization = OrganizationFactory()
             dataset = DatasetFactory(organization=organization)
 
-        dataset_url = url_for(
-            'api.dataset', dataset=dataset.id, _external=True)
-        response = self.get(url_for('api.oembed', **{'url': dataset_url}))
+        response = self.get(
```
noirbizarre: same as above

udata/tests/api/test_oembed_api.py
```diff
@@ -34,66 +34,52 @@ def test_oembed_dataset_api_get(self):
...
-    def test_oembed_dataset_api_get_without_good_item(self):
-        '''It should fail at fetching an oembed without a wrong item.'''
+    def test_oembeds_dataset_api_get_without_good_item(self):
+        '''It should fail at fetching an oembed with a wrong item.'''
         with self.autoindex():
             user = UserFactory()
 
-        user_url = url_for('api.user', user=user.id, _external=True)
-        response = self.get(url_for('api.oembed', **{'url': user_url}))
+        response = self.get(
```
noirbizarre: same

Issue #288, opened 2015-12-11
datasets.csv does not download all datasets


vinyll:
When downloading from http://www.data.gouv.fr/en/datasets.csv we expect +21400 datasets and only ~5400 are downloaded.
Here's a local stacktrace of that same call:

```
127.0.0.1 - - [09/Dec/2015 18:56:22] "GET /api/1/notifications/?_=1449669797181 HTTP/1.1" 200 -
Debugging middleware caught exception in streamed response at a point where response headers were already sent.
Traceback (most recent call last):
  File "/home/me/.virtualenvs/udata/lib/python2.7/site-packages/werkzeug/wsgi.py", line 693, in __next__
    return self._next()
  File "/home/me/.virtualenvs/udata/lib/python2.7/site-packages/werkzeug/wrappers.py", line 81, in _iter_encoded
    for item in iterable:
  File "/home/me/.virtualenvs/udata/lib/python2.7/site-packages/flask/helpers.py", line 122, in generator
    for item in gen:
  File "/home/me/Projects/datagouv/udata/udata/frontend/csv.py", line 180, in yield_rows
    for row in adapter.rows():
  File "/home/me/Projects/datagouv/udata/udata/frontend/csv.py", line 76, in <genexpr>
    return (self.to_row(o) for o in self.queryset)
  File "/home/me/Projects/datagouv/udata/udata/frontend/csv.py", line 82, in to_row
    row.append(safestr(getter(obj)))
  File "/home/me/Projects/datagouv/udata/udata/core/dataset/csv.py", line 11, in serialize_spatial_zones
    return ','.join(z.name for z in dataset.spatial.zones)
  File "/home/me/Projects/datagouv/udata/udata/core/dataset/csv.py", line 11, in <genexpr>
    return ','.join(z.name for z in dataset.spatial.zones)
  File "/home/me/.virtualenvs/udata/lib/python2.7/site-packages/bson/dbref.py", line 85, in __getattr__
    raise AttributeError(key)
AttributeError: name
```

Also we should probably skip when an error occurs to retrieve the maximum number of datasets.

PR #289, opened 2015-12-11
fixed dataset csv that crashes after ~4500 datasets. #288

vinyll:


davidbgk:
Might be worth a test?

noirbizarre:
+1 for the test ;)

vinyll:
The model GeoZone has changed to set_name_ as _required_. Therefore we cannot reproduce the same context where a GeoZone has no name.
Should we run a migration and export the _id_ for the default _name_?

vinyll:
Seen with @davidbgk:
The GeoZone should not have empty _name_. The GeoZone json file dump have all _features_ with a _name_ property.
The `load()` task does not drop the collection (see https://github.com/etalab/udata/blob/master/udata/core/spatial/commands.py#L29).
This may result into older zones that do not have any _name_ field now.

noirbizarre:
OK for me, this is fixed by the new GeoZones enforced specifications. :)

vinyll:
indeed it is. Therefore I wonder if we still need this as it is due to old existing data in the data.

vinyll:
We should reset the geozones in prod to check if we still have the issue.
If not we should ignore this PR.

davidbgk:
Let's reset for the territory deployment, we'll need it anyway :)

udata/tests/dataset/test_csv.py
```diff
@@ -0,0 +1,19 @@
...
+class DatasetModelTest(TestCase, DBTestMixin):
+    def test_owned_by_user(self):
+        user = UserFactory()
+        datasets = [
+            DatasetFactory(owner=user, spatial=SpatialFactory(
+                zones=[GeoZoneFactory(name='')])),
+            DatasetFactory(owner=user, spatial=SpatialFactory()),
+        ]
+
+        self.assertEqual(len(result), 1)
```
davidbgk: Unfinished test?

vinyll:
we may leave this unmerged and delete. GeoZone have changed and this bug should not happen with the new GeoZone model.
Clearing out outdated geozone data and rebuilding should fix.

vinyll:
After discussing we'll drop this resolution and go for skipping a line that fails, whatever the error.

davidbgk:
Should be fixed by https://github.com/etalab/udata/pull/363

vinyll:
Yup. I close this

PR #292, opened 2015-12-16
fixed dataset with empty spatial coverage

vinyll:
Before:
empty spatial coverage

After: the empty block is removed


davidbgk:
r+

PR #297, opened 2016-01-07
Link directly to the discussion within emails, fix #296

### LINKED ISSUE #296: Link directly to the discussion within notification emails
As a user, I want to be directed directly to the right portion of the page when I click in a notification email in order to see the discussion.

davidbgk:


udata/core/dataset/models.py
```diff
@@ -235,13 +235,14 @@ def post_save(cls, sender, document, **kwargs):
...
-        return url_for('datasets.show', dataset=self)
+    def url_for(self, *args, **kwargs):
+        return url_for('datasets.show', dataset=self, *args, **kwargs)
+
+    display_url = property(url_for)
 
     @property
     def external_url(self):
-        return url_for('datasets.show', dataset=self, _external=True)
+        return self.url_for(**{'_external': True})
```
vinyll: Why not `return self.url_for(_external=True)`?

noirbizarre:
r+

Issue #293, opened 2015-12-16
Normalize tags


vinyll:
These steps are suggested to have a uniform tag based on sluggification:

- migrate data for existing tags
- create a slugifier in the tag module (core/tag)
- integrate the normalization in the model level
- write documentation in the FAQ explaining about the tag conversion

vinyll:
https://github.com/etalab/udata/commit/b2ec0920deec7cb8bc7641a8f6db78d765ca74ab

PR #298, opened 2016-01-12
updating dataset.tags and resuse.tags to be slugified. #293

vinyll:
Not sure where it is actually used and should therefore be also updated.

### DIFF
--- a/udata/migrations/2016-01-12-add-tag-slug.js
+++ b/udata/migrations/2016-01-12-add-tag-slug.js
@@ -1,0 +1,53 @@
+/*
+ * Add the slug property to Tags with a default value based on the name
+ * property.
+ */
+
+// Taken from underscore.string
+function slugify(str) {
+  var from = 'ąàáäâãåæăćęèéëêìíïîłńòóöôõøśșțùúüûñçżź';
+  var to = 'aaaaaaaaaceeeeeiiiilnoooooosstuuuunczz';
+  var converted = '[%s]'.replace('%s',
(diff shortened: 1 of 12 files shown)


udata/migrations/2016-01-12-add-tag-slug.js
```diff
@@ -0,0 +1,49 @@
+/*
+ * Add the slug property to Tags with a default value based on the name
+ * property.
+ */
+
+ // Taken from underscore.string
+function defaultToWhiteSpace(characters) {
```
davidbgk: Maybe we should move these to `migrations/utils.js`?

udata/migrations/2016-01-12-add-tag-slug.js
```diff
@@ -0,0 +1,49 @@
...
+  str = str.toLowerCase();
+  return str;
+}
+
+var tags = db.tag.find();
+var count = 0;
+
+tags.forEach(function(tag) {
+  tag.slug = slugify(tag.name);
+  db.tag.save(tag);
```
davidbgk: I wonder if re-saving each tag in Python once the new field is added with both `populate_from` and `required` is less prone to bugs.

vinyll:
Dataset and Reuse are the 2 places actually using tags.
They come _dirty_ and are cleaned within the `TagListField`.
The migration is preserved in JS as we should update all _Dataset_ and _Reuse_ instances which would corrupt their datetime.

davidbgk:
r+

PR #302, opened 2016-01-21
Add migrations for consistency between (uu)ids, fix #301

### LINKED ISSUE #301: Fix the duality UUID/Id in resources that lead to bugs for reordering

davidbgk:


udata/migrations/2016-01-21-set-uuid-for-resources.js
```diff
@@ -0,0 +1,25 @@
...
+var count = 0;
+
+datasets.map(function(dataset) {
+    if (dataset.resources) {
+        var resources = dataset.resources.map(function(resource) {
+            if (!resource._id) {
+                resource._id = resource.id;
+                count++;
+            }
+            delete resource.id;
```
noirbizarre: I'm not sure it will work with resources. I think mongoengine expect `id` for embedded resources.

davidbgk: @noirbizarre well, it looks to work. How would you test it to verify that it works?

PR #303, opened 2016-01-25
Webpack build fixes

noirbizarre:
Some babel6 related fixes
Downgrade to Handlebars 3.X to maintain SwaggerUI compatibility
Upgrade Highlight.js and i18next.
Pin vue-router et swagger-client


davidbgk:
r+

PR #305, opened 2016-01-25
Pin map by default (avoid accidental scroll catchs)

noirbizarre:
This PR fixes the coverage map by default avoiding accidental scroll catches.


davidbgk:
My hero :heart: r+

noirbizarre:
;)

PR #306, opened 2016-01-25
Simplify admin search

noirbizarre:
This pull-request make the admin search more readable.

Search with results before:
Search with result before

After:
Search with result after

Search without any results before:
Search with result before

After:
Search with result after

### DIFF
--- a/js/views/search.vue
+++ b/js/views/search.vue
@@ -1,24 +1,19 @@
 <template>
-<layout :title="_('Search')">
+<layout :title="title">
     <div class="row">
-        <datasets class="col-xs-12" :datasets="datasets" :title="datasetsTitle">
-        </datasets>
+        <datasets class="col-xs-12" :datasets="datasets"></datasets>
     </div>
     <div class="row">
-        <communities class="col-xs-12" :communities="communities" :title="communitiesTitle">
-        </communities>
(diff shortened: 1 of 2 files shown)


js/views/search.vue
```diff
@@ -37,29 +37,38 @@ export default {
         discussions: require('components/discussions/list.vue'),
         Layout
     },
+    computed: {
+        title() {
+            return [
+                this._('Search in your data'),
+                this.$route.query.q
+            ].join(': ');
```
davidbgk: You need a translation here for the `:` (takes a non-breaking space before in French).

BTW, I don't why you join and not use a simple concatenation or maybe a template string?

js/views/search.vue
```diff
@@ -37,29 +37,38 @@ export default {
...
+        },
+        no_results() {
+            const collections = [this.datasets, this.communities, this.reuses, this.issues, this.discussions];
+            return !collections.some(function(collection) {
```
davidbgk: Batteries included you said? :trollface:

davidbgk:
Way better, thanks. r+

js/views/search.vue
noirbizarre: This one is great. Equivalent to python `any()`

PR #308, opened 2016-01-26
Avoid indexation for preprod envs

davidbgk:


udata/templates/raw.html
```diff
@@ -3,8 +3,8 @@
   <head>
     {% block raw_head %}
     {% include theme('metadata.html') with context %}
+    {% if config.DISALLOW_INDEXING %}<meta name="robots" content="noindex,nofollow">{% endif %}
```
noirbizarre: Would be event better into metadata.html

noirbizarre:
r+

PR #309, opened 2016-01-26
Docker composer volumes path to relative

vinyll:
`docker-compose up` raises these errors on OSX:

    ERROR: data/search/plugins includes invalid characters for a local volume name, only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed
    ...
    ERROR: data/db includes invalid characters for a local volume name, only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ db:
     image: mongo:2.6
     command: mongod
     volumes:
-      - data/db:/data/db
+      - ./data/db:/data/db
     expose:
       - "27017"
     ports:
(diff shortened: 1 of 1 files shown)


noirbizarre:
I think there was an error on the PR. There is a lot of unrelated things.
For the docker-compose.yml, the fix is fine to me, the issue was introduced in latest versions of docker dompose.

vinyll:
Fixed the PR unrelated files changes.

noirbizarre:
r+ :)

PR #310, opened 2016-01-27
Fix the missing JSONeditor dependency from the last swagger ui update

noirbizarre:
This fix the admin doc view which is broken due to a new SwaggerUI dependency missing (JSONeditor).
Even if the feature is optionnal, the dependency is required so I also activated it.


davidbgk:
r+

PR #311, opened 2016-01-28
Temp fix for export default

noirbizarre:
Temporary resolve `export default` resolution in handlebars helpers.

### DIFF
--- a/js/helpers/commonmark.js
+++ b/js/helpers/commonmark.js
@@ -1,9 +1,9 @@
 import markdownit from 'markdown-it';
 import { options } from 'markdown-it/lib/presets/commonmark';
 
+options.linkify = true;
+const markdown = markdownit(options);
 
 export default function(text) {
-    options.linkify = true;
-    let markdown = markdownit(options);
     return markdown.render(text);
-};
+}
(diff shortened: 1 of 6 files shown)


js/helpers/commonmark.js
```diff
@@ -1,9 +1,9 @@
 import markdownit from 'markdown-it';
 import { options } from 'markdown-it/lib/presets/commonmark';
 
+options.linkify = true;
+const markdown = markdownit(options);
 
 export default function(text) {
```
davidbgk: Having to use `.default()` everywhere is a bit ugly, can't we "just" `export mardown.render` or something in this vein?

noirbizarre: It's not working like that. In ES6 file, there is no default. This is only for legacy files, this is why only handlebars helpers (meant to disappears) here need to use `.default()`.
Changing the export syntax won't change a thing :/

davidbgk: Too bad :(

davidbgk:
r+

js/helpers/commonmark.js
noirbizarre: An alternative to the default() is https://www.npmjs.com/package/babel-plugin-add-module-exports but it will add some more code to the generated output.

PR #312, opened 2016-01-28
Expose check-url configuration and disable client-side check if not enabled

noirbizarre:
This PR disable client-side checks if server-side is not enabled (and so does not produce 500 errors)

### DIFF
--- a/js/config.js
+++ b/js/config.js
@@ -63,6 +63,11 @@ export const title = _meta('site-title');
 export const csrf_token = _meta('csrf-token');
 
 /**
+ * Is the check url feature enabled ?
+ */
+export const check_urls = JSON.parse(_meta('check-urls'));
+
+/**
  * The API root/base URL
  */
 export const api_root = _link('api-root');
@@ -123,4 +128,5 @@ export default {
     auth_url,
     sentry,
     notify_in,
+    check_urls,
 };
--- a/js/dataset/display.js
+++ b/js/dataset/display.js
@@ -48,36 +48,38 @@ function prepare_resources() {
         });
 
         // Check asynchronuously the status of displayed resources
-        $this.find('.format-label').each(function() {
-            const $self = $(this);
-            const url = $self.parent().property('url').first().attr('href');
-            const group = $Dataset.property('alternateName').value(); // This is the slug.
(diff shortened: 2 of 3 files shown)


js/config.js
```diff
@@ -63,6 +63,11 @@ export const title = _meta('site-title');
 export const csrf_token = _meta('csrf-token');
 
 /**
+ * Is the check url feature enabled ?
+ */
+export const check_urls = JSON.parse(_meta('check-urls'));
```
davidbgk: Oh nice, I didn't think about using `JSON.parse` for that in territories, I'll update :)

js/dataset/display.js
```diff
@@ -48,36 +48,38 @@ function prepare_resources() {
         });
 
         // Check asynchronuously the status of displayed resources
-        $this.find('.format-label').each(function() {
```
davidbgk: Did you change anything except the indent? Not sure if that's github messing up with the diff or if there are actual changes.

udata/templates/metadata.html
```diff
@@ -41,4 +41,5 @@
 />
 {% endif %}
 
-<meta name="notify-in" content="{{ meta.notify_in or '#notifications-container' }}">
+<meta name="notify-in" content="{{ meta.notify_in or '#notifications-container' }}" />
+<meta name="check-urls" content="{{ 'true' if config.CROQUEMORT else 'false' }}" />
```
davidbgk: Maybe `str(bool(config.CROQUEMORT)).lower()` mmh typing it I'm not sure anymore it's more explicit…

davidbgk:
r+

js/config.js
noirbizarre: JSON.parse for primitive types works well ;)

js/dataset/display.js
noirbizarre: Only the indent, not anything else.

udata/templates/metadata.html
noirbizarre: :) I tried different syntax. Maybe I should create a filter `boolify` ?

davidbgk: Nevermind, the current version is fine.

PR #313, opened 2016-01-28
Placeholders coherence

noirbizarre:
This pull request give some coherence to the placeholders management with only one JS module ES5/ES6-compatible.


davidbgk:
r+

PR #315, opened 2016-02-01
Dataset empty

vinyll:
Some datasets list have empty (deleted) items in topics.
See https://www.data.gouv.fr/en/topics/territoires-et-transports/ > _featured datasets_.
screen shot 2016-02-01 at 5 10 15 pm

### DIFF
--- a/udata/core/topic/views.py
+++ b/udata/core/topic/views.py
@@ -44,6 +44,7 @@ def display(topic):
     return theme.render(
         'topic/display.html',
         topic=topic,
+        datasets=[d for d in topic.datasets if hasattr(d, 'pk')],
         **dict(zip(keys, results))
     )
 
--- a/udata/tests/dataset/test_dataset_tasks.py
+++ b/udata/tests/dataset/test_dataset_tasks.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+
+from .. import TestCase, DBTestMixin
+
+from udata.models import Dataset, Topic
+from udata.core.dataset import tasks
+
+
+class DatasetTasksTest(TestCase, DBTestMixin):
+    def test_purge_datasets(self):
+        datasets = [
+            Dataset.objects.create(title='delete me', deleted='2016-01-01'),
+            Dataset.objects.create(title='keep me'),
+        ]
+
+        topic = Topic.objects.create(name='test topic', datasets=datasets)
+        tasks.purge_datasets()
+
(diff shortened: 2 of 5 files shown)


udata/core/topic/views.py
```diff
@@ -44,6 +44,7 @@ def display(topic):
     return theme.render(
         'topic/display.html',
         topic=topic,
+        datasets=[d for d in topic.datasets if hasattr(d, 'pk')],
```
davidbgk: Maybe clearing the cache for `topic.datasets` is more interesting?

noirbizarre:
I think we should prevent this from happening again by updating the dataset purge task:
https://github.com/etalab/udata/blob/master/udata/core/dataset/tasks.py#L25-L40

vinyll:
3 steps are done:
- migrating topic.datasets for clearing any obsolete _dataset_ reference
- scheduling the removal of topic.datasets as a task when a dataset is deleted
- preventing from displaying an empty dataset

udata/core/topic/views.py
davidbgk: If we're running the `purge_datasets` command regularly, is that change still useful?

udata/tests/dataset/test_dataset_tasks.py
```diff
@@ -0,0 +1,20 @@
...
+        datasets = [
+            Dataset.objects.create(title='delete me', deleted='2016-01-01'),
+            Dataset.objects.create(title='keep me'),
+        ]
+
+        topic = Topic.objects.create(name='test topic', datasets=datasets)
+        tasks.purge_datasets()
+
+        topic = Topic.objects(name='test topic').first()
+        self.assertListEqual(topic.datasets, [datasets[1]])
```
davidbgk: :+1: nice test

davidbgk:
r+wc

udata/core/topic/views.py
vinyll: purge_datasets is run every hour in our case.
This change covers the time after deleting a dataset in a topic and before the `purge_datasets` is run which otherwise displays the empty card.
Therefore it's not crucial because it's 1h maximum (in the Datagouv project!) and we don't delete them so many times, however it's a bug display.

udata/tests/dataset/test_dataset_tasks.py
vinyll: :smile:

noirbizarre:
r+wc too. Merge ? :)

PR #316, opened 2016-02-01
Send email environment

vinyll:
Added a `SEND_MAIL` settings.
This should be set for the _next_ server to `False` so emails are not actually sent.


vinyll:
Wrong base branch. Redoing.

PR #317, opened 2016-02-01
added send_mail setting for environment control

vinyll:
Added a `SEND_MAIL` settings.
This should be set for the _next_ server to `False` so emails are not actually sent.


davidbgk:
r+ maybe an update of the `intra` repo is necessary?

vinyll:
the _intra_ repo?

noirbizarre:
r+

PR #321, opened 2016-02-03
Disable the automatic update of last_modified, fix #320

### LINKED ISSUE #320: Disable the automatic update of last_modified
As a visitor, I want to see a meaningful last modified date in order to know when a dataset has really been updated.

davidbgk:


udata/models/datetime_fields.py
```diff
@@ -63,5 +63,6 @@ class Datetimed(object):
 
 @pre_save.connect
 def set_modified_datetime(sender, document):
-    if isinstance(document, Datetimed):
+    from udata.models import Dataset
```
davidbgk: Not proud of this, maybe not inheriting from `Datetimed` is more clever after all.

noirbizarre:
r+ (way more explicit !)

PR #323, opened 2016-02-04
Fix the command related to logos fetching for territories, refs #284

davidbgk:
I should have tested that after the geozones refactoring.


noirbizarre:
r+

PR #324, opened 2016-02-04
Use babel transform-runtime plugin for all project  JS files

noirbizarre:
Add transform-runtime plugin everywhere.

NB: babel configuration can also be specified in `package.json` or `.babelrc`, which one do you prefer ?

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,5 @@
+{
+    "presets": ["es2015"],
+    "plugins": ["transform-runtime"],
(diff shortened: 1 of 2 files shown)


davidbgk:
r+ maybe `.babelrc` is more explicit?

noirbizarre:
OK, I prefer too, I'll update the PR ASAP

davidbgk:
Nicer :)

PR #328, opened 2016-02-04
Ensure checksum type is always present

noirbizarre:
This is the first part of fix on checksum handling (and globally on resources saving and validations).
This one fix bad checksum in database and raise ValidationError if not present ensuring it's not happening again.

### DIFF
--- a/udata/core/dataset/forms.py
+++ b/udata/core/dataset/forms.py
@@ -22,27 +22,6 @@ class ChecksumForm(ModelForm):
     value = fields.StringField()
 
 
-class ChecksumWidget(object):
-    def __call__(self, field, **kwargs):
-        html = [
-            '<div class="input-group checksum-group">',
-            '<div class="input-group-btn">',
-            field.form.type(class_='btn btn-default'),
-            '</div>',
-            field.value(class_='form-control'),
-            '</div>'
-        ]
-        return widgets.HTMLString(''.join(html))
-
-
-class ChecksumField(fields.FormField):
-    widget = ChecksumWidget()
-
-    def __init__(self, label=None, validators=None, **kwargs):
-        super(ChecksumField, self).__init__(ChecksumForm, label, validators,
-                                            **kwargs)
-
-
 class BaseResourceForm(ModelForm):
     title = fields.StringField(_('Title'), [validators.required()])
     description = fields.MarkdownField(_('Description'))
@@ -55,7 +34,7 @@ class BaseResourceForm(ModelForm):
         _('URL'), [validators.required()], storage=resources)
     format = fields.StringField(
         _('Format'), widget=widgets.FormatAutocompleter())
-    checksum = ChecksumField(_('Checksum'))
+    checksum = fields.FormField(ChecksumForm)
     mime = fields.StringField(
         _('Mime type'),
         description=_('The mime type associated to the extension. '
--- a/udata/forms/fields.py
+++ b/udata/forms/fields.py
@@ -198,9 +198,10 @@ class FormWrapper(object):
 
 class FormField(FieldHelper, fields.FormField):
     def __init__(self, form_class, *args, **kwargs):
-        self.has_data = False
         super(FormField, self).__init__(FormWrapper(form_class),
                                         *args, **kwargs)
+        self.prefix = '{0}-'.format(self.name)
+        self.has_data = False
 
     def process(self, formdata, data=unset_value):
         self._formdata = formdata
@@ -209,11 +210,11 @@ class FormField(FieldHelper, fields.FormField):
     def validate(self, form, extra_validators=tuple()):
         if extra_validators:
             raise TypeError('FormField does not accept in-line validators, '
-            'as it gets errors from the enclosed form.')
+                            'as it gets errors from the enclosed form.')
 
         # Run normal validation only if there is data for this form
-        prefix = '{0}-'.format(self.name)
-        self.has_data = self._formdata and any(k.startswith(prefix) for k in self._formdata)
+        self.has_data = bool(self._formdata)
+        self.has_data &= any(k.startswith(self.prefix) for k in self._formdata)
         if self.has_data:
             return self.form.validate()
         return True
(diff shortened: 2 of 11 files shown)


davidbgk:
r+

davidbgk:
Note that tests have failed.

udata/tests/__init__.py
```diff
@@ -176,6 +176,7 @@ def init_search(self):
         self._used_search = True
         self.app.config['AUTO_INDEX'] = True
         es.initialize()
+        es.cluster.health(wait_for_status='yellow', request_timeout=3)
```
davidbgk: :+1:

udata/tests/api/test_datasets_api.py
```diff
@@ -93,6 +96,7 @@ def test_dataset_api_get_deleted_but_authorized(self):
         response = self.get(url_for('api.dataset', dataset=dataset))
         self.assert200(response)
 
+    @attr('create')
```
davidbgk: I didn't know about that feature, nice!

udata/core/dataset/forms.py
```diff
@@ -22,27 +22,6 @@ class ChecksumForm(ModelForm):
     value = fields.StringField()
 
 
-class ChecksumWidget(object):
```
davidbgk: We don't need a field/widget anymore?

udata/forms/fields.py
```diff
@@ -200,10 +198,29 @@ def __getattr__(self, name):
...
                                         *args, **kwargs)
 
+    def process(self, formdata, data=unset_value):
+        self._formdata = formdata
+        return super(FormField, self).process(formdata, data=data)
+
+    def validate(self, form, extra_validators=tuple()):
+        if extra_validators:
+            raise TypeError('FormField does not accept in-line validators, '
+            'as it gets errors from the enclosed form.')
```
davidbgk: Better indent at the opened parenthesis?

udata/forms/fields.py
```diff
@@ -200,10 +198,29 @@ def __getattr__(self, name):
...
     def __init__(self, form_class, *args, **kwargs):
+        self.has_data = False
         super(FormField, self).__init__(FormWrapper(form_class),
                                         *args, **kwargs)
 
+    def process(self, formdata, data=unset_value):
+        self._formdata = formdata
+        return super(FormField, self).process(formdata, data=data)
+
+    def validate(self, form, extra_validators=tuple()):
```
davidbgk: Initialize with `None` and check `if extra_validators is None`.

udata/forms/fields.py
```diff
@@ -200,10 +198,29 @@ def __getattr__(self, name):
 
 class FormField(FieldHelper, fields.FormField):
     def __init__(self, form_class, *args, **kwargs):
+        self.has_data = False
```
davidbgk: I'm not familiar enough with WTForm but that looks weird that you have to deal with that manually :/

udata/forms/fields.py
```diff
@@ -200,10 +198,29 @@ def __getattr__(self, name):
...
+        self._formdata = formdata
+        return super(FormField, self).process(formdata, data=data)
+
+    def validate(self, form, extra_validators=tuple()):
+        if extra_validators:
+            raise TypeError('FormField does not accept in-line validators, '
+            'as it gets errors from the enclosed form.')
+
+        # Run normal validation only if there is data for this form
+        prefix = '{0}-'.format(self.name)
```
davidbgk: Move that to the `init` for consistency with `NestedModelList`?

udata/forms/fields.py
```diff
@@ -391,16 +408,29 @@ def __init__(self, model_form, *args, **kwargs):
...
-        if formdata and any(k.startswith(prefix) for k in formdata):
+        self.has_data = formdata and any(k.startswith(self.prefix) for k in formdata)
+        if self.has_data:
             super(NestedModelList, self).process(formdata, data)
         else:
-            super(NestedModelList, self).process(None, data)
+            self.entries = []
+            # super(NestedModelList, self).process(None, data)
+
+    def validate(self, form, extra_validators=tuple()):
```
davidbgk: Same here.

udata/forms/fields.py
```diff
@@ -391,16 +408,29 @@ def __init__(self, model_form, *args, **kwargs):
...
         self.data_submitted = False
         self.initial_data = []
+        self.prefix = '{0}-'.format(self.name)
 
     def process(self, formdata, data=unset_value):
+        self._formdata = formdata
         self.initial_data = data
-        prefix = '{0}-'.format(self.name)
-        if formdata and any(k.startswith(prefix) for k in formdata):
+        self.has_data = formdata and any(k.startswith(self.prefix) for k in formdata)
```
davidbgk: It smells logic duplication.

davidbgk:
So great to have hunt that bug down, r+wc :)

udata/tests/__init__.py
noirbizarre: It was driving me crazy! Such time lost !!!

udata/tests/api/test_datasets_api.py
noirbizarre: I just discovered it ! Very useful !

udata/core/dataset/forms.py
noirbizarre: We don't have front forms any more so all widgets are no longer used (except for the coming login/register)

udata/forms/fields.py
noirbizarre: Sorry, dumb copy past from upstream, I will fix that.

udata/forms/fields.py
noirbizarre: I kept the upstream signature. I think there is some iteration in `super().validate()`

udata/forms/fields.py
noirbizarre: The big problem is:
- WTForms is meant for form validation (so flat structure with MultiDict) and we are dealing with JSON and nested object
- WTForms consider FormField as required. It took me a lot of mailinglists/issues/stackoverflow reading to come to this "clean" solution :/

Conclusion: I think we should migrate from WTForms to something dedicated to JSON/dict

udata/forms/fields.py
noirbizarre: OK

udata/forms/fields.py
noirbizarre: Yep, but I didn't want to factorize yet. It seems overkill to me for 2 attributes.
It's to becoming complex :/

PR #330, opened 2016-02-05
Popin not closing when reporting bug

vinyll:


js/notify.js
```diff
@@ -15,7 +15,7 @@ define([
             level: type == 'error' ? 'danger' : type,
             message: message
         }));
-        Log.debug('Notify:', {type: type, message: message, container: container});
+        log.debug('Notify:', {type: type, message: message, container: container});
```
davidbgk: Are you sure about that because `Log` is the second parameter of the function and equals to `'logger'`?

vinyll: Nope, you're right. Need to investigate further why this causes a `Log.debug is not a function` error.

js/notify.js
```diff
@@ -15,7 +15,7 @@ define([
             level: type == 'error' ? 'danger' : type,
             message: message
         }));
-        Log.debug('Notify:', {type: type, message: message, container: container});
+        Log.default.debug('Notify:', {type: type, message: message, container: container});
```
vinyll: I'm not sure why we have a _default_ property here that access the functions yet it works

davidbgk: Well, I think that turning that file to ES6 would allow you to use directly `import log from 'logger';` and that's the way to go ;)

vinyll: And then the _default_ makes sense now. I'll go that path, thanks :)

noirbizarre: The other alternative is to move from AMD to ES6 syntax. Then, the `default` won't be necessary anymore.

Just exporting `error` and `success` won't change the signature so there won't be copatibility issue for AMD module using the notify module.

vinyll:
I moved _js/notify.js_ to an es6 module.
But this requires all files using `Notify` to import is as es6.
Do you see a better way or should I just migrate everyting?
([amdtoes6](https://github.com/jonbretman/amd-to-as6) to the rescue)

js/notify.js
```diff
@@ -1,30 +1,25 @@
...
  * User notification module
  */
-define([
-    'jquery',
-    'logger',
-    'config',
-    'templates/notification.hbs'
-], function($, Log, config, message_template) {
-    'use strict';
+'use strict';
```
davidbgk: I think we don't need that anymore, do you confirm @noirbizarre?

js/notify.js
```diff
@@ -1,30 +1,25 @@
...
+        message: message
+    }));
+}
 
-        success: function(message, container) {
-            show_message(message, 'success', container);
-        }
-    };
-});
+export const error = function(message, container) {
```
vinyll: vs `export function error(message, container) {`
any preference?

davidbgk: Consistency ;)

vinyll:
I removed `use strict` and moved from `export const error = function` to `export function error` for consistency with other files.

davidbgk:
r+

davidbgk:
@vinyll merge?

js/notify.js
```diff
@@ -1,30 +1,23 @@
...
  */
-define([
-    'jquery',
-    'logger',
-    'config',
-    'templates/notification.hbs'
-], function($, Log, config, message_template) {
-    'use strict';
+import $ from 'jquery';
+import log from 'logger';
```
noirbizarre: Not needed anymore

js/notify.js
```diff
@@ -1,30 +1,23 @@
...
-        container = container || config.notify_in;
-        $(container).prepend(message_template({
-            level: type == 'error' ? 'danger' : type,
-            message: message
-        }));
-        Log.debug('Notify:', {type: type, message: message, container: container});
-    }
+function show_message(message, type, container) {
+    container = container || config.notify_in;
+    $(container).prepend(message_template({
```
noirbizarre: Do we keep using jQuery ?

noirbizarre:
```javascript
const el = document.querySelector(container);
const msg = message_template({});
el.insertBefore(msg, el.childNodes[0]);
```

vinyll: I chose not to use insertBefore because when the element is empty is throws an error and does not insert the element. Plus msg is a string, not a DOM element.
And then I found out this `insertAdjacentHTML` method: https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML

noirbizarre: It must depends on the browser because I tried with an empty element eand it works (the second `insertBefore()` parameter is optional).
If you have a working solution with `insertAdjacenHTML`, just go with it. It was only a suggestion ;)

noirbizarre:
r+ aussi

PR #333, opened 2016-02-05
Use the tmp storage for geozones loading, fix #332

### LINKED ISSUE #332: Use the tmp storage for geozones loading
As done for the geologos load.

davidbgk:


udata/core/spatial/commands.py
```diff
@@ -29,16 +28,14 @@
 @m.command
 def load(filename, drop=False):
     '''Load a GeoZones Bundle'''
-    tmp = tempfile.mkdtemp()
-
     if filename.startswith('http'):
         log.info('Downloading GeoZones bundle: %s', filename)
-        filename, _ = urlretrieve(filename, join(tmp, 'geozones.tar.xz'))
+        filename, _ = urlretrieve(filename, join(tmp.root, 'geozones.tar.xz'))
```
noirbizarre: I think you can write `tmp.path('geozones.tar.xz')` instead of manually using `join()`

udata/core/spatial/commands.py
```diff
@@ -47,47 +44,42 @@ def load(filename, drop=False):
         GeoLevel.drop_collection()
 
     log.info('Loading levels.json')
-    total = 0
-    with open(join(tmp, 'levels.json')) as fp:
+    levels_filepath = join(tmp.root, 'levels.json')
```
noirbizarre: same as above

udata/core/spatial/commands.py
```diff
@@ -47,47 +44,42 @@ def load(filename, drop=False):
...
+    log.info('Loaded {total} levels'.format(total=len(levels)))
 
     if drop:
         log.info('Dropping existing spatial zones')
         GeoZone.drop_collection()
 
     log.info('Loading zones.json')
-    total = 0
-    with open(join(tmp, 'zones.json')) as fp:
+    zones_filepath = join(tmp.root, 'zones.json')
```
noirbizarre: same as above

noirbizarre:
r+wc

davidbgk:
I updated the command for geologos too, thanks!

PR #335, opened 2016-02-08
Fix the public timeline for users and orgs, fix #334

### LINKED ISSUE #334: Fix the public timeline for users (global vs. personal one)

davidbgk:


js/components/activities/timeline.vue
```diff
@@ -43,7 +43,7 @@ function actor(activity) {
 
 export default {
     el: '#activities',
-    props: ['organizationId', 'userId'],
+    props: ['dataOrganizationId', 'dataUserId'],
```
noirbizarre: I think the original version was fine, the problem seems to come from https://github.com/etalab/udata/blob/master/udata/templates/user/activity.html#L17
It should be `user-id="{{ user.id }}">` instead of `data-user-id="{{ user.id }}"`.
From 1.0 changelog: https://github.com/vuejs/vue/releases/tag/1.0.0:

> Props with the `data-` prefix are now longer supported.

noirbizarre:
r+wc

PR #345, opened 2016-02-12
Ensure that map.vue does not fail if GeoJSON has no property

noirbizarre:
When using the geom field from spatial coverage (manualy submited coverage zone), the map.vue fails because not properties are present in the GeoJSON.
This PR avoid this by checking the feature presence.

It raises a question: what do we chose to display when we have both `geom` and `zones` in the spatial coverage. The current behavior is to display the manualy submitted `geom`.

### DIFF
--- a/js/components/widgets/map.vue
+++ b/js/components/widgets/map.vue
@@ -65,29 +65,23 @@ export default {
 
 
         L.tileLayer(TILES_URL, TILES_CONFIG).addTo(this.map);
-        // this.layer.addTo(this.map);
-        // this.map.fitBounds(this.layer.getBounds());
     },
     watch: {
         'geojson': function(json) {
             if (json) {
                 this.layer = L.geoJson(json, {
                     onEachFeature: function (feature, layer) {
-                        layer.bindPopup(feature.properties.name);
-                        layer.on("mouseover", function () {
-                            layer.openPopup();
-                        });
-                        layer.on("mouseout", function () {
-                            layer.closePopup();
-                        });
+                        if (feature.properties && feature.properties.name) {
+                            layer.bindPopup(feature.properties.name);
+                            layer.on("mouseover", function () {
+                                layer.openPopup();
+                            });
+                            layer.on("mouseout", function () {
+                                layer.closePopup();
+                            });
+                        }
                     }
                 });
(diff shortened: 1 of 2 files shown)


js/components/widgets/map.vue
```diff
@@ -65,29 +65,23 @@ export default {
...
-                        layer.on("mouseout", function () {
-                            layer.closePopup();
-                        });
+                        if (feature.properties && feature.properties.name) {
+                            layer.bindPopup(feature.properties.name);
+                            layer.on("mouseover", function () {
+                                layer.openPopup();
+                            });
+                            layer.on("mouseout", function () {
+                                layer.closePopup();
```
davidbgk: Yay! A conflict with my branch :p

I think you should remove the `closePopup` in that PR as it doesn't exist in leaflet and raises an error.

davidbgk:
> what do we chose to display when we have both geom and zones in the spatial coverage. The current behavior is to display the manualy submitted geom.

Maybe we should warn the user and tell him that we'll use our own zones because it's updated etc?

davidbgk:
r+wc

js/components/widgets/map.vue
noirbizarre: I didn't to touch the popup behavior as it has been handled in a separate issue but if you prefer, I can remove it.

js/components/widgets/map.vue
```diff
@@ -65,29 +65,29 @@ export default {
...
-                        layer.on("mouseover", function () {
-                            layer.openPopup();
-                        });
-                        layer.on("mouseout", function () {
-                            layer.closePopup();
-                        });
+                    onEachFeature: (feature, layer) => {
+                        if (feature.properties && feature.properties.name) {
+                            layer.bindPopup(feature.properties.name);
+                            layer.on("mouseover", function() {
```
davidbgk: `=>` here and below? :)

js/dataset/display.js
```diff
@@ -165,7 +165,13 @@ function load_coverage_map() {
                 layer.openPopup();
             });
             layer.on('mouseout', function() {
```
davidbgk: Same here

noirbizarre:
OK, I fixed the close popup instead. Good to go ?

PR #350, opened 2016-02-13
Remove the cached_property from TerritoryDataset class, fix #349

### LINKED ISSUE #349: Remove the cached_property from TerritoryDataset class
Otherwise given that it's a parent of other classes, the cache is not refreshed.

davidbgk:


noirbizarre:
r+

PR #351, opened 2016-02-13
Fix some markdown rendering issues

noirbizarre:
This PR fixes some glitches rendering markdown.

### DIFF
--- a/udata/frontend/markdown.py
+++ b/udata/frontend/markdown.py
@@ -66,7 +66,8 @@ def init_app(app):
             return ''
         if EXCERPT_TOKEN in value:
             value = value.split(EXCERPT_TOKEN, 1)[0]
-        if length > 0:
-            value = do_truncate(value, length, end='…')
         rendered = md(value)
-        return do_striptags(rendered)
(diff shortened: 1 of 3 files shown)


davidbgk:
r+ doesn't solve that bug https://demo.data.gouv.fr/fr/territory/35238/ when an image is the first element it doesn't strip it, probably a bug upstream…

davidbgk:
r++

udata/frontend/markdown.py
```diff
@@ -66,7 +66,8 @@ def mdstrip(value, length=None):
             return ''
         if EXCERPT_TOKEN in value:
             value = value.split(EXCERPT_TOKEN, 1)[0]
-        if length > 0:
-            value = do_truncate(value, length, end='…')
         rendered = md(value)
-        return do_striptags(rendered)
+        text = do_striptags(rendered)
+        if length > 0:
+            text = do_truncate(text, length, end='…')
```
davidbgk: Maybe add the `end` parameter to the `mdstrip` method? There is a use-case we talked about last week but I can't remember where…

PR #359, opened 2016-02-17
Expose search facets in api

noirbizarre:
This PR expose search facets in API:
- expose `facets` parameter in API documentation
- expose `facets` in paginated results

### DIFF
--- a/udata/api/__init__.py
+++ b/udata/api/__init__.py
@@ -120,17 +120,20 @@ class UDataApi(Api):
         # q parameter
         parser.add_argument('q', type=str, location='args',
                             help='The search query')
+        # Expected facets
+        # (ie. I want all facets or I want both tags and licenses facets)
         facets = adapter.facets.keys()
         if facets:
             parser.add_argument('facets', type=str, location='args',
                                 choices=['all'] + facets, action='append',
                                 help='Selected facets to fetch')
         # Add facets filters arguments
+        # (apply a value to a facet ie. tag=value)
         for name, facet in adapter.facets.items():
             parser.add_argument(name, type=str, location='args')
         # Sort arguments
         keys = adapter.sorts.keys()
-        choices = keys + ['-' + k for k in adapter.sorts.keys()]
+        choices = keys + ['-' + k for k in keys]
         parser.add_argument('sort', type=str, location='args', choices=choices,
                             help='The field (and direction) on which sorting apply')
         if paginate:
--- a/udata/api/__init__.py
+++ b/udata/api/__init__.py
@@ -39,8 +39,8 @@ HEADER_API_KEY = 'X-API-KEY'
 
 class UDataApi(Api):
     def __init__(self, app=None, **kwargs):
-        kwargs['decorators'] = ([self.authentify]
-                                + (kwargs.pop('decorators', []) or []))
+        decorators = kwargs.pop('decorators', []) or []
+        kwargs['decorators'] = [self.authentify] + decorators
         super(UDataApi, self).__init__(app, **kwargs)
         self.authorizations = {
             'apikey': {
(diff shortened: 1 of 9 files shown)


udata/api/__init__.py
```diff
@@ -120,13 +120,19 @@ def search_parser(self, adapter, paginate=True):
...
+                                help='Selected facets to fetch')
         # Add facets filters arguments
         for name, facet in adapter.facets.items():
             parser.add_argument(name, type=str, location='args')
         # Sort arguments
-        choices = (adapter.sorts.keys()
-                   + ['-' + k for k in adapter.sorts.keys()])
-        parser.add_argument('sort', type=str, location='args', choices=choices)
+        keys = adapter.sorts.keys()
+        choices = keys + ['-' + k for k in adapter.sorts.keys()]
```
davidbgk: Why not just `keys`?

udata/api/__init__.py
```diff
@@ -120,13 +120,19 @@ def search_parser(self, adapter, paginate=True):
...
                             help='The search query')
+        facets = adapter.facets.keys()
+        if facets:
+            parser.add_argument('facets', type=str, location='args',
+                                choices=['all'] + facets, action='append',
+                                help='Selected facets to fetch')
         # Add facets filters arguments
         for name, facet in adapter.facets.items():
             parser.add_argument(name, type=str, location='args')
         # Sort arguments
```
davidbgk: Add a comment to explain the (-)logic

udata/api/__init__.py
```diff
@@ -120,13 +120,19 @@ def search_parser(self, adapter, paginate=True):
         # q parameter
         parser.add_argument('q', type=str, location='args',
                             help='The search query')
+        facets = adapter.facets.keys()
+        if facets:
+            parser.add_argument('facets', type=str, location='args',
+                                choices=['all'] + facets, action='append',
+                                help='Selected facets to fetch')
         # Add facets filters arguments
         for name, facet in adapter.facets.items():
```
davidbgk: Should be in the `if facets:` block?

udata/api/__init__.py
```diff
@@ -120,13 +120,19 @@ def search_parser(self, adapter, paginate=True):
         # q parameter
         parser.add_argument('q', type=str, location='args',
                             help='The search query')
+        facets = adapter.facets.keys()
+        if facets:
+            parser.add_argument('facets', type=str, location='args',
+                                choices=['all'] + facets, action='append',
+                                help='Selected facets to fetch')
         # Add facets filters arguments
```
davidbgk: Put that comment a few lines above?

udata/search/fields.py
```diff
@@ -145,23 +148,30 @@ def __init__(self, field, model, labelizer=None, field_name='id'):
...
+    def from_response(self, name, response, fetch=True):
         facet = response.get('facets', {}).get(name)
         if not facet:
             return
         ids = [term['term'] for term in facet['terms']]
-        if is_objectid:
-            ids = map(ObjectId, ids)
-        objects = self.model.objects.in_bulk(ids)
+
+        if fetch:
```
davidbgk: All that block is hard to read/understand/debug, it smells a bit too javascriptic to me ;)

davidbgk:
I'm probably missing something but I'm not sure in which cases the `'all'` key is pertinent, isn't it better to only keep facets parameters explicit with the complete list for people willing to have all facets? It introduces some complexity within the code without a big improvement of the developer experience.

udata/api/__init__.py
noirbizarre: The negative values allow to sort in reverse order: `key` or `-key`.
I didn't change that, I just formatted it to more readable.

udata/api/__init__.py
noirbizarre: no, there is 2 aspecst on facets: the filtering was already implemented. This PR only expose the facet querying and results.

These lines were not touched because they are doing the filtering part.

udata/api/__init__.py
noirbizarre: same as above

noirbizarre:
Given that our main use case in the front search is using the `all` parameter, I considered as the main use case. Selecting facets allow to reduce payload for those who already know what specific facets they want.
In fact, it was `facets=True` but neither Swagger nor Flask-restful RequestParser support mixed parameters (string list or boolean) so I added the special `all` value to only have strings.

The only change introduced in the search part is the optional fetch: front need the objects to render the facets, API only need the IDs.

udata/search/fields.py
noirbizarre: The fetch part is the existing code (model are fetched to render the facets in front).
The only ne part is the `else` part which only returns the model class and its identifier.

udata/api/__init__.py
noirbizarre: lol, I just understood. You're right, that was the initial point of storing the `keys` variable.

udata/api/__init__.py
noirbizarre: I don't know why this comment is not disappearing, I Added some comment in the code.

udata/search/fields.py
noirbizarre: Added some comments in the code

noirbizarre:
Ah, I forgot to tell, the issues/discussions part change is included because this change broke some false positive test. I fixed them in here but I can split that into a separate PR.

udata/search/fields.py
davidbgk: :+1:

davidbgk:
r+

PR #363, opened 2016-02-19
Safer CSV export, fix #362

### LINKED ISSUE #362: Safer CSV export
As a user, I want to be able to download a CSV without any error even if there is a technical issue on a given field.

davidbgk:


udata/frontend/csv.py
```diff
@@ -79,7 +89,11 @@ def to_row(self, obj):
         '''Convert an object into a flat csv row'''
         row = []
         for name, getter in self.get_fields():
-            row.append(safestr(getter(obj)))
+            try:
+                row.append(safestr(getter(obj)))
+            except Exception, e:  # Catch all errors intentionally.
+                log.error('Error exporting CSV for {name}: {error}'.format(
```
noirbizarre: I think this will create shorter rows so invalid CSV, we need to append something to the row

udata/frontend/csv.py
```diff
@@ -128,7 +152,11 @@ def nested_row(self, obj, nested):
         '''Convert an object into a flat csv row'''
         row = self.to_row(obj)
         for name, getter in self.get_nested_fields():
-            row.append(safestr(getter(nested)))
+            try:
+                row.append(safestr(getter(nested)))
+            except Exception, e:  # Catch all errors intentionally.
+                log.error('Error exporting CSV for {name}: {error}'.format(
```
noirbizarre: Same as above, will produce shorter row on error.
We need to append an empty value

noirbizarre:
r+

PR #364, opened 2016-02-19
Missing mails

noirbizarre:
This PR:
- add mail testing facilities
- fix the mailer choice (use FakeMailer when debugging or testing and real mailer in production)
- test issues and discussions signals
- test issues and discussions mails
- fix discussions mails not being sent.

I think another PR is required to test all the mails and, even better, group test by modules.

### DIFF
--- a/udata/core/discussions/models.py
+++ b/udata/core/discussions/models.py
@@ -46,5 +46,5 @@ class Discussion(db.Document):
     @property
     def external_url(self):
         return self.subject.url_for(
-            _anchor='discussion-{ id }'.format(id=self.id),
+            _anchor='discussion-{id}'.format(id=self.id),
             _external=True)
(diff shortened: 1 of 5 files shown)


udata/core/discussions/models.py
```diff
@@ -46,5 +46,5 @@ def person_involved(self, person):
     @property
     def external_url(self):
         return self.subject.url_for(
-            _anchor='discussion-{ id }'.format(id=self.id),
+            _anchor='discussion-{id}'.format(id=self.id),
```
davidbgk: So the actual bug was here? :(

udata/mail.py
```diff
@@ -14,11 +16,21 @@
 
 mail = Mail()
 
+mail_sent = signal('mail-sent')
+
+
+class FakeMailer(object):
```
davidbgk: :+1:

davidbgk:
r+

udata/core/discussions/models.py
noirbizarre: Yes, the spaces were used in the key resolution :( (we are too used too Jinja and Vue)

PR #367, opened 2016-02-20
Make the quality indicators collapsible

noirbizarre:
This PR makes the quality indicators foldables/collapsibles in the admin.
The green/ok ones are initialy folded.

screenshot-quality-folding

### DIFF
--- a/js/components/dataset/quality-section.vue
+++ b/js/components/dataset/quality-section.vue
@@ -1,0 +1,69 @@
+<template>
+<div>
+    <h3 class="pointer" :class="textClass" @click="toggle">
+        <span class="fa" :class="iconClass"></span>
+        {{ title }}
+        <!--small :class="textClass">
+        </small-->
+    </h3>
+    <div class="collapse" :class="contentClass">
+        <slot></slot>
+        <p :class="textClass">
+            <!--span class="fa" :class="iconClass"></span-->
+            <strong v-if="condition && !showInfo">{{ ok }}</strong>
+            <strong v-if="!condition && !showInfo">{{ ko }}</strong>
+            <strong v-if="showInfo">{{ info }}</strong>
+        </p>
+    </div>
+</div>
+</template>
+<script>
+export default {
+    props: {
+        title: String,
+        condition: Boolean,
+        ok: String,
+        ko: String,
+        info: [String, null],
+        showInfo: {
+            type: Boolean,
+            default: false
+        },
+    },
+    data() {
+        return {
+            toggled: null
+        }
+    },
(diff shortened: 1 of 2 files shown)


js/components/dataset/quality-section.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+<div>
+    <h3 class="pointer" :class="textClass" @click="toggle">
+        <span class="fa" :class="iconClass"></span>
+        {{ title }}
+        <!--small :class="textClass">
```
davidbgk: Remove comment?

js/components/dataset/quality-section.vue
```diff
@@ -0,0 +1,69 @@
...
+        </small-->
+    </h3>
+    <div class="collapse" :class="contentClass">
+        <slot></slot>
+        <p :class="textClass">
+            <!--span class="fa" :class="iconClass"></span-->
```
davidbgk: Remove comment?

js/components/dataset/quality-section.vue
```diff
@@ -0,0 +1,69 @@
...
+        }
+    },
+    methods: {
+        toggle() {
+            this.toggled = !this.expanded;
+        }
+    }
+};
+</script>
+<style lang="less"></style>
```
davidbgk: Useless?

js/components/dataset/quality-section.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+<div>
+    <h3 class="pointer" :class="textClass" @click="toggle">
```
davidbgk: I wonder how vuejs deal with that kind of double declaration of classes. Maybe we should add static one to the dynamic ones to keep all in the same place?

davidbgk:
:white_check_mark: R+wc

js/components/dataset/quality-section.vue
noirbizarre: Vue.js handle perfectly (now) these cases:
- take static classes from the component (if it's a component)
- merge static classes from outside the component (those declared on call)
- merge dynamic classes from the component
- merge dynamic classes from outside

So you can write something like:
```xml
<!-- my-component.vue -->
<template>
<div class="static-class-inside" :class="computedClasses"></div>
</template>
```
```xml
<!-- caller.vue -->
<template>
<my-component class="static-class-outside" :class="computedClassesOutside"></my-component>
</template>
```
And the 4 classes declaration will be merged !

Until now, I always split static and dynamic classes to enforce the difference, but if it's not that clear we can discuss and change that.

PR #368, opened 2016-02-24
Workaround babel startsWith not being transpiled in some case

noirbizarre:
IE does not have natively `String.prototype.startsWith`.
Given how Babel is working when using babel-runtime, polyfills are not applied to prototype and there is some case where babel-runtime is not able to detected variable type and to transpile the polyfill.

This PR fix this on `variable.startsWith`.

### DIFF
--- a/js/dataset/display.js
+++ b/js/dataset/display.js
@@ -1,6 +1,9 @@
 /**
  * Dataset display page JS module
  */
+
+// ES6 environment
+import 'babel-polyfill';
 import $ from 'jquery';
 import config from 'config';
 import log from 'logger';
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,4 @@
 {
     "presets": ["es2015"],
-    "plugins": ["transform-runtime"],
     "comments": false
 }
--- a/js/admin.js
+++ b/js/admin.js
@@ -6,6 +6,9 @@ import '../less/admin.less';
 // Catch all errors
 import 'raven';
 
+// ES6 environment
+import 'babel-polyfill';
+
 import $ from 'jquery';
 import 'bootstrap';
 
--- a/js/dashboard.js
+++ b/js/dashboard.js
@@ -1,3 +1,6 @@
+// ES6 environment
+import 'babel-polyfill';
+
 // Catch all errors
 import 'raven';
 
--- a/js/home.js
+++ b/js/home.js
@@ -6,6 +6,9 @@ import '../less/home.less';
 // Catch all errors
 // require('raven');
 
+// ES6 environment
+import 'babel-polyfill';
+
 import $ from 'jquery';
 import 'bootstrap';
 
--- a/js/organization/display.js
+++ b/js/organization/display.js
@@ -1,6 +1,10 @@
 /**
  * Default JS module
  */
+
+// ES6 environment
+import 'babel-polyfill';
+
 import $ from 'jquery';
 import log from 'logger';
 import Auth from 'auth';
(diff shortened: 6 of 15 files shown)


js/dataset/display.js
```diff
@@ -54,10 +59,10 @@ function prepare_resources() {
                 const url = $self.parent().property('url').first().attr('href');
                 const group = $Dataset.property('alternateName').value(); // This is the slug.
 
-                if (!url.startsWith(window.location.origin)
+                if (!startsWith(url, window.location.origin)
                 // TODO: temporary fix before we move all statics on the same server
                     && !url.match(/:\/\/(.[^/]+)/)[1].endsWith('data.gouv.fr')) {
```
vinyll: Does `startWith` not work and `endWith` works properly?

noirbizarre: Yes, in this case: there is the `match` before (native on IE) which is probably hinting babel on the resulting type.
But you're right, I might be better to be homogeneous on this one.

vinyll:
r+wc

noirbizarre:
Babel-runtime is not a strict equivalent of babel-polyfill:
- https://phabricator.babeljs.io/T892: `Array.prototype.find` doesn't work in the runtime
- https://phabricator.babeljs.io/T6905: babel-plugin-transform-runtime: handle methods differently?

In our case, each time a function call an expected prototype method only existing in ES6, it fails in IE which does not yet implement most of then.

Conclusion: I will update this PR to use `babel-polyfill` when possible

noirbizarre:
And this article explain the difference between `babel-polyfill` and `babel-runtime`/`babel-runtime-transform-plugin`: https://medium.com/@jcse/clearing-up-the-babel-6-ecosystem-c7678a314bf3

davidbgk:
I'm surprised we have to import `babel-polyfill` in all these files, as I understand it that should only be required in root JS files?

noirbizarre:
Yes, it's imported only in root JS files, but we have one root file for admin + one for each view.

davidbgk:
Ah, makes sense now, I forgot that we have so many root files, my bad. I'm concerned about the fact that we may forgot to add that line for new root files, how can we prevent this?

noirbizarre:
@davidbgk we need to try on IE11 too. I'm using https://dev.windows.com/en-us/microsoft-edge/tools/vms/linux/ to do so.

If it works without, it meeds we don't use ES6 specifics so the forgot is not bad in this case.

Or, we can provide a template root file.

Perosnnaly, I think, we should as an objective, reduce the number of root files :)

davidbgk:
> Perosnnaly, I think, we should as an objective, reduce the number of root files :)

It depends, with HTTP/2 it's not a bad practice anymore to be modular ;)

noirbizarre:
@davidbgk Next step: HTTP2, Python 3, ipv6 ? :)

davidbgk:
Served by GNU Hurd :p

noirbizarre:
:)

After looking, switching to http2 is simply adding a keyword on NGinx, the real question is between reverse-proxy. I don't know if it's possible to have HTTP2 -> NGINX/HAProxy -> HTTP -> Application.

noirbizarre:
so, Can I merge this one ?

davidbgk:
r+, we have to remember that `.babelrc` is not anymore in use on the `dev` branch and update the `webpack.config` accordingly during the merge.

noirbizarre:
@davidbgk yep, it will conflict on merge as a reminder ;)

PR #369, opened 2016-02-24
fixing territory import

vinyll:
When using the _dev_ branch without _udata-gouvfr_ theme we get the following error:

```
web_1    | Traceback (most recent call last):
web_1    |   File "/usr/local/bin/udata", line 9, in <module>
web_1    |     load_entry_point('udata==0.1.0.dev', 'console_scripts', 'udata')()
web_1    |   File "/src/udata/commands/__init__.py", line 83, in console_script
web_1    |     run_manager()
web_1    |   File "/src/udata/commands/__init__.py", line 78, in run_manager
web_1    |     register_commands(manager)
web_1    |   File "/src/udata/commands/__init__.py", line 60, in register_commands
web_1    |     import udata.features.territories.commands
web_1    |   File "/src/udata/features/territories/commands.py", line 15, in <module>
web_1    |     from udata.models import (
web_1    | ImportError: cannot import name TERRITORY_DATASETS
```

Next we should extract Territory out of _udata core_ and get it into _udata-gouvfr_ theme

### DIFF
--- a/udata/features/territories/commands.py
+++ b/udata/features/territories/commands.py
@@ -12,11 +12,12 @@ from urllib import urlretrieve
 
 import requests
 
-from udata.models import (
-    Dataset, TERRITORY_DATASETS, ResourceBasedTerritoryDataset
-)
+from flask import current_app
+
+from udata.models import Dataset, ResourceBasedTerritoryDataset
 from udata.commands import submanager
 from udata.core.storages import logos, references, tmp
+
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


udata/features/territories/commands.py
```diff
@@ -12,12 +12,17 @@
...
-from udata.models import (
-    Dataset, TERRITORY_DATASETS, ResourceBasedTerritoryDataset
-)
+from udata.app import create_app
+from udata.models import Dataset, ResourceBasedTerritoryDataset
 from udata.commands import submanager
 from udata.core.storages import logos, references, tmp
 
+
+app = create_app()
```
vinyll: @noirbizarre, can you see a better way to get the app (using `standalone()` generates an _AssertionError_)

noirbizarre: You shouldn't. The app is already created by the script launcher: https://github.com/etalab/udata/blob/master/udata/commands/__init__.py#L72-L78
You should use `current_app.config` to access the config within a command.

noirbizarre:
r+ until proper solution ;)

PR #370, opened 2016-02-29
Improve the checksum migration

noirbizarre:
This PR improve the badchecksum detection migration.

It will need to unrecord the migration to apply it again (it's safe).

```shell
$ udata db unrecord udata 2016-01-25-fix-resources-checksums.js
-> Removing migration udata:2016-01-25-fix-resources-checksums.js

$ udata db migrate
-> udata:2015-05-06-migrate-issues.js ................................... [Skipped]
-> udata:2015-06-10-public-service-to-badges.js ......................... [Skipped]
-> gouvfr:2015-07-10-dataconnexions-badges.js ........................... [Skipped]
-> udata:2015-08-19-migrate-resource-published.js ....................... [Skipped]
-> gouvfr:2015-09-14-dataconnexions-6.js ................................ [Skipped]
-> udata:2015-09-14-generic-badges.js ................................... [Skipped]
-> udata:2015-09-23-community-resources.js .............................. [Skipped]
-> udata:2015-09-25-migrate-harvest-sources-validation.js ............... [Skipped]
-> gouvfr:2015-09-28-fix-maaf-harvester-doublouns.js .................... [Skipped]
-> udata:2015-10-08-migrate-resources-field-names.js .................... [Skipped]
-> udata:2015-10-19-deduplicate-badges.js ............................... [Skipped]
-> udata:2015-11-16-remove-supplier.js .................................. [Skipped]
-> udata:2015-11-23-set-uuid-as-community-resources-id.js ............... [Skipped]
-> udata:2016-01-21-set-uuid-for-community-resources.js ................. [Skipped]
-> udata:2016-01-21-set-uuid-for-resources.js ........................... [Skipped]
-> udata:2016-01-25-fix-resources-checksums.js .......................... [Apply]
│
│ Candidates: 47
│ 97 resources modified.
│ 47 datasets modified.
│
└──[OK]

-> udata:2016-02-03-clean-topic-datasets.js ............................. [Skipped]
```

### DIFF
--- a/udata/migrations/2016-03-02-fix-resources-checksums.js
+++ b/udata/migrations/2016-03-02-fix-resources-checksums.js
@@ -1,0 +1,36 @@
+/*
+ * Ensure all resources with checksum have a checksum type
+ *
+ * Same migration as udata:2016-01-25-fix-resources-checksums.js just optimized.
+ *
+ * Needed to reapply because of https://github.com/etalab/udata/pull/328
+ * not fixing the creation case  and data corruption was still happening
+ * until https://github.com/etalab/udata/pull/371
+ */
+
+var datasets = db.dataset.find({
+    resources: {$elemMatch: {
+        'checksum.value': {$exists: true},
+        'checksum.type': {$exists: false}
+    }}
+});
+
+print('Candidates: ' + datasets.count());
+
+var counter = 0;
+var count = 0;
+
+datasets.forEach(function(dataset) {
+    var resources = dataset.resources.map(function(resource) {
+        if (resource.checksum && resource.checksum.value && !resource.checksum.type) {
+            resource.checksum.type = 'sha1';
+            count++;
+        }
+        return resource;
+    });
+    var result = db.dataset.update({_id: dataset._id},
+                                   {$set: {resources: resources}});
+    counter += result.nModified;
+});
+print(count, 'resources modified.');
+print(counter, 'datasets modified.');


vinyll:
r+

davidbgk:
Why not a new migration?

noirbizarre:
@davidbgk because this is simply a fix to the previous one which is appliable unlimited of times. I didn't want to have the same migration to appear twice in every log and every migration process.
But I can submit a new one if you find it more explanatory or logical.

davidbgk:
@noirbizarre I think that we should only add new migrations to be able to understand what happened to the data without having to check revisions of migrations. Does it make sense?

noirbizarre:
I agree, but in this case, I find it more easy to understand that there is one migration to fix checksum than twi which reguires to read both PR to understand

davidbgk:
Well, I guess that if we still have an issue with checksums after that change we'll have to read both PR anyway ;)

noirbizarre:
yes, and will have to pass that migration again. Will we add a third same one ? ;)

davidbgk:
Definitely.

noirbizarre:
I duplicated the migration. Can I merge this ?

udata/migrations/2016-01-25-fix-resources-checksums.js
```diff
@@ -3,8 +3,10 @@
  */
 
 var datasets = db.dataset.find({
-    'resources.checksum.value': {$exists: true},
-    'resources.checksum.type': {$exists: false}
+    resources: {$elemMatch: {
```
davidbgk: Revert that part too?

udata/migrations/2016-03-02-fix-resources-checksums.js
```diff
@@ -0,0 +1,30 @@
+/*
+ * Ensure all resources with checksum have a checksum type
```
davidbgk: Maybe document why there is a need for a second migration?

udata/migrations/2016-03-02-fix-resources-checksums.js
```diff
@@ -0,0 +1,30 @@
+/*
```
davidbgk: I would avoid using the exact same name to avoid confusion ;)

davidbgk:
r+wc

noirbizarre:
I think there is a misunderstanding.
Like I said this is the same migration, I just optimized it (iter over less datasets) because I add to pass it again. This is not a fix of the previous migration (which worked perfectly).
The PR was here as a reminder that we have to apply it again.

If I revert the previous commit, I also copy the old migration because for the main point is we have to apply the same migration again.
For me, having 2 differents migrations will be misleading into the fact the first one didn't worked, but it did.

Or I can put this in the header:
```
Same migration as udata:2016-01-25-fix-resources-checksums.js just optimized.
Needed to reapply because of https://github.com/etalab/udata/pull/328 not fixing the case on creation and data corruption was still happening until https://github.com/etalab/udata/pull/371
```

By the way, for adding a new migration, I should have submitted it with #371 :/

If the rule is never touch, existing migrations, we have to discuss about `udata db unrecord` command which should not exist given this policy. If this is only for development purpose, I think manually removing the document should suffice. having a command doing this is exposing this action to the enduser saying "this is part of the process", no ?

davidbgk:
@noirbizarre let's talk about that next week :)

PR #371, opened 2016-03-01
Ensure non submitted nested properties (JSON) are not in the instanciated

noirbizarre:
This PR is the missing part of #328 and fixes the case of nested object (non)instanciation on creation when the data is optionnal and non present (main case: checksum on resources)


davidbgk:
r+

Issue #354, opened 2016-02-16
Add structured data (Schema, OGP, and so on)


davidbgk:
Search Engine Optimization & Social Integration Optimization

PR #372, opened 2016-03-03
Fix structured data (microdata) properties, refs #354

davidbgk:


udata/templates/dataset/display.html
```diff
@@ -52,11 +52,13 @@
 {% endblock %}
 
 {% block content %}
+<div itemscope="itemscope" itemtype="http://schema.org/Dataset" itemid="{{dataset.id}}">
```
noirbizarre: Why an extra div ?

davidbgk: We need to have all sections in it for complex relations (isPartOf, etc)

udata/templates/dataset/display.html
```diff
@@ -387,7 +389,9 @@ <h3 class="text-left">{{ _('Producer') }}</h3>
         <div class="row">
             <div class="col-sm-9 list-group resources-list smaller">
             {% for resource in dataset.community_resources %}
-                {% include theme('dataset/resource/list-item.html') %}
+                <div itemprop="hasPart">
```
noirbizarre: I think this break the markup defined on resource list item which is already using Dataset/distribution aka. DataDownload
https://github.com/etalab/udata/blob/master/udata/templates/dataset/resource/list-item.html

davidbgk: I tried with https://developers.google.com/structured-data/testing-tool/ and embedded concepts/properties looks to work as expected.

udata/templates/reuse/display.html
```diff
@@ -55,8 +55,13 @@
 {% endblock %}
 
 {% block content %}
+<div itemscope="itemscope" itemtype="http://schema.org/Dataset" itemid="{{ reuse.id }}">
```
noirbizarre: I don't think Dataset fits.
Maybe [CreativeWork](https://schema.org/CreativeWork) or one of its children, depending on the reuse type.

davidbgk: You're right.

udata/templates/dataset/display.html
noirbizarre: So it displays a Dataset/disbrutions[DataDownload] ? So I don't understand the point of hasPart

davidbgk: capture d ecran 2016-03-03 a 18 40 24

noirbizarre: OK, je tested, the real error is the resource item template missing the `itemprop="distribution"` here: https://github.com/etalab/udata/blob/master/udata/templates/dataset/resource/list-item.html#L3

This will probable conflict with `hasPart` but should provide a richier integration

davidbgk: And now I realize that references to example.com are inappropriate.

udata/templates/dataset/display.html
noirbizarre: OK. It does not break the style ?

udata/templates/dataset/display.html
noirbizarre: example.com, lol

Yes I know it validate but as you see, you only have the name with part.
With distributions (which is exactly the resource concept), you have:
screenshot-developers google com 2016-03-03 18-55-17

noirbizarre: Hum sorry, I didn't saw it was the Community Resources :/

udata/templates/dataset/display.html
```diff
@@ -327,9 +329,11 @@ <h3 class="text-left">{{ _('Producer') }}</h3>
                         </div>
 
                         {% for key, value in dataset.extras.items() %}
-                        <div itemprop="additionalProperty" itemscope itemtype="http://schema.org/PropertyValue">
-                            <meta itemprop="name" content="{{key}}"/>
-                            <meta itemprop="value" content="{{value.serialize()|to_json if value.serialize else value }}"/>
+                        <div itemscope itemtype="http://schema.org/PropertyValue">
+                            <div itemprop="additionalProperty">
```
noirbizarre: I think the `additionnalProperty` and its div are not needed anymore. We only want to have `name` and `value` defined on `PropertyValue`

noirbizarre:
I think `dataset/display.js` need an update too to follow naming change

udata/templates/dataset/display.html
davidbgk: So is that still relevant to consider it a distribution?

davidbgk:
I reverted the removal of `itemid` for the JS part and now it works :)

udata/templates/dataset/resource/list-item.html
```diff
@@ -1,6 +1,6 @@
 {% set resource_format = resource.format|trim|lower or 'data' %}
 <div id="resource-{{resource.id}}" class="list-group-item"
-    itemscope="itemscope" itemtype="http://schema.org/DataDownload"
+    itemscope itemtype="http://schema.org/DataDownload"
```
noirbizarre: needs `itemprop="distribution"` to be properly attached to the dataset

davidbgk: Done.

noirbizarre:
r+wc

PR #373, opened 2016-03-07
Reuse a unique fake StringIO file for CSV generation

davidbgk:


noirbizarre:
r+ if you're sure that `seek(0) + truncate(0)` is lighter/faster than a new StringIO instance.

davidbgk:
Actually, it's [not faster](http://stackoverflow.com/a/4330829), I'll update with an explicit `del` instead ;)

davidbgk:
I might be onto something, [with croquemort](https://github.com/davidbgk/croquemort/pull/7) it's definitely better to use the truncate/seek pattern otherwise the generator is hanging prior to streaming. Checking.

davidbgk:
The results are almost the same so I let the explicit `StringIO` creation but I added the `del` because I can't test that for sure.

PR #382, opened 2016-03-14
Deal with (geo)zones with the same postal code

davidbgk:


udata/features/territories/__init__.py
```diff
@@ -13,13 +13,13 @@ def check_for_territory(query):
...
     if len(query) == 5 and query.isdigit():
-        # First check INSEE code then postal code.
+        # First check INSEE code then the first postal code.
         try:
             return GeoZone.objects.get(code=query, level='fr/town')
         except GeoZone.DoesNotExist:
             try:
-                return GeoZone.objects.get(
-                    keys__postal__contains=query, level='fr/town')
+                return GeoZone.objects.filter(
```
noirbizarre: The `filter` is optionnal (I know you like the Django way ;))

To avoid long lines, you can factorize:
```python
qs = GeoZone.objects(level=fr/town)
try:
    return qs.get(code=query)
except GeoZone.DoesNotExist:
    try:
        return qs(keys__postal__contains=query).first()
    except:
        #...
```

noirbizarre:
r+wc

udata/features/territories/__init__.py
noirbizarre: or if you want to keep the filter:

```python
qs = GeoZone.objects.filter(level=fr/town)
try:
    return qs.get(code=query)
except GeoZone.DoesNotExist:
    try:
        return qs.filter(keys__postal__contains=query).first()
    except:
        #...
```

noirbizarre:
great ! r+ ;)

PR #387, opened 2016-03-16
From territories to towns, fix #386

### LINKED ISSUE #386: Improve territories after demo
* [x] deal with duplicates
* [ ] better wording
* [x]  fix typos
* [ ] ~~put data from existing towns on top~~ → next iteration
* [x] better order of datasets
* [x] add licenses and retro-actions(?) to/from wikipedia
* [x] more information within typeahead dropdown
* [x] change email address

davidbgk:


js/territory/typeahead.js
```diff
@@ -35,7 +36,7 @@ define([
         source: engine,
         display: 'name',
         templates: {
-            header: header({title: i18n._('Territory')}),
+            header: header({title: i18n._('Territories')}),
```
davidbgk: Towns?

udata/features/territories/__init__.py
```diff
@@ -14,38 +14,11 @@ def check_for_territory(query):
...
-                    # Warning: it means that we're dropping some territories
-                    # here, e.g.: 62760 vs. 62814 vs. 80756.
-                    geozones = qs(keys__postal__contains=query)
-                    geozones = geozones.order_by('-population', '-area')
-                    return geozones.first()
-            except GeoZone.DoesNotExist:
-                return False
+        # Match both INSEE and postal codes
+        qs = qs(db.Q(code=query) | db.Q(keys__postal__contains=query))
+        return qs.order_by('-population', '-area')
```
davidbgk: Useless?

udata/templates/search.html
```diff
@@ -52,11 +52,11 @@
             </a>
         </li>
     {% endif %}
-    {% if territory %}
+    {% if territories %}
         <li>
-            <a href="{{ url_for('territories.territory', territory=territory) }}">
-                {{ _('Territory:') }}
-                <span class="badge">{{ territory.name }}</span>
+            <a href="#territories" data-toggle="tab">
```
davidbgk: I think we should only have a tab when there are more than one result.

udata/core/search/views.py
```diff
@@ -37,5 +37,6 @@ def render_search():
         results_labels = ['datasets', 'reuses', 'organizations', 'users']
     results = search.multiquery(*search_queries)
     context = dict(zip(results_labels, results))
-    context['territory'] = check_for_territory(params.get('q'))
+    territories = check_for_territories(params.get('q'))
+    context['territories'] = territories.paginate(1, search.DEFAULT_PAGE_SIZE)
```
davidbgk: Doesn't work when territories are not activated.

js/territory/typeahead.js
```diff
@@ -9,6 +9,7 @@ define([
     'templates/search/suggestion.hbs',
     'i18n'
 ], function($, API, Bloodhound, header, suggestion, i18n) {
+    const MAX = 2;
```
davidbgk: Maybe extend to more than 2 (5?) given that there are more cases than I thought.

noirbizarre: 5 is way too much (we limit dataset to 3). So maybe 3 ?

udata/templates/search.html
noirbizarre: OK, I will mix the two solution: 1 result, the tab redirects to the territory page, 2 and more, we list results.

udata/templates/search.html
```diff
@@ -52,12 +52,20 @@
...
         </li>
     {% endif %}
-    {% if territory %}
+    {% if territories %}
         <li>
-            <a href="{{ url_for('territories.territory', territory=territory) }}">
-                {{ _('Territory:') }}
-                <span class="badge">{{ territory.name }}</span>
-            </a>
+            {% set total=territories.count() %}
```
davidbgk: Maybe we should extend that logic to other tabs?

PR #390, opened 2016-03-18
Fix datasetless communautary resources deletion

noirbizarre:
This PR:

- [x] add the purge command (avoiding to launch each purge job manually)
- [x] force admin modal display for communautary resources without dataset (has been deleted)


davidbgk:
r+

PR #392, opened 2016-03-18
Fix the harvester list on user and org admin pages

noirbizarre:
Ensure that the harvester listing does not fetch all harvesters on user and org admin page.


davidbgk:
r+

PR #394, opened 2016-03-21
Geospatial and API localization

noirbizarre:
This pull-request:

- [x] internationalize spatial granularities and territories
- [x] ensure Admin API calls are localized
- [x] ensure spatial APIs and completer are localized (ie .World vs "Monde" for french localization)

### DIFF
--- a/js/api.js
+++ b/js/api.js
@@ -3,6 +3,30 @@ import SwaggerClient from 'swagger-client';
 import config from 'config';
 import log from 'logger';
 
+
+function updateQueryParam(uri, key, value) {
+    const re = new RegExp(`([?&])${key}=.*?(&|$)`, 'i');
+    const separator = uri.indexOf('?') !== -1 ? '&' : '?';
+    if (uri.match(re)) {
+        return uri.replace(re, `$1${key}=${value}$2`);
+    } else {
+        return `${uri}${separator}${key}=${value}`;
+    }
+}
+
+/**
+ * This object inject the lang query parameter in every API request
+ */
+const LangParameter = function(name) {
+  this.name = name;
+};
+
+LangParameter.prototype.apply = function(obj) {
+    obj.url = updateQueryParam(obj.url, 'lang', config.lang);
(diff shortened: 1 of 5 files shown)


js/api.js
```diff
@@ -3,6 +3,30 @@ import SwaggerClient from 'swagger-client';
 import config from 'config';
 import log from 'logger';
 
+
+function updateQueryParam(uri, key, value) {
+    const re = new RegExp(`([?&])${key}=.*?(&|$)`, 'i');
```
davidbgk: Add comments for both the function and the regexp ;)

udata/core/spatial/api.py
```diff
@@ -67,7 +68,7 @@ def get(self):
         return [
             {
                 'id': opt['text'],
-                'name': opt['payload']['name'],
+                'name': _(opt['payload']['name']),
```
davidbgk: Wow, does it work dynamically like this?! I'm impressed :+1:

udata/core/spatial/translations.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+# Here to force translations by gettext
+_ = lambda s: s
+
+TRANSLATIONS = (
```
davidbgk: Oh OK I get it, so it's not exactly dynamic. How can we maintain that list? Is there a way to check that we have all entries with a command?

davidbgk:
r+wc

udata/core/spatial/api.py
noirbizarre: It works only if the string are in the po/mo files.

udata/core/spatial/translations.py
noirbizarre: There was two alternative: the easy one in which we manually maintain a label list in uData.
The other one is to import translations from GeoZones.
I add to manually test the translations (on card) to check that every string is translated.

PR #395, opened 2016-03-21
Fix bad reuse-discussion route name

noirbizarre:
This PR fix wrong and duplicate route name for `reuse-discussion` in admin


davidbgk:
r+

PR #397, opened 2016-03-21
Improve Croquemort error handling

noirbizarre:
This PR use `requests.ConnectTimeout` in addition to `requests.ConnectionError` in croquemort error handling to catch all raised errors including timeouts and add a logging to sentry in each case.

### DIFF
--- a/udata/core/dataset/croquemort.py
+++ b/udata/core/dataset/croquemort.py
@@ -1,12 +1,20 @@
 import json
+import logging
 import time
 
 import requests
 from flask import current_app
 
+log = logging.getLogger(__name__)
+
 # See http://docs.python-requests.org/en/latest/user/advanced/#timeouts
 # We are waiting 3 sec for the connexion and 9 for the response.
 TIMEOUT = (3.1, 9.1)
+
+CONNECTION_ERROR_MSG = 'Unable to reach the URL checker'
+
+ERROR_LOG_MSG = 'Unable to connect to croquemort'
+TIMEOUT_LOG_MSG = 'Timeout connectin to Croquemort'
 
 
 def check_url(url, group=None):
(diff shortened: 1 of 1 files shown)


udata/core/dataset/croquemort.py
```diff
@@ -64,8 +64,8 @@ def check_url_from_cache(url, group=None):
         response = requests.get(retrieve_url,
                                 params={'url': url, 'group': group},
                                 timeout=TIMEOUT)
-    except requests.ConnectionError:
-        return {'error': 'URL checker not found'}, {}
+    except requests.RequestException:
```
davidbgk: I think that the `ConnectionError` is more specific here and that we should keep it to make the difference between the explicit checked not found and other errors.

udata/core/dataset/croquemort.py
```diff
@@ -85,8 +85,8 @@ def check_url_from_group(group):
         response = requests.get(retrieve_url,
                                 params={'group': group},
                                 timeout=TIMEOUT)
-    except requests.ConnectionError:
-        return {'error': 'URL checker not found'}, {}
+    except requests.RequestException:
```
davidbgk: Same here.

davidbgk:
r+wc

udata/core/dataset/croquemort.py
noirbizarre: The "not found" case is tested on the line below (`response.status_code == 404`).
As the `timeout` parameter is given as parameter, `TimeoutException` can be raised.
Am I wrong ?

udata/core/dataset/croquemort.py
noirbizarre: Same here ;)

udata/core/dataset/croquemort.py
davidbgk: A 404 is returned when the URL is not found on the server, not when the server is not up and running. I think that we should differentiate `ConnectionError` and `ReadTimeout` (and maybe fallback on `RequestException`?). The first one happens when there is a server error, the second when there is a network error. Having that information can help us debug the issue more fast.

noirbizarre: I know, but 404 don't raise exception and is handled just the line after

davidbgk:
r+

PR #398, opened 2016-03-22
Synchronize dev with master

davidbgk:


noirbizarre:
Just needs the last translations and then r+ ;)

PR #401, opened 2016-03-22
Better territories

davidbgk:
This is a 3 steps process:

First the user is not logged in

capture d ecran 2016-03-22 a 14 42 27

Then we encourage him to publish datasets for that zone

capture d ecran 2016-03-22 a 14 43 07

Finally ask him to claim that town to be referenced first

capture d ecran 2016-03-22 a 14 43 36

Labels and styles must be updated but it's more a discussion on the workflow ATM.


davidbgk:
Don't forget to review the gouvfr counterpart too for tests: https://github.com/etalab/udata-gouvfr/pull/97

udata/core/organization/commands.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import logging
+
+from udata.commands import manager
```
noirbizarre: better use a submanager for namespacing on organization.
See: https://github.com/etalab/udata/blob/master/udata/harvest/commands.py#L10-L17

udata/core/organization/commands.py
```diff
@@ -0,0 +1,34 @@
...
+from udata.models import GeoZone, Organization
+
+log = logging.getLogger(__name__)
+
+
+@manager.command
+def attach_zone(insee_code, organization_id_or_slug, level='fr/town'):
```
noirbizarre: Optionnal: limit level values to granularities, with `@manager.option(choices=...)` (or enum)

udata/core/user/models.py
```diff
@@ -169,6 +169,12 @@ def generate_api_key(self):
     def clear_api_key(self):
         self.apikey = None
 
+    def can_modify(self, datasets):
```
noirbizarre: can work on reuses too, so maybe generalize `dataset` to `model`

noirbizarre: In fact, there is the dataset permission module that centralize these rules, better use it.

noirbizarre:
r+wc

udata/core/organization/commands.py
davidbgk: It introduces a bit of complexity, let's do that when we'll introduce more levels.

noirbizarre:
r+

Issue #399, opened 2016-03-22
Better territories


davidbgk:
Update the current territories from users’ feedback.

PR #402, opened 2016-03-22
Add territories (towns) to the sitemap, refs #399

davidbgk:


noirbizarre:
r+

davidbgk:
I screwed up here: the sitemap should only be activated when territories are activated and so tests should move to -gouvfr, it will be fixed in https://github.com/etalab/udata/pull/401

(It has no incidence on the production given that we have activated territories.)

PR #406, opened 2016-03-23
Fix harvester deletion

noirbizarre:
This PR ensure that deleted harvest sources are not shown in the API listing (and so in the admin) and display a badge on deleted sources pages

### DIFF
--- a/udata/harvest/actions.py
+++ b/udata/harvest/actions.py
@@ -30,9 +30,10 @@ def list_backends():
 
 def list_sources(owner=None):
     '''List all harvest sources'''
+    sources = HarvestSource.objects.visible()
     if owner:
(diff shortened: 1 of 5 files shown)


udata/harvest/actions.py
```diff
@@ -30,9 +30,10 @@ def list_backends():
 
 def list_sources(owner=None):
     '''List all harvest sources'''
+    sources = HarvestSource.objects(deleted=None)
```
davidbgk: Maybe we should add a `visible()` method for consistency?

davidbgk:
r+wc

udata/harvest/actions.py
noirbizarre: Yes, I will ASAP

PR #407, opened 2016-03-23
Ensure tags are unique in dataset tag test

noirbizarre:
This PR fix this annoying randomly failing test du to words collisions.


davidbgk:
r+

PR #409, opened 2016-03-24
Admin forms

noirbizarre:
This pull request is proposal for an admin improvement:
- full page forms
- new page menu
- display some missing data 

screenshot-admin-dataset-new

screenshot-www data dev 2016-03-24 11-52-19

screenshot-admin-dataset-form

Pages to migrate:

- [x] Dataset page and form
- [x] Organisation page and form
- [x] Reuse page and form
- [x] Harvester view and form
- [x] Topic page and form
- [x] Profile page and forms
- [x] Post page and form

### DIFF
--- a/js/views/dataset.vue
+++ b/js/views/dataset.vue
@@ -159,9 +159,7 @@ export default {
         },
         territories_labels() {
             if (this.geojson && this.geojson.features) {
-                return this.geojson.features.map(function(feature) {
-                    return feature.properties.name;
-                }).join(', ');
+                return this.geojson.features.map(feature => feature.properties.name).join(', ');
             }
         },
         map_footer() {
--- a/js/views/dataset.vue
+++ b/js/views/dataset.vue
@@ -58,6 +58,10 @@ export default {
     name: 'DatasetView',
     data: function() {
         var actions = [{
+                label: this._('Edit'),
+                icon: 'edit',
+                method: this.edit
+            },{
                 label: this._('Transfer'),
                 icon: 'send',
                 method: this.transfer_request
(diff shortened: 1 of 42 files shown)


davidbgk:
Looks good so far! I think we need to iterate on buttons placement but let's do that IRL next week ;)

noirbizarre:
I agree, there's still somm rooms for improvement on the header.
The button placement is not optimal.
I'd like to put it on the top right and move le site on site elsewhere

noirbizarre:
Proposal:

screenshot-www data dev 2016-03-24 15-01-10

davidbgk:
Better! Can we unstack the button on the right to a bar without any dropdown?

noirbizarre:
I think this is too large. When there is only one action, there is no dropdown.
right now, extra actions are:
- delete
- transfert
- admin actions (badges...)

js/views/dataset.vue
```diff
@@ -131,13 +156,17 @@ export default {
                 color: 'purple',
                 target: '#trafic'
             }];
+        },
+        territories_labels() {
+            if (this.geojson && this.geojson.features) {
+                return this.geojson.features.map(function(feature) {
```
davidbgk: `feature => feature.properties.name` ;)

noirbizarre: Oh, a shortcut, I'll try !

noirbizarre:
Added scroll-to submenu (only one active, scroll to a given section):

screenshot-www data dev 2016-03-29 17-15-11

davidbgk:
r+

PR #410, opened 2016-03-24
Fix datepicker

noirbizarre:
This pull request fix date pickers clic outside detection.
Also fix the daterange-picker current calendar value display.


davidbgk:
Nice, r+

PR #413, opened 2016-03-25
Downgrade Vue.js to 1.0.17

noirbizarre:
In Vue.js 1.0.18 `PageList` pagination doesn't work as expected: the datatable widget doesn't detect the page changed and is locked on the first page (used on resource list and harvest job list).

This PR downgrade to Vue.js 1.0.17 where it works fine.


davidbgk:
r+ ;)

PR #414, opened 2016-03-25
Fix unexplained freezes

noirbizarre:
Don't know why, but it makes the UI freeze !!


davidbgk:
I guess it's some kind of double interpolation of the string. r+

PR #422, opened 2016-04-04
Better titles for towns, fix #421

### LINKED ISSUE #421: Better titles for towns
## Detailed Description

The current title is confusing with “Town name (postal code)” because some users consider the postal code as an INSEE code.

## Context

When I display a town page, I want to identify which one is that from duplicates, so I can be assured that I consult the one I'm looking for.

## Possible Implementation

We'll try “Town name (department)” which is more explicit.

davidbgk:
Update typeahead results:

capture d ecran 2016-04-04 a 09 27 23

Page details:

capture d ecran 2016-04-04 a 09 27 37

And search results:

capture d ecran 2016-04-04 a 11 26 40

### DIFF
--- a/udata/frontend/markdown.py
+++ b/udata/frontend/markdown.py
@@ -49,25 +49,27 @@ class UDataMarkdown(object):
         return Markup(html)
 
 
+def mdstrip(value, length=None, end='…'):
+    '''
+    Truncate and strip tags from a markdown source
+
+    The markdown source is truncated at the excerpt if present and
+    smaller than the required length. Then, all html tags are stripped.
+    '''
+    if not value:
+        return ''
+    if EXCERPT_TOKEN in value:
+        value = value.split(EXCERPT_TOKEN, 1)[0]
+    rendered = md(value)
+    text = do_striptags(rendered)
+    if length > 0:
+        text = do_truncate(text, length, end=end)
+    return text
+
+
 def init_app(app):
     parser = CommonMark.DocParser  # Not an instance because not thread-safe(?)
     renderer = CommonMark.HTMLRenderer()
     app.extensions['markdown'] = UDataMarkdown(app, parser, renderer)
 
-    @app.template_filter()
-    def mdstrip(value, length=None, end='…'):
-        '''
-        Truncate and strip tags from a markdown source
-
-        The markdown source is truncated at the excerpt if present and
-        smaller than the required length. Then, all html tags are stripped.
-        '''
-        if not value:
-            return ''
-        if EXCERPT_TOKEN in value:
-            value = value.split(EXCERPT_TOKEN, 1)[0]
-        rendered = md(value)
-        text = do_striptags(rendered)
-        if length > 0:
-            text = do_truncate(text, length, end=end)
-        return text
+    app.add_template_filter(mdstrip)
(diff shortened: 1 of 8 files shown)


udata/templates/search.html
```diff
@@ -78,6 +80,7 @@
 {% endblock %}
 
 {% block toolbar_container %}
+{% if datasets %}
```
noirbizarre: Toolbars are always rendered, just not visible.
The current visible is linked in JS in https://github.com/etalab/udata/blob/master/js/search.js#L16-L27

noirbizarre: OK, can be removed as dataset tab is removed too. But the tab activation should still be triggered

davidbgk: New version, I verified and the toolbar changes accordingly.

udata/frontend/markdown.py
```diff
@@ -49,25 +49,27 @@ def __call__(self, stream):
...
-        if not value:
-            return ''
-        if EXCERPT_TOKEN in value:
-            value = value.split(EXCERPT_TOKEN, 1)[0]
-        rendered = md(value)
-        text = do_striptags(rendered)
-        if length > 0:
-            text = do_truncate(text, length, end=end)
-        return text
+    app.template_filter()(mdstrip)
```
noirbizarre: I think you can write: `app.add_template_filter(mdstrip)` (which the decorator wraps)

udata/templates/search.html
```diff
@@ -134,7 +138,15 @@
 {% block main_content %}
 <div class="tab-content">
 
-    <div class="tab-pane row active" id="datasets">
+    {% if not datasets and not reuses and not organizations and not users and not territories %}
```
noirbizarre: When not any results, the tabs are still displayed. Doesn't it break the layout ?

udata/frontend/markdown.py
davidbgk: Nop, I can't ;)

The decorator needs to be instanciated because it accepts parameters (there is a way to avoid that but that's not our code).

udata/templates/search.html
davidbgk: It looks fine

capture d ecran 2016-04-04 a 11 14 01

I wasn't sure about the wording though.

noirbizarre: Yes, the "typos" is very developper centric I think :)

udata/frontend/markdown.py
noirbizarre: It's not the same function in my example:
- `app.template_filter()(mdstrip)`  the decorator currently
- `app.add_template_filter(mdstrip)` the method to add (my suggestion)

udata/templates/search.html
davidbgk: Not sure, it's mentioned on the [wikipedia page](https://en.wikipedia.org/wiki/Typographical_error)

udata/frontend/markdown.py
davidbgk: Oh I read it too faaast, testing!

udata/templates/search.html
noirbizarre: "fix your query" ?

noirbizarre: OK, :+1:

noirbizarre:
r+wc

vinyll:
:cool:

PR #423, opened 2016-04-04
Tasks cleanup

noirbizarre:
This pull request cleanup some dev tasks:
- remove unused tasks
- removed useless `nrun()` helper
- use `npm run` for every node related task
- fix lint tasks
- fix lint rules (ESLint 2.0 and Flake8)
- make flake8 apply `# noqa` for PEP8 rules too
- removed some useless files
- added missing translations in Vue.js bindings
- added the report dir to `.gitignore`


davidbgk:
Why not against `master`? That way the documentation can be improved accordingly ;)

noirbizarre:
OK, I will resubmit ont master

PR #425, opened 2016-04-04
Dataset Vue.js

noirbizarre:
This PR migrate the dataset view to Vue.js and jQuery free.

This is a proposal for a migration pattern:
- expose `fetch()` through a custom api plugin (with preconfigured settings)
- reuse some components (leaflet-map is extracted from map widget)
- port modals from handlebars templates to Vue.js components
- use a minimalist microdata module
- use Velocity.js instead of jQuery.animate
- use balloon-css instead of bootstrap tooltips
- module is `front/dataset`

Legacy jQuery components are left untouched (allowing to migrate them one by one later)


js/front/dataset/index.js
```diff
@@ -0,0 +1,224 @@
+/**
+ * Dataset display page JS module
+ */
+// Catch all errors
+import 'raven';
+
+// ES6 environment
+import 'babel-polyfill';
+
+import $ from 'jquery'; // Only needed for tooltips :(
```
davidbgk: May I suggest https://github.com/kazzkiq/balloon.css here? (once tested with embeds maybe?)

davidbgk: There is `this.$http` too that we can replace with [fetch](https://github.com/github/fetch)

js/helpers/text.js
```diff
@@ -1,29 +1,42 @@
...
-    'use strict';
+/**
+ * Ellipsis a text given a length.
+ * @param  {string} text   The text to truncate
+ * @param  {int} length The truncate length
+ * @return {string}        The truncated text
+ */
+export function truncate(text, length) {
+    if (text && text.length > length) {
+        return text.substr(0, length - 3) + '...';
```
davidbgk: `…` instead of `...`?

js/microdata.js
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Match  and extract microdata of a given itemtype.
+ * @param  {String} itemtype    The itemtype URL
+ * @param  {HTMLElement} el     An optionnal root element to restrict the search to
+ * @return {Array}              A list of extracted object
+ */
+function microdata(itemtype, el) {
+    el = el || document;
+    const els = el.querySelectorAll(`[itemscope][itemtype="${itemtype}"]`);
+    return [...els].map(function(el) {
```
davidbgk: `.map(extract)` should work here ;)

udata/settings.py
```diff
@@ -104,6 +104,8 @@ class Defaults(object):
     #     'delay': 1,
     #     'retry': 10,
     # }
+    #
+    CROQUEMORT_WHITELIST = []
```
davidbgk: What do we plan to add here?

js/front/dataset/index.js
noirbizarre: I think `this.$http` is not using fetch but it provide good vue.js integration (promise, loading state...)

noirbizarre: And :+1: for balloon.css, thanks for the tip by the way.

js/helpers/text.js
noirbizarre: Yes, good one !

js/components/leaflet-map.vue
```diff
@@ -0,0 +1,75 @@
...
+        L.tileLayer(TILES_URL, TILES_CONFIG).addTo(this.map);
+    },
+    watch: {
+        'geojson': function(json) {
+            if (json) {
+                this.layer = L.geoJson(json, {
+                    onEachFeature: (feature, layer) => {
+                        if (this.popup && feature.properties && feature.properties.name) {
+                            layer.bindPopup(feature.properties.name);
+                            layer.on('mouseover', function() {
```
davidbgk: Arrows here and below?

js/front/dataset/index.js
```diff
@@ -0,0 +1,224 @@
...
+                this.$refs.map.geojson = response.data;
+            });
+        },
+
+        /**
+         * Asynchronuously check all resources status
+         */
+        checkResources() {
+            if (config.check_urls) {
+                this.dataset.resources.forEach(resource => this.checkResource(resource));
```
davidbgk: `forEach(this.checkResource)` might work.

js/plugins/api.js
```diff
@@ -0,0 +1,25 @@
...
+        Vue.http.options.root = config.api_root.substring(0, config.api_root.length - 1);
+    } else {
+        Vue.http.options.root = config.api_root;
+    }
+
+    ['post', 'put', 'patch', 'delete'].forEach(method => {
+        Vue.http.headers[method]['X-CSRFToken'] = config.csrftoken;
+    });
+
+    // Vue.http.interceptors.push({
```
davidbgk: Dead code?

davidbgk:
I use to put my `preventDefault`s sooner in the function but we can discuss that tomorrow. r+wc

js/front/dataset/index.js
```diff
@@ -203,20 +203,25 @@ new Vue({
...
                         }
-                    }, response => {
-                        if (response.status !== 503) {
-                            el.classList.add('format-label-danger');
-                            addTooltip(el, this._('The server cannot be found.'));
-                        }
-                    });
+                    })
+                    // May need some handling.
+                    .catch(error => {});
```
davidbgk: At least `.catch(console.log.bind(console))`.

js/plugins/api.js
```diff
@@ -1,25 +1,173 @@
...
+
+        /**
+         * Perform a POST upload operation (form/multipart)
+         * @param  {String} url     An absolute or API root relative URL
+         * @param  {Object} data    The body payload
+         * @param  {Object} headers Optionnal HTTP headers
+         * @return {Promise}        An instanciated fetch promise
+         */
+        upload(url, data, headers) {
+            const body = data;
```
davidbgk: Why?

js/plugins/api.js
```diff
@@ -1,25 +1,173 @@
...
+         * @param  {Object} data    The body payload
+         * @param  {Object} headers Optionnal HTTP headers
+         * @return {Promise}        An instanciated fetch promise
+         */
+        put(url, data, headers) {
+            return fetch(this.build(url), {
+                method: 'put',
+                credentials: 'include',
+                headers: Object.assign({}, DEFAULT_HEADERS, WRITE_HEADERS, headers || {}),
+                body: JSON.stringify(DATA)
```
davidbgk: Not consistent with `post`.

js/plugins/api.js
```diff
@@ -1,25 +1,173 @@
...
+         * @param  {String} url     An absolute or API root relative URL
+         * @param  {Object} headers Optionnal HTTP options
+         * @return {Promise}        An instanciated fetch promise
+         */
+        delete(url, headers) {
+            return fetch(this.build(url), {
+                method: 'delete',
+                credentials: 'include',
+                headers: Object.assign({}, DEFAULT_HEADERS, WRITE_HEADERS, headers || {}),
+            }).then(this.jsonHandler);
```
davidbgk: Not needed, should be empty.

js/plugins/api.js
noirbizarre: API use JSON everywhere, upload is a special case using `multipart/form-data` content type.
I just don't have implented the body yet.

js/components/leaflet-map.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+<div></div>
+</template>
+<script>
+import L from 'leaflet';
+
+const ATTRIBUTIONS = `&copy;<a href="http://openstreetmap.org">OpenStreetMap</a>
+                      /<a href="http://open.mapquest.com/">MapQuest</a>`;
+const TILES_PREFIX = location.protocol === 'https:' ? '//otile{s}-s' : '//otile{s}';
```
davidbgk: Maybe it's time to force `https`?

js/config.js
```diff
@@ -69,6 +69,11 @@ export const csrf_token = _meta('csrf-token');
 export const check_urls = _jsonMeta('check-urls');
 
 /**
+ * List of whitelisted domain for url check
+ */
+export const check_urls_whitelist = _jsonMeta('check-urls-whitelist');
```
davidbgk: `whitelist` or `ignorelist`?

davidbgk:
Is the empty `js/front/discussions/thread.vue` file necessary?

js/helpers/text.js
```diff
@@ -1,29 +1,42 @@
...
-            }
-            return text;
-        },
+/**
+ * Titleize a string
+ * @param  {string} text  The input text to transform
+ * @return {string}       The titleized string
+ */
+export function title(text) {
+    return text.replace(/\w\S*/g, function(txt) {
```
davidbgk: Arrow?

js/microdata.js
```diff
@@ -0,0 +1,104 @@
+/*
```
davidbgk: Nice module :+1:

js/plugins/api.js
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    /**
+     * Encode paramaters as Object to querystring
+     * @param  {Object} params The parameters to encode
+     * @return {String}        The URL encoded querystring
+     */
+    encodeParams(params) {
+        return Object.keys(params).map(key => {
+            return [key, params[key]].map(encodeURIComponent).join('=');
```
davidbgk: That `return` is probably unnecessary.

js/components/leaflet-map.vue
noirbizarre: :+1:

js/plugins/api.js
```diff
@@ -0,0 +1,213 @@
...
+
+    /**
+     * Perform a POST upload operation (multipart/form-data)
+     * @param  {String} url     An absolute or API root relative URL
+     * @param  {Object} data    The body payload
+     * @param  {Object} headers Optionnal HTTP headers
+     * @return {Promise}        An instanciated fetch promise
+     */
+    upload(url, data, headers) {
+        const body = data; // TODO handle body
```
davidbgk: From another project, I have this:
```
  this.on('file:new', (file) => {
    const reader = new FileReader()
    reader.onloadend = event => {
      if (event.target.readyState === FileReader.DONE) {
        fetch(this.filesUrl, {
          method: 'post',
          headers: {
            'Content-Type': file.type,
            'Content-Length': file.size
          },
          body: event.target.result
        })
        .then(response => response.headers.get('Location'))
        .then(location =>
          RiotControl.trigger('form:add_file', file, location))
      }
    }
    reader.readAsArrayBuffer(file)
  })
```

js/plugins/api.js
```diff
@@ -0,0 +1,213 @@
...
+
+    /**
+     * Default handler for JSON response
+     * Parse the response if OK otherwise throw it
+     * (let the developper handle it in the catch() handler)
+     * @param  {Response} response A response with JSON expected body.
+     * @return {Promise}           A promise for data handling
+     */
+    onResponse(response) {
+        if (response.ok) {
```
davidbgk: Are you sure that `response.ok` deal with "invalid" status codes, the [fetch polyfill documentation](https://github.com/github/fetch#handling-http-error-statuses) is quite explicit on the fact that you have to do that by hand(!)

davidbgk:
r+wc, that's a big one :sweat:

js/config.js
noirbizarre: Good question. But you're right, `ignorelist` is more neutral :+1:

js/front/dataset/index.js
```diff
@@ -0,0 +1,225 @@
...
+        /**
+         * Display a resource in a modal
+         */
+        showResource(id, e) {
+            // Ensure edit button work
+            if ([e.target, e.target.parentNode].some((el) => {el.classList.contains('btn-edit');})) {
+                return;
+            }
+            e.preventDefault();
+            const resource = this.dataset.resources.filter(resource => resource.id === id)[0];
```
noirbizarre: OMG ! Thanks !

js/plugins/api.js
noirbizarre: Yes:
https://developer.mozilla.org/en-US/docs/Web/API/Response/ok
https://github.com/github/fetch/pull/82
And there is a test case.

js/plugins/api.js
noirbizarre: Interesting.
We also have FineUploader as dependency, used both in front and admin.
It was chosen for resumable chunked upload but this feature is not yet used (should allow big file upload).

I think upload experience is a key feature that need it's own PR (and right now it's only used in admin)
In this PR, I haven't the need for upload method so I left a TODO. Maybe I should throw an error instead or remove the method

js/plugins/api.js
noirbizarre: nop :cry:.  I just tested and `return` is mandatory using the `{}` form.

davidbgk: then shave those ugly mustaches! :D

js/plugins/api.js
davidbgk: :+1: for removing it for now if it's not in use.

PR #426, opened 2016-04-04
Python lints

noirbizarre:
Fix flake8 lints.

Also fix the `TERRITORY_DATASETS` not working without `udata-gouvfr`.
See https://github.com/etalab/udata-gouvfr/pull/99 for the `udata-gouvfr` counterpart.


udata/templates/macros/metadata.html
```diff
@@ -12,7 +12,7 @@
 <meta name="description" content="{{ description }}" />
 <meta property="og:description" content="{{ description }}" />
 <meta property="og:image" content="{{ image }}" />
-<meta name="keywords" content="{{ ', '.join(current_site.keywords + (keywords)) }}" />
+<meta name="keywords" content="{{ ', '.join(current_site.keywords + keywords) }}" />
```
davidbgk: Be careful here, did you try both cases where there is a list and a unique item for `keywords`?

davidbgk:
Yay for dropping `lambda`s in the process :+1: r+wc

udata/templates/macros/metadata.html
noirbizarre: keyword**s** should always be an array. It defaults to an empty array (https://github.com/noirbizarre/udata/blob/python-lints/udata/templates/macros/metadata.html#L5).

If it's not the case, a coma is missing because `(x) = x` and it should be `(keywords, )` but in this case it won't work with list :/

noirbizarre:
Yes but sometimes, mostly in tests, it's easier to read lambdas so I just put a `# noqa`

PR #427, opened 2016-04-04
Make upload file argument required

noirbizarre:
This PR ensure that the `file` argument is submited on file file upload:
- no more server error when not submited
- understandable error message for API users


davidbgk:
r+

PR #429, opened 2016-04-04
Add rollover for embeds, fix #428

### LINKED ISSUE #428: Add rollover for embeds
## Detailed Description

Some icons are hard to understand when you're viewing a rendered embed on an external page.

## Context

When I move my mouse on top of an embed dataset, I want to get extra infos, so I can better understand what it means.

## Possible Implementation

Using https://github.com/kazzkiq/balloon.css

davidbgk:
territoires


noirbizarre:
r+

PR #436, opened 2016-04-09
User registration form (to replace YouCKAN)

grischard:
This uses flask-security and babel, confirms email addreses, and adapts the YouCKAN templates.

### DIFF
--- a/udata/auth/__init__.py
+++ b/udata/auth/__init__.py
@@ -3,18 +3,19 @@ from __future__ import unicode_literals
 
 import logging
 
-from flask import render_template, current_app, Blueprint, request, redirect
-from flask.ext.principal import (  # noqa
-    Permission as BasePermission, PermissionDenied, identity_loaded, RoleNeed,
-    UserNeed
-)
+from flask import current_app
+from flask import render_template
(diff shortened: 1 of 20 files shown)


udata/auth/__init__.py
```diff
@@ -3,27 +3,28 @@
...
 
-from flask import render_template, current_app, Blueprint, request, redirect
-from flask.ext.principal import (  # noqa
-    Permission as BasePermission, PermissionDenied, identity_loaded, RoleNeed,
-    UserNeed
-)
+from flask import current_app
+from flask import render_template
+
+from flask.ext.principal import identity_loaded
```
davidbgk: Not exactly our import style, pros?

grischard: Axel and I were chatting about this this afternoon. It's recommended by [the OpenStack style guidelines](https://pypi.python.org/pypi/hacking), and we think it's because it makes diffs easier to read, so we thought why not?

davidbgk: In that case: 1. update the related documentation 2. update all imports for consistency ;)

PR #437, opened 2016-04-11
Upgrade dependencies

noirbizarre:
This PR upgrade to latest Python dependencies (all excluding flask-restplus) with the following changes:
- Flask-Login 0.3+ transform functions `is_*` to properties (See the related PRs: etalab/udata-gouvfr#/104 and etalab/udata-youckan#2)
- factory-boy provide a simpler and integrated way to use faker
- CommonMark upgraded to latest specs (and rename 2 classes)

### DIFF
--- a/udata/core/jobs/forms.py
+++ b/udata/core/jobs/forms.py
@@ -32,6 +32,12 @@ class PeriodicTaskForm(ModelForm):
     task = fields.StringField(_('Tasks'))
     enabled = fields.BooleanField(_('Enabled'))
 
+    def save(self, commit=True, **kwargs):
+        '''PeriodicTask is now dynamic and save behavior changed'''
+        if not self.instance:
+            self.instance = self.model_class()  # Force populate_obj in super()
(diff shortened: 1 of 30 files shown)


udata/core/jobs/forms.py
```diff
@@ -32,6 +32,12 @@ class PeriodicTaskForm(ModelForm):
     task = fields.StringField(_('Tasks'))
     enabled = fields.BooleanField(_('Enabled'))
 
+    def save(self, commit=True, **kwargs):
+        '''PeriodicTask is now dynamic and save behavior changed'''
```
davidbgk: `now` should not appear in a comment ;)

davidbgk:
r+wc

udata/core/jobs/forms.py
noirbizarre: You're right, I'l link to the changeset in celerybeat-mongo et to the comming PR in mongoengine

PR #438, opened 2016-04-11
Upgrade js dependencies

noirbizarre:
This PR upgrade to latest JS versions and fixes page change detection in job list


js/components/harvest/job.vue
```diff
@@ -93,44 +87,43 @@ export default {
...
-                    return STATUS_CLASSES[status];
-                },
-                label_func: function(status) {
-                    return STATUS_I18N[status];
-                }
-            }]
+                label_type: (status) =>  STATUS_CLASSES[status],
+                label_func: (status) => STATUS_I18N[status],
+            }],
+            p: new PageList({data: this.job.items}),
```
davidbgk: Should not be computed anymore?

davidbgk:
r+wc

js/components/harvest/job.vue
noirbizarre: The computed in this case was not working: the widgets were binded on the first instance and cannot detect deep changes

PR #440, opened 2016-04-11
Add license and sources for embeds, fix #439

### LINKED ISSUE #439: Include license and links to sources in embeds
## Detailed Description

Add licenses for all datasets embed within territories and provide link to sources for the ones we split manually.

## Context

When I browse a town page, I want to quickly see the license and sources, so I can choose wisely my datasets.

davidbgk:
capture d ecran 2016-04-11 a 17 56 52

### DIFF
--- a/udata/frontend/markdown.py
+++ b/udata/frontend/markdown.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+from urlparse import urlparse
+
 import bleach
 import CommonMark
 from html5lib.sanitizer import HTMLSanitizer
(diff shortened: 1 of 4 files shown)


udata/frontend/markdown.py
```diff
@@ -1,18 +1,52 @@
...
+
 md = LocalProxy(lambda: current_app.extensions['markdown'])
 
 EXCERPT_TOKEN = '<!--- --- -->'
 
 
+def avoid_mailto_callback(attrs, new=False):
+    """Remove completely the link containing a `mailto`."""
+    if attrs['href'].startswith('mailto:'):
+        return None
```
noirbizarre: Why not reuse https://github.com/etalab/udata/blob/master/udata/frontend/helpers.py#L116-L118

davidbgk: Mostly because I can't use the `safe` filter after content posted by user and I *guess* that 90% of the time the user doesn't want a link at that place. But I may be wrong. Actually I'm not even sure if we have that case today…

udata/frontend/markdown.py
```diff
@@ -1,18 +1,52 @@
...
+    otherwise add `nofollow`.
+    That callback is not splitted in order to parse the URL only once.
+    """
+    parsed_url = urlparse(attrs['href'])
+    if parsed_url.netloc in ('', current_app.config['SERVER_NAME']):
+        attrs['href'] = '{scheme}://{netloc}{path}'.format(
+            scheme=current_app.config['USE_SSL'] and 'https' or 'http',
+            netloc=current_app.config['SERVER_NAME'],
+            path=parsed_url.path)
+        attrs['data-tooltip'] = _('Source')  # For embeds.
```
noirbizarre: Isn't it another way to do this ? Having an hardcoded `data-tooltip` on all markdown links seems violent.

davidbgk: It's only for our internal links but I agree that's it's far from being perfect… suggestions? Callbacks don't have any extra parameter :/

noirbizarre:
r+wc

udata/frontend/markdown.py
noirbizarre: I'm sure to remember it was added because we had cases, it was asked. The use case was to easily allow to put an official contact point in the dataset description. But things evolved since, there are discussions (but discussions are only on the site dataset page which assume the contact point is a registered user).
But let's merge it like that.

udata/frontend/markdown.py
noirbizarre: Yes but `UdataMarkdown.__call__(stream)` can have extra parameters.
If I understand the use case, it's only for territory datasets in embed. Maybe the template can be:
```html+jinja
{{ description|markdown(tooltip=_('Source')) }}
```
and if `tooltip` is set we add an extra callback adding the tooltip.

davidbgk: Interesting, I'll try that :heart:

davidbgk:
@noirbizarre actually I can't pass a parameter to the callback so I had to use a boolean, if you see a better approach…

noirbizarre:
:+1:

PR #442, opened 2016-04-12
fixed header template html formatting

vinyll:
removed an extra `</div>`
unindented the `<ul>`

> this diff in github is confusing. Check the line numbers if you wish to compare. I just removed line 99 and unindended line 35 to 98.


noirbizarre:
r+

PR #444, opened 2016-04-14
Refactoring factories to individual modules

vinyll:
1. Moved test.factories.* to individual udata.core.*.factories modules


davidbgk:
Nice, don't forget to update udata-gouvfr too ;-)

vinyll:
I'm not sure how you want to review it.
I'd suggest that as tests are now passing we'd review this commit for merge and work on a next branch for implement the true fixture functionalities.
What say you my friends?

udata/tests/dataset/test_dataset_model.py
```diff
@@ -5,13 +5,14 @@
 from mongoengine import post_save
 
 from udata.models import db, Dataset
+from udata.core.dataset.factories import (ResourceFactory, DatasetFactory,
```
davidbgk: I prefer:

```
from udata.core.dataset.factories import (
    ResourceFactory, DatasetFactory, DatasetDiscussionFactory, DatasetDiscussionFactory
)
```

davidbgk:
:+1: for merging before improving :)

Note that there are some conflicts for now.

udata/tests/dataset/test_dataset_model.py
vinyll: I chose consciously the first option as:
- it is more readable to me when it's blocked with other `from udata.core… import …` lines
- it saves a line

If you still prefer the second version I change it at the several places where this is applied.

vinyll:
conflicts are resolved. Reviews are welcome :)

udata/tests/dataset/test_dataset_model.py
davidbgk: I'm not sure it really saves a line and each new import will add a new line, you end up having some kind of flag code indentation which looks weird *to me*. 

When I made the project pep8 consistent, I tried to use the indentation I mentioned in my comment above so if there are still inconsistencies we have to fix those. My point is not to discuss one version vs. the other but just to be consistent and to chose the path of least resistance for that.

vinyll:
I've updated the commit to display long import direclty on a new line (eg: https://github.com/etalab/udata/pull/444/files#diff-1dbb8d79fdfb5d5c3ecd4268c5b4f457R11)

davidbgk:
r+gouvfr ;)

vinyll:
Thanks, I was waiting to get this approval before working on gouvfr ;) On my way!

PR #445, opened 2016-04-19
Move the /terms route to the core, and move terms.html to the theme templates

grischard:
References https://github.com/etalab/udata-gouvfr/pull/109

### DIFF
--- a/udata/core/site/views.py
+++ b/udata/core/site/views.py
@@ -163,7 +163,13 @@ class SiteDashboard(SiteView, DetailView):
         return context
 
 
+@blueprint.route('/terms/')
+def terms():
+    return theme.render('terms.html')
+
+
 @sitemap.register_generator
 def site_sitemap_urls():
     yield 'site.home_redirect', {}, None, 'daily', 1
     yield 'site.dashboard_redirect', {}, None, 'weekly', 0.6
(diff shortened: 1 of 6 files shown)


udata/templates/security/register_user.html
```diff
@@ -18,7 +18,7 @@
                     </label>
                 </div>
                 <div class="col-xs-10 col-md-8 col-lg-9 text-left">
-                    <label>{{ _('I have read and accept <a href="%(url)s"> the terms and conditions of the service.</a>'), url='/terms' }}</label>
+                    <label>{{ _('I have read and accept <a href="%(url)s"> the terms and conditions of the service.</a>', url=config.SITE_GITHUB_URL) }}</label>
```
noirbizarre: `s/SITE_GITHUB_URL/SITE_TERMS_URL/`

udata/core/site/views.py
```diff
@@ -163,7 +163,13 @@ def get_context(self):
         return context
 
 
+@blueprint.route('/terms/')
```
grischard: This path could also be a setting. Any thoughts?

noirbizarre: It could but I don't think it's really needed and/or worth the complexity for now.
If you don't need it and someone needs it, it can be done in another PR.

noirbizarre:
r+

PR #446, opened 2016-04-19
Init fixtures data

vinyll:
Right now this is implementing fixture creation via a task.
Next step will be to fix `udata init` and prompt for fixtures.

### DIFF
--- a/udata/commands/fixtures.py
+++ b/udata/commands/fixtures.py
@@ -6,37 +6,43 @@ import logging
 
 from udata.commands import manager
 from udata.core.user.factories import UserFactory, AdminFactory
-from udata.core.dataset.factories import (DatasetFactory, VisibleDatasetFactory,
-                                          DatasetDiscussionFactory)
-from udata.core.reuse.factories import ReuseFactory, VisibleReuseFactory
+from udata.core.dataset.factories import (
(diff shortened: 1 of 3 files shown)


udata/commands/fixtures.py
```diff
@@ -0,0 +1,49 @@
...
+        DatasetDiscussionFactory(subject=dataset)
+
+
+def generate_reuses(count, user=None):
+    for _ in range(0, count):
+        VisibleReuseFactory(owner=user)
+
+
+def generate_licenses(count):
+    for _ in range(0, count):
```
davidbgk: Why not `.create_batch(count)`?

davidbgk:
With a fresh update on `dev`, I have this error when I serve the project:

```ERROR: Unable to import init: No module named fixtures```

Not sure if it's related though.

vinyll:
Let me check that a file is not missing

udata/commands/fixtures.py
vinyll: Because it does not exist for factories. That's a model's method.

noirbizarre:
Ready to merge ?

vinyll:
r+?

noirbizarre:
r+ :+1:

PR #447, opened 2016-04-26
Command line: Validate email addresses manually, and when creating users

grischard:
Users can be created through the command line. This automatically validates the email address for users created in this way.

This also adds the 'validate' CLI user command to manually validate an user's email address confirmation request.

### DIFF
--- a/udata/core/user/commands.py
+++ b/udata/core/user/commands.py
@@ -37,12 +37,24 @@ def create():
         data['password'] = encrypt_password(data['password'])
         del data['password_confirm']
         user = datastore.create_user(**data)
+        user.activate_user()
         print '\nUser created successfully'
         print 'User(id=%s email=%s)' % (user.id, user.email)
         return
     print '\nError creating user:'
     for errors in form.errors.values():
         print '\n'.join(errors)
+
+@m.command
+def activate():
+    '''Activate an existing user (validate their email address)'''
+    email = prompt('Email')
+    user = User.objects(email=email).first()
+    if not user:
+        print 'Invalid user'
+        return
+    user.activate_user()
+    print 'User activated successfully'
 
 
 @m.command
--- a/udata/core/user/commands.py
+++ b/udata/core/user/commands.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 
 import logging
 
+from datetime import datetime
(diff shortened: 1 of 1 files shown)


udata/core/user/commands.py
```diff
@@ -37,6 +37,7 @@ def create():
         data['password'] = encrypt_password(data['password'])
         del data['password_confirm']
         user = datastore.create_user(**data)
+        user.confirmed_at = datetime.utcnow()
```
davidbgk: I think that you'll have to add that property to the `User`'s model too?

davidbgk:
The manual `password` command should set the attribute too(?)

udata/core/user/commands.py
grischard: It already is: https://github.com/etalab/udata/blob/3c2aa6ace82da212b30de6012174e866c3df312c/udata/core/user/models.py#L67

davidbgk: Oh I just realized I made a typo, then redeclare the attribute, my bad :D

udata/core/user/commands.py
```diff
@@ -46,6 +47,20 @@ def create():
 
 
 @m.command
+def activate():
+    '''Activate an existing user (validate their email confirmation)'''
+    email = prompt('Email')
```
davidbgk: Interesting that for `delete` we prompt the email and for `set_admin` and `password` we have it as a parameter. Maybe an opportunity to be consistent here? @noirbizarre ?

noirbizarre: :+1: for consistency.
Proposal: they all takes an optional first argument email (login in fact) that will be prompted if not provided

udata/core/user/commands.py
```diff
@@ -46,6 +47,20 @@ def create():
...
+    user = User.objects(email=email).first()
+    if not user:
+        print 'Invalid user'
+        return
+    if user.confirmed_at is not None:
+        print 'User email address already corfirmed'
```
yohanboniface: Typo: "confirmed" ;)

udata/core/user/commands.py
```diff
@@ -46,6 +47,20 @@ def create():
...
+    user.confirmed_at = datetime.utcnow()
```
yohanboniface: Missing a `save`?

PR #450, opened 2016-04-27
Provide an update task and document it

noirbizarre:


tasks.py
```diff
@@ -34,6 +34,19 @@ def clean(bower=False, node=False):
         info('Removing {0}'.format(pattern))
         run('cd {0} && rm -rf {1}'.format(ROOT, pattern))
 
+@task
+def update(data=False):
```
davidbgk: Why not `True` by default?

tasks.py
```diff
@@ -34,6 +34,19 @@ def clean(bower=False, node=False):
...
+    '''Perform a development update'''
+    header('Update all dependencies and data')
+    info('Updating Python dependencies')
+    lrun('pip install -r requirements/develop.pip')
+    lrun('pip install -e .')
+    info('Updating JavaScript dependencies')
+    lrun('npm install')
+    if data:
```
davidbgk: +1 for `migrate`.

davidbgk:
r+wc

tasks.py
noirbizarre: Because it migrate data. Loading a new database can be very long so I prefer the dev to be conscious that it will be migrating its data too by explicitely adding `-d` or `--data` on the command line. What's your opinion ?

davidbgk:
Alright, LGTM!

PR #380, opened 2016-03-14
Update restplus

noirbizarre:
This PR upgrades to latest Flask-RESTPlus (0.9.2) and apply some fixes

PR #451, opened 2016-04-27
Use fields mask

noirbizarre:
This PR make use of fields Mask and GenericModelFields.

Extracted from #380 (so requires #380 )


udata/commands/fixtures.py
```diff
@@ -5,20 +5,19 @@
 import logging
 
 from udata.commands import manager
-from udata.core.user.factories import UserFactory, AdminFactory
-from udata.core.dataset.factories import (
-    VisibleDatasetFactory, DatasetDiscussionFactory, LicenseFactory
-)
-from udata.core.reuse.factories import VisibleReuseFactory
+from udata.core.dataset.factories import VisibleDatasetFactory, LicenseFactory
```
vinyll: Bewarn that it may conflict with https://github.com/etalab/udata/pull/444

udata/tests/forms/test_reference_field.py
```diff
@@ -7,7 +7,7 @@
 from udata.forms import ModelForm, fields
 from udata.models import db
 from udata.tests import TestCase
-from udata.tests.factories import faker
+from udata.core.user.factories import faker
```
vinyll: In https://github.com/etalab/udata/pull/444/commits/1b3226924f7aa46608a6e56aae04d0de4532e91f#diff-718f5796155236b92d81115cf73d992cR208 faker was moved to `udata.utils`

udata/features/oembed/api.py
```diff
@@ -5,7 +5,7 @@
 
 from udata import theme
 from udata.api import api, API, output_json
-from udata.models import db, Dataset, GeoZone
+from udata.models import db, Dataset, GeoZone, TERRITORY_DATASETS
```
vinyll: I guess with should move this out from _udata_ to _datagouv_

js/components/discussions/list.vue
```diff
@@ -37,10 +37,10 @@ export default {
     },
     events: {
         'datatable:item:click': function(discussion) {
-            let prefix = discussion.class.replace('Discussion', '').toLowerCase(),
+            let prefix = discussion.subject.class.toLowerCase(),
```
vinyll: `let` vs `const` (eg: https://github.com/etalab/udata/pull/451/commits/45501bf92e4542163142d50664826b5daeec1941#diff-c0bbbc951175bcb27057d4ec96470f3eR46)

udata/core/organization/api.py
```diff
@@ -87,8 +87,13 @@ def get(self, org):
...
+    @api.response(410, 'Organization has been deleted')
     def put(self, org):
-        '''Update a organization given its identifier'''
+        '''
+        Update a organization given its identifier
+
+        :raises PermissionDenied:
+        '''
         if org.deleted:
             api.abort(410, 'Organization has been deleted')
```
vinyll: Is this still needed with the decorator _response 410_ ?

vinyll: should https://github.com/etalab/udata/pull/451/commits/45501bf92e4542163142d50664826b5daeec1941#diff-6ce9a6ba822a15f73a6b9269b1a7c62aR109 also get a decorator?

udata/core/user/api.py
```diff
@@ -205,25 +210,67 @@ def delete(self):
...
-    model = User
-    fields = user_fields
-    form = UserProfileForm
+@api.response(404, 'User not found')
+@api.response(410, 'User has been deleted')
+class UserAPI(API):
+    @api.doc('get_user')
+    @api.marshal_with(user_fields)
+    def get(self, user):
+        '''Get an user given its identifier'''
```
vinyll: _Delete a user_ (_u_ being pronounced 'you' there's no 'n')
This applies for other '_an user_' comments in this PR

udata/templates/macros/issues.html
```diff
@@ -1,8 +1,8 @@
 {% macro issues_btn(obj, icons=None) %}
-{% set basename = obj.__class__.__name__|lower %}
 <button type="button" class="btn btn-danger btn-issues"
     data-api-url="{{ url_for('api.issues', for=obj.id|string) }}"
-    data-subject="{{ obj.id }}"
+    data-subject-id="{{ obj.id }}"
+    data-subject-class="{{ obj.__class__.__name__ }}"
```
vinyll: shouldn't we have a property in the obj (like `class_name`?) that returns `self.__class__.__name__` and use that instead here?

udata/tests/test_issues.py
```diff
@@ -178,6 +178,36 @@ def test_list_issues_closed_filter(self):
...
+            issues.append(issue)
+        # Creating a closed issue that shouldn't show up in response.
+        user = UserFactory()
+        other_dataset = Dataset.objects.create(title='Other Test dataset')
+        message = Message(content=faker.sentence(), posted_by=user)
+        issue = Issue.objects.create(
+            subject=other_dataset,
+            user=user,
+            title='test issue {}'.format(i),
+            discussion=[message]
```
vinyll: I had never noticed this _discussion_ property name was in singular form.
All other _db.ListField_ properties in the project have a name in the plural form.

udata/commands/fixtures.py
noirbizarre: Already rebased and fixed in last commit ;)

udata/features/oembed/api.py
noirbizarre: Nop, it's only populated with french territories in gouvfr

js/components/discussions/list.vue
noirbizarre: noted :+1:

udata/core/organization/api.py
noirbizarre: The decorator is only here to expose the documentation which can't be extracted.

udata/core/user/api.py
noirbizarre: noted too :+1: (Probably was a given user before ;))

udata/templates/macros/issues.html
noirbizarre: Yes, we can, but I find `__class__.__name__` explicit and consistent with the fact that we are avoiding necessary inheritance (one of the purpose of this PR).
I think a property hide the fact that it's simply the raw class name.
It's debatable, what do you think about it ?

udata/tests/test_issues.py
noirbizarre: `discussion` property is a legacy naming which might need renaming (in another PR) because now we have a `Discussion` model.

udata/core/followers/models.py
```diff
@@ -15,16 +15,16 @@ class FollowQuerySet(db.BaseQuerySet):
     def following(self, user):
         return self(follower=user, until=None)
 
-    def followers(self, user):
-        return self(following=user, until=None)
+    def followers(self, followee):
```
vinyll: I don't mind having `user` here as it represent explicitely the type of object that is expected.

udata/core/organization/api.py
```diff
@@ -296,7 +294,7 @@ def delete(self, org, user):
 
 @ns.route('/<id>/followers/', endpoint='organization_followers')
 class FollowOrgAPI(FollowAPI):
-    model = FollowOrg
+    model = Organization
```
vinyll: :+1:

udata/migrations/2015-12-09-generic-references.js
```diff
@@ -0,0 +1,62 @@
+/*
+ * Migrate disucssions and issues to DBRef/GenericReferenceField
```
vinyll: _discussions_

udata/migrations/2015-12-09-generic-references.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+/**
+ * Maps inherited classes to theiru subject collection
```
vinyll: _their_

udata/migrations/2015-12-09-generic-references.js
```diff
@@ -0,0 +1,62 @@
...
+
+var CLASSES = {
+    'dataset': 'Dataset',
+    'reuse': 'Reuse',
+    'organization': 'Organization',
+    'user': 'User'
+};
+
+/**
+ * Produce a (wweird) GenericReferenceField from a class and an ID
```
vinyll: _weird_ or is it on purpose?

udata/tests/api/test_model_resolution.py
```diff
@@ -1,30 +0,0 @@
...
-        with self.assertRaises(ValueError):
-            api.resolve_model('UDataMongoEngine')
-
-    def test_raise_if_none(self):
-        with self.assertRaises(ValueError):
-            api.resolve_model(None)
-
-    def test_raise_if_missing_class_entry(self):
-        with self.assertRaises(ValueError):
-            api.resolve_model({'field': 'value'})
```
vinyll: Why is this test file deleted?

udata/tests/forms/test_reference_field.py
```diff
@@ -0,0 +1,230 @@
...
+    def test_with_invalid_id(self):
+        Nested.objects.create(name=faker.name())
+        form = WithReferenceForm.from_json({'nested': 'bad'})
+
+        form.validate()
+        self.assertIn('nested', form.errors)
+        self.assertEqual(len(form.errors['nested']), 1)
+
+
+class Nested2(db.Document):
```
vinyll: "_Nested2_"? It would be worth having a short pydoc here. Same for `Nested3`.

js/components/dataset/card.vue
```diff
@@ -92,7 +92,11 @@ import DatasetFilters from 'components/dataset/filters';
 import placeholders from 'helpers/placeholders';
 import config from 'config';
 
-const MASK = 'id,title,description,spatial{zones,granularity},metrics,organization,frequency,temporal_coverage';
+const MASK = [
```
vinyll: As we discussed – and this is subjective – I know that `mask` actually refers to `X-Fields`in the API meta. I would therefore have called it _XFIELDS_ or _FIELDS_ to make it more uniform.
An extra opinion would be welcome.

js/views/dataset.vue
```diff
@@ -168,17 +173,17 @@ export default {
...
         dataset: require('components/dataset/details.vue'),
         quality: require('components/dataset/quality.vue'),
         chart: require('components/charts/widget.vue'),
         resources: require('components/dataset/resource/list.vue'),
         followers: require('components/follow/list.vue'),
         wmap: require('components/widgets/map.vue'),
-        issues: require('components/issues/list.vue'),
-        discussions: require('components/discussions/list.vue'),
-        community: require('components/dataset/communityresource/list.vue'),
+        CommunityList,
```
vinyll: is this some magic import that converts CamelCase to dashed-name?

js/views/editorial.vue
```diff
@@ -3,19 +3,21 @@
     <div class="row">
         <dataset-cards class="col-xs-12 col-md-6"
             :title="_('Featured datasets')"
-            :datasets="home_datasets.items">
+            :datasets="home_datasets.items"
+            :loading="home_datasets.loading">
         </dataset-cards>
-        <reuses class="col-xs-12 col-md-6"
+        <reuse-cards class="col-xs-12 col-md-6"
```
vinyll: `reuse-list`?

vinyll: `<reuse-card-list>`?

js/views/editorial.vue
```diff
@@ -3,19 +3,21 @@
     <div class="row">
         <dataset-cards class="col-xs-12 col-md-6"
```
vinyll: `<dataset-card-list>`?

js/views/site.vue
```diff
@@ -18,7 +18,7 @@
         <reuse-list id="reuses" class="col-xs-12" :reuses="reuses"></reuse-list>
     </div>
     <div class="row">
-        <organizations id="organizations" class="col-xs-12" :organizations="organizations"></organizations>
+        <org-list id="organizations" class="col-xs-12" :organizations="organizations"></org-list>
```
vinyll: `<organization-list>`?

js/views/home.vue
```diff
@@ -40,7 +41,8 @@ export default {
             metrics: new MyMetrics(),
             reuses: new PageList({
                 ns: 'me',
-                fetch: 'my_org_reuses'
+                fetch: 'my_org_reuses',
```
vinyll: _my_organization_reuses_?

js/views/search.vue
```diff
@@ -51,15 +52,16 @@ export default {
             datasets: new PageList({
                 ns: 'me',
                 fetch: 'my_org_datasets',
```
vinyll: _my_organization_datasets_?

js/views/search.vue
```diff
@@ -51,15 +52,16 @@ export default {
...
+                mask: DatasetList.MASK
             }),
             communities: new PageList({
                 ns: 'me',
                 fetch: 'my_org_community_resources'
             }),
             reuses: new PageList({
                 ns: 'me',
-                fetch: 'my_org_reuses'
+                fetch: 'my_org_reuses',
```
vinyll: _my_organization_reuses_?

js/views/editorial.vue
```diff
@@ -1,10 +1,10 @@
 <template>
 <layout :title="_('Editorial')">
     <div class="row">
-        <datasets class="col-xs-12 col-md-6"
+        <dataset-cards class="col-xs-12 col-md-6"
```
vinyll: `<dataset-card-list>`?

js/views/me.vue
```diff
@@ -7,7 +7,7 @@
     </div>
 
     <div class="row">
-        <datasets class="col-xs-12" :datasets="datasets"></datasets>
+        <datasets-list class="col-xs-12" :datasets="datasets"></datasets-list>
```
vinyll: `<dataset-list>`
We have `<reuse-list>` and others with no _s_

js/views/search.vue
```diff
@@ -49,7 +50,8 @@ export default {
         return {
             datasets: new PageList({
                 ns: 'me',
-                fetch: 'my_org_datasets'
+                fetch: 'my_org_datasets',
```
vinyll: _my_organization_datasets_?

udata/templates/macros/issues.html
vinyll: I really don't mind using `__class__.__name__` anywhere in Python but I don't think it's appropriate in a template.

vinyll:
Massively impressive PR. :open_mouth:

udata/templates/macros/issues.html
noirbizarre: But here, to be able to do that properly, the property needs to be declared on each model because right now we don't have a common ancestor.
I agree with you, template should stay simple, but in this specific case, the main point is that it's a standard Python object (a class instance) instead of a specific inherited `*Discussion`.

udata/core/followers/models.py
noirbizarre: The debate is open: I can switch back to user. It depends on what you find clearer.
The vocabulary here is a little tricky: `user`/`followee`/`follower`/`following`.
Until then, I switched back to user but I think the `follow` module need a little bit of naming coherence in another PR.

udata/tests/api/test_model_resolution.py
noirbizarre: Because this part moved on models/db and so the tests moved into `test_models`.

js/views/dataset.vue
noirbizarre: yes, it is :) But it's very well documented:
- https://vuejs.org/guide/components.html#Assets-Naming-Convention
- https://vuejs.org/guide/components.html#camelCase-vs-kebab-case

This is mainly du to the fact that on one side, html is case insensitive and recommend using dash into WebComponents name and properties and on the other side, dash is not legit in JavaScript naming and CamelCase is the convention.

js/views/home.vue
noirbizarre: Yes, maybe in another PR on `org` vs `organization` naming convention (this one require to change on the API side too)

js/views/search.vue
noirbizarre: same as above

js/views/search.vue
noirbizarre: Same as above

js/components/reuse/card.vue
```diff
@@ -64,7 +64,18 @@ import placeholders from 'helpers/placeholders';
 import moment from 'moment';
 import config from 'config';
 
+const MASK = [
```
davidbgk: Consistency? ;)

js/views/reuse.vue
```diff
@@ -38,8 +38,14 @@ import Metrics from 'models/metrics';
 import Vue from 'vue';
 import Issues from 'models/issues';
 import Discussions from 'models/discussions';
+// Widgets
+import DatasetCards from 'components/dataset/card-list.vue';
+import DiscussionList from 'components/discussions/list.vue';
+import IssueList from 'components/issues/list.vue';
 import Layout from 'components/layout.vue';
 
+const MASK = `datasets{${DatasetCards.MASK}},*`;
```
davidbgk: Do we really need all fields for reuses?

udata/api/__init__.py
```diff
@@ -34,6 +34,23 @@
...
+FACETS_TYPES = {
+    search.fields.BoolFacet: inputs.boolean
+}
+
+
+def description():
+    '''Load description from markdown file'''
+    if 'short' in request.args:
+        return ''
+    global DESCRIPTION
```
davidbgk: Ugh, can't we load that once and for all on server launch?

udata/api/description.md
```diff
@@ -0,0 +1,42 @@
+## {{ _('Authentication') }}
```
davidbgk: That's weird to have that here, maybe it should be part/included from the docs?

udata/core/dataset/api_fields.py
```diff
@@ -101,6 +101,14 @@
 community_resource_page_fields = api.model(
     'CommunityResourcePage', fields.pager(community_resource_fields))
 
+#: Default mask to make it lightweight by default
+DEFAULT_MASK = ','.join((
```
davidbgk: It seconds my thoughts on consistency for masks, we should stick to lists on the js side too.

udata/migrations/2015-12-09-generic-references.js
```diff
@@ -0,0 +1,68 @@
...
+            count++;
+        }
+    });
+    print('Processed ' + count + ' documents in db.' + collection);
+}
+
+process('discussion', 'subject');
+process('issue', 'subject');
+process('follow', 'following');
+// process('activity', 'related_to');
```
davidbgk: Why not activity anymore?

udata/core/dataset/models.py
```diff
@@ -324,7 +322,11 @@ def quality(self):
             * description length
             * and so on
         """
+        from udata.models import Discussion
```
davidbgk: If it's a circular import, add a comment.

js/components/reuse/card.vue
noirbizarre: Different day, different mood, different solution :)
Which one do you prefer: single line or joined array ?

js/views/reuse.vue
noirbizarre: I've chosen the shorter version: required dataset fields and all reuse fields (not so many bu most of them used on the page). The biggest data was the datasets payload (and all cards are now fetch once with only the required fields vs one call by card before with all fields)

js/components/reuse/card.vue
davidbgk: Arrays, joined when necessary, not during the definition otherwise we'll have to split if again if we need to merge/modify it later.

udata/api/__init__.py
noirbizarre: Not meant to be in the PR. I'll extract it. It was an experiment having as goal:
- all the documentation in the swagger (meant for it)
- description into a single templatable and translatable markdown file (one by langue, we translate the whole file)
- description serialized only when needed (not in admin as example)

udata/core/dataset/api_fields.py
noirbizarre: In this PR, the position was:
- mask as string by default
- as array only when needed or better (here only to be able to have a line under 80 char)

On the client side I wanted to avoir array joining every time. But using joined array by default is fine too me too.

udata/migrations/2015-12-09-generic-references.js
noirbizarre: Because right now, inheritance is an entire part of the Activity management algorythm (whereas on other classes it's only there to type the subject which is exactly what `GenericReferenceField` do) and to remove it it needs rewriting the activity handling.
It might be necessary but I think it was too much for this PR and deserve a separate PR.

udata/core/dataset/models.py
noirbizarre: Yes, it's for avoid them in futur.
I can't help wondering it we are not missing a layer here. Having so much computation in a property might need action action layer.

davidbgk: The computation has been moved to `compute_quality_score` and `check_availability`, we are just assembling the dict here. But I'm open to better suggestions, maybe it should be done completely asynchronously, or even externalized…

js/models/mask.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Serialize a mask as String for header
+ * @param  {String|Array} mask The mask to serialize
+ * @return {String}      A Serialized mask ready to be used into HTTP header
+ */
+export function serialize(mask) {
```
davidbgk: 👍

udata/api/__init__.py
```diff
@@ -34,6 +33,11 @@
 
 DEFAULT_PAGE_SIZE = 50
 HEADER_API_KEY = 'X-API-KEY'
+DESCRIPTION = None
```
davidbgk: Still useful?

udata/migrations/2015-12-09-generic-references.js
davidbgk: In that case remove that line and related `TYPES` definitions?

davidbgk:
r+wc

PR #459, opened 2016-05-12
Add a changelog, fix #458

### LINKED ISSUE #458: Add a changelog for the project
## Detailed Description

Add a changelog with breaking changes, new & improved and fixes.

## Context

When I update my instance, I want to be aware of what may broke, so I can update my code accordingly.

## Possible Implementation

* http://elasticsearch-py.readthedocs.io/en/master/Changelog.html
* http://falcon.readthedocs.io/en/stable/changes/index.html
* http://tinydb.readthedocs.io/en/latest/changelog.html

to name a few, the primary goal is to initiate something.

davidbgk:
Given that we don't have versioning (yet?) for the moment it's just labeled as `breaking changes` and `news & improved`. I did not add `fixes` given that it can be cumbersome to add an entry for each bug but that can be discussed :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,13 @@
+# Changelog
+
+## Breaking Changes
+
+* 2016-05-11: Upgrade of ElasticSearch from 1.7 to 2.3 [#449](https://github.com/etalab/udata/pull/449)
+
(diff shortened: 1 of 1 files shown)


noirbizarre:
r+

CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
+
+## Breaking Changes
+
+* 2016-05-11: Upgrade of ElasticSearch from 1.7 to 2.3 [#449](https://github.com/etalab/udata/pull/449)
```
grischard: What needs to be done to not break things? Upgrade ES at the same time? Run a migration?

davidbgk: Good point, I'll update accordingly 👍

PR #463, opened 2016-05-13
Small fixes for udata init to work

yohanboniface:


udata/commands/db.py
```diff
@@ -99,10 +99,11 @@ def execute_migration(plugin, filename, script, dryrun=False):
     return success
 
 
-def record_migration(plugin, filename, script):
+def record_migration(plugin, filename, script, **kwargs):
```
yohanboniface: It get passed `dryrun` kwarg.

udata/commands/db.py
```diff
@@ -99,10 +99,11 @@ def execute_migration(plugin, filename, script, dryrun=False):
...
     '''Only record a migration without applying it'''
     db = get_db(DEFAULT_CONNECTION_NAME)
     db.eval(RECORD_WRAPPER, plugin, filename, script)
+    return True
```
yohanboniface: should we catch errors?

udata/commands/init.py
```diff
@@ -21,7 +21,7 @@ def init():
     generate_fixtures()
 
     log.info('Apply DB migrations if needed')
-    migrate(record=True)
+    migrate(record=True, dryrun=False)
```
yohanboniface: Or make `dryrun` a kwarg defaulting to false in `migrate`?

davidbgk:
r+ I'll let @noirbizarre doing a review of that one because I'm not familiar enough with that part.

udata/commands/db.py
noirbizarre: I think the error is already handled by MongoEngine so in case of misconfiguration you will have the information to correct it.
Do you see something more useful to do ?

udata/commands/init.py
noirbizarre: I think defaulting to False into the upstream `migrate` might be more useful (to avoid this error in other futur usages)

udata/commands/db.py
yohanboniface: Nope. Just wanted to be sure my "return True" was not a bit optimistic ;)

udata/commands/init.py
yohanboniface: Done :)

PR #473, opened 2016-05-23
Migrate all front views to Vue/ES6

noirbizarre:


davidbgk:
Is that ready for review?

noirbizarre:
Not yet, I have to push my last changes and to fix the broken tests

vinyll:
Can you push your `https://github.com/noirbizarre/udata/tree/migration-es6-vue` branch on your repo so we get synchronized? ;)

noirbizarre:
Sorry, here the latest version. Tests should pass.
Still missing some views to be migrated:
- [ ] User display (Need merge with existing Vue.js views)
- [ ] Home page (use carousel)
- [ ] Search pages
- [ ] legacy widgets
  - [ ] "featured" button
  - [ ] "issues" button
  - [ ] "discussions" button
  - [ ] "integrate" button
  - [ ] "site search" bar

js/components/buttons/follow.vue
```diff
@@ -0,0 +1,78 @@
...
+        },
+        url: {
+            type: String,
+            required: true
+        }
+    },
+    computed: {
+        btnClasses() {
+            const classes = {active: this.following};
+            this.classes.forEach(cls => classes[cls] = true);
```
davidbgk: We hit [no-return-assign](http://eslint.org/docs/rules/no-return-assign) with @yohanboniface the other day and given that arrow function implicitly return the statement, StandardJS consider that you must add mustaches in these ambiguous cases. Thoughts?

js/front/organization/membership-request.vue
```diff
@@ -0,0 +1,63 @@
...
+    computed: {
+        canSubmit() {
+            return this.comment.length > 0 && !this.pending;
+        }
+    },
+    methods: {
+        submit() {
+            this.pending = true;
+            this.$api.post(this.url, {comment: this.comment})
+                .catch(error => {
```
davidbgk: I prefer to put the `.catch` at the end like within a `try...catch` statement but YMMV.

js/front/organization/index.js
```diff
@@ -0,0 +1,119 @@
...
+        if (location.hash !== '') {
+            this.$refs.tabs.$children.some((tab, index) => {
+                if (`#${tab.id}` === location.hash) {
+                    this.currentTab = index;
+                    return true;
+                }
+            });
+        }
+
+        // Async membership request
```
vinyll: Why not deleting this code?

js/front/organization/membership-request.vue
```diff
@@ -0,0 +1,63 @@
...
+                    Notify.error(this._('Error while requesting membership'));
+                    log.error(e.responseJSON);
+                })
+                .then(data => {
+                    Notify.success(this._('A request has been sent to the administrators'));
+                    this.$refs.modal.close();
+                })
+                .always(() => this.pending = false);
```
davidbgk: AFAIK, `always` is [not standard](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) and it's very dangerous to mix different kinds of promises implementations. I spent *weeks* on that issue at Mozilla and I don't want to fall on that trap again! :)

davidbgk:
Rewriting the whole thing… you fools :) Keep going 👍

js/front/organization/membership-request.vue
noirbizarre: Sadly, depending on the promise implementation, order matters here :/

js/front/organization/membership-request.vue
noirbizarre: Yes, that drived me crazy: no `always()` or `finally()`. This makes `fetch` a little bit less easier to use than `$.ajax` because we have a lot of `always`.
How do we handle that ?

davidbgk: Moving `this.pending = false` within both `then()` and `catch()` + re-generating a standard promise from `this.$api.post` surrounding it with a `Promise(...)` statement.

js/components/buttons/follow.vue
noirbizarre: Yes: implicit is evil !

js/components/buttons/follow.vue
```diff
@@ -0,0 +1,80 @@
...
+        label() {
+            return this.following ? this._('Unfollow') : this._('Follow');
+        }
+    },
+    methods: {
+        toggle() {
+            if (!Auth.need_user(i18n._('You need to be logged in to follow.'))) {
+                return;
+            }
+            const method = this.following ? this.$api.delete(this.url) : this.$api.post(this.url);
```
davidbgk: Is `this.$api` returning a standard promise? Otherwise you have to do `Promise.resolve(this.$api....)` everywhere to avoid any side effect :/

noirbizarre: Yes this is a standard promise

js/front/organization/membership-request.vue
noirbizarre: Already done. This is a standard promise which is returned so no need to wrap it into another one.

js/components/buttons/follow.vue
davidbgk: In that case how could you be able to add a `.always()` to it?

noirbizarre: No I wasn't, it's something I forgot to remove. I had hope and tried ;)

js/components/buttons/follow.vue
```diff
@@ -0,0 +1,80 @@
...
+            method.then(data => {
+                this.following = !this.following;
+                if (this.followers !== undefined) {
+                    this.followers = data.followers;
+                }
+            });
```
davidbgk: Missing at least `.catch(console.error.bind(console))` or is it caught later? Otherwise we don't have any clue when it fails.

js/components/popover.vue
```diff
@@ -0,0 +1,47 @@
...
+        // An optionnal title
+        title: [String, Element, HTMLElement],
+        // The popover content
+        content: [String, Element, HTMLElement],
+        // Wider popover
+        large: {type: Boolean, default: false},
+    },
+    computed: {
+        classes() {
+            const classes = {large: this.large};
```
davidbgk: I'm curious if it works with just `{this.large}` :)

js/front/dataset/add-reuse-modal.vue
```diff
@@ -42,7 +42,7 @@ export default {
         addToReuse(id) {
             const url = `reuses/${id}/datasets/`;
             const data = {id: this.dataset.id, class: 'Dataset'};
-            this.$http.post(url, data).then(() => {
+            this.$api.post(url, data).then(() => {
                 window.location = `${config.admin_root}reuse/${id}`;
             });
```
davidbgk: Missing error catching too.

js/plugins/api.js
```diff
@@ -181,7 +181,8 @@ export class Api {
             credentials: 'include',
             headers: Object.assign({}, DEFAULT_HEADERS, WRITE_HEADERS, headers || {}),
         })
-        .catch(this.onError);
+        .catch(this.onError)
```
davidbgk: Not exactly related to that PR but I just realized that we [don't handle HTTP errors properly](https://github.com/github/fetch#handling-http-error-statuses) in that plugin. `fetch` requires that you deal with that explicitly and doesn't consider a high status code as an error :(

udata/templates/topic/display_base.html
```diff
@@ -10,7 +10,7 @@
 {% set body_class = 'topic-display' %}
 
 {% block extra_js %}
-<script src="{{ static('topic/display.js') }}"></script>
+<script src="{{ static('topic.js') }}"></script>
```
davidbgk: 👍

davidbgk:
r+wc, I'm surprised that the ratio +953 / −320 is so unbalanced.

js/components/buttons/follow.vue
noirbizarre: Basic error handling is factorized and done in the api module. (previous PR on the topic):
https://github.com/opendatateam/udata/blob/dev/js/plugins/api.js

js/components/popover.vue
noirbizarre: Already tried ! Sadly, it's not working :( Each time I want to write `{this.attribute}` instead of `{attribute: this.attribute}` (I'm a lazy developer :)) but it's not working.

noirbizarre:
Yes, until we dismissed all the legacy components, we have to redefine some in Vue.js and keep the legacy ones for unported components. Just wait for the last step:
- all Handlebars views will disappear
- all legacy factorized commons.js (which by the still have form handling) will disappear
- ...
The ratio will change but right now, we are betwwen 2 states :/

js/plugins/api.js
noirbizarre: yes we do: https://github.com/opendatateam/udata/blob/dev/js/plugins/api.js#L96-L102 ;)
In fact, fetch simplify error handling with [`response.ok`](https://developer.mozilla.org/fr/docs/Web/API/Response/ok). (I think we already talked about it :wink:)

PR #478, opened 2016-06-02
Optimise image files

grischard:
Compressed all the images in the udata tree losslessly by combining various tools:

* Zopfli
* Pngcrush
* PNGOUT
* AdvPNG

* JpegOptim
* Mozjpeg

This saves 131.4 KB out of 1.1MB, 35.4% on average, and up to 66.2%.

While most of these images are placeholders overridden by the theme, it's nice to show what can easily be achieved.


vinyll:
Except for the build that's breaking for different reasons it seems good to me

noirbizarre:
r+ (but I need to fix the build first)

noirbizarre:
I merge it now, no need to pass the test, only images are changed.

PR #479, opened 2016-06-03
Use upstream Sifter lib and activate nesting

noirbizarre:
This PR fix the build by using the upstream Sifter.js with nesting (was a patched version but upstream patch has been accepted)


davidbgk:
r+

PR #482, opened 2016-06-07
Fix uploader mixin import (side-effect of using ES6)

noirbizarre:
This PR fix a side-effect of uploader switching to ES6 syntax.

### DIFF
--- a/js/components/dataset/resource/file-form.vue
+++ b/js/components/dataset/resource/file-form.vue
@@ -57,9 +57,11 @@
 import API from 'api';
 import Dataset from 'models/dataset';
 import CommunityResource from 'models/communityresource';
+import UploaderMixin from 'mixins/uploader';
+
 
 export default {
-    mixins: [require('mixins/uploader')],
+    mixins: [UploaderMixin],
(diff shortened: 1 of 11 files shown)


davidbgk:
There are a couple of other places where we `require` directly within the `mixin` declaration, is that a problem?

noirbizarre:
Yes, it might be if the mixin is exported with `export default`. I'm in trouble: I'm porting everything in ES6 on a branch, but to make the dev branch work until then, I have to backport these fixes :/

davidbgk:
In that case, don't you want to change each `require('components/form/base-field').FieldComponentMixin` in that PR too?

davidbgk:
And the change in `widgets/thumbnailer.vue` is missing too.

noirbizarre:
Hum, I guess you're right. i'm doing this right now

noirbizarre:
OK, fixed the missing `widgets/thumbnailer.vue` and migrated all `require('components/form/base-field').FieldComponentMixin` statements

davidbgk:
r+

PR #467, opened 2016-05-17
Add counties to territories, fix #466

davidbgk:
See https://github.com/etalab/udata-gouvfr/pull/110 too.

Once merged, the command `territories collect_references_files` must be run on subsequent deployments.

PR #483, opened 2016-06-07
Add counties and regions to territories, fix #466

### LINKED ISSUE #466: [France] Add counties to territories
After towns, counties are the geographical level that is the most pertinent to deal with for French users.
## Context

When I access a town, I want to have access to data on the upper level, so I can aggregate those with my local data.

davidbgk:
Replace https://github.com/opendatateam/udata/pull/467 because my git history wasn't clean.

Once merged, the command `territories collect_references_files` must be run on subsequent deployments.

Please review https://github.com/etalab/udata-gouvfr/pull/110 too.


udata/api/fields.py
```diff
@@ -40,7 +40,7 @@ def output(self, key, obj):
 
 class NextPageUrl(String):
     def output(self, key, obj):
-        if not obj.has_next:
+        if not hasattr(obj, 'has_next') or not obj.has_next:
```
noirbizarre: `if not getattr(obj, 'has_next')` ;)

udata/api/fields.py
```diff
@@ -50,7 +50,7 @@ def output(self, key, obj):
 
 class PreviousPageUrl(String):
     def output(self, key, obj):
-        if not obj.has_prev:
+        if not hasattr(obj, 'has_prev') or not obj.has_prev:
```
noirbizarre: same one ;)

udata/core/spatial/api.py
```diff
@@ -32,6 +35,35 @@
     'size', type=int, help='The amount of suggestion to fetch',
     location='args', default=10)
 
+dataset_parser = api.parser()
+dataset_parser.add_argument(
+    'with_dynamic', type=bool, help='Append dynamic datasets',
```
noirbizarre: Is it possible to reduce to a single word ? `dynamic=true` seems sufficient and explicit

udata/core/spatial/api.py
```diff
@@ -48,35 +80,64 @@ def get(self, ids):
...
     def get(self, id):
         '''Fetch a zone'''
         zone = GeoZone.objects.get_or_404(id=id)
-        return zone
+        if zone.level_name == 'country':
+            children = zone.regions
+        elif zone.level_name == 'region':
+            children = zone.counties
+        elif zone.level_name == 'county':
+            children = zone.towns
```
noirbizarre: This is not very scalable. Isn't it possible to factorize that ?

udata/core/spatial/models.py
```diff
@@ -100,15 +111,55 @@ def postal_string(self):
...
+    def towns(self):
+        return self.get_children('fr/town').order_by('-population', '-area')
+
+    @cached_property
+    def counties(self):
+        return self.get_children('fr/county').order_by('-population', '-area')
+
+    @cached_property
+    def regions(self):
+        return self.get_children('fr/region').order_by('-population', '-area')
```
noirbizarre: I think this a very factorizable:
- the `order_by` will always be the same and so can be in get_children
- the levels are already defined and loaded and so the child types list
- this way it does not support at all fetching `town` or `counties` for other countries.

Just exposing the `get_children` method should be sufficient, no ?

udata/core/spatial/api.py
davidbgk: It's a bit hard for a few reasons:

* for the moment we are in a Russian dolls situation but it'll not be the case anymore once we introduce other levels so adding more abstraction without these edge cases is useless
* even if I use data from the `geo_levels` table, we don't have a strong hierarchy with our current levels (towns' parents doesn't contain counties for instance) so it means that we'll perform too many extra queries
* if I hardcode our French hierarchy it'll not be anymore generic, I think that API will move to `-gouvfr` once we introduce custom levels anyway.

Thoughts?

udata/features/territories/__init__.py
```diff
@@ -20,3 +24,37 @@ def check_for_territories(query):
...
+    """
+    Return a geozone queryset of counties given the `query`.
+
+    If it's a 2-digits code try it, otherwise use the name.
+    """
+    if (not query or len(query) < 2 or
+            not current_app.config.get('ACTIVATE_TERRITORIES')):
+        return GeoZone.objects.none()
+    qs = GeoZone.objects(level='fr/county')
+    if len(query) == 2 and query.isdigit():
```
noirbizarre: Does not work for `Corse` (`2a`/`2b`).

udata/features/territories/__init__.py
```diff
@@ -20,3 +24,37 @@ def check_for_territories(query):
...
+    Return a geozone queryset of regions given the `query`.
+    """
+    if (not query or len(query) < 4 or
+            not current_app.config.get('ACTIVATE_TERRITORIES')):
+        return GeoZone.objects.none()
+    qs = GeoZone.objects(level='fr/region')
+    # Check names starting with query or exact match.
+    qs = qs(db.Q(name__istartswith=query) | db.Q(name__iexact=query))
+    # Sort matching results by population and area.
+    return qs.order_by('-population', '-area')
```
noirbizarre: This all can be factorized. I think the query check is not so useful and can be dropped (filtered on a given level, it just won't match any zone. Without the query check, it's always the same function.

udata/features/territories/api.py
```diff
@@ -19,14 +20,21 @@ class SuggestTerritoriesAPI(API):
...
+        # Python sort given that we aggregate two distinct querysets.
+        territories = sorted(territories,
+                             key=lambda t: (t.population, t.area),
+                             reverse=True)
         return [{
             'id': territory.id,
             'title': territory.name,
-            'county': territory.county.name,
+            'county': territory.county and territory.county.name or None,
+            'region': territory.region and territory.region.name or None,
```
noirbizarre: Payload will quickly become heavy when we add level. Can't we just support a single main parent ?

udata/core/spatial/models.py
davidbgk: Where do I know that children of a county should be towns? See my previous comment ;)

udata/features/territories/__init__.py
davidbgk: Being able to query by postal code or county code looks very important to me, don't think (only) as a Parisian :p .

Sometimes factorization leads to a more complex situation.

udata/features/territories/__init__.py
davidbgk: Oops I forgot DROM-COM too ;)

udata/templates/territories/search-result.html
```diff
@@ -1,14 +1,21 @@
 {% cache cache_duration, 'territory-search-result', territory.id, g.lang_code %}
 {% set logo=territory.logo_url() or theme_static('img/placeholder_territory.png') %}
+{% if territory.level_name == 'town' %}
```
davidbgk: I want a better solution here but I'm not sure where to put the `gettext` call in order to be found during the parsing given that it's dynamic.

udata/core/spatial/models.py
```diff
@@ -22,6 +22,9 @@
     ('other', _('Other')),
 ]
 
+# WARNING: the order is important to compute parents/children.
+HANDLED_ZONES = ('fr/town', 'fr/county', 'fr/region')
```
noirbizarre: This might be handy as a setting ;)

udata/features/territories/__init__.py
noirbizarre: My point was the opposite: always query both name and code :)
I'm a suburban ;)

udata/templates/territories/search-result.html
noirbizarre: I think you can use GeoLevels and use the same pattenr as: https://github.com/opendatateam/udata/blob/master/udata/core/spatial/models.py#L130-L138 :
- loaded once a startup
- cached by language

docs/adapting-settings.md
```diff
@@ -66,6 +66,30 @@ The enabled theme name.
...
+
+**default**: `False`
+
+Whether you want to activate pages and API related to territories.
+Don't forget to set the `HANDLED_LEVELS` setting too.
+
+
+### HANDLED_LEVELS
+
+**default**: `()`
```
noirbizarre: Either `tuple()`, `(, )` or `[]`

js/widgets.js
```diff
@@ -1,108 +1,172 @@
-Array.range = function(n) {
-    // Array.range(5) --> [0,1,2,3,4]
-    return Array.apply(null, Array(n)).map((x, i) => i);
-};
+/* This file is using http://standardjs.com/ */
+// `fetch` polyfill added by webpack.
```
noirbizarre: It reminds me that the way we load the polyfill is erroneous.

js/widgets.js
```diff
@@ -1,108 +1,172 @@
...
+/* global fetch */
 
-Object.defineProperty(Array.prototype, 'chunk', {
-    // Chunking an Array given the numeric parameter.
-    value: function(n) {
-        return Array.range(Math.ceil(this.length / n)).map((x, i) => this.slice( i * n, i * n + n));
-    }
-});
+// Constants in use within the HTML.
+const TERRITORY = 'data-udata-territory'
```
noirbizarre: Maybe be use underscores (`data-udata_territory`) as dashes are not accepted in the specification. `el.dataset['udata-territory'] = 'something'` will raise an exception.

js/widgets.js
```diff
@@ -1,108 +1,172 @@
-Array.range = function(n) {
-    // Array.range(5) --> [0,1,2,3,4]
-    return Array.apply(null, Array(n)).map((x, i) => i);
-};
+/* This file is using http://standardjs.com/ */
```
noirbizarre: Even if it may be a good/best practice, It adds another convention. We are still migrating files to ES6 using the eslint rules :/

js/widgets.js
```diff
@@ -1,108 +1,172 @@
...
+const qSA = document.querySelectorAll.bind(document)
+
+// Extract the base URL from the URL of that script.
+const parser = document.createElement('a')
+const scriptURI = qS('script#udata').src
+parser.href = scriptURI
+const baseURL = `${parser.protocol}//${parser.host}`
+
+function chunk (array, n) {
+  // Return the `array` chunked by `n`, useful to retrieve all datasets
```
noirbizarre: Until now, the convention used for JS doctring was:
```js
/**
 * Documentation
 */
function func(param) {}
```
We might need to really discuss those choices and stick with it in the futur.

js/widgets.js
```diff
@@ -1,108 +1,172 @@
...
+    throw error
+  }
 }
-function zip(arrays) {
-    // Equivalent to `zip()` in Python.
-    return arrays[0].map((_, i) => {
-        return arrays.map((array) => array[i]);
-    });
+
+function zip (arrays) {
```
noirbizarre: :+1:

js/widgets.js
```diff
@@ -1,108 +1,172 @@
...
-const territories = [].slice.call(document.querySelectorAll('[data-udata-territory-id]'));
-const datasets = [].slice.call(document.querySelectorAll('[data-udata-dataset-id]'));
-// Extract the base URL from the URL to that script.
-const parser = document.createElement('a');
-const scriptURI = document.querySelector('script#udata').src
-parser.href = scriptURI;
-const baseURL = `${parser.protocol}//${parser.host}`;
-// Keep the chunk > 9 otherwise territories pages will issue more than one query.
-territories.concat(datasets).chunk(9).forEach((elements) => {
+const embedDatasets = (territories, datasets) => {
```
noirbizarre: Why an arrow function here ?

js/widgets.js
```diff
@@ -1,108 +1,172 @@
...
-        .then((response) => response.json())
-        .then((jsonResponse) => {
-            // We match the returned list with the list of elements.
-            zip([elements, jsonResponse, references]).forEach(([element, response, id]) => {
-                element.innerHTML = response.html;
-                document.querySelector(`#${id} .integrate`).addEventListener('click', handleIntegration)
-            });
+      .then(checkStatus)
+      .then((response) => response.json())
+      .then((jsonResponse) => {
```
noirbizarre: I think first parenthesis can be safely removed from these two lines.

js/widgets.js
```diff
@@ -1,108 +1,172 @@
...
+        // We match the returned list with the list of elements.
+        zip([elements, jsonResponse, references])
+          .forEach(([element, response, id]) => {
```
noirbizarre: here too

js/widgets.js
```diff
@@ -1,108 +1,172 @@
...
+      })
+      .catch(console.error.bind(console))
+  })
+}
+
+// If a whole territory is defined, it takes precedence.
+// Otherwise fallback on ids.
+const territoryElement = qS(`[${TERRITORY}]`)
+if (territoryElement) {
+  const territorySlug = territoryElement.dataset.udataTerritory
```
noirbizarre: OMG !!! Is there black magic or is the dash to camel native ?
If it's the case, good to know (and then you can safely ignore the previous comment on the subject)

js/widgets.js
```diff
@@ -1,108 +1,172 @@
...
+  const territoryId = territorySlug.replace(/-/g, '/')
+  const url = `${baseURL}/api/1/spatial/zone/${territoryId}/datasets?dynamic=1`
+  // Polyfill added by webpack.
+  fetch(url)
+    .then(checkStatus)
+    .then((response) => response.json())
+    .then((jsonResponse) => {
```
noirbizarre: Same as above: optional parenthesis

udata/core/spatial/api.py
```diff
@@ -32,6 +35,35 @@
     'size', type=int, help='The amount of suggestion to fetch',
     location='args', default=10)
 
+dataset_parser = api.parser()
+dataset_parser.add_argument(
+    'dynamic', type=bool, help='Append dynamic datasets',
```
noirbizarre: It might be safe to use flask restplus bool input so the boolean is properly parsed when using `0/1`, `false/true`...

udata/core/spatial/api.py
```diff
@@ -32,6 +35,35 @@
...
+    location='args', required=False)
+dataset_parser.add_argument(
+    'size', type=int, help='The amount of datasets to fetch',
+    location='args', default=25)
+
+
+@ns.route('/zones/suggest', endpoint='suggest_zones')
+class SuggestZonesAPI(API):
+    @api.marshal_list_with(zone_suggestion_fields)
+    @api.doc('suggest_zones', parser=suggest_parser)
```
noirbizarre: Depends on what you prefer, but you can also write:
```
@api.expect(parser)
@api.doc('suggest_zones')
```

js/widgets.js
davidbgk: It was mostly a test on that particular file to see how it works. It's easy to revert. Let's discuss that next week ;)

js/widgets.js
davidbgk: Alright, I'll change that, it's not related to standardjs

js/widgets.js
davidbgk: See below, we need a function here.

js/widgets.js
davidbgk: Parenthesis are required with standardjs

js/widgets.js
davidbgk: No black magic, [pretty standard](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#Examples) I think ;)

udata/core/spatial/api.py
davidbgk: From where do I import that? I can't find it in [that doc](http://flask-restplus.readthedocs.io/en/stable/parsing.html).

udata/core/spatial/api.py
davidbgk: That method has just moved within the file but I'll update.

udata/core/spatial/models.py
```diff
@@ -78,6 +94,40 @@ def keys_values(self):
                 keys_values.append(value)
         return keys_values
 
+    @cached_property
+    def level_name(self):
+        """Truncated level name for the sake of readability."""
+        if self.level.startswith('fr/'):
+            return self.level[3:]
+        # Keep the whole level name as a fallback (e.g. `country/fr`)
```
noirbizarre: `fr/county` instead of `county/fr`

udata/core/spatial/models.py
```diff
@@ -96,19 +147,24 @@ def postal_string(self):
...
-        """Representation of a town with optional county."""
-        if self.county:
-            return '{name} <small>({county_name})</small>'.format(
-                name=self.name, county_name=self.county.name)
-        return self.name
+    @property
+    def parent(self):
+        if self.parent_level:
+            for parent in self.parents:
+                if parent.startswith(self.parent_level):
```
noirbizarre: Why `startswith ` ? equality is not  working ?

udata/core/spatial/models.py
davidbgk: No I maintain the `countRy/fr` ;)

udata/core/spatial/models.py
davidbgk: No, parent can be `fr/region/53` and `parent_level` will be `fr/region` for instance.

udata/templates/dataset/display.html
```diff
@@ -370,6 +370,19 @@ <h3 class="text-left">{{ _('Producer') }}</h3>
...
                 </div>
                 {% endif %}
+
+                {# Territories panel #}
+                {% if dataset.spatial.handled_zones %}
+                <div class="panel panel-default">
+                    <div class="panel-body">
+                        <h3>{{ _('Territories')}}</h3>
+                        {% for territory in dataset.spatial.handled_zones %}
+                            <img src="{{ territory.logo_url(external=True) or theme_static('img/placeholder_territory.png', external=True) }}" alt="{{ territory.name }}" class="float-left" width="13px" /> <a href="{{ territory.url }}" title="{{ territory.name }}">{{ territory.name }}</a>{% if not loop.last %}, {% endif %}
```
noirbizarre: you can use the `placeholder` filter which fulfill exactly the same purpose:
```html+jinja
<img src="{{ territory.logo_url(externa=True)|placeholder('territory') }}" />
```
The placeholder has to be in the placholder directory.

udata/templates/territories/home.html
```diff
@@ -0,0 +1,41 @@
...
+        L.geoJson({{ geojson|tojson }}, {
+            onEachFeature: function (feature, layer) {
+                layer.on('click', function () {
+                    window.location = '/fr/territory/region/' + feature.properties.code + '/';
+                })
+            }
+        }).addTo(map);
+        tilelayer.addTo(map);
+        map.setView(defaultCenter, defaultZoom);
+    </script>
```
noirbizarre: Why not reuse the map component like other pages ? Specific needs ?

noirbizarre:
r+wc

js/widgets.js
noirbizarre: :+1: (and what do you think about it ? )

js/widgets.js
noirbizarre: OK, I understand

js/widgets.js
noirbizarre: Good to know. dashes drived me crazy not so much ago

udata/core/spatial/api.py
noirbizarre: OK, I will improve this documentation to mention input parsers :+1: 
The boolean input is here: http://flask-restplus.readthedocs.io/en/stable/api.html#flask_restplus.inputs.boolean

udata/core/spatial/models.py
noirbizarre: lol, I misread

udata/core/spatial/models.py
noirbizarre: OK :+1: Understood

udata/templates/territories/home.html
davidbgk: Quick and dirty here, because I know that we'll have to change that after the demo, having just the map to navigate is not enough.

udata/templates/dataset/display.html
davidbgk: I did that because the placeholder comes from the theme, is there a way to deal with that case? There are placeholders in both the theme and the core which is confusing :/

PR #484, opened 2016-06-08
Ensure empty form field value is treated as undefined

noirbizarre:
Ensure empty string is treated as an empty value in form serialization (and so does not break form validation)

### DIFF
--- a/js/components/form/base-form.js
+++ b/js/components/form/base-form.js
@@ -43,6 +43,31 @@ function empty_schema() {
     return {properties: {}, required: []};
 }
 
+/**
+ * Input type which are text or formated text
+ */
+const TEXT_INPUTS = [
+    'color',
+    'date',
+    'datetime',
+    'datetime-local',
+    'email',
+    'hidden',
+    'month',
+    'range',
+    'search',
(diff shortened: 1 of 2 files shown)


js/components/form/base-form.js
```diff
@@ -168,7 +193,7 @@ export default {
 
             Array.prototype.map.call(elements, function(el) {
                 let value;
-                if (el.tagName.toLowerCase() === 'select') {
+                if (TEXT_TAGS.includes(el.tagName.toLowerCase()) || TEXT_INPUTS.includes(el.type)) {
```
davidbgk: Convert to lowercase for the type too?

js/components/form/base-form.js
```diff
@@ -43,6 +43,31 @@ function empty_schema() {
     return {properties: {}, required: []};
 }
 
+/**
+ * Input type which are text or formated text
+ */
+const TEXT_INPUTS = [
```
davidbgk: Maybe link to the spec where you retrieved that list?

davidbgk:
r+wc

PR #485, opened 2016-06-08
Fix flask 0.11 warnings

noirbizarre:
This PR fix warnings introduced in Flask 0.11.
`flask.ext` is now deprecated and so the current `udata.ext`.

This PR remove all `flask.ext` and `udata.ext` references.

Remaining warnings are:
- flask-mongoengine: fixed upstream, not yet released
- flask-fs: fixed upstream, not yet released
- flask-cache: does not seems actively maintained but there is 2 PRs fixing it


davidbgk:
r+, nothing to do on `udata-*`?

noirbizarre:
Yes: https://github.com/opendatateam/udata-youckan/pull/3

PR #489, opened 2016-06-16
Minor fixes

davidbgk:


js/widgets.js
```diff
@@ -123,7 +123,9 @@ function embedDatasets (territories, datasets) {
       }
     })
     const url = `${baseURL}/api/1/oembeds/?references=${references}`
-    // Polyfill added by webpack.
+    // Warning: if you are tempted to use generators instead of chaining
+    // promises, you'll have to use babel-polyfill which adds 300 Kb
+    // once the file is converted to ES5.
```
noirbizarre: ;)

udata/features/territories/commands.py
```diff
@@ -71,8 +71,13 @@ def collect_references_files():
                     continue
                 filename = resource.url.split('/')[-1]
                 reference_path = references.path(filename)
+                log.info('Found reference: {reference_path}'.format(
+                    reference_path=reference_path))
```
noirbizarre: Or you can use the logging extrapolation:
```python
log.info('Found reference: %s', reference_path)
```
Extrapolation is only evaluated on display

noirbizarre:
r+wc

udata/features/territories/commands.py
davidbgk: Better 👍

PR #490, opened 2016-06-17
Serve widgets minified for production + new API for widgets and optional search

davidbgk:
* From 19.2Kb to 9.43Kb, not bad :) Actually more 14.4Kb with added features, before gzip
* New API to load on demand the datasets, useful for territories' external pages
* New optional search box to filter retrieved datasets on the client side


js/widgets.js
```diff
@@ -18,6 +21,31 @@ parser.href = scriptURI
...
+    {'base': 'a', 'letters': /[\u0061\u24D0\uFF41\u1E9A\u00E0\u00E1\u00E2\u1EA7\u1EA5\u1EAB\u1EA9\u00E3\u0101\u0103\u1EB1\u1EAF\u1EB5\u1EB3\u0227\u01E1\u00E4\u01DF\u1EA3\u00E5\u01FB\u01CE\u0201\u0203\u1EA1\u1EAD\u1EB7\u1E01\u0105\u2C65\u0250]/g},
+    {'base': 'e', 'letters': /[\u0065\u24D4\uFF45\u00E8\u00E9\u00EA\u1EC1\u1EBF\u1EC5\u1EC3\u1EBD\u0113\u1E15\u1E17\u0115\u0117\u00EB\u1EBB\u011B\u0205\u0207\u1EB9\u1EC7\u0229\u1E1D\u0119\u1E19\u1E1B\u0247\u025B\u01DD]/g},
+    {'base': 'i', 'letters': /[\u0069\u24D8\uFF49\u00EC\u00ED\u00EE\u0129\u012B\u012D\u00EF\u1E2F\u1EC9\u01D0\u0209\u020B\u1ECB\u012F\u1E2D\u0268\u0131]/g},
+    {'base': 'o', 'letters': /[\u006F\u24DE\uFF4F\u00F2\u00F3\u00F4\u1ED3\u1ED1\u1ED7\u1ED5\u00F5\u1E4D\u022D\u1E4F\u014D\u1E51\u1E53\u014F\u022F\u0231\u00F6\u022B\u1ECF\u0151\u01D2\u020D\u020F\u01A1\u1EDD\u1EDB\u1EE1\u1EDF\u1EE3\u1ECD\u1ED9\u01EB\u01ED\u00F8\u01FF\u0254\uA74B\uA74D\u0275]/g},
+    {'base': 'oe', 'letters': /[\u0153]/g},
```
jdesboeufs: ❤️

js/widgets.js
```diff
@@ -18,6 +21,31 @@ parser.href = scriptURI
...
+    {'base': 'u', 'letters': /[\u0075\u24E4\uFF55\u00F9\u00FA\u00FB\u0169\u1E79\u016B\u1E7B\u016D\u00FC\u01DC\u01D8\u01D6\u01DA\u1EE7\u016F\u0171\u01D4\u0215\u0217\u01B0\u1EEB\u1EE9\u1EEF\u1EED\u1EF1\u1EE5\u1E73\u0173\u1E77\u1E75\u0289]/g}
+  ]
+  rules.map((rule) => {
```
jdesboeufs: You should use `forEach` in place of `map` to avoid the creation of an `undefined`-fulfilled array.

davidbgk: Good catch, a remaining test. I'm a bit surprised that linters are failing on that.

js/widgets.js
```diff
@@ -203,52 +229,51 @@ function embedDatasets (territories, datasets, dataTerritoryIdAttr, dataDatasetI
...
-      if (filteredResults) {
-        territoryElement.innerHTML = ''
-        filteredResults.forEach((result) => {
-          territoryElement.appendChild(result.dataset)
-        })
-      } else {
-        territoryElement.innerHTML = initialDatasets
+function filterDatasets (territoryElement, event, datasets, initialDatasets) {
+  const searchValue = event.target.value
+  if (searchValue.length >= 3) {
```
rik: Early return here can save an indentation for the reminder of the function.

js/widgets.js
```diff
@@ -203,52 +229,51 @@ function embedDatasets (territories, datasets, dataTerritoryIdAttr, dataDatasetI
...
+    const scoredResults = datasets.map((dataset) => {
+      const title = removeDiacritics(dataset.querySelector('.udata-title').innerText)
+      return {
+        dataset: dataset,
+        score: title.score(removeDiacritics(searchValue))
       }
+    })
+    scoredResults.sort((a, b) => a.score > 0 && a.score < b.score)
+    const filteredResults = scoredResults.filter((result) => result.score > 0)
```
rik: You should probably filter before sorting, that would save some unneeded sort operations.

rik: While also removing the need to check for a positive score in the sort function.

js/widgets.js
```diff
@@ -203,52 +229,51 @@ function embedDatasets (territories, datasets, dataTerritoryIdAttr, dataDatasetI
...
-        territoryElement.innerHTML = ''
-        filteredResults.forEach((result) => {
-          territoryElement.appendChild(result.dataset)
-        })
-      } else {
-        territoryElement.innerHTML = initialDatasets
+function filterDatasets (territoryElement, event, datasets, initialDatasets) {
+  const searchValue = event.target.value
+  if (searchValue.length >= 3) {
+    const scoredResults = datasets.map((dataset) => {
```
rik: I think this would read better as `scoredDatasets` and `filteredScoredDatasets`

js/widgets.js
```diff
@@ -88,7 +153,7 @@ function easeCopyPasting (content) {
 /**
  * Display/hide the integration box for embeds.
  */
-function handleIntegration (event) {
+function toggleIntegration (event) {
   event.preventDefault()
   const element = event.target
```
rik: This whole sequence of DOM traversal tightly couples the code and the DOM order. A good pattern is to use `js-` prefixed classnames to make refactoring easier.

js/widgets.js
```diff
@@ -113,70 +178,149 @@ function handleIntegration (event) {
...
-        return `territory-${el.dataset.udataTerritoryId}`
-      } else if (el.hasAttribute(DATASET_ID)) {
-        return `dataset-${el.dataset.udataDatasetId}`
+      if (el.hasAttribute(dataTerritoryIdAttr)) {
+        return `territory-${el.dataset[camelCaseData(dataTerritoryIdAttr)]}`
+      } else if (el.hasAttribute(dataDatasetIdAttr)) {
+        return `dataset-${el.dataset[camelCaseData(dataDatasetIdAttr)]}`
       }
     })
     const url = `${baseURL}/api/1/oembeds/?references=${references}`
```
rik: Here, you are relying on the cast from array to string to include commas. I think being explicit helps clarity:
```js
const url = `…${references.join(',')}`
```

js/widgets.js
```diff
@@ -113,70 +178,149 @@ function handleIntegration (event) {
  * Main function retrieving the HTML code from the API.
  * Keep the chunk > 12 otherwise territories pages will issue more than one query.
  */
-function embedDatasets (territories, datasets) {
-  chunk(territories.concat(datasets), 12).forEach((elements) => {
+function embedDatasets (territories, datasets, dataTerritoryIdAttr, dataDatasetIdAttr) {
+  const items = territories.concat(datasets)
+  const chunkBy = 12
+  const promises = chunk(items, chunkBy).map((elements) => {
```
rik: I'd split in a name function. Maybe `embedDatasetChunk` ?

rik: Why are you switching from `items` to `elements` ? Also both those names are not very descriptive.

js/widgets.js
```diff
@@ -113,70 +178,149 @@ function handleIntegration (event) {
...
           })
+        return elements
       })
       .catch(console.error.bind(console))
   })
+  Promise.all(promises)
+    .then((chunks) => {
+      // Flatten the array of datasets arrays.
+      const datasets = [].concat(...chunks)
+      if (datasets.length === items.length) {
```
rik: Can you have this not equal? I think you could if one network request fail. To avoid doing this, you can just remove line 207. The fetch promise will stay in the rejected state and you will never enter this then. And you'll get the console logging by line 229.

js/widgets.js
```diff
@@ -113,70 +178,149 @@ function handleIntegration (event) {
...
     })
     const url = `${baseURL}/api/1/oembeds/?references=${references}`
     // Warning: if you are tempted to use generators instead of chaining
     // promises, you'll have to use babel-polyfill which adds 300 Kb
     // once the file is converted to ES5.
-    fetch(url)
-      .then(checkStatus)
-      .then((response) => response.json())
+    return fetchJSON(url)
       .then((jsonResponse) => {
```
rik: I think you could have a better name than just `jsonResponse` here. Something like `datasetsContent` (not happy with this though). You are describing how you this variable got created (by a network request), it's usually better to describe what is in the variable.

js/widgets.js
rik: Some `Promise` theory to explain why: http://jsbin.com/nukazorogi/edit?js,console

js/widgets.js
```diff
@@ -272,6 +278,7 @@ function insertSearchInput (event, territoryElement) {
   searchNode.style.margin = '1em'
   searchNode.style.padding = '.1em'
   territoryElement.parentNode.insertBefore(searchNode, territoryElement)
+  // TODO: add a timeout between keyups.
```
rik: https://css-tricks.com/debouncing-throttling-explained-examples/ :) Use loads :)

package.json
```diff
@@ -102,6 +102,7 @@
     "sortablejs": "^1.4.2",
     "source-sans-pro": "^2.0.10",
     "spin.js": "^2.3.2",
+    "string_score": "^0.1.22",
```
noirbizarre: Hum, a third dependency to perform text sorting :/

tasks.py
```diff
@@ -187,7 +187,7 @@ def assets_build(progress=False):
     if progress:
         cmd += ' --progress'
     lrun(cmd.format('webpack.config.prod'), pty=True)
-    lrun(cmd.format('webpack.widgets.config'), pty=True)
+    lrun(cmd.format('webpack.widgets.config.prod'), pty=True)
```
noirbizarre: In order to be consistent, files should be named: `webpack.config.*.js`. In this case, `webpack.config.widgets.prod.js`.

webpack.widgets.config.js
```diff
@@ -21,10 +20,5 @@ module.exports = {
                 }
             },
         ]
-    },
-    plugins: [
-        new webpack.ProvidePlugin({
-            'fetch': 'imports?this=>global!exports?global.fetch!whatwg-fetch'
-        })
-    ]
```
noirbizarre: :+1:

noirbizarre:
r+wc

js/widgets.js
davidbgk: Adding `lodash.debounce` adds 5Kb to my build vs. 10 lines of unbloated ES6 that I understand. Not worth it IMHO.

package.json
davidbgk: Added after discussing with @jdesboeufs, I wasn't aware we already have two!
Which one is more appropriated here?

noirbizarre: We have the soon to be removed Bloodhound (typeahead underlaying suggester) and Sifter.js (which is used in admin).
In your case I'm not sure they are relevant for embedded, so it's up to you wether you reuse Sifter or keep string_score

js/widgets.js
```diff
@@ -26,11 +87,30 @@ const baseURL = `${parser.protocol}//${parser.host}`
...
     .map((x, i) => array.slice(i * n, i * n + n))
 }
 
 /**
+ * Return a function, that, as long as it continues to be invoked,
+ * will not be triggered. The function will be called after
+ * it stops being called for `wait` milliseconds. Inspired by:
+ * https://davidwalsh.name/javascript-debounce-function
+ */
+function debounce (func, wait = 350) {
```
noirbizarre: We have a `debounce` dependency (I can't say if it's still in use or not). It might be interesting to either drop the useless dependency and use this piece everywhere (or reuse the dependency)

js/widgets.js
```diff
@@ -110,73 +197,183 @@ function handleIntegration (event) {
...
+  const items = territories.concat(datasets)
+  const chunkBy = 12
+  const promises = chunk(items, chunkBy)
+    .map((datasetChunk) => embedDatasetChunk(datasetChunk))
+  Promise.all(promises)
+    .then((chunks) => {
+      // Flatten the array of datasets arrays.
+      const datasets = [].concat(...chunks)
+      window.dispatchEvent(
+        new CustomEvent(
```
noirbizarre: Might not work with IE11: http://caniuse.com/#search=CustomEvent

js/widgets.js
```diff
@@ -110,73 +197,183 @@ function handleIntegration (event) {
...
+   *
+   * The `size` parameter limits the number of loaded datasets.
+   * The `withSearch` parameter optionally loads an input
+   * allowing the user to filter currently displayed datasets.
+   */
+  loadTerritory ({size = 100,
+                  withSearch = false,
+                  dataTerritoryAttr = 'data-udata-territory',
+                  dataTerritoryIdAttr = 'data-udata-territory-id',
+                  dataDatasetIdAttr = 'data-udata-dataset-id'} = {}) {
```
noirbizarre: same

js/widgets.js
```diff
@@ -110,73 +197,183 @@ function handleIntegration (event) {
...
+
+global.udataScript = {
+
+  /**
+   * Load territories and datasets from the API given the set
+   * data-attributes `dataTerritoryIdAttr` and
+   * `dataDatasetIdAttr` within the DOM.
+   */
+  load ({dataTerritoryIdAttr = 'data-udata-territory-id',
+         dataDatasetIdAttr = 'data-udata-dataset-id'} = {}) {
```
rik: Nope, it is needed to be able to call the function without any parameter. http://www.2ality.com/2011/11/keyword-parameters.html

davidbgk: http://exploringjs.com/es6/ch_parameter-handling.html#_simulating-named-parameters-in-javascript argh, not fast enough :p

js/widgets.js
davidbgk: Good catch :)

js/widgets.js
noirbizarre: My bad, I mis expressed myself: I know how destructuring parameters with defaults  works but I was reading `loadTerritory()` when I commented and I can't find the call si I can't say whether we always use all parameters, whether we call it without any...

I'm just wondering:
- Do we use so many combination ?
- Do we need to make the data attributes names configurables ? (It's an embed meant to be copy-pasted so I wonder why so much configurable parameters)(maybe I missed something)

davidbgk: Good questions. We use options in the [future territories SPA](https://github.com/etalab/collectivites.data.gouv.fr/pull/1/files#diff-f2ffb88b86d7d51e6a8abd66598cc301R40) so I made the ids customable too in order to be more flexible on the markup from reusers. But it's maybe a bit too over-engineered, I thought it was a good idea to avoid global consts but I finally redefined those since then to limit arguments passing.

PR #500, opened 2016-06-24
Fix typo

ndamiens:
fchier -> fichier


davidbgk:
Hello @ndamiens!

Thanks a lot for your contribution, unfortunately we translate these string via [Transifex](https://www.transifex.com) and the `js/locales/udata.fr.json` file is automatically generated from there. So I fixed the typo directly there, thanks a lot for pointing that out :-)

PR #469, opened 2016-05-23
Update fine uploader

noirbizarre:
This PR fix fine-uploader new packaging (introduced in 5.8) and migrate the uploader mixin to ES6 (and lint)

PR #502, opened 2016-06-28
Backport dev fixes

noirbizarre:
This PR is backport of the following fixes from dev to master:
- #469
- #478 
- #482 
- #484


js/mixins/uploader.js
```diff
@@ -1,195 +1,210 @@
...
+                    // Microsoft Office
+                    'xls', 'xlsx', 'doc', 'docx', 'pps', 'ppt',
+                    // Archives
+                    'tar', 'gz', 'tgz', 'rar', 'zip', '7z', 'xz', 'bz2',
+                    // Images
+                    'jpeg', 'jpg', 'jpe', 'gif', 'png', 'dwg', 'svg', 'tiff', 'ecw', 'svgz',
+                    // Geo
+                    'shp', 'kml', 'kmz', 'gpx', 'shx', 'ovr',
+                    // Misc
+                    'dbf', 'prj', 'sql', 'epub', 'sbn', 'sbx', 'cpg', 'lyr', 'xsd', 'owl'
```
odtvince: 'gtfs' ?

js/mixins/uploader.js
```diff
@@ -1,195 +1,210 @@
...
+                    // OpenOffice
+                    'ods', 'odt', 'odp', 'odg',
+                    // Microsoft Office
+                    'xls', 'xlsx', 'doc', 'docx', 'pps', 'ppt',
+                    // Archives
+                    'tar', 'gz', 'tgz', 'rar', 'zip', '7z', 'xz', 'bz2',
+                    // Images
+                    'jpeg', 'jpg', 'jpe', 'gif', 'png', 'dwg', 'svg', 'tiff', 'ecw', 'svgz',
+                    // Geo
+                    'shp', 'kml', 'kmz', 'gpx', 'shx', 'ovr',
```
odtvince: 'geojson' ?

js/mixins/uploader.js
odtvince: 'grib2'  ?

noirbizarre: This is simply a backport of dev fixes, no changes. The values here are already been deployed and are extracted from the existing datasets.

js/mixins/uploader.js
```diff
@@ -1,195 +1,210 @@
...
+            });
+            this.$emit('uploader:error', id, name, reason);
+        },
 
-            clear: function() {
-                // this.$progress.addClass('hide');
-                // this.$drop.removeClass('hide');
-                this.$uploader.clearStoredFiles();
-            }
+        clear: function() {
```
vinyll: `clear() {…`

specs/components/form/base.specs.js
```diff
@@ -204,9 +204,9 @@ describe('Common Form features', function() {
             ];
 
             expect(this.vm.serialize()).to.eql({
-                input: '',
+                input: undefined,
```
vinyll: I would use `null` instead of `underfined` here, as it is actually defined.

js/mixins/uploader.js
noirbizarre: As this PR is a backport, I'll change this on the dev branch, fine to you ?

specs/components/form/base.specs.js
noirbizarre: Same as above, simply a backport.

Actually, we are testing the `undefined` value explicitely here because it matters. `null` can't and shouldn't be used because of this:
```javascript
JSON.stringify({key: null})
"{"key":null}"
JSON.stringify({key: undefined})
"{}"
```
We don't want `undefined` value to be serialized.

vinyll:
Yes, it waz fine with me ;)

js/mixins/uploader.js
jphnoel: According to https://developers.google.com/transit/gtfs/examples/gtfs-feed, GTFS is a format but not a file extension. A GTFS should be stored in a .txt file.

js/mixins/uploader.js
jphnoel: But it is defined here: http://fileinfo.com/extension/geojson

OK, let's take it !

PR #503, opened 2016-06-28
Front init

noirbizarre:
This PR:
- ensure all views are Vue.js based (allowing porting shared components)
- ensure all views have a common initialization
- fix and simplify common dependencies and style handling
- compile chunks in `udata/static/chunks` for easier cleanup

This is a start base for smaller PR.


js/front/apidoc.js
```diff
@@ -0,0 +1,63 @@
+/**
+ * Display a SwaggerUI documentation
+ */
+import 'front/bootstrap';
+
+import 'less/udata/swagger.less';
+
+import Vue from 'vue';
+
+import $ from 'expose?$!expose?jQuery!jquery';
```
davidbgk: Just curious about that syntax.

js/front/apidoc.js
```diff
@@ -0,0 +1,63 @@
...
+import 'script!swaggerui/lib/underscore-min';
+import 'script!swaggerui/lib/backbone-min';
+import 'script!swaggerui/lib/jsoneditor.min';
+
+import 'script!swaggerui/swagger-ui.min';
+SwaggerUi = window.SwaggerUi;
+
+// Marked compatibility for SwaggerUI
+window.marked = commonmark;
+marked.setOptions = function() {};
```
davidbgk: `_ => {}` do we turn the file into ES6 too?

js/front/user.js
```diff
@@ -0,0 +1,56 @@
...
+                currentTab: 0,
+                // API,
+            };
+        },
+        ready() {
+            log.debug('User display page');
+
+            // Restore tab from hash
+            if (location.hash !== '') {
+                this.$refs.tabs.$children.some((tab, index) => {
```
davidbgk: I see duplicated code here, can we factorize?

udata/templates/apidoc.html
```diff
@@ -6,20 +6,13 @@
     'keywords': ['API', 'documentation', 'swagger'],
 } %}
 
+{% set bundle = 'apidoc' %}
```
davidbgk: Does the page load/become responsive faster with that version?

davidbgk:
We really need to sprint on adding tests for the frontend 😨

js/front/apidoc.js
noirbizarre: SwaggerUI requires jQuery to be a global. This syntax is from the webpack expose plugin (expose both `$` and `jQuery`)

js/front/apidoc.js
noirbizarre: It's already turned (there is import). The file has only been moved.

js/front/user.js
noirbizarre: OK, but not in this PR

udata/templates/apidoc.html
noirbizarre: Not this page, it has been untouched

noirbizarre:
Yes, there is some unit test (the specs folder) but we need end to end testing.
I think they should live outside the udata core repository as end to end tests relies on rendering and rendering is specific to each instance (theme, special pages...)

PR #504, opened 2016-06-30
Fix territory parent handling

noirbizarre:


udata/core/spatial/models.py
```diff
@@ -116,7 +116,7 @@ def child_level(self):
         HANDLED_LEVELS = current_app.config.get('HANDLED_LEVELS')
         try:
             return HANDLED_LEVELS[HANDLED_LEVELS.index(self.level) - 1]
-        except IndexError:
+        except (IndexError, ValueError):
```
davidbgk: When does it happen? Maybe `current_app.config.get('HANDLED_LEVELS', tuple())` is more explicit if you try to address that case?

noirbizarre: It does happen each time we use these methods with a zone not in `HANDLED_LEVELS`.
It breaks suggestion and all geozone related feature right now.
It's not related to `HANDLED_LEVELS` being set or not but to `index()` raising `ValueError` when value is not found in tuple/list

davidbgk:
r+

PR #506, opened 2016-07-01
Site search component

noirbizarre:
This pull request port the site search as a Vue.js component and drop Typeahead and all legacy site search dependencies (this PR depends on #503).
Doing so it:
- simplify site search:
  - no more Typeahead monkey patching
  - single encapsulated component
  - reuse the API plugin (fetch/promise)
  - start to drop server-side search state/scope handling
- improve style (wider dropdown if possible)
- give visual feedback on querying status

## Before

Without theme (was buggy):
capture d ecran de 2016-07-01 12-07-49

With `gouvfr` theme:
capture d ecran de 2016-07-01 00-53-39

## After

Without theme:
capture d ecran de 2016-07-01 12-02-06

With `gouvfr` theme:
capture d ecran de 2016-06-30 22-56-44


js/components/dashboard/graphs.vue
```diff
@@ -1,32 +1,37 @@
...
+    .graphs-chart .box-title {
+        margin-top: 0;
+    }
 }
 </style>
 <template>
-    <div class="row small-boxes">
+<div class="graphs">
+    <div v-if="metrics.loading" class="text-center"><span class="fa fa-4x fa-cog fa-spin"></span></div>
+    <divv-if="!metrics.loading" class="row small-boxes">
```
davidbgk: Missing space, does it work without?

js/components/site-search.vue
```diff
@@ -0,0 +1,292 @@
...
+            const query = value.trim()
+            if (!query || query.length < this.minLength) {
+                this.reset()
+                return false
+            }
+
+            this.current = -1;
+            this.show = true;
+            
+            this.groups.forEach(group => {
```
davidbgk: I wonder if we should wait with a `Promise.all()` that all data is retrieved prior to display it. From my user experience it can be hard to click on the right item when the list is still loading and sections are jumping down at each new api response.

js/utils.js
```diff
@@ -48,11 +48,28 @@ export function setattr(obj, name, value) {
     obj[name] = value;
 }
 
+/**
+ * Parse a query string into an object
+ * @param  {String} qs  A querystring, with or without the leading '?'
+ * @return {Object}     The parsed query string
+ */
+export function parseQS(qs) {
```
davidbgk: Why not https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams ? :)

davidbgk:
As mentioned in https://github.com/etalab/udata-gouvfr/pull/117 I think that we should find a way to be better at [cutting the mustard](https://sixtwothree.org/posts/cutting-the-mustard-revisited), there are even [more robust patterns](https://www.filamentgroup.com/lab/enhancing-optimistically.html) nowadays. Not sure where to discuss this but I'm scared to make that core functionality JS dependent only.

js/components/site-search.vue
noirbizarre: I tried but right now, territory suggest is too long to respond for the suggestions to show before submitting or typing new chars :/

js/utils.js
noirbizarre: Because I didn't know it ! ;)
Sadly no IE and no Safari support :/

noirbizarre:
Interesting, but if somehow JS don't load, these patterns doesn't apply too.
I was thinking in using a `<noscript>` tag with fallback raw input.

Right now, core features are already mostly JS depedendant (bootstrap, jquery, typeahead...) and the site is totally broken without JS (no modals so no actions, no menus, no suggest, no facets, no admin...)

js/utils.js
davidbgk: Maybe babeljs knows how to deal with it?!

davidbgk:
If JS doesn't load, the pattern works because without JS there is the default form that is hidden by JS once the script is loaded with an additional CSS class for instance.

I know that the site is currently broken without JS and that makes me sad, never to late to provide a decent experience for more people :)

js/utils.js
noirbizarre: BabelJS doesn't know about browser, DOM or anything related to this, it only transpile ES6 to ES5. If we want to use this API, we have to find a polyfill (which might be very close from this snippet and potentially bigger).
I can't wait for browser to have native support for the whole ES6 stack and the new DOM/HTML5 APIs !!

davidbgk: Indeed, it's definitely bigger https://github.com/WebReflection/url-search-params

I wonder if there are things to pick in it though.

noirbizarre:
Hum, I think we should discuss this approach globally and in a stable state.
My first priority before perfection is to reduce complexity and code base (a drop that goddamn jQuery! :wink:)

Then we can see how we can improve the code quality, the user experience...

In this PR, everything is supported by our officially supported browsers, so to me, the only case to handle is "there is no javascript". About this, we can:
- use the dedicated `<noscript>` tag
- use one of the articles' patterns
- assume that javascript is mandatory

In the case of a broken JS build (which should not happen and more over not being deployed in production), I like the idea that this is visible from the start: you won't have a search field!

Now, I think the idea is interesting but might need its own PR to do it properly, globally and once and for all. What do you think ?

PR #507, opened 2016-07-01
Port tasks to invoke 0.13

noirbizarre:
The latest invoke version (0.13.0) make all merge `@task` and `@ctask` so all tasks take the context as parameter


davidbgk:
r+

PR #508, opened 2016-07-01
Drop bootstrap tooltips and use tooltip directive everywhere

noirbizarre:
This PR make use of the v-tooltip directive everywhere and drops bootstrap tooltips

### DIFF
--- a/js/components/harvest/job.vue
+++ b/js/components/harvest/job.vue
@@ -30,18 +30,15 @@
             </dd>
             <dt v-if="job.items && job.items.length">{{ _('Items') }}</dt>
             <dd v-if="job.items && job.items.length">
-                <span class="text-warning"
-                    data-toggle="tooltip" data-placement="top"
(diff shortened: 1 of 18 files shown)


js/components/harvest/job.vue
```diff
@@ -60,7 +57,6 @@ import {
 } from 'models/harvest/job';
 import {STATUS_CLASSES, STATUS_I18N} from 'models/harvest/item';
 import {PageList} from 'models/base';
-import $ from 'jquery';
```
davidbgk: A kitten has just been resurrected! 💛

davidbgk:
r+❤️

js/components/harvest/job.vue
noirbizarre: There will be a lot of resurrected kittens until this evening :wink:  :tada: :confetti_ball:

PR #511, opened 2016-07-04
Post cleanup merge fixes

noirbizarre:
Small fixes:
- missing bundles (were stashed)
- loading state on timeline and graphs
- typo on graphs
- nested bootstrap rows and columns

### DIFF
--- a/udata/templates/organization/dashboard.html
+++ b/udata/templates/organization/dashboard.html
@@ -14,14 +14,10 @@
 {% endblock %}
 
 {% block main_content %}
-<section>
-    <div class="container">
-        <dashboard-graphs object-id="{{ org.id }}"></dashboard-graphs>
-        <div class="col-sm-9" id="graphs" object-id="{{ org.id }}">
-            <div class="text-center"><span class="fa fa-4x fa-cog fa-spin"></span></div>
-        </div>
-    </div>
-</section>
+<dashboard-graphs object-id="{{ org.id }}">
+    {# Load spinner until JS is ready #}
+    <div class="text-center"><span class="fa fa-4x fa-cog fa-spin"></span></div>
+</dashboard-graphs>
 {% endblock %}
 
 {% block after_content %}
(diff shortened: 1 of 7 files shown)


udata/templates/organization/dashboard.html
```diff
@@ -14,14 +14,10 @@
...
-<section>
-    <div class="container">
-        <dashboard-graphs object-id="{{ org.id }}"></dashboard-graphs>
-        <div class="col-sm-9" id="graphs" object-id="{{ org.id }}">
-            <div class="text-center"><span class="fa fa-4x fa-cog fa-spin"></span></div>
-        </div>
-    </div>
-</section>
+<dashboard-graphs object-id="{{ org.id }}">
+    {# Load spinner until JS is ready #}
```
davidbgk: 👍

udata/templates/organization/display_base.html
```diff
@@ -1,6 +1,7 @@
 {% extends theme("layouts/1-column.html") %}
 {% from theme('macros/follow.html') import follow_btn with context %}
 
+{% set bundle = 'organization' %}
```
davidbgk: Isn't the `bundle` declaration redundant with the `body_class` one? Can we merge these statements into one?

davidbgk:
r+wc

udata/templates/organization/dashboard.html
```diff
@@ -14,14 +14,10 @@
...
+    <div class="text-center"><span class="fa fa-4x fa-cog fa-spin"></span></div>
```
vinyll: This being used in multiple places, couldn't we do a `<div class=spinner></div>` instead of this long bootstrap combo?

davidbgk: Or better, a `<spinner>` component ;)

udata/templates/organization/display_base.html
noirbizarre: It will need some work but it can improve readability. Right now it's not connected:
- `bundle` specify JS/style dependencies
- `body_class` only add a class.
Some pages share a common `body_class` with different `bundle`.

udata/templates/organization/dashboard.html
vinyll: I guess a component does not make much sense here as it is waiting for JS to be ready and the component is a JS

noirbizarre: @davidbgk A spinner component is possible, but the way it's used (as a noscript fallback) we can't use a component.
@vinyll The `<div class="spinner">` might be a good idea, but I don't know if centering + icon can be performed with a single class. I'll try

PR #512, opened 2016-07-04
Upgrade js dependencies

noirbizarre:
This PR upgrade some JS dependencies.
Not updated to latest version:
- Swagger client: needs deeper testing because of one failing test (add badge)
- vue-loader 2.x & dependencies (only meant for vue.js 2.0) upgraded to latest 1.x
- Chart.js 2.0: API changes, upgraded to latest 1.x

PR #513, opened 2016-07-04
Improve Swagger Api lazy handling

noirbizarre:
This PR:
- drop jquery event for ready status
- remove the need to wait API being ready on views initialization
- ensure models empty models are correctly set on API ready
- ensure API calls wait for API being ready

This PR is built on #512

### DIFF
--- a/js/api.js
+++ b/js/api.js
@@ -44,7 +44,9 @@ const API = new SwaggerClient({
     url: config.api_specs,
     useJQuery: true,
     success: function() {
-        $(this).trigger('built');
+        this.readyCallbacks.forEach(callback => {
+            callback(this);
+        });
     },
     progress: function(msg) {
         log.debug(msg);
@@ -54,6 +56,8 @@ const API = new SwaggerClient({
     }
 });
 
+API.readyCallbacks = [];
+
 /**
  * Resolve a definition from a $ref
  * @param  {String} $ref The reference string
@@ -62,6 +66,15 @@ const API = new SwaggerClient({
 API.resolve = function($ref) {
     const def = $ref.split('#')[1].replace('/definitions/', '');
     return this.definitions[def];
+};
+
+/**
+ * Register a callback to be called when API is ready.
+ * The callback should have the following signature: function(API)
+ * @param  {Function} callback A callback to call when API is ready.
+ */
+API.onReady = function(callback) {
+    this.readyCallbacks.push(callback);
 };
 
 export default API;
--- a/js/api.js
+++ b/js/api.js
@@ -43,12 +43,12 @@ const API = new SwaggerClient({
     debug: config.debug,
     url: config.api_specs,
     useJQuery: true,
-    success: function() {
+    success() {
         this.readyCallbacks.forEach(callback => {
             callback(this);
         });
     },
-    progress: function(msg) {
+    progress(msg) {
         log.debug(msg);
     },
     authorizations: {
--- a/js/api.js
+++ b/js/api.js
@@ -74,7 +74,13 @@ API.resolve = function($ref) {
  * @param  {Function} callback A callback to call when API is ready.
  */
 API.onReady = function(callback) {
-    this.readyCallbacks.push(callback);
+    if (this.ready && this.isBuilt) {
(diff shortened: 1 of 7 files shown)


js/front/organization/index.js
```diff
@@ -19,61 +19,55 @@ import Tab from 'components/tab';
...
-            log.debug('Organization display page');
-
-            // Restore tab from hash
-            if (location.hash !== '') {
-                this.$refs.tabs.$children.some((tab, index) => {
-                    if (`#${tab.id}` === location.hash) {
-                        this.currentTab = index;
-                        return true;
-                    }
+new Vue({
```
noirbizarre: Misleading diff, only removed the `$(API).on('built'...` and so deindent.

js/front/user.js
```diff
@@ -13,44 +13,35 @@ import FollowButton from 'components/buttons/follow.vue';
...
-        },
-        watch: {
-            /**
-            * Set current tab ID as location hash
-            * @param  {Number} index The new tab index
-            */
-            currentTab(index) {
-                location.hash = this.$refs.tabs.$children[index].id;
-            },
+new Vue({
```
noirbizarre: Misleading diff, only removed the `$(API).on('built'...` and so deindent.

js/api.js
```diff
@@ -44,7 +44,9 @@ const API = new SwaggerClient({
     url: config.api_specs,
     useJQuery: true,
     success: function() {
```
davidbgk: `success()`?

js/api.js
```diff
@@ -44,7 +44,9 @@ const API = new SwaggerClient({
...
-        $(this).trigger('built');
+        this.readyCallbacks.forEach(callback => {
```
davidbgk: `this.readyCallbacks.forEach(callback)` might work

js/models/base.js
```diff
@@ -99,17 +99,19 @@ export class Base {
...
-        const namespace = parts[0];
-        const method = parts[1];
-        const operation = API[namespace][method];
-
-        if (this.$options.mask && !('X-Fields' in data)) {
-            data['X-Fields'] = mask(this.$options.mask);
-        }
+    $api(endpoint, data, on_success, on_error = () => {}) {
+        API.onReady(() => {
+            const parts = endpoint.split('.');
```
davidbgk: `const {namespace, method} = endpoint.split('.')` should do the deconstruction just fine.

js/models/base.js
```diff
@@ -99,17 +99,19 @@ export class Base {
...
+            const namespace = parts[0];
+            const method = parts[1];
+            const operation = API[namespace][method];
+
+            if (this.$options.mask && !('X-Fields' in data)) {
+                data['X-Fields'] = mask(this.$options.mask);
+            }
 
-        return operation(data, on_success.bind(this), on_error.bind(this));
+            operation(data, on_success.bind(this), on_error.bind(this));
```
davidbgk: I wonder if the `this` binding is required here. It probably depends on the way `on_success` has been declared, tricky :)

js/models/base.js
```diff
@@ -156,8 +158,10 @@ export class Model extends Base {
 
     on_fetched(data) {
         for (const prop in data.obj) {
-            const value = data.obj[prop];
-            this._set(prop, value);
+            if (data.obj.hasOwnProperty(prop)) {
+                const value = data.obj[prop];
```
davidbgk: Do we really need the `const` definition?

davidbgk:
r+wc, you're a machine. 😱

js/models/base.js
noirbizarre: Oh, I forgot it's possible now :tada: !

js/api.js
noirbizarre: No, the callback is not defined here :/

js/models/base.js
noirbizarre: Sadly it's required because Swagger apply a scope if not binded.

js/api.js
davidbgk: But `readyCallbacks` is a list of callback ready to be called when passing through the `forEach` right?

PR #517, opened 2016-07-08
Small dataset display JS fixes

noirbizarre:
This PR:
- fix `ReferenceError` on map
- fix extras display
- prevent checkUrl failure on non-resource list item


davidbgk:
r+

PR #524, opened 2016-07-13
Switch to Elasticsearch DSL

noirbizarre:
This pull request switch from a [custom DSL for ElasticSearch](https://github.com/opendatateam/udata/tree/dev/udata/search) to official [Elasticsearch DSL](https://elasticsearch-dsl.readthedocs.io/).

Elasticsearch DSL allows progressive migration with `Search.from_dict()` and `Search.to_dict()` which is the chosen way here.

Tasks to perform:
- [x]  Switch from [`SearchAdapter`](https://github.com/opendatateam/udata/blob/dev/udata/search/adapter.py) to [ElasticDSL persitence](https://elasticsearch-dsl.readthedocs.io/en/latest/persistence.html)
- [x] Switch from [`SearchQuery`](https://github.com/opendatateam/udata/blob/dev/udata/search/query.py) to [`Search` object](https://elasticsearch-dsl.readthedocs.io/en/latest/search_dsl.html#the-search-object)
  - [x] aggregations
  - [x] pagination
  - [x] sorting
  - [x] filtering
  - [x] request/query string parsing
  - [x] Custom scoring (decay functions, boosts...)
- [x] Switch from [`SearchResult`](https://github.com/opendatateam/udata/blob/dev/udata/search/result.py) to [Response object](https://elasticsearch-dsl.readthedocs.io/en/latest/search_dsl.html#response)
- [x] Move `SearchIterator` on top of ElasticSearch DSL `scan()`
- [x] Switch from [`multisearch` ](https://github.com/opendatateam/udata/blob/dev/udata/search/__init__.py#L128-L142) to [`MultiSearch`](https://elasticsearch-dsl.readthedocs.io/en/latest/search_dsl.html#multisearch)
- [x] Switch from [`suggest`](https://github.com/opendatateam/udata/blob/dev/udata/search/__init__.py#L145-L157) to [Suggestions](https://elasticsearch-dsl.readthedocs.io/en/latest/search_dsl.html#suggestions)
- [x] update search templates and macros
- [x] switch to Elasticsearch DSL facets:
  - [x] Terms facet
  - [x] Model terms facet
  - [x] Range facet
  - [x] Temporal coverage facet
  - [x] Bool facet
- [x] metrics support
- [x] API search parser
- [x] group all parts into a single search package
- [x] remove label parameter from `labelize`
- [x] Handle request parameters errors:
  - [x] Fix `to_url`
  - [x] error 400 on front
  - [x] error 400 on API (aka. request parser format validation ;))


davidbgk:
Great, thanks for pushing and making tests pass with aggregations ;)

noirbizarre:
You're welcome ! There's still a lot of work (I'm still working on it)! ;)

udata/search/fields.py
```diff
@@ -129,6 +130,12 @@ def to_query(self, size=20, args=None, **kwargs):
                 [args] if isinstance(args, basestring) else args)
         return query
 
+    def to_aggregations(self, name, size=20, *args):
```
davidbgk: Maybe we should use an explicit `exclude` parameter here?

udata/search/query.py
```diff
@@ -63,14 +66,19 @@ def iter(self):
...
+        qs = self.get_sort(qs)
+        # Pagination
+        start = (self.page - 1) * self.page_size
+        qs = qs[start:start + self.page_size]
+        # Aggregations
+        qs = self.build_aggregations(qs)
+        # don't return any fields, just the metadata
+        qs = qs.fields([])
+
+        body = qs.to_dict()
```
davidbgk: Instead of the `to_dict`/`from_dict`, we can use the [update_from_dict](http://elasticsearch-dsl.readthedocs.io/en/latest/search_dsl.html#serialization-and-deserialization) method here to optionally add scoring functions.

noirbizarre:
Fixed ;)

udata/search/analyzers.py
```diff
@@ -72,3 +72,18 @@
 
 simple = analyzer('simple')
 standard = analyzer('standard')
+
+
+def analysis_settings():
```
davidbgk: Do we really need a function here?

udata/core/search/commands.py
```diff
@@ -80,11 +76,15 @@ def init(name=None, delete=False, force=False):
...
+    for adapter_class in adapter_catalog.values():
+        index.doc_type(adapter_class)
+    index.create()
+
     adapters = adapter_catalog.items()
     # Ensure that adapters are indexed in the same order.
     adapters.sort()
-    for model, adapter_class in adapters:
-        adapter_class.init(using=es.client, index=es.index_name)
+    for model, adapter_class in reversed(adapters):
```
davidbgk: If we keep the `reversed` and that may be a good thing to init geozones at last, we have to document that line.

udata/search/__init__.py
```diff
@@ -108,14 +108,24 @@ def get_i18n_analyzer():
...
+            adapter = adapter_class.from_model(obj)
+            adapter.save(using=es.client, index=es.index_name)
+        except:
+            log.exception('Unable to index %s "%s"',
+                          model.__name__, str(obj.id))
+    elif adapter_class.get(obj.id, index=es.index_name):
+        log.info('Unindexing %s (%s)', model.__name__, obj.id)
+        try:
+            adapter = adapter_class.from_model(obj)
+            adapter.delete(using=es.client, index=index_name)
```
davidbgk: Are you sure of the `index_name`? Where is it coming from?

noirbizarre: No, it's an error, here it should be `es.index_name`

udata/search/analyzers.py
noirbizarre: You're right, exposing only analyzers and filters should be sufficient

udata/core/search/commands.py
noirbizarre: I just pushed the work as it was. If we want to pin an order, maybe we should be more specific and explicit.

udata/core/search/views.py
```diff
@@ -9,34 +9,29 @@
...
-            search.SearchQuery(Dataset, **params),
-            search.SearchQuery(Reuse, **params),
-            search.SearchQuery(Organization, **params),
-            search.SearchQuery(User, **params)
-        ]
-        results_labels = ['datasets', 'reuses', 'organizations', 'users']
-    results = search.multisearch(*search_queries)
-    context = dict(zip(results_labels, results))
+        types = ['datasets', 'reuses', 'organizations', 'users']
+    models = [MAPPING[typ] for typ in types]
```
davidbgk: To avoid the mapping, why don't we use directly the models in the lists and then zip with the `__class__.__name__`?

udata/search/__init__.py
```diff
@@ -195,32 +186,38 @@ def facets_kwargs(adapter, facets=None):
...
-def multisearch(*queries):
+def multisearch(*models, **params):
     ms = MultiSearch(using=es.client, index=es.index_name)
-    for query in queries:
-        qs = query.build_query()
-        ms = ms.add(qs)
+    queries = []
+    for model in models:
+        s = search_for(model, **params)
+        ms = ms.add(s._s)
```
davidbgk: Isn't it weird that we have to use the `_s`?

udata/search/__init__.py
```diff
@@ -195,32 +186,38 @@ def facets_kwargs(adapter, facets=None):
...
+        queries.append(s)
     responses = ms.execute()
-    return responses
+    return [
+        SearchResult(query, response)
+        for response, query in zip(responses, queries)
```
davidbgk: What about swapping values within the zip to be consistent with the order of params for SearchResult?

udata/core/search/views.py
noirbizarre: I thought about this, but right no it means changing all templates variables names (or simply construct the plural by adding an "s").

udata/search/__init__.py
noirbizarre: I think we will end by submitting some PRs to elasticsearch-dsl project. Exposing the constrcuted search object seems logical to me too.

udata/search/__init__.py
noirbizarre: You're right. It can be shorter too:
```python
return [SearchResult(*params) for params in zip(queries, responses)]
```

noirbizarre:
18 failures, we are getting close to the goal ! 🎉 🎊

udata/search/__init__.py
```diff
@@ -158,19 +158,50 @@ def register(adapter):
...
+                 for s in sorts]
+        self.sorts = [
+            {self.adapter.sorts[s].field: d}
+            for s, d in sorts if s in self.adapter.sorts
+        ]
+
+    def extract_pagination(self, params):
+        '''Extract and build pagination from parameters'''
+        try:
+            self.page = max(int(params.pop('page', 1) or 1), 1)
```
davidbgk: Not a huge fan of that line, proposal: `self.page = 'page' in params and int(params.pop('page')) or 1` without the `expect` that catches all.

udata/search/__init__.py
```diff
@@ -158,19 +158,50 @@ def register(adapter):
...
+        except:
+            self.page = 1
+        try:
+            self.page_size = int(
```
davidbgk: Same here and I guess that we don't need to cast `DEFAULT_PAGE_SIZE` to an `int` too?

udata/search/__init__.py
```diff
@@ -158,19 +158,50 @@ def register(adapter):
 
 
 class UdataFacetedSearch(FacetedSearch):
+    model = None
+    adapter = None
+
     def __init__(self, params):
-        self.sort = params.pop('sort', None)
+        self.extract_sort(params)
```
davidbgk: 👍 for extracting extractions (so meta!)

noirbizarre: :)
I wanted to highlight the fact that parameters are removed

udata/search/__init__.py
noirbizarre: The casting is not for `DEFAULT_PAGE_SIZE` but for the params value which might be a string.

udata/search/__init__.py
noirbizarre: OK, it was just a copy paste from the previous code. I can improve its readability 👍

udata/search/result.py
```diff
@@ -8,31 +8,49 @@
...
+from elasticsearch_dsl.result import Response
+
 from udata.utils import Paginable
 
 
 log = logging.getLogger(__name__)
 
 
-class SearchResult(Paginable):
+class SearchResult(Paginable, Response):
```
davidbgk: Rename to `SearchResponse` for consistency?

udata/search/query.py
```diff
@@ -17,182 +19,83 @@
...
-                'function_score': {
-                    'query': self.get_query(),
-                    'functions': self.get_score_functions(),
-                }
-            }})
-        else:
-            qs.update_from_dict({'query': self.get_query()})
-
-        return qs
+class SearchQuery(FacetedSearch):
```
davidbgk: Rename to `CustomSearch` for consistency?

udata/search/query.py
```diff
@@ -17,182 +19,83 @@
...
+
+    def search(self):
+        """
+        Construct the Search object.
+        """
+        s = Search(doc_type=self.doc_types, using=es.client, index=es.index_name)
+        s = s.fields([])  # don't return any fields, just the metadata
+        s = s.sort(*self.sorts)
+        s = s[self.page_start:self.page_end]
+        return s.response_class(partial(SearchResult, self))
```
davidbgk: The use of `partial` always needs a comment ;)

noirbizarre: Added ;)

udata/tests/test_search.py
```diff
@@ -122,6 +123,11 @@ def get_query(facet_search):
 class SearchQueryTest(SearchTestMixin, TestCase):
     maxDiff = None
 
+    def assert_dict_equal(self, d1, d2, *args, **kwargs):
```
davidbgk: I think that we should stick to the `TestCase` case with `assertDictEqual` for consistency. By the way, do we need the `*args, **kwargs` parameters?

What about putting this helper at the `TestCase` level for it to be reuse elsewhere?

udata/core/topic/views.py
```diff
@@ -52,42 +61,27 @@ def display(topic):
 @blueprint.route('/<topic:topic>/datasets')
 def datasets(topic):
     kwargs = multi_to_dict(request.args)
-    kwargs.update(topic=topic)
-    query = TopicSearchQuery(Dataset, facets=True, **kwargs)
+    s = topic_search_for(topic, DatasetSearch, facets=True, **kwargs)
```
davidbgk: Not fond of these one letter vars. Proposition `topic_search` (same above)

udata/core/search/commands.py
```diff
@@ -1,120 +0,0 @@
-# -*- coding: utf-8 -*-
```
davidbgk: Moving these files doesn't display the related diff 😭

udata/search/__init__.py
```diff
@@ -64,97 +82,155 @@ def index_name(self):
...
+    queries = []
+    for model in models:
+        s = search_for(model, **params)
+        ms = ms.add(s._s)
+        queries.append(s)
+    responses = ms.execute()
     return [
-        SearchResult(query, response)
-        for response, query in zip(result['responses'], queries)
+        SearchResult(query, response._d_)
```
davidbgk: La réponse \_d\_.

udata/search/adapter.py
```diff
@@ -72,19 +58,67 @@ def completer_tokenize(cls, value, min_length=3):
         ]))
         return list(set([value] + tokens + [' '.join(tokens)]))
 
+    @classmethod
+    def facet_search(cls, *facets):
```
davidbgk: Add a comment to explain why we have to do that.

udata/search/analysis.py
```diff
@@ -0,0 +1,83 @@
...
+fr_elision = token_filter(
+    'fr_elision',
+    type='elision',
+    articles=[
+        'l', 'm', 't', 'qu', 'n', 's', 'j', 'd', 'c',
+        'jusqu', 'quoiqu', 'lorsqu', 'puisqu'
+    ]
+)
+
+# Still in use?
```
davidbgk: Doesn't look to, let's remove it!

udata/search/fields.py
```diff
@@ -2,355 +2,250 @@
...
+            return value
+        try:
+            return ObjectId(value)
+        except Exception:
+            raise ValueError('"{0}" is not valid identifier'.format(value))
+
+
+
+class RangeFacet(Facet, DSLRangeFacet):
+    def __init__(self, **kwargs):
```
davidbgk: We need to document that labels/ranges split on why and how to use it.

udata/search/fields.py
```diff
@@ -2,355 +2,250 @@
...
+
+
+    def get_value_filter(self, filter_value):
+        '''
+        Fix here until upstream PR is merged
+        https://github.com/elastic/elasticsearch-dsl-py/pull/473
+        '''
+        self.validate_parameter(filter_value)
+        f, t = self._ranges[filter_value]
+        limits = {}
```
davidbgk: Add a comment on the reason why `gte` on one side and `lt` on the other.

udata/search/views.py
```diff
@@ -0,0 +1,37 @@
...
+    params = multi_to_dict(request.args)
+    params['facets'] = True
+    # We only fetch relevant data for the given filter.
+    if 'tag' in params:
+        types = ['datasets', 'reuses']
+    elif 'badge' in params:
+        types = ['datasets', 'organizations']
+    else:
+        types = ['datasets', 'reuses', 'organizations', 'users']
+    models = [MAPPING[typ] for typ in types]
```
davidbgk: We have to find a consistent `type_` vs. `_type` vs. `typ` for that one 😉

udata/templates/search.html
```diff
@@ -71,9 +71,9 @@
 </ul>
 <div class="label-list">
     {{ s.filter_label('tag', _('Tag'), ficon('tag')) }}
-    {{ s.filter_label('badge', _('Badge'), ficon('bookmark'),
+    {#{ s.filter_label('badge', _('Badge'), ficon('bookmark'),
```
davidbgk: Remove or update?

udata/search/__init__.py
noirbizarre: :)

udata/search/views.py
noirbizarre: Agreed: might need another PR as it's application wide

udata/search/fields.py
noirbizarre: Done 👍

udata/core/search/commands.py
noirbizarre: Mostly rewritten to use Elasticsearch DSL tooling

davidbgk:
👍 for review comments, now can we have an independent review @opendatateam/etalab ? :)

vinyll:
Hey guys! Great job for this crazy work, you just made it!!! :D 🎂

PR #528, opened 2016-07-26
Disable production flag until a better solutions is found

noirbizarre:
This PR fix lazy bindings not being triggered in production mode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "main": "index.js",
   "scripts": {
     "assets:watch": "webpack -c -d  --progress --watch",
-    "assets:build": "webpack -p --progress",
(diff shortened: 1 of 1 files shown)


davidbgk:
What does it do exactly? What is the goal of these lazy bindings?

noirbizarre:
Lazy bindings is what permit to Vue.js to update values when values are set after component initialization.
I can't explain why but on the `master` branch, the webpack `-p/--production` flag breaks it.
Vue.js is compiled differently with the production flag.
The issue is not present on the `dev` branch

davidbgk:
Alright, let's remove that :/ r+

PR #530, opened 2016-08-03
Ensure content type is never None

noirbizarre:


davidbgk:
Maybe add a test for that use-case?

noirbizarre:
Yes but before I'll need to fix the html5lib breakage

noirbizarre:
I spent 2 hours trying to simulate the case for testing and now I can say: there is no way to test it!
But this solution is a double safety:

- mimetype should never be `None`
- if it's the case for any reason, we ensure it's a string

PR #488, opened 2016-06-09
Add libxml2 and LZMA in system-dependencies.md

michelbl:
Solve python local dependencies install issue on Ubuntu/Debian.

PR #531, opened 2016-08-03
List a few more needed packages in the documentation

odtvince:


noirbizarre:
Can you integrate/merge #488 ? Same topic and the full dependency list is in the issue.

PR #533, opened 2016-08-31
Minor improvements on the installation doc and on .gitignore

jphnoel:


.gitignore
```diff
@@ -65,3 +65,9 @@ coverage.xml
 !.gitignore
 .noseids
 Procfile
+
+# Local configuration file of udata
+udata.cfg
+
+# Local data directory
+data
```
davidbgk: I use to put the configuration file and data directory one level up in my filesystem but if it can ease the first installation I'm OK with that.

docs/system-dependencies.md
```diff
@@ -41,7 +41,7 @@ $ docker-compose up
 On the very first run it will download and install Docker images which takes a while depending of your connection.
 
 !!! warning
-    Test your _docker-compose_ is running successfully entering `curl http://locahost:9200`.
+    Test your _docker-compose_ is running successfully entering `curl http://localhost:9200`.
```
davidbgk: 👍

davidbgk:
r+wc ([explanation](http://webdev-bootcamp-osmose.readthedocs.io/en/overhaul/reference/glossary.html), not sure it's documented somewhere, if you have the time you can add it to https://github.com/opendatateam/udata/blob/dev/docs/contributing-guide.md ?)

.gitignore
noirbizarre: Same as @davidbgk for me. The default values should be in `.gitignore`. Custom/specific files for each of us should be in your `.git/ignore` file.
I can't remember, are `data` and `udata.cfg` the file created by default ?

jphnoel: As far as I remember, the 'data' directory has been created by default when I installed the platform. And it is explicitly referenced in docker-compose.yml.

'udata.cfg' is not really created by default, but, if you do not specify UDATA_SETTINGS, it is the file that udata looks for. It is then the default user configuration file.

For me, it is reasonable to put them in the global .gitignore.

noirbizarre: OK 👍

PR #535, opened 2016-09-12
Fix js/models/geogranularities double export default

noirbizarre:
This PR fix double "export default" in `js/models/geogranularities`
(Was breaking admin)


davidbgk:
How can we check that it's not broken again like this in the future?

noirbizarre:
Manual testing on deployment.
I think we can fail webpack the build (it shouldn't be seen as succesful when an entry doesn't compile)

PR #537, opened 2016-09-16
Add a link in the doc to a glossary of the PR-related terms we use

jphnoel:


davidbgk:
Testing the new approved stuff from Github in the meantime, nice :)

davidbgk:
Note that with the new “approved” thing from Github that's probably not relevant anymore, thoughts?

noirbizarre:
@davidbgk I just pust "r+" as an approval comment but as you said it's a little bit redundant.
Maybe we need drop this practice. In the meantime, I agree for this PR and we should remove it when we change our review process (its wording)

jphnoel:
I am a bit confused...
I understand that we are using the new GitHub approval process.
Why should we then add this link to a glossary that we do not want to use anymore ?
For me, we should not merge this PR, should we ?

Issue #536, opened 2016-09-14
Switch from microdata to JSON-LD


davidbgk:
## Context
Today, we embed some metadata with the microdata format, i.e. directly within the DOM. We use these metadata both for internal use (in the JS of the dataset frontend) and for external use (SEO). 

## Detailed Description
When we display the dataset page, we need to parse the DOM to retrieve all the metadata, for the modals, and to be able to check the availibility of the resources with Croquemort. 

Let us consider now a dataset with many resources (typically 300, which is a real use cas, encountered in Luxembourg): it does work, from a functional point of view, but it is much too slow.

## Possible Implementation
With [JSON-LD](https://developers.google.com/schemas/formats/json-ld) we should have much better performance.

Moreover, for W3C and Google, microdata is considered as deprecated, and should be replaced by JSON-LD.

Lastly, using microdata makes the templates verbose and harder to maintain.

davidbgk:
The performance issue was found by @opendatalu when there are more that 300 resources for a given dataset.

davidbgk:
Relevant files:

* https://github.com/opendatateam/udata/blob/dev/js/microdata.js
* https://github.com/opendatateam/udata/blob/dev/js/front/dataset/index.js
* https://github.com/opendatateam/udata/blob/dev/udata/templates/dataset/display.html#L58

jphnoel:
Other relevant files:
- https://github.com/opendatateam/udata/tree/dev/specs/fixtures/microdata
- https://github.com/opendatateam/udata/blob/dev/specs/microdata.specs.js

jphnoel:
Even more other relevant files:
- https://github.com/opendatateam/udata/blob/dev/udata/templates/dataset/resource/list-item.html
- https://github.com/opendatateam/udata/blob/dev/udata/templates/organization/display.html
- https://github.com/opendatateam/udata/blob/dev/udata/templates/reuse/display.html
- https://github.com/opendatateam/udata/blob/dev/udata/templates/user/base.html

PR #545, opened 2016-09-29
JSON-LD

jphnoel:
This PR is related to #536.

General remarks:
- **itemid** has been set in **@id**
- I removed the JS test of microdata and I did not add a new one. Indeed, it does not seem very useful to test **JSON.parse(document.querySelector(selector).text)**... Nevertheless, we could try to add a test to validate the modals, but it is probably too long in the frame of this issue. 
- I removed an **itemprop** in *post/display.html* which, apparently, was not used

Dataset:
- I removed **hasPart** and I replaced it by **distribution** (which used to be at the **resource** level, which was not consistent with what is mentioned in schema.org)
- I removed **isPartOf**
- I put the **PropertyValue** data directly in the dataset. It is not consistent with what is on schema.org, but it is the same thing for **checkum** and **checksumType** anyway.
- I fill the **author** field, even if the author is just a user, and not an organization

Reuse:
- I fill the **author** field, even if the author is just a user, and not an organization


js/front/dataset/index.js
```diff
@@ -171,13 +164,14 @@ new Vue({
...
-         * Asynchronuously check a displayed resource status
-         * @param  {Object} resource A resource as extracted from microdata
+         * Asynchronously check a displayed resource status
+         * @param  {Object} resource A resource as extracted from JSON-LD
          */
         checkResource(resource) {
             const url = parseUrl(resource.url);
-            const el = resource.$el.querySelector('.format-label');
-            const checkurl = resource.$el.dataset.checkurl;
+            const resource_el = document.querySelector("#resource-" + resource['@id'])
```
noirbizarre: With ECMA6, you can write:
```javascript
document.querySelector(`#resource-${resource['@id']}`)
```

js/front/dataset/resource-modal.vue
```diff
@@ -10,20 +10,16 @@
...
-          <dd v-if="resource.checksum">{{checksum}}</dd>
+          <dd v-if="resource.checksum">{{ resource.checksum }}</dd>
           <dt v-if="resource.dateCreated">{{ _('Created on') }}</dt>
           <dd v-if="resource.dateCreated"> {{ resource.dateCreated|dt }}</dd>
           <dt v-if="resource.dateModified">{{ _('Modified on') }}</dt>
           <dd v-if="resource.dateModified"> {{ resource.dateModified|dt }}</dd>
           <dt v-if="resource.datePublished">{{ _('Published on') }}</dt>
           <dd v-if="resource.datePublished"> {{ resource.datePublished|dt }}</dd>
-          <dt v-if="resource.downloads">{{ _('Downloads') }}</dt>
-          <dd v-if="resource.downloads"> {{ resource.downloads }}</dd>
```
noirbizarre: We still need a way to display this value

js/front/dataset/resource-modal.vue
```diff
@@ -10,20 +10,16 @@
           <dd><a :href="resource.url" @click="onClick">{{resource.url}}</a></dd>
           <dt v-if="resource.encodingFormat">{{ _('Format') }}</dt>
           <dd v-if="resource.encodingFormat">{{resource.encodingFormat}}</dd>
-          <dt v-if="resource.mimetype">{{ _('MimeType') }}</dt>
-          <dd v-if="resource.mimetype">{{resource.mimetype}}</dd>
```
noirbizarre: We still need a way to display this value

udata/core/dataset/views.py
```diff
@@ -86,8 +88,86 @@ def get_context(self):
         context['can_edit'] = DatasetEditPermission(self.dataset)
         context['can_edit_resource'] = ResourceEditPermission
         context['discussions'] = Discussion.objects(subject=self.dataset)
+
         return context
 
+    def get_json_ld(self):
```
noirbizarre: I'm not sure the views should host the JSON-LD construction as this is related to models.

udata/core/dataset/views.py
```diff
@@ -86,8 +88,86 @@ def get_context(self):
...
+                  "distribution": map(self.get_json_ld_resource, dataset.resources),
+                  # This value is not standard
+                  "extras": map(self.get_json_ld_extra, dataset.extras.items())
+        }
+
+        if dataset.license and dataset.license.url:
+            result["license"] = dataset.license.url
+
+        if dataset.organization:
+            view = OrganizationDetailView()
```
noirbizarre: Now, I'm sure it can't.
If the model have the method, you could directly write:
```python
if dataset.organization:
    author = dataset.organization.__jsonld__
```
This is an example with models having:
```python
@property  # even better with @cachedproperty
def __jsonld__(self):
    return {
        #JSONLD
    }
```
This is simply a proposal.

udata/core/dataset/views.py
```diff
@@ -86,8 +88,86 @@ def get_context(self):
...
+            author = view.get_json_ld(dataset.organization)
+        elif dataset.owner:
+            view = UserActivityView()
+            author = view.get_json_ld(dataset.owner)
+        else:
+            author = None
+
+        if author:
```
noirbizarre: You can directly white:
```python
if dataset.organization:
    result['author'] = dataset.organization.__jsonld__
```

udata/core/dataset/views.py
```diff
@@ -86,8 +88,86 @@ def get_context(self):
...
+            result['author'] = author
+
+        return result
+
+    @staticmethod
+    def get_json_ld_resource(resource):
```
noirbizarre: As stated before, should be on the `Resource` model directly

udata/core/organization/views.py
```diff
@@ -95,6 +95,19 @@ def get_context(self):
         })
         return context
 
+    def get_json_ld(self, org=None):
+        if org is None:
+            org = self.organization
+        prefix = 'Government' if org.public_service else ''
```
noirbizarre: Or, to avoid a string format later, directly:
```python
type = 'GovernmentOrganization' if org.public_service else 'Organization'
```

udata/core/reuse/views.py
```diff
@@ -97,11 +99,37 @@ def get_context(self):
         context.update(
             followers=followers,
             can_edit=ReuseEditPermission(self.reuse),
-            discussions=Discussion.objects(subject=self.reuse)
+            discussions=Discussion.objects(subject=self.reuse),
```
noirbizarre: No, keep the trailing coma.
When possible, we keep extra comas as Python allows it and it produces cleaner diffs when we add a line.

udata/core/organization/views.py
```diff
@@ -95,6 +95,19 @@ def get_context(self):
...
+        return {"@context": "http://schema.org",
```
noirbizarre: Insert a line break and indent only by 4 spaces:
```python
return {
    "@context": "..."
    # ...
}
```

udata/core/organization/views.py
```diff
@@ -95,6 +95,19 @@ def get_context(self):
...
+                "@type": '{}Organization'.format(prefix),
+                "alternateName": org.slug,
+                "logo": org.logo(external=True),
+                "url": url_for('organizations.show', org=org, _external=True),
+                "name": org.name,
+                "description": org.description
```
noirbizarre: Add a trailing coma

udata/core/reuse/views.py
```diff
@@ -97,11 +99,37 @@ def get_context(self):
...
         )
 
         return context
 
+    def get_json_ld(self):
+        reuse = self.reuse
+        result = {'@context': "http://schema.org",
```
noirbizarre: Same as above: line break and trailing coma

udata/core/dataset/views.py
```diff
@@ -86,8 +88,86 @@ def get_context(self):
...
+
         return context
 
+    def get_json_ld(self):
+        dataset = self.dataset
+
+        result = {"@context": "http://schema.org",
+                  "@type": "Dataset",
+                  "@id": str(dataset.id),
+                  "description": dataset.description,
```
noirbizarre: Careful: `description` is markdown so we should render it and striptags

udata/core/dataset/views.py
```diff
@@ -86,8 +88,86 @@ def get_context(self):
...
+                "interactionType": {
+                    "@type": "DownloadAction"
+                },
+                # We take resource.metrics.views if it exists
+                "userInteractionCount": getattr(resource.metrics, 'views', None) or ''
+            }
+        }
+
+        if resource.description:
+            result["description"] = resource.description
```
noirbizarre: `description` is markdown so we should render it and strip tags

udata/core/organization/views.py
noirbizarre: and  `description` is markdown so we should render it and strip tags

udata/core/reuse/views.py
```diff
@@ -97,11 +99,37 @@ def get_context(self):
...
+                  "alternateName": reuse.slug,
+                  "dateCreated": reuse.created_at.isoformat(),
+                  "dateModified": reuse.last_modified.isoformat(),
+                  "url": url_for('reuses.show', reuse=reuse, _external=True),
+                  "name": reuse.title,
+                  "description": reuse.description,
+                  "isBasedOnUrl": reuse.url
+            }
+
+        if reuse.organization:
```
noirbizarre: Same as above: Models should have the JSON-LD serialization, and you can directly affect to `result['author']`

udata/core/user/views.py
```diff
@@ -81,6 +82,22 @@ def get_context(self):
         return context
 
 
+    def get_json_ld(self, user=None):
+        if user is None:
+            user = self.user
+
+        result = {"@type": 'Person',
```
noirbizarre: Same: line break and trailing coma.

udata/core/user/views.py
```diff
@@ -81,6 +82,22 @@ def get_context(self):
...
+                  "@context": "http://schema.org",
+                  "image": placeholder(user.avatar, 'user'),
```
noirbizarre: Placeholders are only for visual comfort and consistence. They should not be used when no image is available for JSON-LD. Simply output `image` only when it's present.

udata/core/user/views.py
```diff
@@ -81,6 +82,22 @@ def get_context(self):
...
+                  "name": user.fullname,
+                  "description": user.about
```
noirbizarre: It's markdown: render it and strip tags

udata/templates/macros/metadata.html
```diff
@@ -43,4 +43,11 @@
 <meta name="check-urls" content="{{ 'true' if config.CROQUEMORT else 'false' }}" />
 <meta name="check-urls-ignore" content="{{ config.CROQUEMORT_IGNORE|tojson|urlencode }}" />
 <meta name="territory-enabled" content="{{ 'true' if config.ACTIVATE_TERRITORIES else 'false' }}">
+
+{% if json_ld %}
+<script type="application/ld+json">
+  {{ json_ld|safe }}
```
noirbizarre: It can bo one line instead of 3 (spaces and line breaks are not stripped)

udata/templates/organization/display.html
```diff
@@ -15,11 +15,8 @@
 {% endblock %}
 
 {% block main_content %}
-<div itemscope itemtype="http://schema.org/{% if org.public_service %}Government{% endif %}Organization">
-    <meta itemprop="alternateName" content="{{ org.slug }}" />
-    <meta itemprop="logo" content="{{ org.logo(external=True) }}" />
-    <meta itemprop="url" content="{{ url_for('organizations.show', org=org, _external=True) }}" />
-    <h1 itemprop="name">
+<div>
```
noirbizarre: Need to check, but I think the extras divs can be safely removed as they where present only for microdata scoping

udata/templates/reuse/display.html
```diff
@@ -54,18 +54,14 @@
 {% endblock %}
 
 {% block content %}
-<div itemscope itemtype="http://schema.org/CreativeWork">
+<div>
```
noirbizarre: Same, I think the extra div can be safely removed

udata/tests/frontend/__init__.py
```diff
@@ -12,3 +15,11 @@ def create_app(self):
         api.init_app(app)
         frontend.init_app(app)
         return app
+
+    def get_json_ld(self, response):
```
noirbizarre: Nice 👍

udata/tests/frontend/test_dataset_frontend.py
```diff
@@ -50,9 +51,28 @@ def test_render_list_empty(self):
...
+        resource = ResourceFactory()
+        author = UserFactory()
+        dataset = DatasetFactory(tags=['foo', 'bar'],
+                                 resources=[resource],
+                                 description='a&éèëù$£',
+                                 owner=author)
+        url = url_for('datasets.show', dataset=dataset)
+        response = self.get(url)
         self.assert200(response)
+        json_ld = self.get_json_ld(response)
```
noirbizarre: It would be a lot easier and faster to test on model instance and to only test presence in views

js/front/dataset/index.js
vinyll: const resource_el = document.querySelector(`#resource-${resource['@id']}`)

udata/core/dataset/views.py
```diff
@@ -86,8 +88,86 @@ def get_context(self):
         context['can_edit'] = DatasetEditPermission(self.dataset)
         context['can_edit_resource'] = ResourceEditPermission
         context['discussions'] = Discussion.objects(subject=self.dataset)
+
         return context
 
+    def get_json_ld(self):
+        dataset = self.dataset
+
+        result = {"@context": "http://schema.org",
```
vinyll: single quotes all the way

udata/core/dataset/views.py
```diff
@@ -86,8 +88,86 @@ def get_context(self):
...
+                  "url": url_for('datasets.show', dataset=dataset, _external=True),
+                  "name": dataset.title,
+                  "keywords": ', '.join(dataset.tags),
+                  "distribution": map(self.get_json_ld_resource, dataset.resources),
+                  # This value is not standard
+                  "extras": map(self.get_json_ld_extra, dataset.extras.items())
+        }
+
+        if dataset.license and dataset.license.url:
+            result["license"] = dataset.license.url
```
vinyll: single quotes all the way

js/front/dataset/index.js
```diff
@@ -72,21 +71,15 @@ new Vue({
...
-            }
+            const selector = "script[type='application/ld+json']";
+            const dataset = JSON.parse(document.querySelector(selector).text)
+            dataset.resources = dataset.distribution;
             delete dataset.distribution;
-            if (utils.isString(dataset.keywords)) {
-                dataset.keywords = [dataset.keywords];
-            }
-            dataset.extras = microdata('http://schema.org/PropertyValue');
+            dataset.keywords = dataset.keywords.split(',')
```
noirbizarre: They should be stripped too to avoid spaces in strings:
```javascript
dataset.keywords.split(',').map((keyword) => keyword.trim())
```

udata/core/dataset/views.py
```diff
@@ -86,8 +88,86 @@ def get_context(self):
...
+        if author:
+            result['author'] = author
+
+        return result
+
+    @staticmethod
+    def get_json_ld_resource(resource):
+
+        result = {
+            "@type": "DataDownload",
```
vinyll: guess what… single quotes. Ok, you got it ;)

udata/core/dataset/views.py
```diff
@@ -86,8 +88,86 @@ def get_context(self):
...
+            "datePublished": resource.published.isoformat(),
+            "contentSize": resource.filesize or '',
+            "fileFormat": resource.mime or '',
+            "interactionStatistic": {
+                "@type": "InteractionCounter",
+                "interactionType": {
+                    "@type": "DownloadAction"
+                },
+                # We take resource.metrics.views if it exists
+                "userInteractionCount": getattr(resource.metrics, 'views', None) or ''
```
vinyll: If you want an empty string by default, why not just:

    getattr(resource.metrics, 'views', '')

js/front/dataset/index.js
```diff
@@ -72,21 +71,15 @@ new Vue({
...
          * @return {Object} The parsed dataset
          */
         extractDataset() {
-            const dataset = microdata('http://schema.org/Dataset')[0];
-            if (Array.isArray(dataset.distribution)) {
-                dataset.resources = dataset.distribution;
-            } else {
-                dataset.resources = dataset.distribution ? [dataset.distribution] : [];
-            }
+            const selector = "script[type='application/ld+json']";
```
noirbizarre: Maybe we can speed this a little bit by using an ID.
Having this helper outside of the view would allows to test it and reuse this mecanic in other views.

udata/core/dataset/views.py
noirbizarre: 👍 for single quotes

udata/templates/user/base.html
```diff
@@ -34,14 +34,14 @@
...
-                        alt="{{ user.fullname }}" class="scalable" itemprop="image"/>
+                        alt="{{ user.fullname }}" class="scalable"/>
                 </div>
-                <h2 itemprop="name">{{ user.fullname }}</h2>
-                <div itemprop="description">
+                <h2>{{ user.fullname }}</h2>
+                <div>
                     {{ user.about|markdown }}
                 </div>
                 {# <hr/> #}
```
vinyll: shouldn't we remove this?

js/front/dataset/resource-modal.vue
noirbizarre: It should be `{{ resource.fileFormat }}` as provided in the JSON-LD

js/front/dataset/resource-modal.vue
noirbizarre: It should be `{{ resource.interactionStatistic.userInteractionCount }}` as provided in the JSON-LD, if present

udata/core/dataset/views.py
```diff
@@ -86,8 +88,86 @@ def get_context(self):
...
+    @staticmethod
+    def get_json_ld_resource(resource):
+
+        result = {
+            "@type": "DataDownload",
+            "@id": str(resource.id),
+            "url": resource.url,
+            "name": resource.title or _('Nameless resource'),
+            "contentUrl": resource.url,
+            "encodingFormat": resource.format or '',
```
noirbizarre: I think the keys should only be present if not empty (I comment it, but it's valid for all empty values)

udata/templates/user/base.html
noirbizarre: Yes, we need to check, but I think we can remove a lot of useless divs which were present only for microdata scoping.

jphnoel: I tried, I promise !
Nevertheless, it requires long, manual and painful tests.
Moreover, I do not have on my dev environment the data allowing me to complete these tests, which means that there could be undetected regressions.
I prefer not to do it, and to postpone it to the next refactoring of the templates.

udata/tests/frontend/test_dataset_frontend.py
jphnoel: It is not possible (or at least vey difficul) to put the tests at the model level. Indeed, we need url_for to generate the JSON/LD.

js/front/dataset/index.js
jphnoel: I put the ID.

For the move to a separate file, I am bit more dubious. Testing it is not very useful (actually, we would basically test JSON.parse and querySelector) and so far, it is used for dataset only.

udata/core/dataset/models.py
```diff
@@ -148,6 +149,48 @@ def check_availability(self, group):
...
+            '@id': str(self.id),
+            'url': self.url,
+            'name': self.title or _('Nameless resource'),
+            'contentUrl': self.url,
+            'dateCreated': self.created_at.isoformat(),
+            'dateModified': self.modified.isoformat(),
+            'datePublished': self.published.isoformat(),
+        }
+
+        if 'views' in self.metrics.keys():
```
davidbgk: You can omit the `.keys()`.

udata/core/dataset/models.py
```diff
@@ -416,6 +459,48 @@ def update_resource(self, resource):
...
+        else:
+            author = None
+
+        if author:
+            result['author'] = author
+
+        return result
+
+    @staticmethod
+    def get_json_ld_extra((key, value)):
```
davidbgk: Remove parenthesis.

udata/templates/macros/metadata.html
```diff
@@ -43,4 +43,9 @@
 <meta name="check-urls" content="{{ 'true' if config.CROQUEMORT else 'false' }}" />
 <meta name="check-urls-ignore" content="{{ config.CROQUEMORT_IGNORE|tojson|urlencode }}" />
 <meta name="territory-enabled" content="{{ 'true' if config.ACTIVATE_TERRITORIES else 'false' }}">
+
+{% if json_ld %}
+<script id="json_ld" type="application/ld+json">{{ json_ld|safe }}</script>
```
davidbgk: Why `safe`? (and is it really safe? ;) )

udata/core/dataset/models.py
jphnoel: I know, but I think it is a bit more self-understanding like this.
When you read it, you immediately see that metrics is a dictionary. Otherwise, you do not know whether it is a list, a dictionary or any other kind of iterable.
I am ready to correct it if you confirm that this is the current coding guideline of the project.
Let me know !

udata/core/dataset/models.py
jphnoel: I need the parenthesis because self.extras.items() return a list of tuples.

udata/templates/macros/metadata.html
jphnoel: For the same reason why there is this comment (https://github.com/opendatateam/udata/blob/dev/udata/frontend/helpers.py#L332).

If you do not put safe, the " are escaped and it does not generate a valid JSON.

Since the JSON is fully generated by the server, I think it can be considered as safe.

udata/core/dataset/models.py
```diff
@@ -416,6 +459,48 @@ def update_resource(self, resource):
...
+            'description': mdstrip(self.description),
+            'alternateName': self.slug,
+            'dateCreated': self.created_at.isoformat(),
+            'dateModified': self.last_modified.isoformat(),
+            'url': url_for('datasets.show', dataset=self, _external=True),
+            'name': self.title,
+            'keywords': ','.join(self.tags),
+            'distribution': [resource.json_ld for resource in self.resources],
+            # This value is not standard
+            'extras': map(self.get_json_ld_extra, self.extras.items()),
```
jphnoel: Pros: it removes indeed 2 character ('(' and ')').
Cons:
* It adds 14 characters
* It adds a star operator which, at least for me, makes the code harder to read

Why do you hate the parenthesis that much ;-) ?

udata/templates/macros/metadata.html
davidbgk: Oh good point, you're right.

udata/core/dataset/models.py
noirbizarre: I agree with @davidbgk : `'string' in dict` is used everywhere accross the project.
Given the fact that it's `self.metrics` it means you inside the class so you know which structure is `metrics`.
In udata, `self.metrics` is always a dict.

udata/core/dataset/models.py
jphnoel: OK, I will fix it.

udata/tests/frontend/test_user_frontend.py
```diff
@@ -50,9 +50,19 @@ def test_render_list_empty(self):
 
     def test_render_profile(self):
         '''It should render the user profile'''
-        user = UserFactory()
+        user = UserFactory(first_name=' John', last_name='Doe   ',
```
noirbizarre: The attribute are automatically filled by the factory

udata/tests/frontend/test_user_frontend.py
```diff
@@ -50,9 +50,19 @@ def test_render_list_empty(self):
...
+                           about='* Title 1\n* Title 2',
+                           website='http://www.datagouv.fr/user',
+                           avatar_url='http://www.datagouv.fr/avatar')
         response = self.get(url_for('users.show', user=user))
         self.assert200(response)
+        json_ld = self.get_json_ld(response)
+        self.assertEquals(json_ld['@context'], 'http://schema.org')
```
noirbizarre: You can also test directly the entire dict equality

udata/core/dataset/models.py
jphnoel: I agree and, from a personal point of view, I prefer free function to class methods, since it is simpler and easier to test.

But:
- In the file, there are only classes, and this function is obviously related to a class. So, it made sense to 'attach' it to the class
-  I am ready to change it to a free function if we decide, as a coding guideline, that we should **always** use free functions instead of static methods.

udata/tests/frontend/test_reuse_frontend.py
```diff
@@ -48,9 +48,26 @@ def test_render_list_empty(self):
 
     def test_render_display(self):
         '''It should render the reuse page'''
-        reuse = ReuseFactory()
-        response = self.get(url_for('reuses.show', reuse=reuse))
+        author = UserFactory()
+        reuse = ReuseFactory(url='http://www.datagouv.fr/reuse',
```
jphnoel: OK, I removed all the attributes which were populated by the factory (except the markdown description).

udata/templates/user/base.html
jphnoel: OK, I removed the now useless DOM. I checked as carefully as I could in the web interface that there was no regression.

udata/tests/frontend/test_user_frontend.py
jphnoel: That is right ! It is more complete and, while trying to do it, I even discovered two micro-bugs.

But:
* When the equality is not satisfied, it is harder to debug
* I did not succeed in doing it for the dates, especially 'last_modified'
* It makes sometimes the assertion much heavier

All in all, after having tried to do it, I prefer to keep the existing version.

PR #546, opened 2016-09-29
Refactoring

jphnoel:
I made these few minor modifications during my initial training period, while I was discovering the code.


udata/core/dataset/api.py
```diff
@@ -196,7 +196,7 @@ def post(self, dataset):
     @api.secure
     @api.doc('update_resources', **common_doc)
     @api.expect([resource_fields])
-    @api.marshal_with(resource_fields)
+    @api.marshal_list_with(resource_fields)
```
davidbgk: It could be useful to add a test for that to verify what was broken before.

docs/running-project.md
```diff
@@ -24,6 +24,8 @@ You need to initialize some data before pushing uData to it's full potential:
 $ udata init
 # Fetch and load licenses
 $ udata licenses https://www.data.gouv.fr/api/1/datasets/licenses
+# Build the search index
+$ udata search init
```
noirbizarre: There is two choices:
- either you loaded the fixture during the init and so the init should perform the search indexation
- either you didn't and so you don't have data to index here.

So, proposal: include the search indexation after the fixture loading in the `udata init` command

udata/core/dataset/api.py
noirbizarre: 👍

udata/core/dataset/models.py
```diff
@@ -68,6 +68,8 @@
 
 
 class License(db.Document):
+    # We need to declare id explicitly since we do not use the default
+    # value set by Mongo.
```
noirbizarre: I don't know if we need to explain here a documented and common feature of MongoEngine/MongoDB.
But, right now, if it helps, let's keep it 👍

udata/core/dataset/models.py
```diff
@@ -150,6 +152,8 @@ def is_available(self):
 
 
 class Resource(ResourceMixin, WithMetrics, db.EmbeddedDocument):
+    ''' A Resource is a resource provided by the original provider of the
+    dataset'''
```
noirbizarre: It's repeating itself: "a resource is a resource".
I'l change for "a resource is a file (remote or locale) or an API provided by..."

udata/core/dataset/models.py
```diff
@@ -421,6 +425,8 @@ def community_resources(self):
 
 
 class CommunityResource(ResourceMixin, WithMetrics, db.Document):
+    ''' A CommunityResource is a resource added by the community of the users
+    to the original dataset'''
```
noirbizarre: I'll change it here too to be consistent

udata/core/organization/factories.py
```diff
@@ -13,8 +13,8 @@ class OrganizationFactory(MongoEngineFactory):
     class Meta:
         model = models.Organization
 
-    name = factory.LazyAttribute(lambda o: faker.sentence())
-    description = factory.LazyAttribute(lambda o: faker.text())
+    name = factory.LazyAttribute(lambda o: faker.text())
```
noirbizarre: I don't think so. `faker.text()` output an entire paragraph whereas `faker.sentence()` output a single sentence,  which is more consistent with a title or a name.

udata/core/organization/factories.py
```diff
@@ -13,8 +13,8 @@ class OrganizationFactory(MongoEngineFactory):
...
+    description = factory.LazyAttribute(lambda o: faker.sentence())
```
noirbizarre: Reverse comment of the above one: `faker.sentence()` output a single sentence whereas `faker.text()` give us an entire paragraph which is what we expect for a description.

udata/tests/test_activity.py
```diff
@@ -52,29 +52,30 @@ def test_create_and_retrieve_for_related(self):
...
+    def callback(self, sender, activity):
         self.assertEqual(sender, FakeActivity)
         self.assertIsInstance(activity, FakeActivity)
-        self.emitted = True
+        self.callback_called = True
 
-    def test_emit_signal(self):
-        '''It should emit a signal on new activity'''
+    def test_callback_registration(self):
+        '''Check that the registered callback is executed when a new activity
```
noirbizarre: Test description are meant to be read on run in the console output. As a convention, we express what the tested code **should** do, not what the test itself checks (as test always checks something).
To keep the change, it could be:

     (it should) execute registered callbacks when a new activity is issued

(The `it` or `it should` are optionnals)

udata/tests/test_activity.py
```diff
@@ -52,29 +52,30 @@ def test_create_and_retrieve_for_related(self):
...
             FakeActivity.objects.create(actor=self.user, related_to=self.fake)
 
-        self.assertTrue(self.emitted)
+        self.assertTrue(self.callback_called)
 
-    def test_class_shortcut(self):
-        '''It should emit a signal on new activity'''
+    def test_emit(self):
+        '''Check that the registered callback is executed when a new activity
+        is saved via the 'emit' shortcut'''
```
noirbizarre: Same as above

udata/tests/test_activity.py
```diff
@@ -52,29 +52,30 @@ def test_create_and_retrieve_for_related(self):
...
-        with FakeActivity.on_new.connected_to(self.check_emitted):
+        with FakeActivity.on_new.connected_to(self.callback, sender=FakeActivity):
             FakeActivity.objects.create(actor=self.user, related_to=self.fake)
 
-        self.assertTrue(self.emitted)
+        self.assertTrue(self.callback_called)
 
-    def test_class_shortcut(self):
-        '''It should emit a signal on new activity'''
+    def test_emit(self):
```
noirbizarre: I think the method_name was good, it was meant to test the `connect()` class method. But you're right, the `emit()` class method wasn't tested. So I think we should have 2 tests.

udata/tests/frontend/test_swagger.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from flask import url_for
+import json
```
davidbgk: We put imports from standard lib before the ones from external dependencies following the [PEP8 on imports](https://www.python.org/dev/peps/pep-0008/#imports).

udata/core/dataset/models.py
```diff
@@ -150,6 +152,8 @@ def is_available(self):
 
 
 class Resource(ResourceMixin, WithMetrics, db.EmbeddedDocument):
+    ''' Local file, remote file or API provided by the original provider of the
```
noirbizarre: 👍 for the docstring but we use this convention:
- No leading space for one liners
- Begins and stops with a line break for multiple lines

udata/core/dataset/models.py
```diff
@@ -421,6 +425,9 @@ def community_resources(self):
 
 
 class CommunityResource(ResourceMixin, WithMetrics, db.Document):
+    ''' Local file, remote file or API added by the community of the users
+    to the original dataset'''
```
noirbizarre: Same as above.
I think the next empty line can also be removed.

udata/tests/frontend/test_swagger.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import json
+
+from . import FrontTestCase
+
+
+class SwaggerBlueprintTest(FrontTestCase):
```
noirbizarre: It's related to the API not the front so it should inherit from `ApiTestCase` (with proved an helper to test JSON) and located in `udata/tests/api`.

udata/tests/frontend/test_swagger.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def test_swagger(self):
```
noirbizarre: Be more specific: you're not testing the entire `swagger.json`, only the resource type

PR #547, opened 2016-10-03
Huge performance improvement on notifications querying

noirbizarre:
This PR greatly improve performance and memory usage on notifications retrieval by:
- fetching only the required fields from MongoDB
- turning off GenericReferenceField dereferencing

The response time is divided by approximatively 5 on my computer (from more than 5 seconds to 800ms with an administrator account with 64 pending notifications).

The master branch need a different fix because of bc93fc233dde16753078e5d2eeafbfd398477024 being only on dev branch

### DIFF
--- a/udata/core/discussions/actions.py
+++ b/udata/core/discussions/actions.py
@@ -14,8 +14,8 @@ def discussions_for(user, only_open=True):
 
     :param bool only_open: whether to include closed discussions or not.
     '''
-    datasets = Dataset.objects.owned_by(user.id, *user.organizations)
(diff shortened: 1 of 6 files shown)


petzlux:
@noirbizarre Thanks for this, I will try and use your pull request on the dev branch. As we based ourselves before on the master branch, I will have to try and get the dev branch to work first though.

noirbizarre:
Tests are already here: having all the tests pass without any change other than optimizations is the main test ;)
These kind of optimization are great to develop when you have a lot of test with great coverage: no new test required and a global point a view of the impacted code !

PR #548, opened 2016-10-03
Huge performance improvement on notifications querying (master edition)

noirbizarre:
Same as #547 for master branch.

Should raise conflict on merge.


petzlux:
@noirbizarre Super, on my dev machine I have at least a 10x decrease ! Great job

noirbizarre:
Great !
It might be interesting to find other places where the same simple optimization can be applied !

PR #550, opened 2016-10-05
Add the missing community resource API

noirbizarre:
This PR add the missing community resource API creation endpoint and fixes permissions handling.

### DIFF
--- a/udata/core/dataset/api.py
+++ b/udata/core/dataset/api.py
@@ -394,6 +394,25 @@ class CommunityResourcesAPI(API):
         return (community_resources.order_by(args['sort'])
                                    .paginate(args['page'], args['page_size']))
 
+    @api.secure
+    @api.doc('create_community_resource')
+    @api.expect(community_resource_fields)
+    @api.marshal_with(community_resource_fields)
+    def post(self):
+        '''Create a new community resource'''
+        form = api.validate(CommunityResourceForm)
+        resource = CommunityResource()
+        form.populate_obj(resource)
+        if not resource.dataset:
+            api.abort(400, errors={
+                'dataset': 'A dataset identifier is required'
+            })
+        if not resource.organization:
+            resource.owner = current_user._get_current_object()
+        resource.modified = datetime.now()
+        resource.save()
+        return resource, 201
+
 
 @ns.route('/community_resources/<crid:community>/',
           endpoint='community_resource', doc=common_doc)
--- a/udata/core/dataset/api.py
+++ b/udata/core/dataset/api.py
@@ -1,4 +1,22 @@
 # -*- coding: utf-8 -*-
+'''
+TODO: We need to cleanup and give more coherence to these APIs
+- upload enpoints should be singular instead of plural
+- community resource APIs should be consistent
+    (single root instead of the current state splitted
+    between /dataset/<id>/community and /community_resource/)
+- Swagger operation IDs needs to be explicits
+- make use of latest RESTPlus decorator:
+  - @api.expect
+  - @api.param
+- permissions needs to be more comprehensible
+- some mutualizations may be removed (common_doc ?)
+- some API need to renaming (ie. featured vs moderated...)
+
+These changes might lead to backward compatibility breakage meaning:
+- new API version
+- admin changes
+'''
 from __future__ import unicode_literals
 import os
 from datetime import datetime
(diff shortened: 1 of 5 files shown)


udata/core/dataset/api.py
```diff
@@ -288,12 +288,13 @@ def post(self, dataset):
 @api.doc(parser=upload_parser, **common_doc)
 class UploadCommunityResources(UploadMixin, API):
```
davidbgk: We have two `UploadCommunityResource` class declared, I think we should remove that one? Note that in the other case the `owner` is not set.

udata/core/dataset/permissions.py
```diff
@@ -7,31 +7,31 @@
...
-        super(ResourceEditPermission, self).__init__(*needs)
+        super(OwnablePermission, self).__init__(*needs)
+
+
+class DatasetEditPermission(OwnablePermission):
+    '''Permissions to edit a Dataset'''
+    pass
+
+
+class ResourceEditPermission(OwnablePermission):
```
davidbgk: Maybe we should make the distinction between permissions for Resources and CommunityResources here to be explicit?

udata/templates/dataset/resource/list-item.html
```diff
@@ -41,7 +41,7 @@ <h4 class="list-group-item-heading ellipsis">
     {{ resource.owner|avatar(52, 'resource-owner') }}
     {% endif %}
 
-    {% if can_edit_resource(resource, dataset) %}
+    {% if can_edit_resource(resource if resource.from_community else dataset) %}
```
davidbgk: I prefer `if can_edit_resource(dataset) and can_edit_community_resource(resource)`, does it make sense?

udata/tests/api/test_datasets_api.py
```diff
@@ -967,3 +957,54 @@ def test_community_resource_api_update_with_file(self):
                          'new description')
         self.assertTrue(
             CommunityResource.objects.first().url.endswith('test.txt'))
+
+    @attr('create')
```
davidbgk: 👍 for tests

udata/templates/dataset/resource/list-item.html
noirbizarre: Yes it does, but as discussed earlier, I made the minimal change to make it work until a proper cleanup and refactoring is done.

udata/core/dataset/permissions.py
noirbizarre: I asked myself the same question. In order to keep the ResourceEdit permission (which is used by views), I made the change in this way, but to me the proper way is to have only `DatasetEditPermission` and `CommunityRessourcePermission` as there is no permissions for ressources, only for dataset.

udata/core/dataset/api.py
noirbizarre: I've just checked, these are different APIs and names differs with the trailing `s`:
- one allows to create by upload (this is why owner is set)
- the other allows to update by upload (owner is already set but we check permission)

I propose we let like this on master as it will be a lot more than the initial fix and we rename it in refactoring on dev because:
- Swagger operation IDs, which are used by the admin, need to change in consequence (and so the admin code
- we need to do a big cleanup/coherence check on these APIs (which will probably breaks backward compatibility and so need a new API version)

davidbgk: Alright, can we at least rename the classes to be more explicit or at least add a comment related to that?

udata/core/dataset/permissions.py
davidbgk: In that case, can we rename that class to be more explicit too?

udata/templates/dataset/resource/list-item.html
davidbgk: Got it 👍

udata/core/dataset/api.py
noirbizarre: OK, I will 👍

udata/core/dataset/api.py
```diff
@@ -1,4 +1,22 @@
 # -*- coding: utf-8 -*-
+'''
+TODO: We need to cleanup and give more coherence to these APIs
```
davidbgk: 👍 maybe we should declare an issue related to that and link to it?

udata/core/dataset/api.py
noirbizarre: Done in the last commit

udata/core/dataset/api.py
noirbizarre: 👍

udata/core/dataset/permissions.py
```diff
@@ -7,31 +7,31 @@
...
+    '''Permissions to edit a Resource (aka. its dataset) or community resource'''
```
noirbizarre: No, the point is that this permission is currently used for both community resource and resource vie its dataset (explaining the docstring and the exception I added to prevent misuse until then)

noirbizarre: But I agree, after the refactoring we need to have clearly separated  permissions for Dataset and CommunityResource and maybe keep the type check in each to be a little more defensive

davidbgk:
LGTM once Circle-CI’s ES is up…

noirbizarre:
OK, rebuild triggered 👍 
You need to approve the review to allow the merge.

PR #551, opened 2016-10-05
Fix admin daterange picker

noirbizarre:
This pull request fix the temporage coverage field:

- ensure a proper attribute nesting in the JSON (in and out)
- localize the date display

It also fix the localization on frequency field.

### DIFF
--- a/js/components/form/daterange-picker.vue
+++ b/js/components/form/daterange-picker.vue
@@ -15,14 +15,14 @@
         v-el:start-input :placeholder="_('Start')"
         @focus="onFocus"
         :required="required"
-        :value="start_value"
+        :value="start_value|dt date_format ''"
         :readonly="readonly">
     <span class="input-group-addon">{{ _('to') }}</span>
     <input type="text" class="input-sm form-control"
         v-el:end-input :placeholder="_('End')"
         @focus="onFocus"
         :required="required"
-        :value="end_value"
+        :value="end_value|dt date_format ''"
         :readonly="readonly">
     <div class="dropdown-menu dropdown-menu-right">
         <calendar :selected="current_value"></calendar>
@@ -73,11 +73,14 @@ export default {
             if (this.hiddenField) {
                 return this.hiddenField.value;
             }
+        },
+        date_format() {
+            return this.field.format || DEFAULT_FORMAT;
         }
     },
     events: {
         'calendar:date:selected': function(date) {
-            this.pickedField.value = date.format(this.field.format || DEFAULT_FORMAT);
+            this.pickedField.value = date.format(this.date_format);
             this.hiddenField.value = date.format(ISO_FORMAT);
             this.picking = false;
             return true;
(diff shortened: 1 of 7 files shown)


js/components/form/daterange-picker.vue
```diff
@@ -102,6 +105,9 @@ export default {
             });
         }
     },
+    filters: {
```
davidbgk: Why?

js/components/form/daterange-picker.vue
```diff
@@ -15,25 +15,25 @@
...
-        :value="end_value"
+        :value="end_value|dt date_format ''"
         :readonly="readonly">
     <div class="dropdown-menu dropdown-menu-right">
         <calendar :selected="current_value"></calendar>
     </div>
     <input type="hidden" v-el:start-hidden
-        :id="field.id + '-start'"
-        :name="field.id + '-start'"
+        :id="field.id + '.start'"
```
davidbgk: Why?

js/plugins/i18next.js
```diff
@@ -8,8 +8,9 @@ export function install(Vue) {
         return i18n._(value, options);
     });
 
-    Vue.filter('dt', function(value, format) {
-        return value ? moment(value).format(format || 'LLL') : '-';
+    Vue.filter('dt', function(value, format, empty) {
+        empty = empty !== undefined ? empty : '-';
```
davidbgk: Why not make the empty string the default value for `empty` and that way not having to pass it each time?

js/components/form/daterange-picker.vue
noirbizarre: Because the dot is made for nesting. All the fields with dot in names are nested into the json (as HTML forms are flat).
See in [the dataset form definition](https://github.com/opendatateam/udata/blob/dev/js/components/dataset/form.vue), spatial fields are nested so there are already dotted.
By the way I kept the IDs, but I think they are  useless with vue.js as there is the `v-el` directive

js/plugins/i18next.js
noirbizarre: Because I wanted to be backward compatible. If I change the signature, I need to modify every calls in the admin.
Datpicker are the only places where we need empty value to trigger placeholder display.

js/components/form/daterange-picker.vue
noirbizarre: Sorry, I forgot to remove this

noirbizarre: Removed in the last commit

js/components/form/daterange-picker.vue
davidbgk: Alright

js/plugins/i18next.js
davidbgk: OK I see, thanks for the explanation.

noirbizarre: ;)

PR #553, opened 2016-10-06
JSON-LD

jphnoel:
This follows #545 and fixes the last comments made by @noirbizarre.

### DIFF
--- a/udata/core/organization/models.py
+++ b/udata/core/organization/models.py
@@ -229,10 +229,12 @@ class Organization(WithMetrics, BadgeMixin, db.Datetimed, db.Document):
 
     @cached_property
     def json_ld(self):
+        type_ = 'GovernmentOrganization' if self.public_service \
+                else 'Organization'
 
         result = {
             '@context': 'http://schema.org',
-            '@type': 'GovernmentOrganization' if self.public_service else 'Organization',
+            '@type': type_,
             'alternateName': self.slug,
             'url': url_for('organizations.show', org=self, _external=True),
             'name': self.name,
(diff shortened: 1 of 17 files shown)


udata/core/user/models.py
```diff
@@ -196,9 +196,12 @@ def json_ld(self):
             '@type': 'Person',
             '@context': 'http://schema.org',
             'name': self.fullname,
-            'description': mdstrip(self.about),
         }
 
+        description = mdstrip(self.about)
+        if description != '':
```
davidbgk: Is there a case where `if description` doesn't work?

jphnoel: If self.about is None, mdstrip returns an empty string. I realized that when re-writing a unit test.

davidbgk: But `bool('')` returns `False` so is the current comparison needed?

jphnoel: No, indeed ! I am going to fix this.

udata/core/organization/models.py
```diff
@@ -229,15 +229,26 @@ def check_availability(self):
...
-            'logo': self.logo(external=True),
             'url': url_for('organizations.show', org=self, _external=True),
             'name': self.name,
-            'description': mdstrip(self.description),
         }
 
+        if self.description:
+            result['description'] = mdstrip(self.description)
+
+        logo = self.logo(external=True)
```
noirbizarre: I think, self.logo is directly bool-testable aka.:
```
if self.logo:
    result['logo'] = self.logo(external=True)
```

PR #558, opened 2016-10-12
Fix erroneous user statistics in the admin (fixes #539)

### LINKED ISSUE #539: User statistics are erroneous in the admin interface
The statistics are typically the number of reuses, datasets and followers.
safariscreensnapz323

jphnoel:


udata/tests/api/test_user_api.py
```diff
@@ -170,3 +170,22 @@ def test_suggest_users_api_by_id(self):
...
+        '''It should provide a list of users'''
+        with self.autoindex():
+            user = UserFactory(
+                about='* Title 1\n* Title 2',
+                website='http://www.datagouv.fr/user',
+                avatar_url='http://www.datagouv.fr/avatar',
+                metrics={'datasets': 10})
+        response = self.get(url_for('api.users'))
+        self.assert200(response)
+        json = response.json['data'][0]
```
noirbizarre: I'll test that `json['data']` length is 1. It seems obvious but when it's not the case anymore, the error will be more explicit.

udata/tests/api/test_user_api.py
```diff
@@ -170,3 +170,22 @@ def test_suggest_users_api_by_id(self):
...
         suggestion = response.json[0]
         self.assertEqual(suggestion['id'], str(user.id))
+
+    def test_users(self):
+        '''It should provide a list of users'''
+        with self.autoindex():
+            user = UserFactory(
+                about='* Title 1\n* Title 2',
+                website='http://www.datagouv.fr/user',
+                avatar_url='http://www.datagouv.fr/avatar',
```
noirbizarre: Use random values with faker (I think `faker.paragraph()` for `about` and `faker.url()` for `website` and `avatar_url` might be perfect)).

udata/core/site/metrics.py
```diff
@@ -114,7 +114,7 @@ class MaxDatasetFollowersMetric(SiteMetric):
     def get_value(self):
         dataset = (Dataset.objects(metrics__followers__gt=0).visible()
                           .order_by('-metrics.followers').first())
-        return dataset.metrics.get('followers', 0)
+        return dataset.metrics.get('followers', 0) if dataset else 0
```
noirbizarre: Your fix is right but I believe that `metrics_followers__gt=0` ensure there is always a `followers` metric if there is a matching dataset so you can write:
```python
return dataset.metrics['followers'] if dataset else 0
```

udata/core/site/metrics.py
```diff
@@ -127,7 +127,7 @@ class MaxDatasetReusesMetric(SiteMetric):
     def get_value(self):
         dataset = (Dataset.objects(metrics__reuses__gt=0).visible()
                    .order_by('-metrics.reuses').first())
-        return dataset.metrics.get('reuses', 0)
+        return dataset.metrics.get('reuses', 0) if dataset else 0
```
noirbizarre: Same as above

PR #561, opened 2016-10-13
Fixes an error message when creating a dataset for an org (fixes #560)

### LINKED ISSUE #560: Error in a development environment when creating a new dataset for an org
I run a development platform locally, with the branch dev.

As a user, I belong to an organization.

I submit a new dataset on behalf of this organization.

I get the following traces in the shell where I run `inv work`:

```
[2016-10-13 10:54:58,805: ERROR/MainProcess] Task udata.search.reindex[4f90421b-1876-4f34-be8c-21b31c2061e3] raised unexpected: RuntimeError('Application was not able to create a URL adapter for request independent URL generation. You might be able to fix this by setting the SERVER_NAME config variable.',)
Traceback (most recent call last):
  File "/home/jpnoel/.virtualenvs/udata/local/lib/python2.7/site-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/jpnoel/udata/udata/tasks.py", line 20, in __call__
    return super(ContextTask, self).__call__(*args, **kwargs)
  File "/home/jpnoel/.virtualenvs/udata/local/lib/python2.7/site-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/jpnoel/udata/udata/search/__init__.py", line 105, in reindex
    id=obj.id, body=adapter.serialize(obj))
  File "/home/jpnoel/udata/udata/core/dataset/search.py", line 181, in serialize
    image_url = dataset.organization.logo(40)
  File "/home/jpnoel/.virtualenvs/udata/local/lib/python2.7/site-packages/flask_fs/mongo.py", line 174, in best_url
    return self.fs.url(filename, external=external) if filename else None
  File "/home/jpnoel/.virtualenvs/udata/local/lib/python2.7/site-packages/flask_fs/storage.py", line 156, in url
    return url_for('fs.get_file', fs=self.name, filename=filename, _external=external)
  File "/home/jpnoel/.virtualenvs/udata/local/lib/python2.7/site-packages/flask/helpers.py", line 298, in url_for
    raise RuntimeError('Application was not able to create a URL '
RuntimeError: Application was not able to create a URL adapter for request independent URL generation. You might be able to fix this by setting the SERVER_NAME config variable.

```

Workaround: set SERVER_NAME = 'localhost:7000' in udata.cfg

jphnoel:


udata/settings.py
```diff
@@ -37,6 +37,8 @@ class Defaults(object):
     CACHE_KEY_PREFIX = 'udata-cache'
     CACHE_TYPE = 'redis'
 
+    SERVER_NAME = 'localhost:7000'
```
noirbizarre: I wonder if it might not be better in `Debug` settings as it's requirement to set this value in production; (Open to debate, just wondering)

jphnoel: I got your point.

But, if you put it in `Debug`, it is taken into account only if you run `python manage.py serve`, not if you run `inv serve`, while this is what is mentioned in the installation doc (actually, `honcho` is mentioned in the doc, but it leads the reader to user `inv serve`).

Thus, it would still not work out of the box.

davidbgk: Maybe we should change that at the `inv serve` command level?

jphnoel: If you use `inv serve`, even without `honcho`, you have a problem with the traces on the standard output.

Therefore, personally, I prefer to use `python manage.py serve`, not `inv serve`.

If you want to solve the problem in the two situations, without using Defaults, I think we should define an intermediate structure between Debug and Defaults.

Alternatively, we could make that `inv serve` (which is used in development only, as far as I know) uses `Debug` instead of `Defaults`.

vinyll: I'd also rather have the same config with `inv serve` and `manage.py serve` (ie debug)

davidbgk:
After team discussion, too specific to dev env.

PR #564, opened 2016-10-17
Slugify the tags and warn the user if tags are too short/long (fixes …

jphnoel:
…563)

### DIFF
--- a/js/models/reuse.js
+++ b/js/models/reuse.js
@@ -21,19 +21,17 @@ export default class Reuse extends Model {
     /**
      * Create or update the given reuse
      */
-    save() {
+    save(on_error) {
+        let ep, args;
         if (this.id) {
-            this.$api('reuses.update_reuse', {
-                reuse: this.id,
-                payload: this
-            },
-            this.on_fetched);
-        } else {
-            this.$api('reuses.create_reuse', {
-                payload: this
-            },
-            this.on_fetched);
+            ep = 'reuses.update_reuse';
+            args = {payload: this, reuse: this.id,};
         }
+        else {
+            ep = 'reuses.create_reuse';
+            args = {payload: this,};
+        }
+        this.$api(ep, args, this.on_fetched, on_error);
     }
 
     update(data, on_success, on_error) {
--- a/udata/forms/fields.py
+++ b/udata/forms/fields.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
-import re
 import uuid
 
 from dateutil.parser import parse
@@ -21,13 +20,10 @@ from udata.models import db, User, Organization, Dataset, Reuse, datastore
 from udata.core.storages import tmp
 from udata.core.organization.permissions import OrganizationPrivatePermission
 from udata.i18n import lazy_gettext as _
+from udata import tags
 from udata.utils import to_iso_date, get_by
 
 # _ = lambda s: s
-
-RE_TAG = re.compile('^[\w \-.]*$', re.U)
-MIN_TAG_LENGTH = 2
-MAX_TAG_LENGTH = 128
 
 
 class FieldHelper(object):
@@ -312,11 +308,9 @@ class TagField(StringField):
 
     def process_formdata(self, valuelist):
         if valuelist and len(valuelist) > 1:
-            self.data = valuelist
+            self.data = [tags.slug(value) for value in valuelist]
         elif valuelist:
-            self.data = list(set([
-                x.strip().lower()
-                for x in valuelist[0].split(',') if x.strip()]))
+            self.data = tags.tags_list(valuelist[0])
         else:
             self.data = []
 
@@ -324,15 +318,12 @@ class TagField(StringField):
         if not self.data:
             return
         for tag in self.data:
-            if not MIN_TAG_LENGTH <= len(tag) <= MAX_TAG_LENGTH:
+            if not tags.MIN_TAG_LENGTH <= len(tag) <= tags.MAX_TAG_LENGTH:
                 message = _(
                     'Tag "%(tag)s" must be between %(min)d '
                     'and %(max)d characters long.',
-                    min=MIN_TAG_LENGTH, max=MAX_TAG_LENGTH, tag=tag)
-                raise validators.ValidationError(message)
-            if not RE_TAG.match(tag):
-                message = _('Tag "%(tag)s" must be alphanumeric characters '
-                            'or symbols: -_.', tag=tag)
+                    min=tags.MIN_TAG_LENGTH,
+                    max=tags.MAX_TAG_LENGTH, tag=tag)
                 raise validators.ValidationError(message)
 
 
--- a/udata/harvest/tests/test_ckan_backend.py
+++ b/udata/harvest/tests/test_ckan_backend.py
@@ -237,13 +237,15 @@ class CkanBackendTest(DBTestMixin, TestCase):
         # dataset-2 has geo feature
         self.assertIn('dataset-2', datasets)
         d = datasets['dataset-2']
-        self.assertEqual(d.tags, ['africa',
-                                  'bandwidth',
-                                  'broadband',
-                                  'cables',
-                                  'fibre',
-                                  'optic',
-                                  'terrestrial'])
+        self.assertEqual(
+            d.tags,
+            ['africa',
+             # 'b' must be absent,
+             'b' * 32,  # 'b' * 33 must be truncated
+             'cables-cables',  # the tag must be slugified
+             'fibre',
+             'optic',
+             'terrestrial'])
         self.assertEqual(len(d.resources), 1)
         self.assertEqual(d.spatial.geom['coordinates'],
                          [[[[50.8, -34.2],
--- a/udata/migrations/2016-11-08-truncate-tags.js
+++ b/udata/migrations/2016-11-08-truncate-tags.js
@@ -1,0 +1,42 @@
+/*
+ * Ensure all tags have a correct length
+ *
+ */
+
+const MIN_TAG_LENGTH = 3
+const MAX_TAG_LENGTH = 32
+
+function migrate(collection) {
+
+    var objects = collection.find({tags: {$gt: []}})
+    var name = collection.getName()
+
+    print(name + ' candidates : ' + objects.count());
+
+    var objectsCounter = 0;
+    var tagsCounter = 0;
+
+    objects.forEach(function(object) {
+        var tags = object.tags.map(function(tag) {
+            if (tag.length > MAX_TAG_LENGTH)  {
+                tagsCounter++;
+                return tag.substring(0, MAX_TAG_LENGTH)
+            }
+            if (tag.length < MIN_TAG_LENGTH)  {
+                tagsCounter++;
+                return null
+            }
+            return tag;
+        });
+        tags = tags.filter(function(tag) { return tag !== null })
+        var result = collection.update({_id: object._id},
+                                       {$set: {tags: tags}});
+        objectsCounter += result.nModified;
+    });
+
+    print(tagsCounter, name + ' tags modified.');
+    print(objectsCounter, name + ' modified.');
+}
+
+migrate(db.dataset);
+migrate(db.reuse);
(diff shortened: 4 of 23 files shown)


davidbgk:
Looks good to me, I let @noirbizarre approve it given that I'm not sure of myself on the vuejs part.

js/models/reuse.js
```diff
@@ -21,19 +21,17 @@ export default class Reuse extends Model {
     /**
      * Create or update the given reuse
      */
-    save() {
+    save(on_error) {
+        var ep, args;
```
noirbizarre: Use `let` instead (ES6)

udata/forms/fields.py
```diff
@@ -297,7 +296,7 @@ def process_formdata(self, valuelist):
             self.data = valuelist
         elif valuelist:
             self.data = list(set([
-                x.strip().lower()
+                slugify(x, to_lower=True)
```
noirbizarre: I think we should factorize that in order to have the same tag rules everywhere
(ie. [here too](https://github.com/opendatateam/udata/blob/dev/udata/harvest/filters.py#L61-L62))

udata/forms/fields.py
```diff
@@ -25,9 +25,8 @@
 
 # _ = lambda s: s
 
-RE_TAG = re.compile('^[\w \-.]*$', re.U)
-MIN_TAG_LENGTH = 2
-MAX_TAG_LENGTH = 128
+MIN_TAG_LENGTH = 3
+MAX_TAG_LENGTH = 32
```
noirbizarre: We need to check that all existing tags (or at least pertinent ones) are shorter than 32

udata/tests/api/test_datasets_api.py
```diff
@@ -156,6 +156,38 @@ def test_dataset_api_create_tags(self):
         self.assertEqual(dataset.tags, sorted(data['tags']))
 
     @attr('create')
+    def test_dataset_api_fail_to_create_too_short_tags(self):
```
noirbizarre: 👍 for the tests

udata/forms/fields.py
jphnoel: In [filters.py](https://github.com/opendatateam/udata/blob/dev/udata/harvest/filters.py), slug and taglist seem not to be used in the rest of the code.

Should we not remove them rather than factorize the code ?

noirbizarre: They are used by harvesters:
- https://github.com/opendatateam/udata/blob/dev/udata/harvest/backends/ckan.py#L54
- https://github.com/etalab/udata-gouvfr/blob/master/udata_gouvfr/harvesters/maaf.py#L96

udata/forms/fields.py
jphnoel: OK, I added a migration script

udata/tests/forms/test_tag_field.py
```diff
@@ -0,0 +1,20 @@
...
+        self.assertEquals([], tags_list(''))
+        self.assertEquals(['a'], tags_list('a'))
+        self.assertEquals(['a', 'ac', 'b'],
+                          sorted(tags_list('a, b ,  ,,, a,  ac')))
+        self.assertEquals(['a-b', 'c', 'd-e'],
+                          sorted(tags_list('a b, c , d e')))
+
+    def test_slug(self):
+        self.assertEquals('', slug(''))
+        self.assertEquals('ab-c-d', slug('ab c  d '))
```
noirbizarre: I think we need more test case, in order to ensure the awesome slugify configuration match our expectations.
Some expected cases:
- accents slugification (ie. `é` becomes `e`) (may be subject to discussion, do we want to strip accents ?)
- case lowering (ie. `aBc` becomes `abc`)
- inside spaces deduplicate (ie `a  b` becomes `a-b` and not `a--b`)
- max length handling
- special chars treatment (do we keep `.`, `_`...)

To be, the following cases are subject to debate:
- do we keep inner spaces (`ecole-publique` vs. `ecole publique`) ?
- do we keep accents (`école` vs. `ecole`) ?

My position: all in or all out, ie. `école publique` or `ecole-publique`. But I can't chose between the two, both have pros and cons.

udata/tests/forms/test_tag_field.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
noirbizarre: As I prefer to have tags handling in a seaparate module, I prefer to have tests in `udata/tests/test_tags.py`.

udata/tests/forms/test_tag_field.py
```diff
@@ -0,0 +1,20 @@
...
+from __future__ import unicode_literals
+
+from udata.tests import TestCase
+from udata.forms.fields import tags_list, slug
+
+
+class TagFieldTest(TestCase):
```
noirbizarre: It might be even better with a single case by method, ie. `test_accents_are_stripped`, `test_case_is_lowered`...

udata/tests/forms/test_tag_field.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def test_tags_list(self):
+        self.assertEquals([], tags_list(''))
+        self.assertEquals(['a'], tags_list('a'))
+        self.assertEquals(['a', 'ac', 'b'],
+                          sorted(tags_list('a, b ,  ,,, a,  ac')))
+        self.assertEquals(['a-b', 'c', 'd-e'],
+                          sorted(tags_list('a b, c , d e')))
```
noirbizarre: We also need to test deduplication (ie. `a, b, c, b, a` becomes `a, b, c`)

udata/migrations/2016-11-08-truncate-tags.js
```diff
@@ -0,0 +1,34 @@
...
+
+const MIN_TAG_LENGTH = 3
+const MAX_TAG_LENGTH = 32
+
+var datasets = db.dataset.find({ tags: { $gt: [] } })
+
+print('Candidates: ' + datasets.count());
+
+var counter = 0;
+var count = 0;
```
noirbizarre: Might be easier to read with `datasetsCount(er)` and `tagsCount(er)`

udata/migrations/2016-11-08-truncate-tags.js
```diff
@@ -0,0 +1,34 @@
+/*
+ * Ensure all tags have a correct length
+ *
+ */
+
+const MIN_TAG_LENGTH = 3
+const MAX_TAG_LENGTH = 32
+
+var datasets = db.dataset.find({ tags: { $gt: [] } })
```
noirbizarre: `reuse` collection need the same treatment

udata/forms/fields.py
```diff
@@ -285,6 +284,14 @@ def iter_choices(self):
...
 
 
+def slug(value):
+    return slugify(value.lower())
+
+
+def tags_list(value):
+    return list(set(slug(tag) for tag in value.split(',') if tag.strip()))
+
+
```
noirbizarre: I'd prefer to have it in a separate module because it's not more related to forms than harvest.
I don't like `harvest` having a new dependency on `udata.forms.fields` :/

udata/forms/fields.py
```diff
@@ -296,9 +303,7 @@ def process_formdata(self, valuelist):
         if valuelist and len(valuelist) > 1:
             self.data = valuelist
         elif valuelist:
-            self.data = list(set([
-                x.strip().lower()
-                for x in valuelist[0].split(',') if x.strip()]))
+            self.data = tags_list(valuelist[0])
```
noirbizarre: 👍

udata/forms/fields.py
```diff
@@ -296,9 +303,7 @@ def process_formdata(self, valuelist):
         if valuelist and len(valuelist) > 1:
             self.data = valuelist
```
noirbizarre: I think this a candidate normalization too.

udata/forms/fields.py
noirbizarre: Proposal:
- module `udata.tags`
- members:
  - `MIN_LENGTH`
  - `MAX_LENGTH`
  - `normalize`
  - `from_list`
(this only a proposal, names are up to you ;))

udata/forms/fields.py
jphnoel: I do not understand what you mean.

jphnoel: BTW, I kept the MIN_TAG_LENGTH and MAX_TAG_LENGTH in fields.py because it is used here only, i.e. not by the harvesters. Do you think the harvesters should crop the tags when they have not the right length ? Or ignore them ? Or raise an error ?

noirbizarre: Normlization => `self.data = [tags.slug(v) for v in valuelist]`

udata/tags.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from slugify import slugify
+
+
+def slug(value):
```
noirbizarre: I prefer `normalize` because slug means "url part for permalink/pretty URLs".
I think at some we might have some different rules for normalized tags and slugs

udata/tests/api/test_datasets_api.py
```diff
@@ -144,6 +144,38 @@ def test_dataset_api_create_tags(self):
...
+        self.assertStatus(response, 400)
+
+    @attr('create')
+    def test_dataset_api_create_and_slugify_tags(self):
+        '''It should create a dataset from the API and slugify the tags'''
+        data = DatasetFactory.attributes()
+        data['tags'] = [' Aaa bBB $$ $$-µ  ']
+        with self.api_user():
+            response = self.post(url_for('api.datasets'), data)
+        self.assertStatus(response, 201)
```
noirbizarre: or `self.assert201` ;)

udata/migrations/2016-11-08-truncate-tags.js
```diff
@@ -0,0 +1,42 @@
...
+    var name = collection.getName()
+
+    print(name + ' candidates : ' + objects.count());
+
+    var objectsCounter = 0;
+    var tagsCounter = 0;
+
+    objects.forEach(function(object) {
+        var tags = object.tags.map(function(tag) {
+            if (tag.length > 32)  {
```
noirbizarre: Use `MAX_TAG_LENGTH` defined above

udata/migrations/2016-11-08-truncate-tags.js
```diff
@@ -0,0 +1,42 @@
...
+                tagsCounter++;
+                return tag.substring(0, 32)
```
noirbizarre: Use `MAX_TAG_LENGTH` defined above

udata/migrations/2016-11-08-truncate-tags.js
```diff
@@ -0,0 +1,42 @@
...
+            }
+            if (tag.length < 3)  {
```
noirbizarre: Use `MIN_TAG_LENGTH` defined above

js/models/reuse.js
noirbizarre: Last remaining fix to apply

udata/harvest/tests/test_ckan_backend.py
```diff
@@ -237,13 +237,15 @@ def package_show_responses(request, uri, headers):
...
-                                  'broadband',
-                                  'cables',
-                                  'fibre',
-                                  'optic',
-                                  'terrestrial'])
+        self.assertEqual(
+            d.tags,
+            ['africa',
+             # 'b' must be absent,
+             'b' * 32,  # 'b' * 33 must be truncated
```
noirbizarre: Might be even more explicit with `MAX_LENGTH` instead of 32 ;)

udata/migrations/2016-11-08-truncate-tags.js
```diff
@@ -0,0 +1,42 @@
...
+                tagsCounter++;
+                return tag.substring(0, MAX_TAG_LENGTH)
+            }
+            if (tag.length < MIN_TAG_LENGTH)  {
+                tagsCounter++;
+                return null
+            }
+            return tag;
+        });
+        tags = tags.filter(function(tag) { return tag !== null })
```
noirbizarre: Not necessary but just so you know: Mongo 3.2 support ES6 arrow functions so you can write:
```javascript
tags = tags.filter(tag => tag)  // Works because filter cast as boolean
```

PR #598, opened 2016-10-25
Manage the users in the admin interface (fixes #541)

### LINKED ISSUE #541: As an administrator, I want to manage the users from the admin interface
As a user, I want to :
- [x] Edit the web site URL in the user form

As an administrator, I want to manage the users from the admin interface.
- [x] Ability to update the parameters of a user via a form
- [x] Ability to select the roles of a user in the form
- [x] Display the roles of the users in the users list view (no sort, no search)
- [x] Ability to toggle the active status of a user in the form

As an administrator, I want that:
- [x] When receiving a request for the public user profile API for an inactive user, send a 410 if the current user is anonymous or not admin and answer normally otherwise
- [x] When receiving a request for the user activity page for an inactive user, send a 410 if the current user is anonymous or not admin and answer normally otherwise

jphnoel:


js/components/user/form.vue
```diff
@@ -7,24 +7,44 @@ import User from 'models/user';
...
 export default {
     props: {
+        // Edited user
         user: {
             type: User,
             default: function() {return new User();}
-        }
+        },
+        // Admin property of the editing user (i.e. NOT of the edited one)
+        admin: {
```
noirbizarre: I is already available with `$root.me.is_admin` in templates and `this.$root.me.is_admin` in Vue code

js/components/user/form.vue
```diff
@@ -7,24 +7,44 @@ import User from 'models/user';
...
+                label: this._('First name'),
+            },
+            {
+                id: 'last_name',
+                label: this._('Last name'),
+            },
+            {
+                id: 'about',
+                label: this._('About'),
+            },
```
noirbizarre: I think it's missing some fields (ex: website)

js/components/user/form.vue
```diff
@@ -7,24 +7,44 @@ import User from 'models/user';
...
+        ]
+        if (this.admin === true) {
```
noirbizarre: use `this.$root.me.is_admin`

js/views/user-edit.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+<form-layout icon="user" :title="_('Edit profile')" :save="save" :cancel="cancel" footer="true" :model="user">
```
noirbizarre: As there is already an edit profile, I think the title should be "Edit user"

js/views/user-edit.vue
```diff
@@ -0,0 +1,40 @@
...
+    <user-form v-ref:form :user="user" :admin="admin"></user-form>
```
noirbizarre: `admin` can be removed given the previous comments

js/views/user-edit.vue
```diff
@@ -0,0 +1,40 @@
...
+</template>
+
+<script>
+import UserForm from 'components/user/form.vue';
+import User from 'models/user';
+import FormLayout from 'components/form-layout.vue';
+
+export default {
+    data: function() {
+        return {user: new User(), admin: true};
```
noirbizarre: `admin` can be removed given the previous comments

js/components/user/list.vue
```diff
@@ -58,6 +58,13 @@ export default {
                 align: 'center',
                 type: 'metric',
                 width: 95
+            }, {
+                label: this._('Roles'),
+                key: 'roles',
+                sort: 'roles',
+                align: 'center',
+                type: 'metric',
```
noirbizarre: Hum, it's not a metric. This type is meant to change color given the value (and so expect a number).
You should use type `label`, a `label_type` function and maybe a `label_func` (see: https://github.com/opendatateam/udata/blob/dev/js/components/datatable/cells/label.vue)

udata/core/user/search.py
```diff
@@ -48,6 +49,7 @@ class UserSearch(ModelSearchAdapter):
         'followers': Sort('metrics.followers'),
         'views': Sort('metrics.views'),
         'created': Sort('created'),
+        'roles': Sort('roles'),
```
noirbizarre: I think filtering is sufficient, sorting on `roles` does not seems relevant to me:
- a user can have many roles: which one is taken in account ?
- this is not a very variable value: there will be an unordered block of admins then others (and the opposite if the sort is reversed)

udata/core/user/search.py
```diff
@@ -79,5 +81,6 @@ def serialize(cls, user):
                     'slug': user.slug,
                 },
             },
-            'visible': user.visible
+            'visible': user.visible,
+            'roles': [role.name for role in user.roles],
```
noirbizarre: Allow filtering on roles, usefull 👍

js/components/user/list.vue
```diff
@@ -58,6 +58,13 @@ export default {
                 align: 'center',
                 type: 'metric',
                 width: 95
+            }, {
+                label: this._('Roles'),
+                key: 'roles',
+                sort: 'roles',
```
noirbizarre: Sorting on roles does not seems relevant to me. Which role is the sort key when many roles ?

js/components/user/list.vue
jphnoel: I agree it is not a metric.
But it is not really a label, since there can be, for one user, several roles.
Is there something similar in the code, for tags or badges for instance ?

js/components/user/list.vue
jphnoel: For me, in production, a user has either one role (admin) or none.
With the sort, we can, with one clic on column header, view all the users with the admin privilege.
With the filtering, you would have to type explicitly 'admin' in the search control.
Sorting also allows you to group the users having the same roles.

noirbizarre: OK, right now this is the case, but more roles are coming very very soon (it's not the place to discuss thoses). So the question is still legit: what happens when a user have 2 roles ?

udata/forms/fields.py
```diff
@@ -80,6 +80,19 @@ class IntegerField(FieldHelper, fields.IntegerField):
     pass
 
 
+class RolesField(FieldHelper, fields.StringField):
+    def process_formdata(self, valuelist):
+        self.data = []
+        if valuelist:
```
davidbgk: Does the default `valuelist` return `None`? Otherwise iterating over an empty list is OK.

js/components/form/base-form.js
```diff
@@ -195,7 +195,15 @@ export default {
 
             Array.prototype.map.call(elements, function(el) {
                 let value;
-                if (TEXT_TAGS.includes(el.tagName.toLowerCase()) || TEXT_INPUTS.includes(el.type.toLowerCase())) {
+                if (el.tagName.toLowerCase() === 'select' && el.multiple === true) {
```
davidbgk: Do you have to specifically test `=== true` here and below?

js/components/user/form.vue
```diff
@@ -7,24 +7,53 @@ import User from 'models/user';
...
+            fields.push(
+                {
+                    id: 'roles',
+                    label: this._('Roles'),
+                    widget: 'select-input',
+                    multiple: true,
+                    values: [{id: 'admin', name: 'admin'},
+                             {id: 'editor', name: 'editor'},
+                             {id: 'moderator', name: 'moderator'}],
+                    map: function(item) {
```
davidbgk: `map: item => {value: item.id, text: item.name};` should work here.

udata/core/user/api.py
```diff
@@ -238,14 +238,26 @@ def post(self):
...
+        current_user.sysadmin is False)
+
+
 @ns.route('/<user:user>/', endpoint='user')
 @api.response(404, 'User not found')
-@api.response(410, 'User has been deleted')
+@api.response(410, 'User is not active or has been deleted')
 class UserAPI(API):
+
+    def check_availability(self, user):
```
davidbgk: I think it's a good candidate for a decorator.

js/components/user/form.vue
jphnoel: map: item => {return {value: item.name, text: item.name}} works

If I remove the first level of {}, it does not work any more. And if I remove the return, it does not work too, probably because the {{ }} are misinterpreted by vuejs.

udata/core/user/api.py
jphnoel: I agree, it smells like a decorator and that is actually what I started doing... before I stopped !

Indeed, if you put it as a decorator, it actually does not reduce the size of the code and it make it less readable.

js/components/form/base-form.js
jphnoel: No, indeed...

udata/forms/fields.py
jphnoel: Very true !

udata/core/user/views.py
```diff
@@ -3,8 +3,9 @@
 
 import logging
 
-from flask import g
+from flask import abort, g
 
+from udata.core.user.api import is_not_available
```
noirbizarre: No, there shouldn't be a dependency from `views` to `api`, so it might either be a `user.models.User` method or `user.models` function.
**Proposal**: a `is_visible()` method on `User` model

udata/forms/fields.py
```diff
@@ -115,7 +127,13 @@ class FileField(FieldHelper, fields.FileField):
 
 
 class URLField(FieldHelper, EmptyNone, html5.URLField):
-    pass
+    def pre_validate(self, form):
```
noirbizarre: Are you sure it's not already checked by `html5.URLField` ? If it's not the case, it might be better to stick with WTForms validator, ie. [`wtforms.validators.URL`](http://wtforms.readthedocs.io/en/latest/validators.html#wtforms.validators.URL)

udata/tests/frontend/test_user_frontend.py
```diff
@@ -64,6 +64,12 @@ def test_render_profile(self):
         self.assertEquals(json_ld['url'], 'http://www.datagouv.fr/user')
         self.assertEquals(json_ld['image'], 'http://www.datagouv.fr/avatar')
 
+    def test_render_profile_of_an_inactive_user(self):
```
noirbizarre: `test_cant_render` or `test_dont_render` ?
Might be great to also test that admins can still display it.

udata/tests/frontend/test_user_frontend.py
```diff
@@ -64,6 +64,12 @@ def test_render_profile(self):
...
+        '''It should raise a 410'''
```
noirbizarre: ... when the user is inactive

js/components/form/base-form.js
```diff
@@ -195,7 +195,15 @@ export default {
 
             Array.prototype.map.call(elements, function(el) {
                 let value;
-                if (TEXT_TAGS.includes(el.tagName.toLowerCase()) || TEXT_INPUTS.includes(el.type.toLowerCase())) {
+                if (el.tagName.toLowerCase() === 'select' && el.multiple) {
+                    value = []
+                    for (const index in el.options) {
```
jphnoel: Actually, that is what I wanted to do.

But el.options is not really an arry, so I had to call Array.prototype.map.call and Array.prototype.filter.call, which made the code a bit hard to read.

udata/core/user/api.py
```diff
@@ -226,48 +228,62 @@ def get(self):
...
         '''Create a new object'''
-        form = api.validate(UserProfileForm)
+        form = api.validate(UserProfileAdminForm)
         user = form.save()
         return user, 201
 
 
+def is_not_available(user):
+    return user.active is False and (
+        current_user.is_anonymous or
```
jphnoel: If the user is anonymous, it has not even the property sysadmin.

I could check the existence of the property, but it seems less readable to me.

udata/forms/fields.py
jphnoel: Yes, I am sure. I implemented this behavior because the error was not detected at the form level, but at the database level, which raised a 500.

It is easy to use wtforms.validators.URL but I am not in favor of it:
- The validation is much lighter than at the Mongo level. For foobarbaz fails, but htt://www.lemonde.fr passes. If the data submitted by the client is not invalidated by the form, the client receives a 500 and does not understand where the problem comes from.
- 5 forms uses a URLField, which means that we would have to copy times the same validation code.

Let me know what you prefer.

js/views/user-edit.vue
```diff
@@ -0,0 +1,42 @@
...
+import FormLayout from 'components/form-layout.vue';
+
+export default {
+    data() {
+        return {user: new User()};
+    },
+    components: {FormLayout, UserForm},
+    methods: {
+        save() {
+            let form = this.$refs.form;
```
noirbizarre: form is not reaffected so `const` instead of `let`

udata/core/user/views.py
```diff
@@ -75,6 +75,8 @@ class UserActivityView(UserView, DetailView):
     template_name = 'user/activity.html'
 
     def get_context(self):
+        if not self.user.visible:
+            abort(410)
```
noirbizarre: Add an error message to be more explicit to the user

udata/forms/fields.py
noirbizarre: 👍

udata/tests/api/test_me_api.py
```diff
@@ -35,12 +35,15 @@ def test_update_profile(self):
         '''It should update my profile from the API'''
         self.login()
         data = self.user.to_dict()
+        self.assertEqual(self.user.active, True)
```
noirbizarre: So `assertTrue` ;)

udata/tests/api/test_me_api.py
```diff
@@ -35,12 +35,15 @@ def test_update_profile(self):
...
         data['about'] = 'new about'
+        data['active'] = False
         response = self.put(url_for('api.me'), data)
         self.assert200(response)
         self.assertEqual(User.objects.count(), 1)
         self.user.reload()
         self.assertEqual(self.user.about, 'new about')
+        self.assertEqual(self.user.active, True)
```
noirbizarre: `assertTrue`

udata/tests/api/test_user_api.py
```diff
@@ -1,9 +1,11 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+import json
```
noirbizarre: No need to manually parse `response.data` every time, you can directly use `response.json`

udata/tests/api/test_user_api.py
```diff
@@ -189,3 +191,88 @@ def test_users(self):
...
+    def test_user_api_update(self):
+        '''It should update a user'''
+        self.login(AdminFactory())
+        user = UserFactory()
+        data = user.to_dict()
+        data['active'] = False
+        response = self.put(url_for('api.user', user=user), data)
+        self.assert200(response)
+        response = json.loads(response.data)
+        self.assertEqual(response['active'], False)
```
noirbizarre: `self.assertFalse(response.json['active'])`

udata/tests/api/test_user_api.py
```diff
@@ -189,3 +191,88 @@ def test_users(self):
...
+    def test_user_api_update_with_an_existing_role(self):
+        '''It should update a user'''
+        self.login(AdminFactory())
+        user = UserFactory()
+        data = user.to_dict()
+        data['roles'] = ['admin']
+        response = self.put(url_for('api.user', user=user), data)
+        self.assert200(response)
+        response = json.loads(response.data)
+        self.assertEqual(response['roles'], ['admin'])
```
noirbizarre: `self.assertEqual(response.json['roles'], ['admin'])`

udata/tests/api/test_user_api.py
```diff
@@ -189,3 +191,88 @@ def test_users(self):
...
+        response = self.put(url_for('api.user', user=user), data)
+        self.assert400(response)
+
+    def test_user_roles(self):
+        '''It should list the roles'''
+        self.login(AdminFactory())
+        response = self.get(url_for('api.user_roles'))
+        self.assert200(response)
+        response = json.loads(response.data)
+        self.assertEqual(response, [{u'name': u'admin'}])
```
noirbizarre: `self.assertEqual(response.json, [{'name': 'admin'}])`
No need for `u`, there is already `from __future__ import unicode_literals`, we are using unicode be default

udata/core/user/models.py
```diff
@@ -120,9 +122,14 @@ def external_url(self):
...
-    def visible(self):
+    def visible_for_search(self):
         count = self.metrics.get('datasets', 0) + self.metrics.get('reuses', 0)
-        return count > 0
+        return count > 0 and self.active
+
+    @property
+    def visible(self):
+        return self.active or (
+            not current_user.is_anonymous and current_user.sysadmin)
```
jphnoel: I am lost...
Do you think that:
1) the code I propose (which uses current_user), is correct. That would be great ;-)
2) whe should check that current_user is defined. In that case, I have not been able to reproduce this situation. Can you give me a scenario to reproduce it ?

udata/core/user/api.py
```diff
@@ -226,51 +227,57 @@ def get(self):
...
 
 @ns.route('/<user:user>/', endpoint='user')
 @api.response(404, 'User not found')
-@api.response(410, 'User has been deleted')
+@api.response(410, 'User is not active or has been deleted')
 class UserAPI(API):
+
+    def check_permissions(self, user, readonly=False):
+        if not user.visible:
+            api.abort(410, 'User is not active')
```
jphnoel: It cannot be user.active is already a field in the database.
'visible' : means the user is inactive and you are not admin. But when the error is raised, the message must be 'user is not active'. It should not be 'user is not active and you are not admin'.

udata/core/user/api.py
```diff
@@ -226,51 +227,63 @@ def get(self):
...
+            # We cannot modify a deleted user
+            not readonly or (
+                # We cannot get a deleted user if we do not have the
+                # permissions
+                readonly and not UserEditPermission(user).can())
+        ):
+            api.abort(410, 'User has been deleted')
+        elif not readonly:
+            # We must have the permission to modify the user
+            UserEditPermission(user).test()
```
davidbgk: I'm always experimenting on that kind of `elif`-based permissions, here is a proposition:

```python
def check_permissions(self, user, readonly=False):
    user_edit_permission = UserEditPermission(user)
    is_active = not user.visible
    is_editable = not readonly
    admin_access = readonly and not user_edit_permission.can()
    has_been_deleted = user.deleted and (is_editable or admin_access)
    if is_active:
        api.abort(410, 'User is not active')
    elif has_been_deleted:
        api.abort(410, 'User has been deleted')
    elif is_editable:
        user_edit_permission.test()
```
The goal is to clarify what is done in the second part without having too much imbricated tests that are blurring the algorithm. Let's discuss this, as I said I'm not sure that it's more readable at the end.

davidbgk:
Impressed by the amount of work done since my last check 👍

udata/core/user/models.py
noirbizarre: 1. I think this only correct in views because it needs a `current_user`. By example, [here](https://github.com/opendatateam/udata/blob/dev/udata/core/user/search.py#L100) it will always be `False`).
2. no, I think model properties should not involve `current_user` which is only views related
To simplify, the clause should only be `return self.active` and we will ad more rules later if needed. The `current_user` should only happen in views (ie. `user.visible and current_user...`)

PR #606, opened 2016-11-08
Resources permalink

noirbizarre:
This PR adds permalinks to resources ie. an URL always redirecting to the latest version of a resource.

This URL is shown in the front resource details modal et in the `permalink` resource's API field.

In the distribution/resource JSON-LD, `url` is  the permalink whereas `contentUrl` is the direct link.

### DIFF
--- a/js/front/dataset/resource-modal.vue
+++ b/js/front/dataset/resource-modal.vue
@@ -7,6 +7,8 @@
 
         <dl class="dl-horizontal dl-wide">
           <dt>{{ _('URL') }}</dt>
+          <dd><a :href="resource.contentUrl" @click="onClick">{{resource.contentUrl}}</a></dd>
+          <dt>{{ _('Permalink') }}</dt>
           <dd><a :href="resource.url" @click="onClick">{{resource.url}}</a></dd>
           <dt v-if="resource.encodingFormat">{{ _('Format') }}</dt>
           <dd v-if="resource.encodingFormat">{{resource.encodingFormat}}</dd>
--- a/js/front/dataset/resource-modal.vue
+++ b/js/front/dataset/resource-modal.vue
@@ -8,7 +8,7 @@
         <dl class="dl-horizontal dl-wide">
           <dt>{{ _('URL') }}</dt>
           <dd><a :href="resource.contentUrl" @click="onClick">{{resource.contentUrl}}</a></dd>
-          <dt>{{ _('Permalink') }}</dt>
+          <dt>{{ _('Latest URL') }}</dt>
           <dd><a :href="resource.url" @click="onClick">{{resource.url}}</a></dd>
           <dt v-if="resource.encodingFormat">{{ _('Format') }}</dt>
           <dd v-if="resource.encodingFormat">{{resource.encodingFormat}}</dd>
--- a/udata/core/dataset/api_fields.py
+++ b/udata/core/dataset/api_fields.py
@@ -44,6 +44,10 @@ resource_fields = api.model('Resource', {
         required=True, enum=RESOURCE_TYPES.keys()),
     'format': fields.String(description='The resource format', required=True),
     'url': fields.String(description='The resource URL', required=True),
+    'permalink': fields.UrlFor('datasets.resource',
+                               lambda o: {'id': o.id},
+                               description='The resource permanent URL',
+                               readonly=True),
     'checksum': fields.Nested(
         checksum_fields, allow_null=True,
         description='A checksum to validate file validity'),
--- a/udata/core/dataset/api_fields.py
+++ b/udata/core/dataset/api_fields.py
@@ -44,11 +44,11 @@ resource_fields = api.model('Resource', {
         required=True, enum=RESOURCE_TYPES.keys()),
     'format': fields.String(description='The resource format', required=True),
     'url': fields.String(description='The resource URL', required=True),
-    'permalink': fields.String(description='The permanent URL redirecting to '
-                               'the latest version of the resource. When the '
-                               'resource data is updated, the URL will '
-                               'change, the permalink won\'t.',
-                               readonly=True),
+    'latest': fields.String(description='The permanent URL redirecting to '
+                            'the latest version of the resource. When the '
+                            'resource data is updated, the URL will '
+                            'change, the latest URL won\'t.',
+                            readonly=True),
     'checksum': fields.Nested(
         checksum_fields, allow_null=True,
         description='A checksum to validate file validity'),
--- a/udata/core/dataset/api_fields.py
+++ b/udata/core/dataset/api_fields.py
@@ -44,7 +44,8 @@ resource_fields = api.model('Resource', {
         required=True, enum=RESOURCE_TYPES.keys()),
     'format': fields.String(description='The resource format', required=True),
     'url': fields.String(description='The resource URL', required=True),
-    'permalink': fields.String(description='The resource permanent URL',
+    'permalink': fields.String(description='The resource permanent URL to '
+                               'the latest version (won\'t change on update)',
                                readonly=True),
     'checksum': fields.Nested(
         checksum_fields, allow_null=True,
--- a/udata/core/dataset/api_fields.py
+++ b/udata/core/dataset/api_fields.py
@@ -44,9 +44,7 @@ resource_fields = api.model('Resource', {
         required=True, enum=RESOURCE_TYPES.keys()),
     'format': fields.String(description='The resource format', required=True),
     'url': fields.String(description='The resource URL', required=True),
-    'permalink': fields.UrlFor('datasets.resource',
-                               lambda o: {'id': o.id},
-                               description='The resource permanent URL',
+    'permalink': fields.String(description='The resource permanent URL',
                                readonly=True),
     'checksum': fields.Nested(
         checksum_fields, allow_null=True,
--- a/udata/core/dataset/api_fields.py
+++ b/udata/core/dataset/api_fields.py
@@ -44,8 +44,10 @@ resource_fields = api.model('Resource', {
         required=True, enum=RESOURCE_TYPES.keys()),
     'format': fields.String(description='The resource format', required=True),
     'url': fields.String(description='The resource URL', required=True),
-    'permalink': fields.String(description='The resource permanent URL to '
-                               'the latest version (won\'t change on update)',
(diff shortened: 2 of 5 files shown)


noirbizarre:
Yes, it was discussed, both URL are relavant:
- I want to have a specific version and not being impacted by new ones => I use the direct link
- I want to always have the latest version of a resource => I use the permalink (which also adds an extra 302 redirect
Choices is to the end user but both use cases exists.

Yes, there should not be any collision between IDs, they are UUID generated client-side.
If there is a risk, it also exists within the same collection so this PR do not increase it.

davidbgk:
> Choice is to the end user but both use cases exists.

Where and how to document that?

noirbizarre:
Good question. Too me it looks like a FAQ entry, but sadly FAQ is not included in udata.
On the API side, it's documented into the description.

noirbizarre:
BTW, this PR is backward compatible: it adds a field, do not touch the existing ones

js/front/dataset/resource-modal.vue
```diff
@@ -7,6 +7,8 @@
 
         <dl class="dl-horizontal dl-wide">
           <dt>{{ _('URL') }}</dt>
+          <dd><a :href="resource.contentUrl" @click="onClick">{{resource.contentUrl}}</a></dd>
+          <dt>{{ _('Permalink') }}</dt>
```
davidbgk: Maybe document the difference here too?

js/front/dataset/resource-modal.vue
```diff
@@ -7,6 +7,8 @@
 
         <dl class="dl-horizontal dl-wide">
           <dt>{{ _('URL') }}</dt>
+          <dd><a :href="resource.contentUrl" @click="onClick">{{resource.contentUrl}}</a></dd>
```
davidbgk: Isn't it `.url` here and `.permalink` below? Where does that `contentUrl` come from?

udata/core/dataset/api_fields.py
```diff
@@ -44,6 +44,9 @@
         required=True, enum=RESOURCE_TYPES.keys()),
     'format': fields.String(description='The resource format', required=True),
     'url': fields.String(description='The resource URL', required=True),
+    'permalink': fields.String(description='The resource permanent URL to '
+                               'the latest version (won\'t change on update)',
```
davidbgk: Maybe we should be more explicit here, proposition:

> The permanent URL to the latest version of the resource (returns a 302 redirect). When that resource is updated, the data behind the URL will change.

udata/core/dataset/models.py
```diff
@@ -151,13 +151,22 @@ def check_availability(self, group):
     def is_available(self):
         return self.check_availability(group=None)
 
+    @property
+    def permalink(self):
+        '''
+        Permanent link to the latest version of this resource.
+
+        If this resource is updated and `url` change, this property won't.
```
davidbgk: Typo: changeS

udata/core/dataset/models.py
```diff
@@ -151,13 +151,22 @@ def check_availability(self, group):
...
+        return url_for('datasets.resource', id=self.id, _external=True)
+
     @cached_property
     def json_ld(self):
 
         result = {
             '@type': 'DataDownload',
             '@id': str(self.id),
-            'url': self.url,
+            'url': self.permalink,
```
davidbgk: Alright, now I see where it comes from! Let's try to be consistent here and rename `contentUrl` to `url` and `url` to `permalink`.

udata/core/dataset/views.py
```diff
@@ -103,6 +104,19 @@ def get_context(self):
...
+@blueprint.route('/r/<uuid:id>', endpoint='resource')
+def resource_redirect(id):
+    '''
+    Redirect to the latest version of a resource given its identifier.
+    '''
+    dataset = Dataset.objects(resources__id=id).first()
+    if dataset:
+        resource = get_by(dataset.resources, 'id', id)
+    else:
+        resource = CommunityResource.objects(id=id).first()
```
davidbgk: Why not a `.get()`?

js/front/dataset/resource-modal.vue
noirbizarre: these are JSON-LD attributes

js/front/dataset/resource-modal.vue
noirbizarre: I thought about it but there is not enough space this is ugly :/
I'll readd it

udata/core/dataset/models.py
noirbizarre: No we can't, it follows the schema specifications:
- https://schema.org/Dataset
- https://schema.org/DataDownload

udata/core/dataset/views.py
noirbizarre: Because `get()` raise an exception if the object does not exists and I needed to have a `none` value in this case. I don't wanted to  add an extra `try...catch` just to set it to `None`

js/front/dataset/resource-modal.vue
davidbgk: Maybe `Latest URL` is more explicit?

udata/core/dataset/views.py
davidbgk: What about performances?

udata/core/dataset/api_fields.py
noirbizarre: We can be more explicit (I will try), but:
- I won't specify the 302 behavior which seems an internal to me
- On the second sentence, this is not exactly true: URL change only on data update. If someone update the resource description, data behind the URL won't change.

Proposal:

> The permanent URL to the latest version of the resource.
> When the resource data is updated, the URL will change, the permalink won't.

?

davidbgk: As a consumer of the API, I'd be interested in knowing that I'll have to follow the redirection to actually get the file. OK for the second part.

js/front/dataset/resource-modal.vue
noirbizarre: To me `Latest URL` seems dynamic, meaning that it's changing over the time.
Whereas permalink is totally defined: https://en.wikipedia.org/wiki/Permalink

udata/core/dataset/models.py
noirbizarre: We need to have a way to display it. I don't wanted to add an undocumented attribute.

udata/core/dataset/views.py
noirbizarre: Same, unique index

js/front/dataset/resource-modal.vue
davidbgk: To me a permalink is stable both in the URI *and* in the content delivered, see a quote from that wikipedia page:

> In the context of systems that support versioning, such as most wikis, a permalink is commonly understood as a link to a *specific version*.

So semantically, we should name the `url` permalink and the redirecting one `latest` 😉

udata/core/dataset/models.py
noirbizarre: BTW, this is exactly the use case, the reason why there is an extra `contentUrl`: `url` is e generic URL also used as a an object identifier (can be a presentation page of a music track) whereas `contentUrl` is a direct access to the content (in case of a music track, it will be the direct access to the binary file ie. mp3 or whatever)

js/front/dataset/resource-modal.vue
noirbizarre: OK OK OK! 😫

udata/core/dataset/api_fields.py
```diff
@@ -44,6 +44,11 @@
         required=True, enum=RESOURCE_TYPES.keys()),
     'format': fields.String(description='The resource format', required=True),
     'url': fields.String(description='The resource URL', required=True),
+    'permalink': fields.String(description='The permanent URL redirecting to '
```
davidbgk: I think we should rename that attr to `latest` because `permalink` is confusing given that it is not a permalink.

udata/core/dataset/models.py
```diff
@@ -151,13 +151,22 @@ def check_availability(self, group):
     def is_available(self):
         return self.check_availability(group=None)
 
+    @property
+    def permalink(self):
```
davidbgk: Same here with `latest`.

udata/core/dataset/views.py
```diff
@@ -103,6 +104,19 @@ def get_context(self):
         return context
 
 
+@blueprint.route('/r/<uuid:id>', endpoint='resource')
```
davidbgk: What about `'/latest/<uuid:id>'` too?

noirbizarre: I renamed the property but I'd prefer to keep this pattern because it produce shorter URLs and I find it more coherent (I would have agree if patterns where resource related):
- `/datasets/some-dataset-slug/r/cc3cb0d0-87ed-4eb6-aced-470c083741cd`
- vs.
- `/datasets/some-dataset-slug/latest/cc3cb0d0-87ed-4eb6-aced-470c083741cd`

Are you OK with this ?

udata/tests/frontend/test_dataset_frontend.py
```diff
@@ -140,6 +141,30 @@ def test_not_found(self):
         response = self.get(url_for('datasets.show', dataset='not-found'))
         self.assert404(response)
 
+    def test_resource_permalink(self):
```
davidbgk: For consistency, maybe rename the tests too?

udata/core/dataset/views.py
davidbgk: Alright, I misinterpreted where that route was issued, I thought it was at the root like directly `data.gouv.fr/r/cc3cb0d0-87ed-4eb6-aced-470c083741cd` I wonder if it is required to nest it. What about a more general shortcut to access orgs, datasets and resources (probably out of the scope of that PR)

noirbizarre: Right now, with the way this is structured, yes (or I need to declare another blueprint).

I agree with the naming pattern (and the fact this outside this PR scope). At the beginning, the URLs were:
* `/d/dataset-slug/ for datasets
* `/o/organization-slug/ for organizations
* `/r/reuses-slug/ for for reuse

There might be something to to, at least for datasets.

davidbgk: 👌 LGTM!

PR #608, opened 2016-11-09
Ability to perform ORs on the API facets, fix #525

### LINKED ISSUE #525: Add the ability to filter on several organizations
The API should be able to make a filter on several organizations in the API.
It would permit to create a child catalog based on https://github.com/DepthFrance/udata-js in which we could search for datasets filtered on a selection of organizations.

The API datasets request could be extended to accept several organization id in the organization option such as 534fffb3a3a7292c64a78129,551959f6c751df6682057c91,558bef88c751df1fd9a453b9

davidbgk:


udata/search/fields.py
```diff
@@ -56,7 +58,20 @@ def get_value_filter(self, value):
 
 
 class TermsFacet(Facet, DSLTermsFacet):
-    pass
+
+    def add_filter(self, filter_values):
+        if not filter_values:
```
noirbizarre: I think that given than the only case we want to handle is `OR`, it could be refactored as:
```python
def add_filter(self, filter_values:
    if OR_OPERATOR not in filter_values:
        return super(TermsFacet, self).add_filter(filter_values)
    values = filter_values.split(OR_OPERATOR)
    return any(super(TermsFacet, self).add_filter(v) for v in values)
```
And so it preserve the parent logic.
What do you think about ?

udata/tests/api/test_datasets_api.py
```diff
@@ -47,6 +47,44 @@ def test_dataset_api_list(self):
         self.assertEqual(len(response.json['data']), len(datasets))
         self.assertFalse('quality' in response.json['data'][0])
 
+    def test_dataset_api_list_filtered_by_org(self):
+        '''It should fetch a dataset list for a given org'''
+        self.login()
+        with self.autoindex():
+            member = Member(user=self.user, role='editor')
+            org = OrganizationFactory(members=[member])
+            DatasetFactory(resources=[ResourceFactory()])
```
noirbizarre: You can also use `VisibleDatasetFactory`: it prevent manual resource creation and if the visibility rule evolve the test stay valid.

noirbizarre: (apply for all dataset instanciations)

udata/search/fields.py
noirbizarre: I forgot to explain: the `must`/`should`/`shouldnot` is the default ES `AND`/`OR`/`NOT` operations and ES-DSL automatically use it with boolean operations so:
```python
q1, q2 = Q(), Q()
q1 or q2 == Q('bool', should=[q1, q2])
```

davidbgk: We are looking for the `OR` operator within each item of the `filter_values` list so it's sadly not that simple.

noirbizarre: Sorry, I misread the plural.
```python
def add_filter(self, filter_values):
    # Perform a classic AND on values wihtout OR operator
    and_values = (v for v in filter_values if OR_OPERATOR not in v)
    and_filter = super(TermsFacet, self).add_filter(and_values)
    # Build a OR query for each value with OR
    or_values = (v for v in filter_values if OR_OPERATOR in v)
    or_filters = (
        any(Q('term',  , **{field_name: v}) for v in values.split(OR_OPERATOR))
        for values in or_values
    )  # Could be wrapped in all() but we do it once and for with and_filter
    return all(and_filter, *or_filters)
```
I tried to use generators to limit memory usage (but if Python complains, just switch to lists).
Attention, its initially a `terms` operator, not a `match` and to operate an OR, it should become multiple `term`.

What changed:
- a single `terms` query for values without `OR` operator instead of a `match` for each
- preserve the initial `AND` between filter_values (ie. `t=a|b t=c|d` should be `(a OR b) AND (c OR d)` and not `a OR b OR c OR d`

It put to my attention that dataset search `organization` field (and any field storing an ID) should have `index='not_analyzed'`

udata/tests/api/test_user_api.py
```diff
@@ -161,15 +161,16 @@ def test_suggest_users_api_by_id(self):
...
 
-        self.assertGreaterEqual(len(response.json), 1)
-
-        suggestion = response.json[0]
-        self.assertEqual(suggestion['id'], str(user.id))
+        # The batch factory generates ids that might be too close
+        # which then are found with the fuzzy search.
+        suggested_ids = [u['id'] for u in response.json]
+        self.assertGreaterEqual(len(suggested_ids), 1)
+        self.assertIn(str(first_user.id), suggested_ids)
```
noirbizarre: You choose to keep the fuzzy search, OK 👍
(We will have to this carefully on completer to ensure fuzzy does not create too much noise)

udata/search/fields.py
davidbgk: After a few attempts, it doesn't work because you can't return lists/iterators here.

Let's do that the other way:

1. which cases should we cover? addressed to whom?
2. add tests related to these cases
3. implement the solution 😄

noirbizarre: It doesn't coerce to a list, `all` perform a boolean product (ie. apply `AND` on all members) and `any same with `OR`.
I made (at least) 2 mistakes in my sample code (I wrote it without testing it):
- `all` and `any` works in shortcircuit evalutation (ie. stops whenever it can so doesn't evaluate all proposition).
- both expect a list or an iterator, not variables args

I'll make a PR to your PR with a tested proposal

davidbgk: Alright, I'm interested in concrete tests that are not addressed by the current PR. Be aware that the return of the `add_filter` function should be something `clone`able like a `Q()` object.

davidbgk: Actually, if you plan to make a PR against this one, why not merge that one directly and then improve?

noirbizarre: Actually, the PR is already done and it adds the tests:
davidbgk/udata#2

davidbgk:
Still TODO: understand why 
`/fr/datasets/?organization=534fff8ba3a7292c64a77ed4|552cc6bec751df224637dddb&tag=usage-des-sols|specifications-partagees`
and 
`/fr/datasets/?organization=534fff8ba3a7292c64a77ed4|552cc6bec751df224637dddb&tag=usage-des-sols&tag=specifications-partagees`
return the same number of elements, the second should be an `AND` and as such have fewer results.

davidbgk:
@noirbizarre I tried to address that issue in the last commit, can you take a look at it? I'm not sure of myself (especially on perfs!) but the result is more accurate for the given example.

udata/search/fields.py
```diff
@@ -96,40 +123,34 @@ def get_values(self, data, filter_values):
...
+            id = self.model_field.to_mongo(value)
+            value = self.model.objects.get(id=id)
+        return str(value)
 
     def validate_parameter(self, value):
         if isinstance(value, ObjectId):
             return value
         try:
-            return ObjectId(value)
+            return [
```
vinyll: Is the `return`ed value ever used?

udata/tests/test_search.py
```diff
@@ -31,8 +33,10 @@ class Fake(db.Document):
     tags = db.ListField(db.StringField())
     other = db.ListField(db.StringField())
 
+    meta = {'allow_inheritance': True}
+
     def __unicode__(self):
-        return 'fake'
+        return self.title
```
vinyll: Shouldn't we ensure it is unicoded?

vinyll:
My review is more about grammar and global logic than ES patterns (which I'm not familiar with).

udata/core/reuse/search.py
```diff
@@ -29,9 +29,8 @@ def max_followers():
     return max(current_site.metrics.get('max_reuse_followers'), 10)
 
 
-class ReuseTypeFacet(TermsFacet):
-    def default_labelizer(self, value):
-        return REUSE_TYPES[value]
+def reuse_type_labelizer(value):
+    return REUSE_TYPES[value]
```
davidbgk: Yes, it should not happen.

udata/core/dataset/search.py
```diff
@@ -76,7 +76,7 @@ class Meta:
         properties={
```
davidbgk: I updated a few indents to be consistent.

udata/search/__init__.py
```diff
@@ -102,7 +102,9 @@ class Index(ESIndex):
     def _get_mappings(self):
         mappings, _ = super(Index, self)._get_mappings()
         return mappings, {
-            'analyzer': {a._name: a.get_definition() for a in analysis.analyzers},
+            'analyzer': {
+                a._name: a.get_definition() for a in analysis.analyzers
```
davidbgk: Because otherwise it takes more than 79 chars 😉

PR #610, opened 2016-11-14
Authorize the user to delete his account (fixes #543)

### LINKED ISSUE #543: As a user, I want to delete my account
- [x] The function must be accessed under the _Edit_ button of the _Administration_ interface
- [x] The user must be asked a confirmation in the web interface
- [x] The _deleted_ field of the user must be set to _now_
- [x] The other data fields of the user must be anonymized
- [x] The references to the user in the followed and following users must be deleted
- [x] The content of the discussions of the user must be anonymized
- [x] The user must be removed from all its organizations
- [x] The datasets must be kept unchanged
- [x] The reuses must be kept unchanged
- [x] The issues must be kept unchanged
- [x] The activities must be kept unchanged
- [x] A mail must be sent to the user
- [x] The user must be redirected to the home page
- [x] The user must be logged out
- [x] As admin, I want to access to the same functionalities, with a different API

jphnoel:


udata/core/user/models.py
```diff
@@ -210,6 +218,33 @@ def json_ld(self):
...
+        self.save()
+        for organization in self.organizations:
+            organization.members = [member
+                                    for member in organization.members
+                                    if member.user != self]
+            organization.save()
+        for discussion in self.involving_discussions:
+            for message in discussion.discussion:
+                if message.posted_by == self:
+                    message.content = 'Deleted'
```
davidbgk: Are we sure about this? It can lead to discussions very hard to understand.

udata/templates/mail/account_deleted.txt
```diff
@@ -0,0 +1,6 @@
+{% extends theme('mail/base.txt') %}
+
+{% block body %}
+{{ _('Your account has now been deleted') }}.
```
davidbgk: It would be great to have a recap of what happened for orgs/datasets without any remaining member. Do we allow that by the way?

udata/core/user/models.py
noirbizarre: Yes, it was discussed before starting this story. But I agree, just putting 'Deleted' is a little bit hardcore (even more for french based portals). Maybe we can add a deleted flag and display deleted in the front.

udata/templates/mail/account_deleted.txt
noirbizarre: We don't recap in this story because datasets and organizations (if there is some), are not deleted. Right now it's already allowed.
We only delete the account. Another story might prompt the admin for what to do with the related content (more like a moderation)

js/components/user/delete-me-modal.vue
```diff
@@ -0,0 +1,60 @@
...
+
+<script>
+import API from 'api';
+
+export default {
+    components: {
+        modal: require('components/modal.vue')
+    },
+    methods: {
+        confirm: function() {
```
noirbizarre: `confirm(): {`

js/components/user/delete-user-modal.vue
```diff
@@ -0,0 +1,74 @@
...
+    components: {
+        modal: require('components/modal.vue')
+    },
+    data() {
+        return {
+            user: {}
+        };
+    },
+    methods: {
+        confirm: function() {
```
noirbizarre: `confirm(): {`

js/components/user/delete-user-modal.vue
```diff
@@ -0,0 +1,74 @@
...
+            API.users.delete_user(
+                {user: this.user.id},
+                response => {
+                    this.$refs.modal.close();
+                    this.$dispatch('notify', {
+                        icon: 'exclamation-triangle',
+                        title: this._('The user {first_name} {last_name} has been successfully deleted',
```
noirbizarre: Might be shorter and easier to maintain with `fullname` instead of `first_name` and `last_name`.

js/components/user/delete-user-modal.vue
```diff
@@ -0,0 +1,74 @@
+<template>
+<modal :title="_('Confirm deletion')"
+    class="modal-danger"
+    v-ref:modal>
+
+    <div class="modal-body">
+        <p class="lead text-center">
+            {{ _('You are about to delete the profile of the user {first_name} {last_name}.',
```
noirbizarre: Might be shorter and easier to maintain with `fullname` instead of `first_name` and `last_name`.

udata/core/dataset/factories.py
```diff
@@ -58,3 +59,8 @@ class Meta:
 
     id = factory.Sequence(lambda n: '{0}-{1}'.format(faker.word(), n))
     title = factory.LazyAttribute(lambda o: faker.sentence())
+
+
+class UserCreatedDatasetFactory(MongoEngineFactory):
```
noirbizarre: Given the fact that it needs real objects that you give as parameters and that the factory does not provide any attribute, it might be better to not use a factory.

udata/core/followers/factories.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from factory.mongoengine import MongoEngineFactory
+
+from udata import models
+
+
+class FollowFactory(MongoEngineFactory):
```
noirbizarre: Given the fact that it needs real objects that you give as parameters and that the factory does not provide any attribute, it might be better to not use a factory.

udata/core/user/models.py
```diff
@@ -106,6 +109,11 @@ def organizations(self):
         from udata.core.organization.models import Organization
         return Organization.objects(members__user=self)
 
+    @cached_property
+    def involving_discussions(self):
```
noirbizarre: I'm not sure about this: it is used only once, the query can be inlined into `mark_as_deleted`

udata/tests/api/test_me_api.py
```diff
@@ -307,3 +315,100 @@ def test_clear_apikey(self):
...
+
+        # The following are deleted
+        with self.assertRaises(Follow.DoesNotExist):
+            following.reload()
+        # The followers are deleted
+        with self.assertRaises(Follow.DoesNotExist):
+            followed.reload()
+
+        # The personal data of the user are anonymized
+        self.assertEqual(user.email, 'Deleted-{}'.format(user.id))
```
noirbizarre: It might be safer to have something that validates as a mail (ex: `{user.id}@deleted`).

udata/tests/api/test_me_api.py
```diff
@@ -307,3 +315,100 @@ def test_clear_apikey(self):
...
+        self.assertEqual(organization.members[0].user.id, other_user.id)
+
+        # The discussions are kept but the messages are anonymized
+        self.assertEqual(len(discussion.discussion), 2)
+        self.assertEqual(discussion.discussion[0].content, 'Deleted')
+        self.assertEqual(discussion.discussion[1].content,
+                         other_disc_msg_content)
+
+        # The issues are kept and the messages are not anonymized
+        self.assertEqual(len(discussion.discussion), 2)
```
noirbizarre: You are testing discussions twice, given the comment, it should be issues.

udata/core/followers/factories.py
jphnoel: In the tests, it is nicer to call factories. Otherwise, I have to call factories and then, for some other objects, I have to call a constructor and .save().

But it is up to you. Do you want me to suppress these factories ?

js/components/user/delete-user-modal.vue
noirbizarre: Not necessary, it's also present in the User JS model ;)

udata/core/dataset/factories.py
noirbizarre: In this case (empty factories only depending on real parameters), I prefer that we use the model directly:
- do not hide anything (less complexity when not needed)
- do not add code

I you want to avoid constructor+save(), just use `Model.objects.create()`, it will do the same thin in one call

udata/core/user/api.py
```diff
@@ -252,7 +260,7 @@ def get(self, user):
             if not user.active:
                 api.abort(410, 'User is not active')
             if user.deleted:
-                api.abort(410, 'User has been deleted')
+                api.abort(410, 'User has already been deleted')
```
noirbizarre: As it's a GET operation, I think the 'already' is not consitent.

PR #618, opened 2016-11-30
Mark active users as confirmed

noirbizarre:
This PR add a migration marking already active users but without the `confirmed_at` attribute as confirmed

### DIFF
--- a/udata/migrations/2016-11-29-confirm-active-users.js
+++ b/udata/migrations/2016-11-29-confirm-active-users.js
@@ -1,0 +1,12 @@
+/*
+ * As email confirmation is required to be active
+ * this migration confirm all active users.
+ */
+
+var result = db.user.update(
+    {active: true, confirmed_at: {$exists: false}},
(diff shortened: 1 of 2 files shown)


davidbgk:
A bit meta but I think that every migration should have an entry in the changelog to explain what the problem was and how we fixed it, thoughts?

noirbizarre:
Changelog added

udata/migrations/2016-11-29-confirm-active-users.js
```diff
@@ -0,0 +1,12 @@
+/*
+ * As email confirmation is required to be active
+ * this migration confirm all active users.
+ */
+
+var result = db.user.update(
```
jphnoel: If MongoEngine supports ES6, why not use let instead of var ?

noirbizarre: sadly it doesn't support `let` this is not a full ES6 support yet

PR #621, opened 2016-11-30
Fix security issue on py-bcrypt by switching on bcrypt as recommanded

noirbizarre:
Each time a password is encrypted with Flask-Login/Passlib/py-bcrypt, the following warning is issued:

```
UserWarning: passlib.hash.bcrypt: Your installation of the 'pybcrypt' backend is vulnerable to the bsd wraparound bug, and should be upgraded or replaced with another backend (this warning will be fatal under passlib 1.7)
  "(this warning will be fatal under passlib 1.7)" % backend)
```

See <https://passlib.readthedocs.io/en/stable/lib/passlib.hash.bcrypt.html#bsd-wraparound-bug> for the security explanation.

The fix is to simply switch from the unmaintained [py-bcrypt](https://pypi.python.org/pypi/py-bcrypt) backend to the maintained and fixed [bcrypt](https://pypi.python.org/pypi/bcrypt) backend.


davidbgk:
> This does mean that existing hashes suffering this vulnerability will no longer verify using their correct password

If I interpret correctly that, users with passwords' lengths > 256 will have to change their passwords? I'm concerned about the workflow for that kind of user.

noirbizarre:
Yes it does, but:
- when not using an SSO, Flask-Security limit passwords to 128 chars: https://github.com/mattupstate/flask-security/blob/develop/flask_security/forms.py#L69
- when using an SSO, passwords are not stored on udata side

PR #631, opened 2016-12-20
Dublin core frequencies

### LINKED ISSUE #273: Support all dublin core frequencies
The DCAT schema uses frequencies defined by dublin core : http://dublincore.org/groups/collections/frequency/

For better interoperability between data catalogs published on the Web, is it possible to support all these frequencies ?

Triennial [freq:triennial]
Biennial [freq:biennial]
Annual [freq:annual]
Semiannual [freq:semiannual]
Three times a year [freq:threeTimesAYear]
Quarterly [freq:quarterly]
Bimonthly [freq:bimonthly]
Monthly [freq:monthly]
Semimonthly [freq:semimonthly]
Biweekly [freq:biweekly]
Three times a month [freq:threeTimesAMonth]
Weekly [freq:weekly]
Semiweekly [freq:semiweekly]
Three times a week [freq:threeTimesAWeek]
Daily [freq:daily]
Continuous [freq:continuous]
Irregular [freq:irregular]

noirbizarre:
This PR:
- make use of [Dublin Core Frequencies](http://dublincore.org/groups/collections/frequency/) (fix #273)
- add some missing but requested frequencies (mostly between continuous and daily)
- ensure legacy frequencies are still handled and well mapped

To achieve the third point, it also allows form fields to perform data processing before validation (mostly because some fields perform validation in the `pre_validate` phase and using validators to transform data does not work in this case)

### DIFF
--- a/udata/core/dataset/forms.py
+++ b/udata/core/dataset/forms.py
@@ -10,6 +10,7 @@ from udata.core.spatial.forms import SpatialCoverageField
 from .models import (
     Dataset, Resource, License, Checksum, CommunityResource,
     UPDATE_FREQUENCIES, DEFAULT_FREQUENCY, RESOURCE_TYPES, CHECKSUM_TYPES,
+    LEGACY_FREQUENCIES
 )
 
 __all__ = ('DatasetForm', 'ResourceForm', 'CommunityResourceForm')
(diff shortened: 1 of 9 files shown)


udata/core/dataset/forms.py
```diff
@@ -60,6 +61,12 @@ class CommunityResourceForm(BaseResourceForm):
     organization = fields.PublishAsField(_('Publish as'))
 
 
+def map_legacy_frequencies(form, field):
```
davidbgk: I don't get why we still need this given that we migrate old data?

noirbizarre: To ensure backward compatibility: it allows to submit legacy frequencies, it's still working.
- avoid new version in API
- avoid to change tools using API

Was discussed before with @odtvince

davidbgk: In that case, we should document it within the CHANGELOG and set an associated deprecation timeline somehow.

noirbizarre: Added but the Deprecation timeline is (too me) out of the scope of this PR because:
- it needs to be discussed
- it needs some documentation
- it might needs some tooling
So I only added a skeleton section in the documentation

PR #632, opened 2016-12-21
Improve Croquemort error handling

noirbizarre:
This PR improve Croquemort error handling:
- catch more exceptions (and ensure these exception exists in all requests versions)
- ensures timeout is handled with a specific message (before any other error)
- adds tests for these cases
- ensure tests are always run


davidbgk:
Did `requests` change its error handling?

noirbizarre:
Not exactly:
- the tested exception (ConnectTimeout) wasn't imported correctly
- some exception weren't avaiable at the package level in some versions (but Timeout and RequestException always were)
- chosen exception weren't covering all the cases

And as the tests weren't up to date and never executed, we never noticed.

PR #674, opened 2017-01-09
Remove some dead code

jdesboeufs:
Let's begin!


Procfile.sample
```diff
@@ -1,16 +0,0 @@
-## Uncomment lines given your setup and workflow.
```
noirbizarre: I think this file is referenced in the documentation to quickstart.
It is used by Honcho.

We should remove it in another PR and propose another way to quickstart (without using Honcho and Procfile)

jdesboeufs:
Done! 🍸

vinyll:
_+0 −555 _ ❤️

PR #677, opened 2017-01-09
Fix featured button

noirbizarre:
This PR fix the new featured button widget:
- load it where it should be (dataset and reuse pages)
- url deduced from type (so type is configurable)
- remove the wrapping `div.btn-group` (was breaking some integration)
- added compact mode (used on reuse cards)
- added the missing tooltip

Compact mode is this case:
screenshot-www data dev 2017-01-09 15-17-41


udata/templates/reuse/display.html
```diff
@@ -24,35 +24,6 @@
     <li class="active">{{ reuse.title }}</li>
 {% endblock %}
 
-{% block toolbar %}
```
vinyll: I'm not sure why removing the whole actions toolbar though

noirbizarre: It's not used anymore (legacy from dark ages long time ago ;) ).
I'll submit another PR to remove it in other templates.

noirbizarre:
Thanks ;)

PR #678, opened 2017-01-09
Remove more JS dead code

jdesboeufs:
These files or rules are never used.
We should remove them for current codebase.


js/utils/ellipsis.js
```diff
@@ -1,17 +1,5 @@
 define(['jquery', 'bootstrap', 'jQuery.dotdotdot'], function($) {
-
     $(function() {
         $('.ellipsis-dot').dotdotdot();
-
-        $('.ellipsis-tooltip').dotdotdot({
```
noirbizarre: I can't remember if we already clean up the code base from this class but I think this is the case.
Have you checked ?

jdesboeufs:
I checked ;)
Le lun. 9 janv. 2017 à 16:38, Axel Haustant <notifications@github.com> a
écrit :

PR #520, opened 2016-07-09
Converted discussion-btn to Vue

vinyll:
TODO:
- [x] fix _threads-form.vue_ error 500 when submitting (empty slug)
- [x] click title to toggle visibility of a thread
- [x] activate buttons "reply to this discussion" and "create a new discussion"

PR #688, opened 2017-01-12
Merge dev into master

jdesboeufs:
Replaced by #694


jdesboeufs:
We need to know if #520 and #560 can be deferred to `1.1`.

PR #690, opened 2017-01-12
Drop title limits when only slug is indexed (fix #676)

### LINKED ISSUE #676: Silent fail when trying to create a reuse with a long title
When I try to create a reuse with a very long name, I'm not able to click on the next button.
Behind the scene, the POST request ends with a 500 error (without any detail).

- [ ] API should respond with a 400 error, with validation error details.
- [x] Admin should display some information to the end-user.

Env: next

noirbizarre:
Slug truncation was already here, only added a test to ensure it works.
Slug are still limited to 255 to avoid very very long URLs

### DIFF
--- a/udata/tests/test_model.py
+++ b/udata/tests/test_model.py
@@ -22,7 +22,7 @@ class UUIDAsIdTester(db.Document):
 
 class SlugTester(db.Document):
     title = db.StringField()
-    slug = db.SlugField(populate_from='title')
+    slug = db.SlugField(populate_from='title', max_length=1000)
     meta = {
         'allow_inheritance': True,
     }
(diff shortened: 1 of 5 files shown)


udata/tests/test_model.py
```diff
@@ -153,6 +153,12 @@ def test_work_accross_inheritance(self):
         inherited = InheritedSlugTester.objects.create(title='title')
         self.assertNotEqual(obj.slug, inherited.slug)
 
+    def test_crop(self):
+        '''SlugField should truncate itself on save if not set'''
+        obj = SlugTester(title='x' * SlugTester.slug.max_length * 2)
```
davidbgk: May I suggest `+ 1` instead of `* 2` and then later test the title length too?

PR #691, opened 2017-01-13
Prevent croquemort to fail on bad JSON

noirbizarre:


noirbizarre:
Because some cases where already handled with 503 status code (which is consistent with service unavailable).
I think we need to rethink that for next releases

udata/core/dataset/croquemort.py
```diff
@@ -19,6 +19,15 @@
 TIMEOUT_LOG_MSG = 'Timeout connectin to Croquemort'
 
 
+def has_checked(response):
+    if response.status_code == 404:
```
davidbgk: I just realized that maybe we should use requests codes:

```python
if response.status_code == requests.status_codes.codes.not_found:
```

Same for 503 => `codes.service_unavailable`

udata/core/dataset/croquemort.py
```diff
@@ -53,7 +65,7 @@ def check_url(url, group=None):
         log.error(ERROR_LOG_MSG, exc_info=True)
         return {}, 503
     attempts = 0
-    while response.status_code == 404 or 'status' not in response.json():
+    while not has_checked(response):
```
davidbgk: What about `while is_pending(response)` to try to avoid the negation here?

udata/tests/api/test_url_api.py
```diff
@@ -122,3 +122,29 @@ def test_connection_error(self):
         response = self.get(url_for('api.checkurl'),
                             qs={'url': url, 'group': ''})
         self.assertStatus(response, 503)
+
```
davidbgk: 👍

udata/core/dataset/croquemort.py
noirbizarre: I modify it, but I use a negation on purpose: I like to enforce that something has not happen or is not done

PR #695, opened 2017-01-16
Fix webpack --watch calls

jdesboeufs:
Remove `-d` option since it breaks incremental builds.
`-c` doesn't exist anymore.


jdesboeufs:
-c exists but is default

PR #698, opened 2017-01-17
Prevent tag submission if constraints are not fullfiled

noirbizarre:
This PR ensure that the tag completer component does not allow to create a tag not fullfilling the min and max length constraints.


js/components/form/tag-completer.vue
```diff
@@ -11,11 +12,16 @@ export default {
...
         valueField: 'text',
         plugins: ['remove_button'],
-        create: function(input) {
+        create(input) {
             return {
                 value: input,
                 text: input
             }
+        },
+        createFilter(value) {
```
vinyll:
```
createFilter(value) {
  return value && config.tags.MIN_LENGTH <= value.length <= config.tags.MAX_LENGTH;
```

udata/tags.py
```diff
@@ -3,6 +3,7 @@
 
 from slugify import slugify
 
+# Constraints. Should be synced with js/config.js
```
davidbgk: Why not now?

js/components/form/tag-completer.vue
noirbizarre: already tried but not working 😢

udata/tags.py
noirbizarre: They are, it's just a reminder for futur modifications

js/components/form/tag-completer.vue
vinyll: ?
What about 
```
if (value) {
  return config.tags.MIN_LENGTH <= value.length <= config.tags.MAX_LENGTH;
}
```

noirbizarre: This is the part that is not working 😞

udata/tags.py
noirbizarre: Updated for a more explanatory comment

js/components/form/tag-completer.vue
jdesboeufs: Comparison operators cannot be chained in JS.
Each part return a `boolean`. And you don't want to compare a `boolean` with the rest of your expression.

PR #705, opened 2017-01-17
Fix tag suggest as discussion

noirbizarre:
Restore the suggest tag behavior
Enable tag suggestion on reuses


davidbgk:
I just remembered that card: https://trello.com/c/RUqf8a3V/312-ameliorer-le-crowdsourcing-mot-cles Might be worth considering proposed UX suggestions on that topic?

noirbizarre:
Yes, in a separate issue. The current behavior is broken et this PR fix it.

(it was also discussed to add community tags)

jdesboeufs:
⚠️ VelocityJS = ~40KB (minified)

PR #716, opened 2017-01-18
Fix the communautary resource modal

noirbizarre:
The communautary resource modal was missing for 2 reasons:
- metadata weren't exposed into json-ld (added to the non standard contributedDitribution to fix that)
- it wasn't handled at all client-side

### DIFF
--- a/js/front/dataset/index.js
+++ b/js/front/dataset/index.js
@@ -9,9 +9,6 @@ import config from 'config';
 import log from 'logger';
 import Velocity from 'velocity-animate';
 
-// Plugins
-import ScrollTo from 'plugins/scroll-to';
-
 // Components
 import AddReuseModal from './add-reuse-modal.vue';
 import DetailsModal from './details-modal.vue';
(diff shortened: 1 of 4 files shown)


udata/templates/dataset/resource/list-item.html
```diff
@@ -1,7 +1,7 @@
 {% set resource_format = resource.format|trim|lower or 'data' %}
 <div id="resource-{{resource.id}}" class="list-group-item"
     data-checkurl="{{ url_for('api.checkurl') }}"
-    @click="showResource('{{resource.id}}', $event)">
+    @click="showResource('{{resource.id}}', $event, {{ resource.owner is defined|tojson }})">
```
davidbgk: I did not know about that syntax, is that standard?

jdesboeufs: Yep: http://flask.pocoo.org/docs/0.12/templating/#standard-filters

davidbgk: I know that but the prior `is defined` returns a boolean?

davidbgk: Oh wait I get it, that what we actually want :)

js/front/dataset/index.js
```diff
@@ -82,21 +79,24 @@ new Vue({
...
             dataset.keywords = dataset.keywords.split(',').map(keyword => keyword.trim());
             return dataset;
         },
 
         /**
-         * Display a resource in a modal
+         * Display a resource or a community ressource in a modal
          */
-        showResource(id, e) {
+        showResource(id, e, community) {
```
davidbgk: Can we switch to `isCommunity` for a more explicit naming?

udata/templates/dataset/resource/list-item.html
noirbizarre: ;) It's the weird but working mix between jinja and Vue.js

js/front/dataset/index.js
noirbizarre: done ;)

PR #720, opened 2017-01-18
Removed users from site search (fix #719)

### LINKED ISSUE #719: Users should not be part of the search
## Detailed Description
User should not be part of the search for 2 reasons:
- it's a privacy issue
- it adds time/weight to the response but does not provide a functionnal

## Context
When I create an account , I don't want to be part of the site search result
When I search for data, I don't care about users matching my criterias

## Possible Implementation
Just remove users from search

## Your Environment
Version 1.0.0.dev1215

noirbizarre:
See #719


noirbizarre:
Doesn't seems to.
But if we want to also remove user listing (which is a search too), it removes tests and templates too.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 You have to re-initialize the index from scratch, not just use the `reindex` command given that ElasticSearch 2+ doesn't provide a way to [delete mappings](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-delete-mapping.html) anymore. The command is `udata search init` and may take some time given the amount of data you are dealing with.
 
+* 2017-01-18: User search and listing has been removed (privacy concern)
```
davidbgk: 👍

PR #725, opened 2017-01-18
Fix $modal method on root admin.vue

### LINKED ISSUE #722: Member modal display is broken
## Detailed Description
The modal displaying the member is broken:

screenshot-www data dev 2017-01-18 20-22-27

## Context
Needed to display current members roles, update it...

## Possible Implementation
Need a fix

## Your Environment
Version 1.0.0.dev1215

noirbizarre:
Fix `$modal` method making use of `propsData` in `admin.vue` and  use only properties in component (fix #722)
There is a lot of ES6 changes noise on members and members-modal.

### DIFF
--- a/js/components/organization/members.vue
+++ b/js/components/organization/members.vue
@@ -145,23 +145,26 @@
 <script>
 import Vue from 'vue';
 import log from 'logger';
+
+import placeholders from 'helpers/placeholders';
+
 import User from 'models/user';
 import Requests from 'models/requests';
 
+import BoxContainer from 'components/containers/box.vue';
+import MemberModal from 'components/organization/member-modal.vue';
+import PaginationWidget from 'components/pagination.vue';
+import UserCompleter from 'components/form/user-completer.vue';
+
 export default {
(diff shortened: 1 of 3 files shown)


js/components/organization/members.vue
```diff
@@ -145,87 +145,84 @@
...
+            avatar_placeholder: placeholders.user,
             placeholder: this._('Type an user name'),
             requests: new Requests(),
             adding: false,
             validating: false,
         };
     },
     events: {
-        'completer:item-add': function(user_id, $item) {
+        'completer:item-add'(user_id, $item) {
```
davidbgk: Wow, are you sure of that syntax or is it a failed global find/replace?

noirbizarre: A failed find & replace :/

davidbgk: That's why reviews are useful ;) 👍

noirbizarre: I fixed it because it's weird but it seems to work with this syntax 😱

PR #728, opened 2017-01-19
Last discussions fixes

noirbizarre:
This PR applies a lot of small fixes on the new discussion component:
- Fix the missing new discussion button
- Hide the form when the discussion has been created
- Focus (and expand) the newly created discussion after submit
- Display avatars in thread
- Fix missing first message
- Triggers auth earlier, when user the new comment or new discussion form instead of the submission (avoid loosing big messages on login)
- Fix some styling between the discussion header and the body
- Fix deep linking (on click and on load)
- Allows to close the comment and new discussion forms
- Removes some console warnings

(The first commit is not related and fixes the `clean` task)

### DIFF
--- a/js/components/discussions/thread-form.vue
+++ b/js/components/discussions/thread-form.vue
@@ -24,7 +24,8 @@ export default {
       return {
           // Flags to `true` when the form is in sending process. This prevents
           // duplicate POSTing.
-          sending: false
+          sending: false,
+          comment: ''
       }
   },
   methods: {
--- a/js/components/discussions/thread-form.vue
+++ b/js/components/discussions/thread-form.vue
@@ -37,7 +37,7 @@ export default {
           this.$api
           .post(`discussions/${this.discussionId}/`, {comment: this.comment})
           .then(response => {
-              this.$dispatch('discussion-load', response);
+              this.$dispatch('discussion:updated', response);
               this.comment = '';
               this.sending = false;
           })
--- a/js/components/discussions/thread-form.vue
+++ b/js/components/discussions/thread-form.vue
@@ -11,8 +11,6 @@
 </template>
 
 <script>
-import i18n from 'i18n';
-import Auth from 'auth';
 import log from 'logger';
 import Notify from 'notify';
 
@@ -30,9 +28,6 @@ export default {
   },
   methods: {
       submit() {
-          if (!Auth.need_user(i18n._('You need to be logged in to comment.'))) {
-              return;
-          }
           this.sending = true;
           this.$api
           .post(`discussions/${this.discussionId}/`, {comment: this.comment})
@@ -42,7 +37,7 @@ export default {
               this.sending = false;
           })
           .catch(err => {
-              const msg = i18n._('An error occured while submitting your comment')
+              const msg = this._('An error occured while submitting your comment')
               Notify.error(msg)
               log.error(err)
               this.sending = false;
--- a/js/components/discussions/thread-form.vue
+++ b/js/components/discussions/thread-form.vue
@@ -2,7 +2,7 @@
 <form role="form" class="clearfix animated" @submit.prevent="submit">
     <div class="form-group">
         <label for="comment-new-message">{{ _('Comment') }}</label>
-        <textarea id="comment-new-message" v-model="comment" class="form-control" rows="3" required></textarea>
+        <textarea v-el:textarea id="comment-new-message" v-model="comment" class="form-control" rows="3" required></textarea>
     </div>
     <button @click="submit" :disabled="this.sending || !this.comment" class="btn btn-primary btn-block pull-right submit-new-message">
         {{ _('Submit your comment') }}
@@ -27,6 +27,15 @@ export default {
       }
   },
   methods: {
+      /**
+       * Prefill the form and focus the comment area
+       */
+      prefill(comment) {
+          comment = comment || '';
+          this.comment = comment;
+          this.$els.textarea.setSelectionRange(comment.length, comment.length);
+          this.$els.textarea.focus();
+      },
       submit() {
           this.sending = true;
           this.$api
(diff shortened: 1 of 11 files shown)


js/components/discussions/threads-form.vue
```diff
@@ -36,15 +35,13 @@ export default {
          * Prefill the form and focus the comment area
          */
         prefill(title, comment) {
+            comment = comment || '';
             this.title = title || '';
-            this.comment = comment || '';
+            this.comment = comment;
```
vinyll: I would move this line right below `comment = comment || '';`

js/components/discussions/threads.vue
```diff
@@ -1,21 +1,19 @@
...
+    <a class="list-group-item add new-discussion" @click="displayForm" v-show="!formDisplayed">
         <div class="format-label pull-left">+</div>
-        <h4 class="list-group-item-heading">
-            {{ _('Start a new discussion') }}
-        </h4>
+        <h4 class="list-group-item-heading">{{ _('Start a new discussion') }}</h4>
     </a>
-    <div class="list-group-item list-group-form list-group-form-discussion animated"
-        v-show="formDisplayed" v-el:form>
+    <div v-el:form id="discussion-create" v-show="formDisplayed" v-if="currentUser"
```
vinyll: Should we then inform the anonymous user he can create a discussion if he were logged in?

js/components/discussions/threads.vue
```diff
@@ -29,46 +27,121 @@
...
+         * Get the thread compoent for a given discussion id
+         */
+        threadFor(id) {
+            return this.$refs.threads.find($thread => $thread.discussion.id == id);
+        },
+
+        /**
+         * Handle deep linking
+         */
+        jumpToHash(hash) {
```
vinyll: Can't we use a simple `document.location.href`?

tasks.py
```diff
@@ -21,7 +21,7 @@
     'jpg', 'gif'
 )
 STATIC_ASSETS_DIRS = (
-    'dataset', 'organization', 'post', 'reuse', 'site', 'topic', 'user'
+    'dataset', 'organization', 'post', 'reuse', 'site', 'topic', 'user', 'chunks',
```
vinyll: Is this related to this PR?

js/components/discussions/thread-form.vue
```diff
@@ -24,24 +22,32 @@ export default {
...
-          sending: false
+          sending: false,
+          comment: ''
       }
   },
   methods: {
+      /**
+       * Prefill the form and focus the comment area
+       */
+      prefill(comment) {
```
noirbizarre: These are not the same forms and there were already quite similar.
I can but it will be overly complex for nothing (a mixin or an intermediate component for both).

js/components/discussions/thread-form.vue
```diff
@@ -24,24 +22,32 @@ export default {
...
-          }
           this.sending = true;
           this.$api
           .post(`discussions/${this.discussionId}/`, {comment: this.comment})
           .then(response => {
-              this.$dispatch('discussion-load', response);
+              this.$dispatch('discussion:updated', response);
               this.comment = '';
               this.sending = false;
+              document.location.href = `#discussion-${this.discussionId}-${response.discussion.length -1}`;
```
noirbizarre: 👍

js/components/discussions/threads-form.vue
noirbizarre: it was written like that before but the comment is reused below twice and it needs to be a string everywhere so I factorised.

js/components/discussions/thread-form.vue
vinyll: Ok to have some duplication over complexity.

js/components/discussions/threads.vue
noirbizarre: Nop, the component is loaded after the hash trigger and so the ids are not present for resolution and scroll. And for comment or new comment links, the component needs to be expanded because you can't scroll to a non visible component.

tasks.py
noirbizarre: Nop, this is why I precised it in the description. But I can split it into another PR.

js/components/discussions/threads-form.vue
vinyll: I mean I'd rather read:
```
comment = comment || '';
this.comment = comment;
this.title = title || '';
```

js/components/discussions/threads.vue
noirbizarre: Yes, but we should think/work on how we display and make it the same on every component because this is something new.
Until now, no component works this way, we never announce it needs login, we display on the login screen the reason that triggers it.

Right now, I just changed the moment where the auth is triggered.
Using YouCKAN, the login screen displays:
- you need to be authentified to post a comment
- you need to be authentified to create a discussion

tasks.py
vinyll: Oh ok. It's a different commit anyways, so it's fine :)

js/components/discussions/threads.vue
vinyll: Ok to consider this later in a more global overview

js/components/discussions/threads.vue
vinyll: ok

js/components/discussions/thread-form.vue
noirbizarre: 👍 👍

js/components/discussions/threads-form.vue
noirbizarre: Ah OK ! I understand. I'll change it right now 👍

js/components/discussions/threads.vue
noirbizarre: 👍 
We need to track the idea somewhere (issue ?)

tasks.py
noirbizarre: ;)

PR #738, opened 2017-02-01
Properly handle deleted datasets and reuses on home

### LINKED ISSUE #731: Featured dataset which is deleted is not deleted from featured list
A featured dataset, which is deleted, still shows up in the featured dataset list as a blank dataset.

## Detailed Description
Obviously, if a featured dataset is deleted, it also needs to be deleted from all places where it could be referenced inside udata, for example, the featured datasets list.

noirbizarre:
This PR ensures:
- deleted datasets or reuses are not displayed on the home page
- datasets and reuses dispatch the proper `on_delete` signal when flagged as deleted
- deleted datasets or reuses are pulled from home featured

Should fix #731

### DIFF
--- a/udata/core/reuse/models.py
+++ b/udata/core/reuse/models.py
@@ -105,6 +105,8 @@ class Reuse(db.Datetimed, WithMetrics, BadgeMixin, db.Document):
             cls.on_create.send(document)
         else:
             cls.on_update.send(document)
+        if document.deleted:
+            cls.on_delete.send(document)
 
     def url_for(self, *args, **kwargs):
         return url_for('reuses.show', reuse=self, *args, **kwargs)
--- a/udata/core/site/models.py
+++ b/udata/core/site/models.py
@@ -1,7 +1,12 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+from flask import g, current_app
+from werkzeug.local import LocalProxy
+
 from udata.models import db, WithMetrics
+from udata.core.dataset.models import Dataset
+from udata.core.reuse.models import Reuse
 
 
 __all__ = ('Site', 'SiteSettings')
@@ -11,8 +16,8 @@ DEFAULT_FEED_SIZE = 20
 
 
 class SiteSettings(db.EmbeddedDocument):
-    home_datasets = db.ListField(db.ReferenceField('Dataset'))
-    home_reuses = db.ListField(db.ReferenceField('Reuse'))
+    home_datasets = db.ListField(db.ReferenceField(Dataset))
+    home_reuses = db.ListField(db.ReferenceField(Reuse))
 
 
 class Site(WithMetrics, db.Document):
@@ -23,3 +28,30 @@ class Site(WithMetrics, db.Document):
     configs = db.DictField()
     themes = db.DictField()
     settings = db.EmbeddedDocumentField(SiteSettings, default=SiteSettings)
+
+
+def get_current_site():
+    if getattr(g, 'site', None) is None:
+        site_id = current_app.config['SITE_ID']
+        g.site, _ = Site.objects.get_or_create(id=site_id, defaults={
+            'title': current_app.config.get('SITE_TITLE'),
+            'keywords': current_app.config.get('SITE_KEYWORDS', []),
+        })
+    return g.site
+
+
+current_site = LocalProxy(get_current_site)
+
+
+@Dataset.on_delete.connect
+def remove_from_home_datasets(dataset):
+    if dataset in current_site.settings.home_datasets:
+        current_site.settings.home_datasets.remove(dataset)
+        current_site.save()
+
+
+@Reuse.on_delete.connect
+def remove_from_home_reuses(reuse):
+    if reuse in current_site.settings.home_reuses:
+        current_site.settings.home_reuses.remove(reuse)
+        current_site.save()
--- a/udata/tests/site/test_site_model.py
+++ b/udata/tests/site/test_site_model.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from datetime import datetime
+
+from udata.core.dataset.factories import DatasetFactory
+from udata.core.site.models import current_site
+from udata.core.reuse.factories import ReuseFactory
+
+from udata.tests import TestCase, DBTestMixin
+
+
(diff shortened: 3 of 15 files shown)


udata/tests/site/test_site_model.py
```diff
@@ -0,0 +1,41 @@
...
+        '''Should pull home datasets on deletion'''
+        current_site.settings.home_datasets = DatasetFactory.create_batch(3)
+        current_site.save()
+
+        dataset = current_site.settings.home_datasets[1]
+        dataset.deleted = datetime.now()
+        dataset.save()
+
+        current_site.reload()
+        print(current_site.settings.home_datasets)
```
davidbgk: To remove.

udata/templates/home.html
```diff
@@ -92,7 +92,7 @@ <h1 class="page-header">
                                 <h3 class="page-header">{{ _('Used datasets') }}</h3>
                             </div>
 
-                            {% for dataset in reuse.datasets[:4] %}
+                            {% for dataset in (reuse.datasets|rejectattr('deleted')|list)[:4] %}
```
davidbgk: Not fond of that much logic in filters.

udata/templates/home.html
```diff
@@ -57,7 +57,7 @@ <h1 class="page-header">
 
                 <div class="row">
                     <ul class="card-list">
-                        {% for dataset in current_site.settings.home_datasets %}
+                        {% for dataset in current_site.settings.home_datasets|rejectattr('deleted') %}
```
davidbgk: I don't get why we have to do that given that we have `current_site.settings.home_datasets.remove(dataset)` above. Is that for the time between the delete action and the propagation of the signal? How much time do we talk about?

udata/core/reuse/models.py
```diff
@@ -105,6 +105,8 @@ def post_save(cls, sender, document, **kwargs):
             cls.on_create.send(document)
         else:
             cls.on_update.send(document)
+        if document.deleted:
```
davidbgk: Do we want to send the `on_update` just above? To me we should check at the start if it has been deleted and not perform other updates in that case.

udata/core/site/models.py
```diff
@@ -23,3 +28,30 @@ class Site(WithMetrics, db.Document):
     configs = db.DictField()
     themes = db.DictField()
     settings = db.EmbeddedDocumentField(SiteSettings, default=SiteSettings)
+
+
+def get_current_site():
```
davidbgk: Why moving it here? It's hard to see if something has changed during the move.

udata/core/site/models.py
```diff
@@ -23,3 +28,30 @@ class Site(WithMetrics, db.Document):
...
+            'keywords': current_app.config.get('SITE_KEYWORDS', []),
+        })
+    return g.site
+
+
+current_site = LocalProxy(get_current_site)
+
+
+@Dataset.on_delete.connect
+def pull_home_dataset(dataset):
```
davidbgk: Naming proposition: `remove_from_home_datasets_cache`.

udata/core/reuse/models.py
noirbizarre: Yes, the update is sent on any modification. Indexing/Deindexing listen on `on_update`

udata/templates/home.html
noirbizarre: It's just one [builtin filter](http://jinja.pocoo.org/docs/2.9/templates/#rejectattr), I don't get what's bothering you, it's meant for that, no ?

udata/templates/home.html
noirbizarre: Because of the stock, new deletion willl trigger the pull, but this avoid failure on existing ones.

udata/core/site/models.py
noirbizarre: Because of:
- coherence: `current_site` only refers to the Site model
- circular dependencies: avoid models to import from views

Nothing changed beside the location.
The related import statements are all changed into this PR and test are passing.
Given the dependency we have on it, they would have failed if there was an issue.

udata/core/site/models.py
noirbizarre: But it's not a removal from cache, this is simply the reference that is pulled from site model.
[Pull is a term used by Mongo for removing an item from a list](https://docs.mongodb.com/manual/reference/operator/update/pull/) but I can name it `remove_from_home_datasets` if you prefer.

udata/core/site/models.py
```diff
@@ -11,8 +16,8 @@
 
 
 class SiteSettings(db.EmbeddedDocument):
-    home_datasets = db.ListField(db.ReferenceField('Dataset'))
-    home_reuses = db.ListField(db.ReferenceField('Reuse'))
+    home_datasets = db.ListField(db.ReferenceField(Dataset))
+    home_reuses = db.ListField(db.ReferenceField(Reuse))
```
vinyll: Could we not filter out deleted datasets here instead of filtering in the template?

noirbizarre: This is not a dynamic part, this is simply a model declaration. There is no filtering possibilities at this point.

vinyll: I would have imagined a method for this class that would filter out deleted datasets.

udata/templates/home.html
davidbgk: Mmmh the stock should be quite low and not contain that much removed datasets or am I wrong? Putting a piece of code to handle that temporary issue doesn't look worth it but I'm not sure I get the whole picture.

udata/templates/home.html
davidbgk: To me we should put as few logic as possible within templates and here we're chaining a bit too much filters for my taste. Why don't we store that in a variable at the view level?

udata/core/reuse/models.py
davidbgk: Alirght.

udata/core/site/models.py
davidbgk: Yes please, I prefer a universal language on what is achieved by the method.

davidbgk:
Thinking again about the whole mechanism, I'm not sure the added complexity to handle the cache of that part is worth it. Can't we use a regular cache on the home view?

noirbizarre:
Again: this PR does not involve any cache

davidbgk:
Alright, let's get this merged!

PR #747, opened 2017-02-07
Fix image upload from admin and some ES6 (fix #743)

### LINKED ISSUE #743: Error 405 on avatar upload
## Detailed Description
On the user admin (admin/me/), when I try to upload a new avatar, I have the following 405 error:

screenshot-next data gouv fr 2017-02-07 08-55-10

The same happen on organization logo upload.

## Your Environment
Version 1.0.0.dev1295

noirbizarre:
See #743

The fix is to use `props` instead of `data`.

### DIFF
--- a/js/components/widgets/image-picker.vue
+++ b/js/components/widgets/image-picker.vue
@@ -73,11 +73,13 @@ import $ from 'jquery';
 import API from 'api';
 import log from 'logger';
 import UploaderMixin from 'mixins/uploader';
+import Thumbnailer from 'components/widgets/thumbnailer.vue';
 
 export default {
     autoUpload: false,
     mixins: [UploaderMixin],
-    data: function() {
+    components: {Thumbnailer},
+    data() {
         return {
             src: null,
             resizing: false,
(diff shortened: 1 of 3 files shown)


js/components/widgets/image-picker.vue
```diff
@@ -129,9 +128,9 @@ export default {
         }
     },
     methods: {
-        save: function() {
+        save() {
             if (this.HAS_FILE_API) {
-                let data = {};
+                const data = {};
```
vinyll: I'm only a little puzzle by this as it is declared as a constant though will be treated as a viariable object (`data.bbox =`). I know JS does not throw an error yet but I'd rather keep `let` in this case.

noirbizarre: `const` doesn't mean this a constant but that the variable reference is constant ie. the identifier cannot be reassign.
So in case of `Array` or `Object`, if you affect inner members, you still should use `const` if you don't reaffect the variable itself. It would have been wrong if there were `data = {}` somewhere.

I'm just applying the linter recommandations ;)

vinyll: That's right: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const 

> in case the content is an object, this means the object itself can still be altered

👍

noirbizarre: ;)

PR #750, opened 2017-02-09
Fix modals properties

noirbizarre:
This PR fix modals data that should be properties.

This bug is (positive) side-effect of #725 

Same fix than #747 and 744 (ie. use `props` instead of `data` when required).

Touched files have been linted/ES6ified

### DIFF
--- a/js/components/badges/modal.vue
+++ b/js/components/badges/modal.vue
@@ -41,34 +41,34 @@
 <script>
 import API from 'api';
 import {badges} from 'models/badges';
+import Modal from 'components/modal.vue';
 
 export default {
-    name: 'BadgesModal',
-    components: {
-        modal: require('components/modal.vue')
+    components: {Modal},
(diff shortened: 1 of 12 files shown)


noirbizarre:
Noticed too.
I tried to make changes minimal (even with ES6).

I propose we make a coherence check for 1.1 milestone

noirbizarre:
See #753

js/components/badges/modal.vue
```diff
@@ -80,8 +80,8 @@ export default {
         }
     },
     methods: {
-        confirm: function() {
-            let to_add = this.selected.filter((badge) => {
+        confirm() {
+            const to_add = this.selected.filter((badge) => {
```
davidbgk: I wonder if `const to_add = !this.selected.filter(this.initial.indexOf)` would work :)

PR #765, opened 2017-02-16
Fix common packaging

noirbizarre:
This PR start to fix common.js packaging and side-effects and ensure:
- optimization plugins are loaded once with a non breaking configuration (`mangle.keep_fnames = true`)
- locales are not bundled into the common chunk
- vue-strap components are cherry picked instead of all being included into the common chunk

To reduce the size of the common chunk, we will have to:
- fix mangling (probably fixed with Vue.js 2.0 :/)
- ends the jQuery/Bootstrap(.js) drop

### DIFF
--- a/js/i18n.js
+++ b/js/i18n.js
@@ -12,7 +12,10 @@ export const lang = config.lang;
 const resources = {};
 
 resources[lang] = {};
-resources[lang][NAMESPACE] = require('locales/' + NAMESPACE + '.' + lang + '.json');
+// Force split and async loading of locale (and so remove them from common.js)
+require.ensure([], function() {
+  resources[lang][NAMESPACE] = require('locales/' + NAMESPACE + '.' + lang + '.json');
+});
 
 moment.locale(lang);
 i18next.init({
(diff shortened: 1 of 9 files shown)


js/i18n.js
```diff
@@ -12,7 +12,10 @@ export const lang = config.lang;
 const resources = {};
 
 resources[lang] = {};
-resources[lang][NAMESPACE] = require('locales/' + NAMESPACE + '.' + lang + '.json');
+// Force split and async loading of locale (and so remove them from common.js)
```
davidbgk: 👍

webpack.config.prod.js
```diff
@@ -1,26 +1,33 @@
...
-    mangle: {
-        screw_ie8: true
-    },
-    compress: {
-        warnings: false,
-        screw_ie8: true
-    }
-}));
+config.plugins.push(
+    new webpack.DefinePlugin({
```
davidbgk: Why is that plugin required now?

noirbizarre: Setting this environment variable drop some Vue.js code (warnings and some debug helpers)  in production.

jdesboeufs:
I would not have approved this PR.
I will explain you on Monday! :)

PR #776, opened 2017-02-17
Fallback if ‘status’ absent for croquemort

davidbgk:
Sadly, I wasn't able to reproduce locally.


noirbizarre:
Isn't it related to the ftp part ?

davidbgk:
We'll try to reproduce it on `next`.

PR #777, opened 2017-02-18
Added some meta[name=robots] markup to handler crawlers behavior (fix #771)

### LINKED ISSUE #771: TransportError: TransportError(500, u'search_phase_execution_exception', u'Result window is too large, from + siz...
https://sentry.data.gouv.fr/share/issue/332e3935333635/

```
TransportError: TransportError(500, u'search_phase_execution_exception', u'Result window is too large, from + size must be less than or equal to: [10000] but was [21560]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level parameter.')
(16 additional frame(s) were not displayed)
...
  File "udata/frontend/views.py", line 93, in get
    return self.render()
  File "udata/frontend/views.py", line 23, in render
    context = context or self.get_context()
  File "udata/frontend/views.py", line 89, in get_context
    context[self.context_name] = self.get_queryset()
  File "udata/frontend/views.py", line 84, in get_queryset
    result = search.query(adapter, **params)
  File "udata/search/__init__.py", line 198, in query
    return search.execute()

TransportError: TransportError(500, u'search_phase_execution_exception', u'Result window is too large, from + size must be less than or equal to: [10000] but was [21560]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level parameter.')
```

noirbizarre:
This PR adds support for an optionnal `meta[name=robots]` in templates
and define some nofollow/noindex on some pages to handle crawlers behavior:
- prevent pagination crawling
- prevent search results from being indexed but allows the first page of result to give some weight 
- prevent all user related pages to be indexed
- prevent admin from being indexed

As a side effect, some pages that where missing metadata (title, description...) gains at least a title and a description (for proper preview when sharing on social network)

### DIFF
--- a/udata/templates/macros/metadata.html
+++ b/udata/templates/macros/metadata.html
@@ -8,8 +8,10 @@
 <title>{{ title }} - {{ current_site.title }}</title>
 <meta property="og:title" content="{{ title }} - {{ current_site.title }}" />
 <link rel="author" href="{{ config.SITE_AUTHOR_URL }}" />
-{% if config.DISALLOW_INDEXING %}<meta name="robots" content="noindex,nofollow" />{% endif %}
 <meta name="description" content="{{ description }}" />
+{% if config.DISALLOW_INDEXING %}<meta name="robots" content="noindex,nofollow" />
+{% elif meta.robots %}<meta name="robots" content="{{ meta.robots }}" />
+{% endif %}
 <meta property="og:description" content="{{ description }}" />
 <meta property="og:image" content="{{ image }}" />
 <meta name="keywords" content="{{ ', '.join(current_site.keywords + keywords) }}" />
--- a/udata/templates/topic/datasets.html
+++ b/udata/templates/topic/datasets.html
@@ -3,6 +3,13 @@
 {% from theme('macros/paginator.html') import paginator with context %}
 {% import theme('macros/search.html') as s %}
 
+{% set meta = {
+    'title': _('%(topic)s datasets', topic=topic.name),
+    'description': _("%(site)s %(topic)s related datasets", site=config['SITE_TITLE'], topic=topic.name),
+    'keywords': [_('search'), _('datasets'), _('topic')] + topic.tags,
+    'robots': 'noindex',
+} %}
+
 {% set toolbar_class='search-toolbar' %}
 {% block breadcrumb %}
     <li>
--- a/udata/templates/user/base.html
+++ b/udata/templates/user/base.html
@@ -3,11 +3,12 @@
 
 {% set bundle = 'user' %}
 
-{% block extra_head %}
-{{ super() }}
-<meta name="robots" content="noindex,follow">
-{% endblock %}
-
+{% set meta = {
+    'title': user.fullname,
+    'description': _("%(site)s %(username)s profile", site=config['SITE_TITLE'], username=user.fullname),
+    'keywords': [_('user'), _('profile')],
+    'robots': 'noindex',
+} %}
 
 {% block breadcrumb %}
     <li>{{ _('Users') }}</li>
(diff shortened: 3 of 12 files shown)


udata/templates/macros/metadata.html
```diff
@@ -12,6 +12,7 @@
 <meta name="description" content="{{ description }}" />
 <meta property="og:description" content="{{ description }}" />
 <meta property="og:image" content="{{ image }}" />
+{% if meta.robots %}<meta name="robots" content="{{ meta.robots }}">{% endif %}
```
davidbgk: Cannot we merge that with `DISALLOW_INDEXING` a few lines above?

udata/templates/topic/datasets.html
```diff
@@ -3,6 +3,13 @@
 {% from theme('macros/paginator.html') import paginator with context %}
 {% import theme('macros/search.html') as s %}
 
+{% set meta = {
+    'title': _('%(topic)s datasets', topic=topic.name),
+    'description': _("%(site)s %(topic)s related datasets", site=config['SITE_TITLE'], topic=topic.name),
+    'keywords': [_('search'), _('datasets'), _('topic')] + topic.tags,
+    'robots': 'noindex',
```
davidbgk: I'm not sure about deindexing that one.

udata/templates/user/base.html
```diff
@@ -3,6 +3,13 @@
...
+{% set meta = {
+    'title': {{ user.fullname }},
+    'description': _("%(site)s %(username)s profile", site=config['SITE_TITLE'], username=user.fullname),
+    'keywords': [_('user'), _('profile')],
+    'robots': 'noindex',
+} %}
+
 {% block extra_head %}
 {{ super() }}
 <meta name="robots" content="noindex,follow">
```
davidbgk: Remove that one?

udata/templates/macros/metadata.html
noirbizarre: Merged: `DISALLOW_INDEXING` now overrides by page setting.

udata/templates/topic/datasets.html
noirbizarre: The topic itself is indexed, this is only the topic datasets listing which is not indexed

davidbgk:
That's the moment I wonder if we should instead whitelist pages that we want to index.

Pros:

* more explicit
* avoid pages we don't think about

Cons:

* might be dangerous if we “forget” to index some critical pages…

I think we already discussed that but at least if there are counter-arguments we can refer to that discussion later. Thoughts?

noirbizarre:
Given the fact this an open data portal, I prefer the open by default.
In this PR I blacklisted the 2 cases I identified as not wanted:
- privacy concern (user related pages)
- nonsense: indexed pages that don't needs (paginated results, admin)

The cons I see with a whitelist:
- needs to think of every pages we want to index and big chances that we never ever see if we miss a page to whitelist
- it's the opposite of `robots.txt` behavior => will be more complex to handle
- unless we use `DISALLOW_INDEXING`, it's very defensive and it's the opposite of the 'being totally transparent' philosophy we had until now
- it doesn't cover the inner links case (ex: pagination) on which we can't have a whitelits behavior => 2 different strategies to handle a same topic

udata/templates/macros/metadata.html
```diff
@@ -8,8 +8,10 @@
 <title>{{ title }} - {{ current_site.title }}</title>
 <meta property="og:title" content="{{ title }} - {{ current_site.title }}" />
 <link rel="author" href="{{ config.SITE_AUTHOR_URL }}" />
-{% if config.DISALLOW_INDEXING %}<meta name="robots" content="noindex,nofollow" />{% endif %}
 <meta name="description" content="{{ description }}" />
+{% if config.DISALLOW_INDEXING %}<meta name="robots" content="noindex,nofollow" />
+{% elif meta.robots %}<meta name="robots" content="{{ meta.robots }}">
```
davidbgk: Missing end `/` to be consistent?

davidbgk:
Alright, let's keep it indexed by default.

PR #782, opened 2017-02-20
Raise maximum tag size to 96

noirbizarre:
Raise maximum tag length to at least support [official INSPIRE tags](http://inspire.ec.europa.eu/theme)

### DIFF
--- a/udata/utils.py
+++ b/udata/utils.py
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals
 
 import hashlib
+import math
 import re
 
 from uuid import uuid4
@@ -29,6 +30,7 @@ def multi_to_dict(multi):
 
 FIRST_CAP_RE = re.compile('(.)([A-Z][a-z]+)')
 ALL_CAP_RE = re.compile('([a-z0-9])([A-Z])')
+UUID_LENGTH = 36
 
 
 def camel_to_lodash(name):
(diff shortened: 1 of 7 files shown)


udata/utils.py
```diff
@@ -199,9 +200,10 @@ def recursive_get(obj, key):
     return recursive_get(value, parts) if parts else value
 
 
-def unique_string(length=None):
+def unique_string(length=36):
```
davidbgk: Let's name that magic number `UUID_LENGTH`?

udata/utils.py
```diff
@@ -199,9 +200,10 @@ def recursive_get(obj, key):
...
     '''Generate unique string'''
-    string = str(uuid4())
+    # We need a string longer than length and an UUID is 36 chars length
+    string = str(uuid4()) * int(math.ceil(length / 36.0))
     return string[:length] if length else string
```
davidbgk: `string` is the name of a standard module.

udata/utils.py
```diff
@@ -199,9 +200,10 @@ def recursive_get(obj, key):
     return recursive_get(value, parts) if parts else value
 
 
-def unique_string(length=None):
+def unique_string(length=36):
     '''Generate unique string'''
-    string = str(uuid4())
+    # We need a string longer than length and an UUID is 36 chars length
+    string = str(uuid4()) * int(math.ceil(length / 36.0))
```
davidbgk: Given our use-case, I would just add `* 2`. Less clever but more readable :)

noirbizarre: `* 2` is not sufficient, we need at least 3.
But I spent too much time searching the cause of the test failure so lets make it work with any length.

udata/utils.py
noirbizarre: Which is not imported and as it's not a builtin, there is no collision.
I changed for `uuid` but it's also a standard module and the result is not an uuid but just a string,
so I propose to keep string as it's explicit

PR #789, opened 2017-02-21
Optimized imgs for territories placeholders, fix #788

### LINKED ISSUE #788: Placeholders for territories are not optimized in size and weight
We should deal with pertinent sizes given the context.

davidbgk:
See gouvfr counterpart: https://github.com/etalab/udata-gouvfr/pull/158

### DIFF
--- a/udata/templates/territories/search-result.html
+++ b/udata/templates/territories/search-result.html
@@ -1,11 +1,11 @@
 {% cache cache_duration, 'territory-search-result', territory.id, g.lang_code %}
-{% set logo=territory.logo_url() or theme_static('img/placeholder_territory.png') %}
+{% set logo=territory.logo_url() or theme_static('img/placeholder_territory_medium.png') %}
 <li class="search-result territory-result">
     <a href="{{ territory.url }}" title="{{ territory.name }}">
         <div class="result-logo pull-left">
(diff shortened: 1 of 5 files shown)


udata/templates/territories/search-result.html
```diff
@@ -1,11 +1,11 @@
 {% cache cache_duration, 'territory-search-result', territory.id, g.lang_code %}
-{% set logo=territory.logo_url() or theme_static('img/placeholder_territory.png') %}
+{% set logo=territory.logo_url() or theme_static('img/placeholder_territory_medium.png') %}
 <li class="search-result territory-result">
     <a href="{{ territory.url }}" title="{{ territory.name }}">
         <div class="result-logo pull-left">
             <img alt="{{ territory.name }}"
                 src="{{ logo }}"
-                width="70" height="70">
+                width="64" height="70">
```
noirbizarre: Is the ratio working with all fetched logos ?

noirbizarre:
(Add a changelog entry before merge)

udata/templates/territories/search-result.html
davidbgk: All logos have different sizes and the general case is the placeholder which matches blazons so I think it's a good trade off.

PR #792, opened 2017-02-21
 Restore placeholders in search suggestions, fix #790

### LINKED ISSUE #790: Restore placeholders in search suggestions
capture d ecran 2017-02-20 a 22 31 53

davidbgk:
Note that it doesn't work for now. Some trickery is removing the image once displayed (after one thick?). Help @noirbizarre ?


noirbizarre:
You need to expose `placeholders` into the component scope. The easiest would be to add it into `data()`

noirbizarre:
(Add a changelog entry)

davidbgk:
@noirbizarre thanks, it works well for territories where I have both existing logos and the placeholder fallback. **But** for datasets and organizations I still have the flashing image this time for the existing logo! I guess it's somehow a caching issue that doesn't happen with territories because that's the last item of the list?

noirbizarre:
What do you mean by flashing issue ?

noirbizarre:
Is it mergeable for 1.0.4 and then we find a solution for the flashing issue or do you prefer to wait for a solution ?

davidbgk:
@noirbizarre I mean that it appears and disappears immediately after that, like if it was somehow overridden by another VueJS update?

davidbgk:
> Is it mergeable for 1.0.4 and then we find a solution for the flashing issue or do you prefer to wait for a solution ?

It's better but far from ideal, I think it lacks a minor fix to make it works.

noirbizarre:
Proposition: let's merge it and try to find a fix a soon as possible, agreed ?

davidbgk:
Alright, let's do it.

PR #793, opened 2017-02-21
Fix share popover

noirbizarre:
Currently the share popover is broken on the production build:

screenshot-next data gouv fr 2017-02-20 23-17-13

This PR refactor a little bit the `popover` transclusion to work in production mode.

This current implementation was relying on Vue.js debug symbols (ie. `_vue_directives` attribute on elements).

The gain is double:
- one directive deleted
- now rely on native `querySelector`

### DIFF
--- a/js/plugins/tooltips.js
+++ b/js/plugins/tooltips.js
@@ -93,6 +93,12 @@ export function install(Vue) {
                 }
             }, popover));
 
+            // Transclude child content if found
+            const content = this.el.querySelector('[data-popover-content]');
+            if (content) {
+                this.popover.content = content;
+            }
+
             switch(this.trigger) {
                 case 'hover':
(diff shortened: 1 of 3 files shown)


js/plugins/tooltips.js
```diff
@@ -93,6 +93,12 @@ export function install(Vue) {
                 }
             }, popover));
 
+            // Transclude child content if found
+            const content = this.el.querySelector('[popover-content]');
```
davidbgk: Why not using a `class` or a `data-` attribute?

noirbizarre: because I simply removed the `v-` prefix and used an attribute.
I can move it to a data-attribute if you prefer

davidbgk: As you wish, consistency is more important in that case and I don't know what is the current best practice for others vuejs files.

PR #801, opened 2017-02-23
DB tuning on GenericReferenceField

noirbizarre:
This PR apply some little tuning on disucssions and issues retrieval:
- optimize the DB query when filtering on a single ID
- ensure indexes and default order match the one provided as default from the API
- manualy create the indexes that monoengine can't express (`subject.id`)


davidbgk:
Were you able to measure the optimization?

noirbizarre:
Yes, ~200ms on a 950ms initial time

PR #802, opened 2017-02-24
Add caches for discussions and zones API, fix #796

### LINKED ISSUE #796: Add cache on zones and discussions views call
Otherwise a repeated view refetch these resources:
https://www.webpagetest.org/result/170222_C5_MSWY/1/details/cached/#waterfall_view_step1

davidbgk:
Submitted for discussion, there is a lot to challenge in there :)


udata/api/cache.py
```diff
@@ -0,0 +1,124 @@
...
+            cache_policy = check_serverside and 'no-cache' or 'must-revalidate'
+            if personal:
+                cache_policy += ', private'
+            else:
+                cache_policy += ', public'
+            cache_policy += ', max-age={}'.format(client_timeout)
+            headers = {}
+            headers['Cache-Control'] = cache_policy
+            client_etag = request.headers.get('If-None-Match')
+            cache_key = key % request.full_path  # Include querystring.
```
noirbizarre: Sadly this will generate an infinite of cache entries never reused on the admin part because to ensure there is not cache, each request is sent with`_=<timestamp>` query string.
I think we need a finer cache key generation mecanism. Maybe something like `f(endpoint, method, request.view_args, request.values)`
In falsk-restplus, it would be easier because we have the `@api.expect` decorator which is listing the varying input content so cache kay would be `f(endpoint, method, request.view_args, parsed exectations)`

udata/api/cache.py
```diff
@@ -0,0 +1,124 @@
...
+                    headers['X-Cache'] = 'HIT from Server'
+                cached_etag = resp.headers.get('ETag')
+                if client_etag and cached_etag and client_etag == cached_etag:
+                    if x_cache_header:
+                        headers['X-Cache'] = 'HIT from Client'
+                    headers['Last-Modified'] = \
+                        resp.headers.get('Last-Modified')
+                    resp = make_response('', httplib.NOT_MODIFIED)
+            else:
+                resp = make_response(serializer(func(*args, **kwargs)))
```
noirbizarre: I think this part is already handled on the flask-resplus part

udata/api/cache.py
```diff
@@ -0,0 +1,124 @@
...
+        return wrapper
+    return decorator
+
+
+@on_new_discussion.connect
+@on_new_discussion_comment.connect
+@on_discussion_closed.connect
+@on_discussion_deleted.connect
+def invalidate_discussions_cache(discussion, **kwargs):
+    cache_key = 'discussions{url}?for={id}'.format(
```
noirbizarre: this i why we need a cache key factory I think

davidbgk:
> I think we need to split applicative cache handling and http cache.

I'm not sure we can do that and proper handle the hard refresh from a browser on the applicative cache side. I think that feature is key for us to be able to understand what happens in case of bugs.

@noirbizarre I refactored the code a bit and I'll be happy to get a new review.

udata/api/__init__.py
```diff
@@ -9,9 +10,9 @@
 from flask import (
     current_app, g, request, url_for, json, make_response, redirect, Blueprint
 )
-from flask_restplus import Api, Resource, inputs, cors
+from flask_restplus import Api, Resource, cors
 
-from udata import search, theme, tracking
+from udata import search, theme, tracking  # NOQA: search must be imported.
```
vinyll: Wouldn't it be more interesting to say why `search` must be imported?

udata/api/__init__.py
```diff
@@ -128,6 +130,7 @@ def page_parser(self):
                             help='The page size to fetch')
         return parser
 
+    cache_page = cache_page
```
vinyll: What is this for?

udata/api/cache.py
```diff
@@ -0,0 +1,117 @@
...
+ONE_HOUR = ONE_MINUTE * 60
+ONE_DAY = ONE_HOUR * 24
+ONE_WEEK = ONE_DAY * 7
+ONE_MONTH = ONE_DAY * 30
+ONE_YEAR = ONE_DAY * 365
+
+
+def cache_page(class_instance=None, check_serverside=True, personal=False,
+               client_timeout=0, server_timeout=ONE_HOUR,
+               cache_pattern='cache%s', make_response=flask_make_response):
```
vinyll: I would call `cache_pattern` something like `key_pattern`.

udata/api/cache.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            headers = Headers()
+            headers.add('Cache-Control', _generate_cache_control())
+            cache_key = _generate_cache_key(cache_pattern)
+            resp = cache.get(cache_key)
+            # Respect the hard-refresh of the browser.
+            if resp is not None and not request.cache_control.no_cache:
```
vinyll: why not just `if resp and not…`?

udata/api/__init__.py
davidbgk: I don't know exactly, I just tried to remove it and it wasn't working anymore so I prefer it to be explicit :)

udata/api/__init__.py
davidbgk: To be able to decorate with `api.cache_page` likewise other api-related decorators.

udata/api/cache.py
davidbgk: Alright 👍

udata/api/cache.py
davidbgk: Valid point.

davidbgk:
@vinyll about turning it into its own project/snippet, let's see if it's useful enough for us. But it has been developed with that in mind :)

noirbizarre:
The checklist to be mergeable:
- [ ] should not drop performance on first load or when applicative cache is disabled (testing, development...)
- [ ] admin should works with it
- [ ] we need to be able to flush/invalidate
- [ ] memory usage does not raise too much (no ghost keys in cache)

udata/api/cache.py
```diff
@@ -0,0 +1,118 @@
...
+            except TypeError:
+                return str_or_callable
+
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            headers = Headers()
+            headers.add('Cache-Control', _generate_cache_control())
+            cache_key = _generate_cache_key(key_pattern)
+            resp = cache.get(cache_key)
```
noirbizarre: I think the decorator should only handle the HTTP caching part.
The `@cache.memoize` decorator can be used on demand on/in any method for the applicative cache part.

udata/core/discussions/caches.py
```diff
@@ -0,0 +1,21 @@
...
+    on_discussion_deleted
+)
+
+
+@on_new_discussion.connect
+@on_new_discussion_comment.connect
+@on_discussion_closed.connect
+@on_discussion_deleted.connect
+def invalidate_discussions_cache(discussion, **kwargs):
+    cache_key = 'discussions{url}?for={id}'.format(
```
noirbizarre: Given the previous cache key implementation, I think the key is missing the `/api/1/` part.
Why not use directly `url_for('api.discussions', qs={'for': discussion.subject.id})` ?

udata/core/spatial/api.py
```diff
@@ -71,11 +70,14 @@ def get(self):
...
              params={'ids': 'A zone identifiers list (comma separated)'})
+    @api.cache_page(check_serverside=False, key_pattern='zones%s',
+                    client_timeout=ONE_DAY, server_timeout=ONE_WEEK,
+                    make_response=api.make_response)
     @api.marshal_with(feature_collection_fields)
     def get(self, ids):
         '''Fetch a zone list as GeoJSON'''
         zones = GeoZone.objects.in_bulk(ids)
-        zones = [zones[id] for id in ids]
+        zones = [zones[id_] for id_ in ids if id_ in zones]
```
noirbizarre: Given the scope, I think the `_` is not necessary here.

udata/api/cache.py
```diff
@@ -0,0 +1,118 @@
...
+ZERO = 0
+ONE_MINUTE = 60
+ONE_HOUR = ONE_MINUTE * 60
+ONE_DAY = ONE_HOUR * 24
+ONE_WEEK = ONE_DAY * 7
+ONE_MONTH = ONE_DAY * 30
+ONE_YEAR = ONE_DAY * 365
+
+
+def cache_page(class_instance=None, check_serverside=True, personal=False,
```
davidbgk: But I use `cache` shortcut when added to the API class.

udata/api/cache.py
davidbgk: As I said above, both are linked if you want to be able to flush the applicative cache on browser's hard refresh.

udata/api/cache.py
```diff
@@ -0,0 +1,118 @@
...
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            headers = Headers()
+            headers.add('Cache-Control', _generate_cache_control())
+            cache_key = _generate_cache_key(key_pattern)
+            resp = cache.get(cache_key)
+            # Respect the hard-refresh of the browser.
+            if resp and not request.cache_control.no_cache:
+                headers.add('X-Cache', 'HIT from Server')
```
davidbgk: This is more a debug header, useful to understand what happens. `X-` headers are non-standard ones.

udata/api/cache.py
```diff
@@ -0,0 +1,118 @@
...
+                etag, is_weak = resp.get_etag()
+                if request.if_none_match.contains(etag):
+                    headers.add('X-Cache', 'HIT from Client')
+                    resp = make_response('', httplib.NOT_MODIFIED)
+            else:
+                result = func(*args, **kwargs)
+                resp = make_response(result)
```
davidbgk: We want a Flask response here to be able to compute ETag for instance + to cache the full response to be returned by the cache without further processing.

udata/api/cache.py
```diff
@@ -0,0 +1,118 @@
...
+                if (request.method in ('GET', 'HEAD') and
+                        resp.status_code == httplib.OK):
+                    headers.add('X-Cache', 'MISS')
+                    resp.add_etag()
+                    resp.last_modified = datetime.utcnow()
+                    cache.set(cache_key, resp, timeout=server_timeout)
```
davidbgk: I don't get it?

udata/core/discussions/caches.py
davidbgk: You have to take into account the `key_pattern` passed but I'll double check that.

I think we can find a better way to generate cache keys but I'm still hesitating on the right place to put it and how to deal with GET params everywhere.

udata/core/spatial/api.py
davidbgk: As you wish, I'll not fight on this. Note that it fixes a bug we have in production ;)

udata/api/__init__.py
```diff
@@ -145,7 +148,7 @@ def page_parser(self):
 
 
 @api.representation('application/json')
-def output_json(data, code, headers=None):
+def output_json(data, code=httplib.OK, headers=None):
```
davidbgk: You need to keep the same signature as [Flask's `make_response`](http://flask.pocoo.org/docs/0.12/api/#flask.make_response) which accepts a unique parameter. I just noticed that it should allow an empty call but in our case it doesn't make sense.

udata/api/cache.py
```diff
@@ -0,0 +1,118 @@
...
+        cache_control.max_age = client_timeout
+        return cache_control
+
+    def _generate_cache_key(str_or_callable):
+        if callable(str_or_callable):
+            return str_or_callable()
+        else:
+            try:
+                # Include querystring with `full_path`.
+                return str_or_callable % request.full_path
```
davidbgk: The GET parameter (`discussions/?for=<id>` in our case) are not available withing `view_args`.

davidbgk:
@noirbizarre (+ inline comments):

> should not drop performance on first load or when applicative cache is disabled (testing, development...)

How do you want to measure that? Which metric is a drop? Setting a cache key in Redis is insignificant compared to the current time allowed to generate the response.

> admin should works with it

It works, but it doesn't take into account the cache given that the `full_path` is different for each user (not sure why we need this on the first place but it may be interesting to challenge).

> we need to be able to flush/invalidate

When and how? Using signals as it's currently done is not enough? Can you develop?

> memory usage does not raise too much (no ghost keys in cache)

`too much` is quite hard to measure and thus validate before merging. The current implementation has ghosts keys given the admin unique parameter.

udata/api/cache.py
noirbizarre: This is why:
> explicit list of signficative parameters
or 
> `@api.expect` stores the `RequestParser` when there is handled parameters
If not, this is an easy DOS:
```python
i = 0
while True:
    response = requests.get('any/api/cached/url', params={'any_non_handled_parameter': i})
    i += 1
    if response.status_code != 200:
        print('Cache is filled after %n requests, no space left' % i)
```

davidbgk: Even with an explicit/expected list, once you understand which keys generate the cache (let's say `for` GET parameter) you can achieve this easily.

About DOS, the current situation is way worse to me given that it consumes a lot of resources for each call…

davidbgk: Solving this issue needs work on API throttling which is not part of that PR but we can add an issue related to this.

udata/api/cache.py
noirbizarre: I don't think this is what happens: browser hard refresh does not force server side cache flush and shouldn't (unless I misunderstood).
HTTP headers says to the browser to keep cache and it won't perform any other request on this URL unless TTL is expired, even if the server-side cache is flushed.
Client side can't see applicative cache and, most of the time, it's unrelated:
- some entries are already cached and this create a double entry (in fact `n+1` where `n` is the combination of extras unkown query strings)
- Static API entries (reference list, translations...), like reference lists, that are faster that an applicative cache access will be slower because it will trigger a cache rountrip, but these entries are typically those we want to cache on the HTTP part
- Caching the full reponse instead of the data means caching any session/user related data that are embedded in it (headers/cookies...) => exploit
- caching the full response does not benefit from/to other cached data access

This is why I think it should focus on doing proper HTTP caching and let the applicative cache part to Flask-Caching which allows more fine tuned server side caching.
If we want to cache responses, this require something more complex to be secure and performant, either something like Varnish if we want to handle it ourself or something like Cloudflare if we want an external service.

udata/api/cache.py
noirbizarre: ok 👍

udata/api/cache.py
noirbizarre: Then it should be a middleware

udata/api/cache.py
davidbgk: Let's refocus on context here… I'm not trying to send a 🚀 to the 🌔 but to cache 2 API calls which are not user-related in order to improve **incrementally** performances for the user experience.

Cache is a poor-man performance solution but we have to stick to it because the server *currently* takes [576 ms](https://www.webpagetest.org/result/170301_TJ_TZB/1/details/cached/#waterfall_view_step1) for each and every request to a dataset view.

We can wait 6 months to release a perfect fine-tuned cache solution with Varnish and all but our users will suffer during that time. I'm just trying to reduce that pain during that interval.

udata/api/cache.py
noirbizarre: ~Response is a class instance => depending on the used backend, it's not serializable like a tuple with primitive types only~
Sorry, I was working on Celery, so I thought the constraint was the same. You can cache anything. But the point on caching the whole response is still valid: it caches any user related headers/cookies

udata/core/discussions/caches.py
noirbizarre: Yes, a reusable factory will ensure cache keys are always the same.

noirbizarre:
> should not drop performance on first load or when applicative cache is disabled (testing, development...)
> How do you want to measure that? Which metric is a drop?
Until now, what I was doing, it's the average of x request with time:
```
$ time curl $API_URL`
```
Just try with and without the decorator to see. With the decorator, you need to flush cache betweend each call.
Without the decorator, you can simply write:
```
$ time `for i in {0..$x}; do curl $API_URL; done`
```
But I think both test should have the same procole.

> Setting a cache key in Redis is insignificant compared to the current time allowed to generate the response.

Not if the response does not query external services like static/hard coded entries (ex: frequencies)

>> admin should works with it
> It works, but it doesn't take into account the cache given that the full_path is different for each user (not sure why we need this on the first place but it may be interesting to challenge).
I think it works because each request is timestamped (which by the way is not wanted for reference list, swagger.json...)

How is it different for each user ? When I'm connected on a dataset page, the query for the related discussions is the same for every users I think. What I want to be sure, is that a submitted discussion or comment in a discussion is visible ASAP in the admin

we need to be able to flush/invalidate
When and how? Using signals as it's currently done is not enough? Can you develop?

>> memory usage does not raise too much (no ghost keys in cache)
> too much is quite hard to measure and thus validate before merging. The current implementation has ghosts keys given the admin unique parameter.

Yes, but we know why, this is why we need to properly handle querystrings, because admin timestamped queries will fill the cache way too fast.

davidbgk:
Without decorator:

```
→ for i in (seq 10); time http http://data.gouv.dev:7000/api/1/discussions/\?for=5862206588ee38254d3f4e5e\&i=$i > /dev/null; end
        0.41 real         0.25 user         0.04 sys
        0.41 real         0.26 user         0.04 sys
        0.39 real         0.25 user         0.04 sys
        0.43 real         0.27 user         0.04 sys
        0.50 real         0.30 user         0.05 sys
        0.58 real         0.36 user         0.05 sys
        0.51 real         0.32 user         0.05 sys
        0.60 real         0.35 user         0.05 sys
        0.54 real         0.33 user         0.05 sys
        0.43 real         0.26 user         0.04 sys
```

With a total of 4s 908ms.

With decorator:

```
→ for i in (seq 10); time http http://data.gouv.dev:7000/api/1/discussions/\?for=5862206588ee38254d3f4e5e\&i=$i > /dev/null; end
        0.46 real         0.28 user         0.04 sys
        0.43 real         0.27 user         0.04 sys
        0.41 real         0.26 user         0.04 sys
        0.50 real         0.30 user         0.04 sys
        0.57 real         0.33 user         0.05 sys
        0.49 real         0.30 user         0.05 sys
        0.48 real         0.29 user         0.04 sys
        0.40 real         0.26 user         0.04 sys
        0.42 real         0.26 user         0.04 sys
        0.44 real         0.27 user         0.04 sys
```

With a total of 4s 678ms.

I think we can check that one ☑️

udata/api/cache.py
noirbizarre: I don't think this is even close:
- actually to make the server down because of the "slow" response (we speak of a 575ms response, which is not good but not even close to be sufficient to make the server down easily), you have to reach the congestion point of the reverse-proxy + the application. Even at this point, traffic will only slow down but it will be served and without any human intervention, it will disappear simply because ressources are freed after each call.. Human intervention will be necessary to make the problem go away faster.
- the parameter problem is already present: the admin will generate a cache entry for each call, it doesn't need to be an intentionnal exploit. And it's not rate related: you don't need X requests/second to make it happen, you only need to make enough requests in `server_timeout` (1 hour by default) to fill the memory. And once this point is reached, redis (in our case) will simply lock and all is down (already happen more than once this is also why I'm really carefull about it). One of the main differences, there is no restart possible without human intervention to empty the cache. Ressources are not freed after each call but after TTL. And by the way, the fact that cache isn't used because the cache key is different also mean that we still have the initial poor performance issue on each call.

Throttling won't solve the issue, it will only limit it. Even with one request/second I think we can fill the cache quite easily in 1 hour because each request add the entire payload in memory for TTL without reuse.

udata/api/cache.py
noirbizarre: It don't have to be an user related API: admin and frontend use a session which is using a cookie to authenticate user so with every request, there is a cookie that identify him. And, I just provided multiple different ways to do it in a safer more compatible way.

I'm totally focused on the goal, this why, if the point is to make it incrementally, I dont understand why you prefer to have something which create a coupling between two things that can be totally decoupled HTTP and introduce an already identified issue. And, as you said, it's to handle 2 APIs, so maybe a generic solution with both HTTP and applicative cache is a little bit over-engineered.

I'm so focused I took the time to deeply details my review and my responses.

To me it's like saying: yes we have a performance issue so let's not resolve it but instead let's try to hide it by adding another issue on top of it.

Why not make the HTTP caching part as one step and use the provided and tested `cache.memoize` to cache the data as an another step ? (the signal part of this PR to invalidate cache is still valid with `cache.memoize`) ? If it's not doing the job, yes, we can incrementally raise the complexity to find another solution.

So, to stay focused, what I'm saying is that to solve this performance issue, we have 3 tools that needs to be handled separately:
- HTTP cache headers (which can be solved by this decorator and can benefit all APIs than need  to and can be cached)
- applicative cache, (which is already provided by Flask-Caching and can already be used with `cache.memoize` and for which the invalidation pattern is this PR should work)
- identify the real bottleneck source of this performance issue (which might be an interesting point to solve as it's the real issue)

This PR try to solve the 2 first point together and I don't think this is a good idea.
So, just so I'm clear, I'm not saying that they shouldn't be solved, I like the idea of having proper HTTP cache handling as well as having more efficient applicative cache use. I just think this is something that need to be splitted because it's 2 distinct problems with 2 distinct solutions.

davidbgk:
Let's wait for a better solution.

PR #807, opened 2017-02-27
Metrics fixes

noirbizarre:
This PR started as an intent to fix metrics not being updated after transfer for initial owner (organization or user).
By doing so I discover some other bug that fixed in the same time.

So, this PR:
- cleanup all factories
  - make use of `factory.Faker`
  - remove duplicate factories
  - use the same factory pattern everywhere (no more global `udata.model` import, use factory as much as possible, 
  - put all testing helpers as `Faker` providers
- Fix mixins order on models => fix user metrics not being initialized and some other undetected bugs
- Provide default metrics for existing users (migration)
- Fix a local dependant test on Me API
- factorize the owning behavior into a tested mixin, ensuring all owned items have the same behavior and allowing to detect owner change
- test and fix transfer acceptation => fix metrics not being updated after transfer

The factory cleanup part is a single commit that I can extract into a separate PR if required.


udata/core/user/metrics.py
```diff
@@ -44,9 +44,21 @@ class ReusesMetric(UserMetric):
     def get_value(self):
         return Reuse.objects(owner=self.user).count()
 
+
 ReusesMetric.connect(Reuse.on_create, Reuse.on_update)
 
 
+@db.Owned.on_owner_change.connect
+def update_downer_metrics(document, previous):
+    print(previous, document)
```
davidbgk: Missing `print` statement.

noirbizarre: fixed

noirbizarre:
Advice: I tried to make coherent commit so you better review commit by commit, it will be way easier !

noirbizarre:
Splitted into #808, #809, #810

PR #808, opened 2017-02-27
Factories cleanup

noirbizarre:
This PR cleanup all factories
  - make use of `factory.Faker`
  - remove duplicate factories
  - use the same factory pattern everywhere (no more global `udata.model` import, use `factory` import as much as possible)
  - put all testing helpers as `Faker` providers


udata/utils.py
```diff
@@ -202,10 +205,29 @@ def recursive_get(obj, key):
...
     return string[:length] if length else string
 
 
+class UDataProvider(BaseProvider):
+    '''
+    A Faker provider for UData missing requirements.
+
+    Might be conributed to upstream Faker project
+    '''
+    def unique_string(self, length=UUID_LENGTH):
```
davidbgk: why not `unique_string = unique_string` ?

noirbizarre: I don't think it works with `method = function`

noirbizarre: Confirmed: just tested `method = function` does not work :/

davidbgk: mmh I see, you'll have to deal with `self` with a partial and add extra complexity. Let's keep it as is.

PR #809, opened 2017-02-27
User metric fixes

noirbizarre:
This PR fix user metrics not being initialized and some other undetected bugs and provides default metrics for existing users (migration).

Also fix a local dependant test on Me API

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
   [#801](https://github.com/opendatateam/udata/pull/801)
 - Cleanup factories
   [#808](https://github.com/opendatateam/udata/pull/808)
+- Fix user default metrics not being set [migration]
+  [#809](https://github.com/opendatateam/udata/pull/809)
 
 ## 1.0.3 (2017-02-21)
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
   [#799](https://github.com/opendatateam/udata/pull/799)
 - Some minor tuning on generic references
   [#801](https://github.com/opendatateam/udata/pull/801)
+- Fix user default metrics not being set
```
davidbgk: Proposition: adding a `[migration]` or something in this vein?

udata/migrations/2017-02-25-missing-user-metrics.js
```diff
@@ -0,0 +1,26 @@
+/*
+ * Set default metric values on users
```
davidbgk: Here we set new defaults but we do not compute metrics for all users again, what about current erroneous ones (if any)?

noirbizarre: This PR doesn't handle the existing values, only default ones.
Existing values are already set and working (excluding those not being updated after transfer, see #810)

davidbgk: Alright.

PR #810, opened 2017-02-27
Update metrics on transfer/owner change

noirbizarre:
This PR fix metrics not being updated after transfer for initial owner (organization or user):
- factorize the owning behavior into a tested mixin, ensuring all owned items have the same behavior and allowing to detect owner change
- split the big `models/__init__.py` into modules (`models/queryset.py`, `models/document.py`)
- test and fix transfer acceptation => fix metrics not being updated after transfer

Tests require #809 to pass

### DIFF
--- a/udata/core/dataset/models.py
+++ b/udata/core/dataset/models.py
@@ -12,9 +12,7 @@ from mongoengine.fields import DateTimeField
 from werkzeug import cached_property
 
 from udata.frontend.markdown import mdstrip
-from udata.models import (
-    db, WithMetrics, BadgeMixin, SpatialCoverage, OwnedByQuerySet
-)
+from udata.models import db, WithMetrics, BadgeMixin, SpatialCoverage
 from udata.i18n import lazy_gettext as _
 from udata.utils import hash_url
 
@@ -107,7 +105,7 @@ class License(db.Document):
         return self.title
 
 
-class DatasetQuerySet(OwnedByQuerySet):
+class DatasetQuerySet(db.OwnedQuerySet):
     def visible(self):
         return self(private__ne=True, resources__0__exists=True, deleted=None)
 
@@ -236,7 +234,7 @@ class Resource(ResourceMixin, WithMetrics, db.EmbeddedDocument):
     on_deleted = signal('Resource.on_deleted')
 
 
-class Dataset(WithMetrics, BadgeMixin, db.Document):
+class Dataset(WithMetrics, BadgeMixin, db.Owned, db.Document):
     created_at = DateTimeField(verbose_name=_('Creation date'),
                                default=datetime.now, required=True)
     last_modified = DateTimeField(verbose_name=_('Last modification date'),
@@ -251,9 +249,6 @@ class Dataset(WithMetrics, BadgeMixin, db.Document):
     resources = db.ListField(db.EmbeddedDocumentField(Resource))
 
     private = db.BooleanField()
-    owner = db.ReferenceField('User', reverse_delete_rule=db.NULLIFY)
-    organization = db.ReferenceField('Organization',
-                                     reverse_delete_rule=db.NULLIFY)
     frequency = db.StringField(choices=UPDATE_FREQUENCIES.keys())
     frequency_date = db.DateTimeField(verbose_name=_('Future date of update'))
     temporal_coverage = db.EmbeddedDocumentField(db.DateRange)
@@ -276,14 +271,12 @@ class Dataset(WithMetrics, BadgeMixin, db.Document):
     }
 
     meta = {
-        'allow_inheritance': True,
         'indexes': [
             '-created_at',
             'slug',
-            'organization',
             'resources.id',
             'resources.urlhash',
-        ],
+        ] + db.Owned.meta['indexes'],
         'ordering': ['-created_at'],
         'queryset_class': DatasetQuerySet,
     }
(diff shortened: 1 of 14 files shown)


udata/models/owned.py
```diff
@@ -0,0 +1,83 @@
...
+    organization = ReferenceField('Organization', reverse_delete_rule=NULLIFY)
+
+    on_owner_change = signal('Owned.on_owner_change')
+
+    meta = {
+        'indexes': [
+            'owner',
+            'organization',
+        ],
+        'queryset_class': OwnedQuerySet,
```
davidbgk: I don't think we use that? Maybe inheriting from `Owned.meta['queryset_class']` is possible but quite ugly :/

udata/models/owned.py
```diff
@@ -0,0 +1,83 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import logging
+
+from blinker import signal
+
```
davidbgk: Useless new line

udata/models/owned.py
```diff
@@ -0,0 +1,83 @@
...
+            'owner',
+            'organization',
+        ],
+        'queryset_class': OwnedQuerySet,
+    }
+
+
+def owned_pre_save(sender, document, **kwargs):
+    '''
+    Owned pre_save handler
```
davidbgk: Not sure that comment is useful? (redundant with method name?)

udata/models/owned.py
```diff
@@ -0,0 +1,83 @@
...
+        else:
+            # Change from owner to another
+            # Need to fetch previous value in base
+            original = sender.objects.only('owner').get(pk=document.pk)
+            document._previous_owner = original.owner
+
+
+def owned_post_save(sender, document, **kwargs):
+    '''
+    Owned post_save handler
```
davidbgk: Same here.

udata/tests/test_transfer.py
```diff
@@ -99,10 +101,89 @@ def test_request_transfer_to_same_organization(self):
             self.assert_transfer_started(dataset, org, org, comment)
 
 
+class TransferAcceptTest(TestCase, DBTestMixin):
```
davidbgk: 👍

udata/models/__init__.py
```diff
@@ -49,6 +46,8 @@ def __init__(self, app=None):
         self.ValidationError = ValidationError
         self.ObjectId = ObjectId
         self.DBRef = DBRef
+        self.Owned = Owned
```
davidbgk: I don't get why we need these two lines?

udata/core/dataset/models.py
```diff
@@ -276,14 +271,12 @@ def __str__(self):
...
     meta = {
-        'allow_inheritance': True,
         'indexes': [
             '-created_at',
             'slug',
-            'organization',
             'resources.id',
             'resources.urlhash',
-        ],
+        ] + db.Owned.meta['indexes'],
```
davidbgk: I wonder if `db.Owned.INDEXES` is easier to understand. Here it looks to be dynamic.

udata/models/owned.py
```diff
@@ -0,0 +1,83 @@
...
+            'organization',
+        ],
+        'queryset_class': OwnedQuerySet,
+    }
+
+
+def owned_pre_save(sender, document, **kwargs):
+    '''
+    Owned pre_save handler
+    Need to fetch original owner before the new one erase it.
```
davidbgk: for `on_owner_change` to transfer metrics too

udata/models/owned.py
```diff
@@ -0,0 +1,83 @@
...
+            # Need to fetch previous value in base
+            original = sender.objects.only('owner').get(pk=document.pk)
+            document._previous_owner = original.owner
+
+
+def owned_post_save(sender, document, **kwargs):
+    '''
+    Owned post_save handler
+    Dispatch the `Owned.on_owner_change` signal
+    once the document has been saved.
```
davidbgk: including the previous owner

udata/models/owned.py
noirbizarre: This is Mongoengine behavior: child class inherit this queryset (at least used in tests), unless this is explicitely overriden.

udata/models/owned.py
noirbizarre: Same coding style used in every module: import grouped by packages if not builtin.

udata/models/owned.py
noirbizarre: I don't understand :/

udata/models/__init__.py
noirbizarre: This allows, as every other db mixins, to be used as `db.Owned` and `db.OwnedQuerySet`.
`db` act as a facade pattern everywhere.

udata/core/dataset/models.py
noirbizarre: But where does it come from ?
The `meta['indexes']` the only way (I know) to get the mixin default indexes.
It relies on mongoengine default (documented) behavior.

davidbgk: Alright, I thought it was udata-related. My bad.

udata/models/owned.py
davidbgk: Okay.

udata/core/dataset/models.py
noirbizarre: ;)

udata/models/document.py
```diff
@@ -0,0 +1,41 @@
...
+        return dict((
+            (key, serialize(value))
+            for key, value in self.to_mongo().items()
+            if key not in excluded_keys
+        ))
+
+
+class DomainModel(UDataDocument):
+    '''Placeholder for inheritance'''
+    pass
```
davidbgk: Useless `pass` ;)

noirbizarre: Just extracted from `models/__init__.py` as it was. I think the only usecase is `isinstance(document, DomainModel)`

davidbgk: In that case, why not directly `isinstance(document, UDataDocument)`?

PR #811, opened 2017-02-28
Make spatial completion ponderation f(level, population) with priority on level

noirbizarre:
This PR try to enhance spatial completion for zones with short name and few population (or no population data available) by taking in account administrative level and area.
The algorithm gives priority to the administrative level and then optionally to the population.
It could be enhanced with the area too which is also available.

### DIFF
--- a/udata/core/spatial/search.py
+++ b/udata/core/spatial/search.py
@@ -9,10 +9,20 @@ from udata.i18n import language, _
 from udata.search import ModelSearchAdapter, register
 from udata.search.analysis import standard
 
-from .models import GeoZone
+from .models import GeoZone, admin_levels, ADMIN_LEVEL_MAX
 
 
 __all__ = ('GeoZoneSearch', )
+
+
+PONDERATION_STEP = 10000
+# Compute weight relative to 10⁸.
+# Only 12 countries + UE have emore population.
+# This is not significative in the ranking algorithm for countries.
+# Given ES indexes weight as an integer,
+# zones start to gain weight at MAX_POPULATION / PONDERATION_STEP
+# Here starts at 10k inhabitants
+MAX_POPULATION = 1E8
 
 
 def labels_for_zone(zone):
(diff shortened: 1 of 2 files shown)


udata/core/spatial/search.py
```diff
@@ -43,15 +46,18 @@ class Meta:
...
-        if 0 <= population <= 2147483647:
-            return population
-        else:
-            if population < 0:  # country/eh population is -99.
-                return 0
-            else:  # World population is 6772425850.
-                return 2147483647
+    def compute_weight(cls, zone):
+        '''
+        Give a weight to the zone according its administrative level first
```
davidbgk: `according to`

udata/core/spatial/search.py
```diff
@@ -43,15 +46,18 @@ class Meta:
...
+        and then its population.
+        '''
+        # Each level give give a step
+        level = min(admin_levels.get(zone.level, ADMIN_LEVEL_MAX), 1)
```
davidbgk: `ADMIN_LEVEL_MIN` instead of `1`?

udata/core/spatial/search.py
```diff
@@ -43,15 +46,18 @@ class Meta:
...
-            if population < 0:  # country/eh population is -99.
-                return 0
-            else:  # World population is 6772425850.
-                return 2147483647
+    def compute_weight(cls, zone):
+        '''
+        Give a weight to the zone according its administrative level first
+        and then its population.
+        '''
+        # Each level give give a step
```
davidbgk: -`give`

udata/core/spatial/search.py
```diff
@@ -43,15 +46,18 @@ class Meta:
...
+        level = min(admin_levels.get(zone.level, ADMIN_LEVEL_MAX), 1)
+        level_weight = (ADMIN_LEVEL_MAX / level) * 10 * PONDERATION_STEP
```
davidbgk: Why not set `PONDERATION_STEP` to `10000` in that case?

udata/core/spatial/search.py
davidbgk: Can you put boundaries like with population to have an idea?

udata/core/spatial/search.py
```diff
@@ -43,15 +46,18 @@ class Meta:
...
+        # Population gives 0 < weight < 1000 to rank between level steps only
+        population = min(max(0, zone.population), POPULATION_MAX)
+        population_weight = (population / POPULATION_MAX) * PONDERATION_STEP
```
davidbgk: Mmh ok got it for the ponderation, in that case why not below directly to the sum?

udata/core/spatial/search.py
noirbizarre: No: here to avoir devision by zero so I prefer explicit 1 over an `ADMIN_LEVEL_MIN` which can be 0.

udata/core/spatial/search.py
noirbizarre: Because, of the use later: for the mathemtical explnation, th `* 10` is here to dilate the `(ADMIN_LEVEL_MAX / level)` which is `1 < f(level) < inf`. The factor is only here to ensure steps are not overlapping too much. If I touch the `PONDERATION_STEP`, the overlapping doesn't change, it is dilated too.

udata/core/spatial/search.py
noirbizarre: Done but I don't think it brings any value because this is relative scoring. What matter is that admin level weight more that population.

udata/core/spatial/models.py
```diff
@@ -24,6 +24,9 @@
 
 ADMIN_LEVEL_MIN = 1
 ADMIN_LEVEL_MAX = 110
+# Max known population for a zone
+# World population is 6772425850.
+POPULATION_MAX = 2147483647
```
davidbgk: I think it comes from an ES limitation given that it's [the maximum positive value for a 32-bit signed binary integer](https://en.wikipedia.org/wiki/2147483647_(number)). So it's not related to the population.

davidbgk: As a consequence, it should be the max for the level weight too.

noirbizarre: OK, so not a problem here because the max computed value is far from it. 
I changed for the world population (approximative) so the number means something and the algo works on any zone

PR #812, opened 2017-03-02
Flask Mongoengine WTForms update

noirbizarre:
This PR supersedes #586, #654, #682 and #767 because they were dependent upgrades.
These PR have been merged into this one and all required fixes/refactorings applied:
- `flask_wtf.Form` is now `flask_wtf.FlaskForm` and has a little signature change
- `flask_mongoengine` has a new connection manager
- `wtform-json` use `StringField` instead of the deprecated `TextField` which breaks every `StringField` instance (inherited or not) using default value (pending bug fix: kvesteri/wtforms-json#49), casting value to something else than a string or expected `None` as default

Consequence:
- make use of `mongoengine.get_db` instead of the removed `flask_mongoengine.connection.get_db`
- cleanup the form fields dependency tree: all fields expecting `None` as default/empty, inheriting from `StringField` but returning another type or using a default value now use `udata.forms.field.Field` as base class
- `Form(formdata, obj)` calls are now `Form(formdata, obj=obj)`

### DIFF
--- a/udata/forms/__init__.py
+++ b/udata/forms/__init__.py
@@ -8,7 +8,7 @@ wtforms_json.init()
 from flask_mongoengine.wtf import model_form  # noqa
 
 from flask_mongoengine.wtf.models import ModelForm as MEModelForm  # noqa
-from flask_wtf import Form as WTForm  # noqa
+from flask_wtf import FlaskForm  # noqa
 
 from udata import i18n  # noqa
 
@@ -22,7 +22,7 @@ class CommonFormMixin(object):
         super(CommonFormMixin, self).process(formdata, obj, data, **kwargs)
 
 
-class Form(CommonFormMixin, WTForm):
+class Form(CommonFormMixin, FlaskForm):
     pass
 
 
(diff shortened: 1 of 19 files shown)


udata/forms/__init__.py
```diff
@@ -8,7 +8,7 @@
 from flask_mongoengine.wtf import model_form  # noqa
 
 from flask_mongoengine.wtf.models import ModelForm as MEModelForm  # noqa
-from flask_wtf import Form as WTForm  # noqa
+from flask_wtf import FlaskForm  # noqa
```
davidbgk: `#noqa` still valid?

udata/tests/forms/test_uuid_field.py
```diff
@@ -52,6 +51,7 @@ def test_with_valid_uuid_string(self):
 
         form.populate_obj(fake)
 
+        print(type(fake.id))
```
davidbgk: Remaining `print` statement.

udata/forms/__init__.py
noirbizarre: I was happy too to remove it but sadly this is here because of the mandatory early `wtforms_json.init()`

PR #817, opened 2017-03-06
Harvesters admin fixes and improvements (first pass)

noirbizarre:
This PR fixes most of the harvesters admin views:
- fix edition (API was also missing)
- fix deletion
- remove owner on harvesters owned by an organization (migration)

There is also some improvements, mostly on the harvester listing:
- is now paginated, wider
- display more details (implementation, owner, last run, validation status...)

screenshot-www data dev 2017-03-06 10-34-46

screenshot-www data dev 2017-03-06 11-54-18

Fixes/changes included that can be extracted into a separate PR:
- Remove jQuery dependency on avatar cell component (and add a `closest()` polyfill): 1c69d87
- Fix `Document.to_dict()` ID handling (required for update): 38c2aa5
- Some Minor models improvements (inheritance and indexing): 30ca20c

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,13 @@
     - allows use of hot components reloading.
 - Pure Vue.js modals. Fix the default membership role. Added contribute modal.
   [#873](https://github.com/opendatateam/udata/pull/873)
+- Lot of fixes and improvements on Harvest admin UI
+  [#817](https://github.com/opendatateam/udata/pull/817):
+    - harvester edition fixed (and missing API added)
+    - harvester deletion fixed
+    - harvester listing is now paginated
+    - more detailed harvesters widgets
+    - ensure harvest source are owned by a user or an organization, not both [migration]
 
 ## 1.0.9 (2017-04-23)
 
--- a/udata/harvest/actions.py
+++ b/udata/harvest/actions.py
@@ -71,6 +71,15 @@ def create_source(name, url, backend,
     return source
 
 
+def update_source(ident, data):
+    '''Update an harvest source'''
+    print(data)
+    source = get_source(ident)
+    source.modify(**data)
+    signals.harvest_source_updated.send(source)
+    return source
+
+
 def validate_source(ident, comment=None):
     '''Validate a source for automatic harvesting'''
     source = get_source(ident)
--- a/udata/harvest/actions.py
+++ b/udata/harvest/actions.py
@@ -22,18 +22,31 @@ from .tasks import harvest
 
 log = logging.getLogger(__name__)
 
+DEFAULT_PAGE_SIZE = 10
+
 
 def list_backends():
     '''List all available backends'''
     return backends.get_all().values()
 
 
-def list_sources(owner=None):
-    '''List all harvest sources'''
+def _sources_queryset(owner=None):
     sources = HarvestSource.objects.visible()
     if owner:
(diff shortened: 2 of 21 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,13 @@
   [Flask-Security-Fork](https://pypi.python.org/pypi/Flask-Security-Fork)
   [#813](https://github.com/opendatateam/udata/pull/813)
 - Ensure URLs are stripped [#823](https://github.com/opendatateam/udata/pull/823)
+- Lot of fixes and improvements on Harvest admin UI
+  [#817](https://github.com/opendatateam/udata/pull/817):
+  - harvester edition fixed (and missing API added)
+  - harvester deletion fixed
+  - harvester listing is now paginated
+  - more detailled harvesters widgets
+  - ensure harvest source are owned by an user or an organization, not both [migration]
```
davidbgk: A user

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@
   [Flask-Security-Fork](https://pypi.python.org/pypi/Flask-Security-Fork)
   [#813](https://github.com/opendatateam/udata/pull/813)
 - Ensure URLs are stripped [#823](https://github.com/opendatateam/udata/pull/823)
+- Lot of fixes and improvements on Harvest admin UI
+  [#817](https://github.com/opendatateam/udata/pull/817):
+  - harvester edition fixed (and missing API added)
+  - harvester deletion fixed
+  - harvester listing is now paginated
+  - more detailled harvesters widgets
```
davidbgk: detailed

udata/harvest/actions.py
```diff
@@ -28,12 +28,16 @@ def list_backends():
...
+def list_sources(owner=None, page=None, page_size=None):
     '''List all harvest sources'''
     sources = HarvestSource.objects.visible()
     if owner:
-        return list(sources.owned_by(owner))
-    return list(sources)
+        sources = sources.owned_by(owner)
+    if page_size:
+        page = max(page or 1, 1)
+        return sources.paginate(page, page_size)
```
davidbgk: No need for a list here? I think it's safer to do `sources = sources.paginate(page, page_size)` here and return a list at the end whatever the case.

noirbizarre: It is also used by cli which doesn't support pagination. Only the API is using pagination.

davidbgk: I think there is a misunderstanding, my suggestion:
```python
def list_sources(owner=None, page=None, page_size=None):
    '''List all harvest sources'''
    sources = HarvestSource.objects.visible()
    if owner:
        sources = sources.owned_by(owner)
    if page_size:
        page = max(page or 1, 1)
        sources = sources.paginate(page, page_size)
    return list(sources)
```

noirbizarre: No misunderstanding: the returned value in case of pagination is not a list but an object like this:
```
{items: [], total: 100, page: 2, page_size: 10}
```
I chose to have only one function with different return depending on pagination active or not but I can split in 2 functions to always have the same return format (with forced pagination in one).

davidbgk: I think we should split, as you previously suggested but I didn't understand at that time. Sorry.

noirbizarre: Done :+1:

PR #823, opened 2017-03-15
Latest url strip spaces

noirbizarre:
This pull request fixes bad URL redirection handling and:
- ensure URL are properly stripped at both form and model level
- resource URL is required and should be an URL

### DIFF
--- a/udata/models/url_field.py
+++ b/udata/models/url_field.py
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from mongoengine.fields import URLField as MEURLField
+
+
+class URLField(MEURLField):
+    '''
+    An URL field that automatically strips extra spaces
+    '''
+    def to_python(self, value):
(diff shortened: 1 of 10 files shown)


davidbgk:
Why do we have to check both at the form and at the model level? Legacy? What about a migration?

noirbizarre:
Because this is 2 separate components that are not always used together:
- when model is directly used, we need the `mongoengine.URLField` to strip
- when a `Form` (not a `ModelForm`) is used without a model, we need the `form.field.URLField` to strip

udata/models/url_field.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from mongoengine.fields import URLField as MEURLField
+
+
+class URLField(MEURLField):
+    '''
+    An URL field that automatically strip extra spaces
```
davidbgk: stripS

PR #826, opened 2017-03-22
Improve images handling

noirbizarre:
This PR supersedes #815 and:
- activate images optimization by default
- add a `images render` command to force (re)rendering of stored images

### DIFF
--- a/udata/commands/images.py
+++ b/udata/commands/images.py
@@ -1,0 +1,73 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import logging
+
+from collections import Counter
+
+from udata.commands import submanager, green, cyan, white
+from udata.core.organization.models import Organization
+from udata.core.post.models import Post
+from udata.core.reuse.models import Reuse
+from udata.core.user.models import User
+
+log = logging.getLogger(__name__)
+
+
+m = submanager(
+    'images',
+    help='Images related operations',
+    description='Handle all images related operations and maintenance'
+)
+
+
+def render_or_skip(obj, attr):
+    try:
+        getattr(obj, attr).rerender()
+        obj.save()
+        return 1
+    except Exception as e:
+        log.warning('Skipped "%s": %e(%s)', obj, e.__class__.__name__, e)
+        return 0
+
+
+@m.command
+def render():
+    '''Force (re)rendering stored images'''
+    print(cyan('Rendering images'))
+
+    count = Counter()
+    total = Counter()
+
+    organizations = Organization.objects(logo__exists=True)
+    total['orgs'] = organizations.count()
+    print(white('Processing {0} organizations logos'.format(total['orgs'])))
+    for org in organizations:
+        count['orgs'] += render_or_skip(org, 'logo')
+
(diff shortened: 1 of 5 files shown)


udata/commands/images.py
```diff
@@ -0,0 +1,73 @@
...
+        obj.save()
+        return 1
+    except Exception as e:
+        log.warning('Skipped "%s": %e(%s)', obj, e.__class__.__name__, e)
+        return 0
+
+
+@m.command
+def render():
+    '''Force (re)rendering sotred images'''
```
vinyll: _sotred_ > _stored_

davidbgk:
If I understand correctly given https://github.com/noirbizarre/flask-fs/commit/af5d5bb4d76a8a89caa6d2d7d02a76adc5497d85 and https://github.com/noirbizarre/flask-fs/commit/7b48c9aa245707026ed034ffffe6702b612a9847 the image is overridden and the original one is lost. Don't we want to keep it in order to have more flexibility for future optimizations?

noirbizarre:
@vinyll: There is no other way to do it. Data migration only apply to mongo data. Here we need to rerender images which is a case not handled by migrations. This also allows to change the image specs (not done in this PR but I plan to do it in a close futur) and force a new rendering.

@davidbgk: no, the original one is always kept so we can rerender in the futur. In this special case (rerendering), I use `overwrite=True` because I overwrite the original with itself.

udata/tests/forms/test_image_field.py
```diff
@@ -77,7 +77,7 @@ def test_with_image(self):
             doc.image.save(img, 'image.jpg')
         doc.save()
         form = self.F(None, obj=doc)
-        self.assertEqual(form.image.filename.data, 'image.jpg')
+        self.assertEqual(form.image.filename.data, 'image.png')
```
davidbgk: Why are you checking the initial image and not the saved one now?

udata/tests/forms/test_image_field.py
```diff
@@ -86,7 +86,7 @@ def test_with_image_and_bbox(self):
             doc.thumbnail.save(img, 'image.jpg', bbox=[10, 10, 100, 100])
         doc.save()
         form = self.F(None, obj=doc)
-        self.assertEqual(form.thumbnail.filename.data, 'image.jpg')
+        self.assertEqual(form.thumbnail.filename.data, 'image.png')
```
davidbgk: Same here.

udata/tests/forms/test_image_field.py
noirbizarre: No, this is because the image are now optimized on `save`, and so converted as png as part of the process. This test the initial form data which is populated from the model field which is now the optimized one.

davidbgk: alright, is that possible to rename the output to avoid the confusion?

noirbizarre: It needs changes at the flask-fs level. I can try to keep the original name and add `.optimized` to the new rendered file.

noirbizarre:
@davidbgk Good to merge ?

PR #833, opened 2017-03-29
Make the allowed extensions list configurable (fix #732)

### LINKED ISSUE #732: Make allowed file extensions to upload a config parameter instead of hardcoded
At the moment, we have to change udata core code to add extensions allowed for upload. Ideally, this list of extensions should become a `udata.cfg` option.

https://github.com/opendatateam/udata/blob/bf836e66a8dc2ad0e6540bc5e6d52d7d47bd6c8d/js/mixins/uploader.js#L70

noirbizarre:
This PR make the allowed extensions list configurable through the `ALLOWED_RESOURCES_EXTENSIONS` setting.
This make the admin file uploader prevent uploading these files AND it prevent these files from manually being uploaded through the API.

It also fixes some file upload bugs:
- properly displays upload errors in the admin
- do not behave like it succeed on upload error in the admin
- unauthorized file type is properly handled as error 400 with a json response in the API

### DIFF
--- a/udata/api/__init__.py
+++ b/udata/api/__init__.py
@@ -9,6 +9,7 @@ from functools import wraps
 from flask import (
     current_app, g, request, url_for, json, make_response, redirect, Blueprint
 )
+from flask_fs import UnauthorizedFileType
 from flask_restplus import Api, Resource, inputs, cors
 
 from udata import search, theme, tracking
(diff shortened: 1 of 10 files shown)


udata/api/__init__.py
```diff
@@ -209,6 +210,13 @@ def handle_value_error(error):
     return {'message': str(error)}, 400
 
 
+@api.errorhandler(UnauthorizedFileType)
+@api.marshal_with(default_error, code=400)
+def handle_unauthorized_file_type(error):
+    '''Error occuring when the user try to upload a non-allowed file type'''
+    return {'message': 'This file type is not allowed'}, 400
```
davidbgk: What about adding a link to `allowed_extensions`?

PR #841, opened 2017-03-31
Allows some needed headers in CORS preflight requests

noirbizarre:
This PR ensure headers used by udata API are authorized in CORS preflight requests.


udata/api/__init__.py
```diff
@@ -33,6 +33,14 @@
 DEFAULT_PAGE_SIZE = 50
 HEADER_API_KEY = 'X-API-KEY'
 
+# TODO: make upstream flask-restplus automatically handle
```
davidbgk: 👍

PR #845, opened 2017-04-03
Fix existing value display on zone completer

noirbizarre:
This PR dix display for selected values on edit.

Before fix:
screenshot-next data gouv fr 2017-04-03 16-07-56

After:
screenshot-www data dev 2017-04-03 16-08-40

### DIFF
--- a/js/components/form/zone-completer.vue
+++ b/js/components/form/zone-completer.vue
@@ -64,22 +64,21 @@ import utils from 'utils';
 
 
 export default {
-    name: 'zone-completer',
     mixins: [BaseCompleter],
     ns: 'spatial',
     endpoint: 'suggest_zones',
-    data: function() {
+    data() {
         return {
             levels: new GeoLevels()
         }
     },
-    selectize: function() {
+    selectize() {
         return {
             valueField: 'id',
             labelField: 'name',
             searchField: ['name', 'code', 'extraKeys'],
             plugins: ['remove_button'],
-            create: function(input, callback) {
+            create(input, callback) {
                 if (utils.isString(input)) {
                     // Need to fetch the real zone
                     // TODO: expose direct full acces to zone by ID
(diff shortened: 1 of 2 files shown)


js/components/form/zone-completer.vue
```diff
@@ -120,31 +119,33 @@ export default {
                     opt.push('</div>');
                     return opt.join('');
```
vinyll: In order to improve readabilty and because you are rewriting into ES6, I would use the ES6 template syntax:

```js
const opt = `
    <div class="selectize-zone">
        <span class="title">
            <span class="name">${escape(data.name)}</span>
            <span class="code">${escape(data.code)}</span>
        </span>
        <span class="level">${this.levelLabel(data)}</span>`;
if (data.keys) {
    opt += '<ul>';
    Object.keys(data.keys).map(key => {
        opt += `
            <li>
                <span class="text-uppercase">${opt.push(escape(key))}</span>: 
                <span class="value">${escape(data.keys[key])}</span>
            </li>`;
    opt += '</ul>';
}
opt += '</div>';
return opt;
```

js/components/form/zone-completer.vue
```diff
@@ -120,31 +119,33 @@ export default {
...
+                    const code = data.hasOwnProperty('properties') ? data.properties.code : data.code;
                     return [
                         '<div class="selectize-zone-item">',
-                        escape(data.name),
+                        escape(name),
                         ' <span class="code">(',
-                        escape(data.code),
+                        escape(code),
                         ')</span></div>'
                     ].join('');
```
vinyll: Same here, moving to ES6 templates

js/components/form/zone-completer.vue
noirbizarre: I agree, but I wanted to limit the diff. These components will be rewritten on the dev branch in pure Vue.js to drop selectize (and so drop jquery)

PR #846, opened 2017-04-03
Fix checksum field not sending values

noirbizarre:
This PR fix checksum field not exporting values to the form (missing field for `checksum.type` and wrong name for `checksum.value`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
   [#850](https://github.com/opendatateam/udata/pull/850)
 - Fix the inline resource edit button not redirecting to admin
   [#852](https://github.com/opendatateam/udata/pull/852)
+- Fix broken checksum component
+  [#846](https://github.com/opendatateam/udata/issues/846)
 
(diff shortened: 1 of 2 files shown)


vinyll:
Seems good with me.
However I'm never sure how a widget would react when using the usual keyboard keys (ie: up/down for combo box and enter for selecting) in form with `<a @click=…` in Vue.

noirbizarre:
This is not Vue.js specific (the syntax is only a shortcut on raw event handlers). Focus+`Enter`, Focus+`space`, touch event should trigger a `click` event on any natively clickable element (anchor, button...) (with the exception of implicite submission, ie. Enter in focused form submit even if the focus is a radio button).
But you're right, in case of an anchor, I need to check but I might need to leave the `href` attribute, even if empty, to make the anchor a clickable element. If so, I think I will be able to remove the `cursor: pointer;` CSS rule 👍

noirbizarre:
After testing, the type picker does not works on keyboard without the `href`. Fixed

PR #849, opened 2017-04-05
Display badge label on dataset and org page

noirbizarre:


udata/templates/dataset/display.html
```diff
@@ -190,7 +190,7 @@ <h3 class="text-left">{{ _('Producer') }}</h3>
                                 </a>
                                 {% for badge in dataset.badges %}
                                     <a href="{{ url_for('front.search', badge=badge) }}">
-                                        {{ badge }}</a>{% if not loop.last %}, {% endif %}
+                                        {{ dataset.__badges__[badge.kind] }}</a>{% if not loop.last %}, {% endif %}
```
davidbgk: Do we want to expose `__private__` in templates? Why iterating over badges is not enough?

noirbizarre: Because the way it's done, right now badge object doesn't has access to the label :/

davidbgk: Oh I get it, too bad :(

noirbizarre:
Changed to hide the private `__badges__` access.

udata/templates/organization/display_base.html
```diff
@@ -44,7 +44,7 @@
                                 <a href="{{ url_for('front.search', badge=badge) }}"
                                     title="{{ _('See all organizations with that badge.') }}">
                                     <span class="glyphicon glyphicon-bookmark"></span>
-                                    {{ badge }}</a>
+                                    {{ org.badge_laebl(badge) }}</a>
```
davidbgk: Typo

davidbgk:
I really prefer that way, thanks!

PR #850, opened 2017-04-05
Remove useless discussions from context

noirbizarre:


noirbizarre:
;) I think we can perform a full review to find others like this

PR #852, opened 2017-04-06
Fix the inline resource edit button not redirecting to admin

noirbizarre:


noirbizarre:
Been there, done that 😉 
Not fond of ruby but arraw functions is something I liked, that and the string interpolations (both are in ES6 which I like)

PR #860, opened 2017-04-12
Allow more headers in CORS preflight request

noirbizarre:
Added some required or common headers to preflight request authorized headers


noirbizarre:
Yep, this is temporary 👍

Issue #526, opened 2016-07-20
Epic: replace JQuery widgets by VueJS components


vinyll:
The following JS widgets should be removed and replaced by their VueJS components:

- [ ] Integrate-Btn: check usage in _territoire_ project prior to refactor this
- [ ] issues-btn: in progress
- [ ] featured: to do
- [ ] share-btn: done, delete widget
- [ ] follow-btn: done, delete widget 
- [ ] ellipsis: to do
- [ ] popover: Vuejs exists, insert directive where old widget is still used (see plugin / tooltip)
- [ ] datarange-picker: use from vuestrap

To do last:
- [ ] modal: delete widget, clear usage in admin

Most of these widgets are located on _/js/widgets_ and may be migrated to _/js/components_

vinyll:
_featured_: https://github.com/opendatateam/udata/pull/527

jdesboeufs:
Share and Follow seem to have been migrated. Cc @vinyll

PR #864, opened 2017-04-12
Fix completer images

noirbizarre:
This PR:
- fixes missing avatars/logos/images in admin completers
- ensure completion API image URLs are absolute (so completion API is usable on another domain)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
   [#860](https://github.com/opendatateam/udata/pull/860)
 - Fix editorialization admin
   [#863](https://github.com/opendatateam/udata/pull/863)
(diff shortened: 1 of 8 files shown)


davidbgk:
Probably not related to this but there are both a `zone-completer` and a `territory-completer`, which is in use? Do we want to uniformize? I don't want to add more to the current territory-related PR :/

noirbizarre:
Good point ! After a quick search, `territory-completer` doesn't seam to be used.
As there is a completer cleanup planed as part of #526 epic, I added the removal point so we wont forget.
(Working on this bugfix remind me why pure vue is easier than a vue/jquery/jquery-plugin mixmatch 😱 😉)

PR #870, opened 2017-04-14
Upgrade jinja to 2.9

l-vincent-l:
The changelog mentions performance gains.

### DIFF
--- a/requirements/install.pip
+++ b/requirements/install.pip
@@ -44,6 +44,4 @@ wtforms==2.1
 xmltodict==0.10.2
 geojson>=1.3.1
 pymongo==3.4.0
-
-# Lock Jinja version because of changes impacting markdown rendering
-Jinja2<2.9
+Jinja2==2.9.6
--- a/udata/frontend/markdown.py
+++ b/udata/frontend/markdown.py
@@ -101,7 +101,7 @@ def mdstrip(value, length=None, end='…'):
     rendered = md(value)
     text = do_striptags(rendered)
     if length > 0:
-        text = do_truncate(text, length, end=end)
+        text = do_truncate(None, text, length, end=end, leeway=2)
     return text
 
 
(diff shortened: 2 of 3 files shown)


requirements/install.pip
```diff
@@ -45,5 +45,4 @@ xmltodict==0.10.2
 geojson>=1.3.1
 pymongo==3.4.0
 
-# Lock Jinja version because of changes impacting markdown rendering
-Jinja2<2.9
+Jinja2==2.9
```
noirbizarre: Can you remove the empty line above and pin it to the real published version ? (2.9.6 right now)

udata/tests/frontend/test_markdown.py
```diff
@@ -155,7 +155,7 @@ def test_mdstrip_does_not_truncate_in_tags(self):
         with self.app.test_request_context('/'):
             result = render_template_string('{{ text|mdstrip(5) }}', text=text)
 
-        self.assertEqual(result.strip(), 'Here …')
+        self.assertEqual(result.strip(), 'Here…')
```
noirbizarre: Thanks, much more logical.

udata/frontend/markdown.py
```diff
@@ -101,7 +101,7 @@ def mdstrip(value, length=None, end='…'):
     rendered = md(value)
     text = do_striptags(rendered)
     if length > 0:
-        text = do_truncate(text, length, end=end)
+        text = do_truncate(text, length, end=end, leeway=2)
```
noirbizarre: I think the filter signature need to be updated to reflect jinja 2.9 changes (and make tests pass):
- decorate the method with `@environmentfilter` and add `env` as first parameter (see https://github.com/pallets/jinja/blob/master/jinja2/filters.py#L466-L467)
- pass `env` as first parameter to `do_truncate`

PR #877, opened 2017-04-23
Fix broken post view

noirbizarre:


js/components/reuse/card-list.vue
```diff
@@ -117,7 +117,7 @@ export default {
     props: {
         title: {
             type: String,
-            default: () => this._('Reuses')
+            default() {return this._('Reuses')}
```
vinyll: Not clear why switching back to ES5 is fixing the post view.
Noob question: it **must** be a function or would an assignment be ok (`default: this._('Reuses')`)?

noirbizarre: It's not switching to ES5, classic functions are still part of ES6. This a case where the difference between a classic function and an arrow functions matters: `this` is not defined at the declaration time so the scoping is wrong when declaring an arrow function. It works as a classique function because Vue.js define `this` as the component when calling propertis functions, methods, event handlers...

The `default: this._('Reuses')` doesn't work for the same reason: there is no `this` here.

vinyll: 👍

PR #879, opened 2017-04-23
Expose both image/and thumbnail in references too (fix #824)

### LINKED ISSUE #824: Unable to retrieve full size organization logo from API
The API returns an URL to a resized image (100x100):
https://www.data.gouv.fr/api/1/organizations/54da1f52c751df59f7467389/

The regular organization page display the full size image:
https://www.data.gouv.fr/fr/organizations/grand-poitiers-open-data/

Consequently we are forced to display a low resolution image in our geo-specific portal:
https://inspire.data.gouv.fr/datasets/e80d3d24c0540868a5656ab8bb91b66af24b80e4

noirbizarre:


noirbizarre:
I tried to stay coherent with the first part of this change. I kept the original name first, but there was some discussions and it has been decided that it is more coherent in term of naming and that it is considered as non-breaking because the image is still right, only bigger for those not applying the migration to `_thumbnail`.

PR #884, opened 2017-04-25
Add MDStripField

l-vincent-l:
This field will populate the field designated by populate with mdstrip(value).

I used it in reuse.description because this takes about 10% of the generation of pages with a lot of reuse cases.

### DIFF
--- a/udata/commands/description_rendered.py
+++ b/udata/commands/description_rendered.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import logging
+
+from udata.commands import manager
+from udata.core.reuse.models import Reuse
+
+from flask import current_app
+from udata.frontend.markdown import mdstrip
+
+log = logging.getLogger(__name__)
+
+@manager.command
+def add_description_rendered():
+    objects = Reuse.objects(description_rendered__exists=False)
+    log.info("Will update %s reuses" % len(objects))
+    for reuse in objects:
+        reuse.description_rendered = mdstrip(reuse.description)
+        reuse.save()
--- a/udata/commands/description_rendered.py
+++ b/udata/commands/description_rendered.py
@@ -3,11 +3,10 @@
 from __future__ import unicode_literals
 
 import logging
+from flask import current_app
 
 from udata.commands import manager
 from udata.core.reuse.models import Reuse
-
-from flask import current_app
 from udata.frontend.markdown import mdstrip
 
 log = logging.getLogger(__name__)
@@ -15,7 +14,8 @@ log = logging.getLogger(__name__)
 @manager.command
 def add_description_rendered():
     objects = Reuse.objects(description_rendered__exists=False)
-    log.info("Will update %s reuses" % len(objects))
+    log.info('Will update %s reuses' % len(objects))
     for reuse in objects:
         reuse.description_rendered = mdstrip(reuse.description)
         reuse.save()
+    log.info('Done.')
--- a/udata/core/reuse/models.py
+++ b/udata/core/reuse/models.py
@@ -41,8 +41,7 @@ class ReuseQuerySet(db.OwnedQuerySet):
                     db.Q(datasets__0__exists=False) |
                     db.Q(deleted__ne=None))
 
-decorator = mdstrip_field('description')
-@decorator
+@mdstrip_field('description')
 class Reuse(db.Datetimed, WithMetrics, BadgeMixin, db.Owned, db.Document):
     title = db.StringField(required=True)
     slug = db.SlugField(
--- a/udata/core/reuse/models.py
+++ b/udata/core/reuse/models.py
@@ -11,6 +11,7 @@ from udata.frontend.markdown import mdstrip
 from udata.i18n import lazy_gettext as _
 from udata.models import db, BadgeMixin, WithMetrics
 from udata.utils import hash_url
+from udata.models.md_fields import mdstrip_field
 
 __all__ = ('Reuse', 'REUSE_TYPES')
 
@@ -40,12 +41,14 @@ class ReuseQuerySet(db.OwnedQuerySet):
                     db.Q(datasets__0__exists=False) |
                     db.Q(deleted__ne=None))
 
-
+decorator = mdstrip_field("description")
+@decorator
 class Reuse(db.Datetimed, WithMetrics, BadgeMixin, db.Owned, db.Document):
     title = db.StringField(required=True)
     slug = db.SlugField(
         max_length=255, required=True, populate_from='title', update=True)
     description = db.StringField(required=True)
+    description_rendered = db.StringField()
     type = db.StringField(required=True, choices=REUSE_TYPES.keys())
     url = db.StringField(required=True)
     urlhash = db.StringField(required=True, unique=True)
--- a/udata/core/reuse/models.py
+++ b/udata/core/reuse/models.py
@@ -41,7 +41,7 @@ class ReuseQuerySet(db.OwnedQuerySet):
                     db.Q(datasets__0__exists=False) |
                     db.Q(deleted__ne=None))
 
-decorator = mdstrip_field("description")
+decorator = mdstrip_field('description')
 @decorator
 class Reuse(db.Datetimed, WithMetrics, BadgeMixin, db.Owned, db.Document):
     title = db.StringField(required=True)
--- a/udata/models/md_fields.py
+++ b/udata/models/md_fields.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
-from udata.frontend.markdown import mdstrip
 from mongoengine import signals
+from udata.frontend.markdown import mdstrip
 
 
 def mdstrip_field(populate_from, populate=None, length=None, end=None):
--- a/udata/models/md_fields.py
+++ b/udata/models/md_fields.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from udata.frontend.markdown import mdstrip
+from mongoengine import signals
+
+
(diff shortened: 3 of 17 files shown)


udata/core/reuse/models.py
```diff
@@ -45,7 +45,8 @@ class Reuse(db.Datetimed, WithMetrics, BadgeMixin, db.Owned, db.Document):
     title = db.StringField(required=True)
     slug = db.SlugField(
         max_length=255, required=True, populate_from='title', update=True)
-    description = db.StringField(required=True)
+    description = db.MDStripField(required=True, populate='description_md')
+    description_md = db.StringField()
```
noirbizarre: Why switch the field ? It breaks the API (we need to keep the unrendered fields at the same place).
**Proposal**: simply add:
```python
description_rendered = db.MDStripField(required=True, populate='description')
```

udata/models/md_fields.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from mongoengine.fields import StringField
+from udata.frontend.markdown import mdstrip
+
+class MDStripField(StringField):
```
noirbizarre: As stripping/truncation behavior is optional, let's call it `MarkdownField` which is more explicit ;)

udata/tests/reuse/test_reuse_model.py
```diff
@@ -10,9 +10,15 @@
 from udata.core.user.factories import UserFactory
 
 from .. import TestCase, DBTestMixin
+from udata.frontend.markdown import init_app
 
 
 class ReuseModelTest(TestCase, DBTestMixin):
+    def create_app(self):
```
noirbizarre: I think we can (we need to) improve that. This change improve run performance but massively reduced test speed (which is already an issue).
Maybe we can disable the behavior on tests and only activate it on demand, when required (in front tests). This is already done for some other behavior with a specific setting like `ENABLE_MARKDOWN_PRERENDERING` which is `True` by default, `False` in testing and ovrriden to `True` when required in tests.

l-vincent-l: I tested the two branches, I couldn't see any difference in test speed with this code.

l-vincent-l:
@davidbgk migration has been handle
@noirbizarre I think the changes you've asked have been done too.

udata/core/reuse/models.py
```diff
@@ -40,12 +41,14 @@ def hidden(self):
                     db.Q(datasets__0__exists=False) |
                     db.Q(deleted__ne=None))
 
-
+decorator = mdstrip_field("description")
+@decorator
```
davidbgk: Why not directly `@mdstrip_field("description")`? (btw, we try to use single quotes when possible)

udata/models/md_fields.py
```diff
@@ -0,0 +1,21 @@
...
+from udata.frontend.markdown import mdstrip
+from mongoengine import signals
+
+
+def mdstrip_field(populate_from, populate=None, length=None, end=None):
+
+    def func_(sender, document, *args, **kwargs):
+        to_ = populate or populate_from + '_rendered'
+        print document._get_changed_fields()
+        print document._created
```
davidbgk: Remove prints :)

udata/tests/test_model.py
```diff
@@ -28,6 +30,17 @@ class SlugTester(db.Document):
     }
 
 
+decorator = mdstrip_field('description', 'description_rendered')
+
+@decorator
```
davidbgk: Same here, use directly the decorator

udata/tests/test_model.py
```diff
@@ -178,6 +191,28 @@ def test_crop(self):
         self.assertEqual(len(obj.slug), SlugTester.slug.max_length)
 
 
+class MDStripFieldTest(DBTestMixin, TestCase):
```
davidbgk: 👍

davidbgk:
@l-vincent-l where is the migration?

udata/core/reuse/models.py
l-vincent-l: I don't know.
For an unknown reason it doesn't work..

l-vincent-l:
I forgot the migration... 
It's here : https://github.com/opendatateam/udata/pull/884/files#diff-f10310a0f27f36b71ccb3522135d9210

udata/commands/description_rendered.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import logging
+
+from udata.commands import manager
+from udata.core.reuse.models import Reuse
+
+from flask import current_app
```
davidbgk: Third-party imports before local ones: https://www.python.org/dev/peps/pep-0008/#imports

udata/commands/description_rendered.py
```diff
@@ -0,0 +1,21 @@
...
+from udata.frontend.markdown import mdstrip
+
+log = logging.getLogger(__name__)
+
+@manager.command
+def add_description_rendered():
+    objects = Reuse.objects(description_rendered__exists=False)
+    log.info("Will update %s reuses" % len(objects))
```
davidbgk: We use to put infos related to the correct execution of a command at the end. At least `Done.`.

davidbgk: (and favor single quotes :) )

udata/core/reuse/models.py
```diff
@@ -40,12 +41,14 @@ def hidden(self):
                     db.Q(datasets__0__exists=False) |
                     db.Q(deleted__ne=None))
 
-
+decorator = mdstrip_field('description')
+@decorator
```
davidbgk: We need to figure out why, what is the error? I guess it's because you're applying it to a class and as such the first parameter of your decorator should be the class instance (even if not in use).

udata/models/md_fields.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from udata.frontend.markdown import mdstrip
+from mongoengine import signals
```
davidbgk: Wrong import order.

udata/templates/home.html
```diff
@@ -85,7 +85,7 @@ <h1 class="page-header">
                         <img class="img-responsive" src="{{reuse.image}}" alt="">
                     </div>
                     <div class="col-md-5">
-                        <p>{{reuse.description | mdstrip(300) }}</p>
+                        <p>{{reuse.description_rendered | truncate(300) }}</p>
```
davidbgk: Do not let spaces around `|` (I know, it was inconsistent before ;) )

udata/tests/reuse/test_reuse_model.py
davidbgk: The impact on circle-ci is more than a minute:

capture d ecran 2017-05-05 a 15 09 13

udata/core/reuse/models.py
l-vincent-l: There's no error.
The decorator seems to have no effect, the func_ code is never run.

davidbgk: I guess that's because `apply` parameters are wrong in this case and thus the signal is not sent.

davidbgk: I'm able to use the decorator as expected (`@mdstrip_field('description')`) with that code:

```python
# -*- coding: utf-8 -*-
from mongoengine.signals import pre_save_post_validation

from udata.frontend.markdown import mdstrip


def mdstrip_field(field_name):

    def apply_mdstrip(sender, document, **kwargs):
        if (field_name in document._get_changed_fields() or
                getattr(document, '_created')):
            initial_value = getattr(document, field_name)
            stripped_value = mdstrip(initial_value)
            document.__setattr__(field_name + '_rendered', stripped_value)

    class Inner(object):

        def __init__(self, cls):
            self.cls = cls

        def __getattr__(self, attr):
            return getattr(self.cls, attr)

        def __call__(self, *args, **kwargs):
            pre_save_post_validation.connect(apply_mdstrip, sender=self.cls)
            return self.cls(*args, **kwargs)

    return Inner
```

BUT, it still requires some work to make all tests pass (only `udata/tests/test_model.py` are OK now). It might be a good starting point.

davidbgk:
My proposal to avoid the custom decorator and use existing signals: https://github.com/davidbgk/udata/commit/e5e2604f999dd10f1bc4971e1633428204a76083

udata/commands/description_rendered.py
```diff
@@ -0,0 +1,21 @@
...
+from udata.frontend.markdown import mdstrip
+
+log = logging.getLogger(__name__)
+
+@manager.command
+def add_description_rendered():
+    objects = Reuse.objects(description_rendered__exists=False)
+    log.info('Will update %s reuses' % len(objects))
+    for reuse in objects:
+        reuse.description_rendered = mdstrip(reuse.description)
```
davidbgk: We should not have to do this but just save the existing reuse would do the conversion for us.

davidbgk: Mmh given that we test on `document._get_changed_fields()` it may not be that simple.

l-vincent-l: Actually it is! How wonderful!

udata/core/reuse/models.py
l-vincent-l: I made something else that is working :)

davidbgk:
Why not using the decorator as in https://github.com/davidbgk/udata/commit/3f6d62ee1de7da6f39ccac48fb0d39eadacf64a2 finally?

l-vincent-l:
Tests were not passing.

oncletom:
Isn't it part of #253 scope? Aka making backend and frontend markdown rendering on par?

abulte:

PR #887, opened 2017-04-27
Drop current badges metrics (fix #885)

### LINKED ISSUE #885: Bug during the task metrics update if there's a badge
## Detailed Description
When i use the task `udata metrics update`, there's an error if there's a dataset with a badge:
Line 16 of udata/udata/core/badges/metrics.py
`AttributeError: 'DatasetBadgesMetric' object has no attribute 'model_badge'`

## Your Environment
UData v1.0.10.dev0

noirbizarre:


noirbizarre:
I think you're right.
After a quick lookup in the code, I think this metric is not really in use.
I'm wondering if counting the number of badge on a dataset is really interesting.
I would be more interested of having the number of object of a given badge (ie. number of organization with public service badge).

@odtvince @davidbgk @vinyll @jdesboeufs I'm thinking in deactivating this on the master for the next 1.0.10 release in order to make the `udata metrics update` command pass and rework the metric into a site metrics for badge mount by model on the dev branch. What do you think about it ?

@ChrisBarriere Would it be OK for you ?

ChrisBarriere:
Ok for deactiving this

PR #892, opened 2017-05-02
Fix topic search

noirbizarre:
This PR ensures dataset or reuse without the required tags are never included in topic search

### DIFF
--- a/udata/tests/frontend/test_topic_frontend.py
+++ b/udata/tests/frontend/test_topic_frontend.py
@@ -3,10 +3,73 @@ from __future__ import unicode_literals
 
 from flask import url_for
 
+from udata import search
+from udata.tests import SearchTestMixin, TestCase
 from udata.tests.frontend import FrontTestCase
 from udata.core.dataset.factories import VisibleDatasetFactory
+from udata.core.dataset.search import DatasetSearch
 from udata.core.reuse.factories import VisibleReuseFactory
 from udata.core.topic.factories import TopicFactory
+from udata.core.topic.views import topic_search_for
+
+
+class TopicSearchTest(SearchTestMixin, TestCase):
+    def test_empty_search_no_match(self):
+        '''Should return no result if no data match the tags'''
+        with self.autoindex():
+            VisibleDatasetFactory.create_batch(2, tags=['whatever'])
+        topic = TopicFactory(tags=['no-match'])
+
+        query = topic_search_for(topic, DatasetSearch)
+        result = search.query(query)
+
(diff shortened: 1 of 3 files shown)


udata/tests/frontend/test_topic_frontend.py
```diff
@@ -3,10 +3,73 @@
...
+            VisibleDatasetFactory.create_batch(2, tags=['whatever'])
+        topic = TopicFactory(tags=['no-match'])
+
+        query = topic_search_for(topic, DatasetSearch)
+        result = search.query(query)
+
+        self.assertEqual(len(result), 0)
+
+    def test_empty_search_with_match(self):
+        '''Should only return no data with at least one tag'''
```
davidbgk: `Should only return data matching topic's tag`? Otherwise I don't get the logic?

udata/tests/frontend/test_topic_frontend.py
```diff
@@ -3,10 +3,73 @@
...
+            match = VisibleDatasetFactory.create_batch(2, tags=[
+                'in', 'filtered'
+            ])
+            # Match the topic condition but not the queried tag
+            no_match = VisibleDatasetFactory.create_batch(2, tags=['in'])
+            # Excluded because not matching one of the topic tag
+            excluded = VisibleDatasetFactory.create_batch(2, tags=[
+                'out', 'filtered'
+            ])
+        topic = TopicFactory(tags=['in', 'no-match'])
```
davidbgk: Is the `no-match` required if not tested?

udata/tests/frontend/test_topic_frontend.py
```diff
@@ -3,10 +3,73 @@
...
+        result = search.query(query)
+
+        self.assertEqual(len(result), 0)
+
+    def test_empty_search_with_match(self):
+        '''Should only return no data with at least one tag'''
+        with self.autoindex():
+            included = VisibleDatasetFactory.create_batch(2, tags=['in'])
+            excluded = VisibleDatasetFactory.create_batch(2, tags=['out'])
+        topic = TopicFactory(tags=['in', 'no-match'])
```
davidbgk: Same here, is `no-match` useful?

noirbizarre: Yes, ensures `at least` is the condition, not `all`

udata/tests/frontend/test_topic_frontend.py
noirbizarre: same

PR #902, opened 2017-05-11
Use tmpfs mongo

l-vincent-l:
This PR makes tests run 30 seconds faster.

You need to add : 
``MONGODB_PORT=27018``
to your configuration to activate the use of tmpfs

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -29,11 +29,11 @@ services:
 
 
 
-        search:
-            image: udata/elasticsearch:2.4.5
-            volumes:
-              - ./data/search/data:/usr/share/elasticsearch/data
-            expose:
-              - "9200"
-            ports:
-              - "9200:9200"
+    search:
+        image: udata/elasticsearch:2.4.5
+        volumes:
+          - ./data/search/data:/usr/share/elasticsearch/data
+        expose:
+          - "9200"
+        ports:
+          - "9200:9200"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,27 +1,31 @@
-db:
-    image: mongo:3.2
-    command: mongod
-    volumes:
-      - ./data/db:/data/db
-    expose:
-      - "27017"
-    ports:
-    - "27017:27017"
+version: "3"
+services:
+    db:
+        image: mongo:3.2
+        command: mongod
+        volumes:
+          - ./data/db:/data/db
+        expose:
+          - "27017"
+        ports:
+        - "27017:27017"
 
-broker:
-    image: redis
-    volumes:
-      - ./data/broker:/data
-    expose:
-      - "6379"
-    ports:
-    - "6379:6379"
+    broker:
+        image: redis
+        volumes:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -1,27 +1,37 @@
...
-    image: redis
-    volumes:
-      - ./data/broker:/data
-    expose:
-      - "6379"
-    ports:
-    - "6379:6379"
+    mongo_test:
+        image: mongo:3.4.0
+        tmpfs: /tmp/db
```
noirbizarre: Is this the right path ? (ie. I think it should be `tmpfs: /data/db` to be effective)

udata/models/__init__.py
```diff
@@ -103,6 +103,7 @@ def resolve_model(self, model):
 def init_app(app):
     if app.config['TESTING']:
         app.config['MONGODB_DB'] = '{MONGODB_DB}-test'.format(**app.config)
+        app.config['MONGODB_PORT'] = 27018
```
noirbizarre: I think this will only cover the case for those using the embedded docker file (and for circleci tests). Some are using a native standalone mongo installation.
Maybe we can define some `MONGODB_TEST_*` settings allowing for those who needs it to run on the same DB (or maybe the opposite, allowing in our case to define the alternate db only we used)

circle.yml
```diff
@@ -9,7 +9,7 @@ dependencies:
     - if [[ ! -d mongodb-linux-x86_64-3.2.4 ]]; then wget http://downloads.mongodb.org/linux/mongodb-linux-x86_64-3.2.4.tgz && tar xvzf mongodb-linux-x86_64-3.2.4.tgz; fi
     - sudo /etc/init.d/mongod stop
     - sudo cp mongodb-linux-x86_64-3.2.4/bin/* /usr/bin
-    - sudo /etc/init.d/mongod start
+    - sudo /etc/init.d/mongod start --port 27018
```
noirbizarre: Maybe we can speedup the circleci build too with this version: https://docs.mongodb.com/manual/core/inmemory/ (ie. by adding the `--storageEngine` parameter)

docker-compose.yml
```diff
@@ -1,27 +1,37 @@
...
-broker:
-    image: redis
-    volumes:
-      - ./data/broker:/data
-    expose:
-      - "6379"
-    ports:
-    - "6379:6379"
+    mongo_test:
+        image: mongo:3.4.0
```
noirbizarre: Use the same version that the other one (`mongo:3.2`)

noirbizarre:
I think you are only gaining 30 seconds because the `tmpfs` is wrong and your are only gaining the fact that there is no filesystem binding on `/data/db`

circle.yml
noirbizarre: Hum, this is an init script and parameters are not taken in account.
We need to put a mongod config file to override the default values

noirbizarre:
Merged in #1148

PR #905, opened 2017-05-15
Make metrics deactivable for tests

l-vincent-l:
We can observe a gain of 10% on the runtime of tests

### DIFF
--- a/udata/core/metrics/__init__.py
+++ b/udata/core/metrics/__init__.py
@@ -14,6 +14,7 @@ metric_catalog = {}
 
 
 from udata.models import db  # noqa: need metrics refactoring
+from flask import current_app
 
 from .tasks import update_metric, archive_metric  # noqa
 
@@ -111,10 +112,12 @@ class Metric(object):
                 'Unsupported format: {0} ({1})'.format(value, type(value)))
 
     def trigger_update(self):
-        self.need_update.send(self)
+        if not current_app.config['TESTING'] or current_app.config.get('FORCE_METRICS'):
+            self.need_update.send(self)
 
     def notify_update(self):
-        self.updated.send(self)
+        if not current_app.config['TESTING'] or current_app.config.get('FORCE_METRICS'):
+            self.updated.send(self)
 
     @classmethod
     def aggregate_monthly(cls, queryset, month):
--- a/udata/core/metrics/__init__.py
+++ b/udata/core/metrics/__init__.py
@@ -112,11 +112,11 @@ class Metric(object):
                 'Unsupported format: {0} ({1})'.format(value, type(value)))
 
     def trigger_update(self):
-        if not current_app.config['TESTING'] or current_app.config.get('FORCE_METRICS'):
+        if current_app.config['USE_METRICS']:
             self.need_update.send(self)
 
     def notify_update(self):
-        if not current_app.config['TESTING'] or current_app.config.get('FORCE_METRICS'):
+        if current_app.config['USE_METRICS']:
             self.updated.send(self)
 
     @classmethod
--- a/udata/core/metrics/__init__.py
+++ b/udata/core/metrics/__init__.py
@@ -5,6 +5,7 @@ import logging
 
 from blinker import Signal
 from datetime import date, datetime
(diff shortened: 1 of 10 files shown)


davidbgk:
Nice 👍 

capture d ecran 2017-05-15 a 16 26 49

I hoped it would be more than that but it's better than nothing :)

udata/core/metrics/__init__.py
```diff
@@ -14,6 +14,7 @@
 
 
 from udata.models import db  # noqa: need metrics refactoring
+from flask import current_app
```
davidbgk: Import with third-parties dependencies above.

udata/core/metrics/__init__.py
```diff
@@ -111,10 +112,12 @@ def iso(self, value):
                 'Unsupported format: {0} ({1})'.format(value, type(value)))
 
     def trigger_update(self):
-        self.need_update.send(self)
+        if not current_app.config['TESTING'] or current_app.config.get('FORCE_METRICS'):
```
davidbgk: Worth an entry in [settings' documentation](http://udata.readthedocs.io/en/latest/adapting-settings/)?

udata/core/metrics/__init__.py
```diff
@@ -111,10 +112,12 @@ def iso(self, value):
...
+            self.need_update.send(self)
 
     def notify_update(self):
-        self.updated.send(self)
+        if not current_app.config['TESTING'] or current_app.config.get('FORCE_METRICS'):
```
davidbgk: I'm not sure we want to restrict that to `TESTING`, I'm thinking about local load of backups for instance but that might be another issue. @noirbizarre thoughts?

udata/tests/test_discussions.py
```diff
@@ -37,6 +37,7 @@ def create_app(self):
         return app
 
     def test_new_discussion(self):
+        self.app.config['FORCE_METRICS'] = True
```
davidbgk: Why not in the `setUp` of that class if you have to set it everywhere?

udata/tests/test_issues.py
```diff
@@ -37,6 +37,7 @@ def create_app(self):
         return app
 
     def test_new_issue(self):
+        self.app.config['FORCE_METRICS'] = True
```
davidbgk: Same here.

udata/core/metrics/__init__.py
noirbizarre: Instead of testing both `TESTING` and `FORCE_METRICS`, I think it would be cleaner to have a single `USE_METRICS` defined to `True` in `udata.settings.Defaults` and overriden to `False` in `udata.settings.Testing` so:
- we avoid runtime code to perform to have a view on testing
- we have all testing parameters in the same place
- activating metrics in test is simply setting `config.USE_METRICS` to True
@davidbgk I think local load of backups is not performed inside the application so metrics are not triggered.

udata/core/metrics/__init__.py
noirbizarre: Yep, it should be documented, but I prefer a `USE_METRICS` or `ENABLE_METRICS` instead of force metrics (see next comment)

udata/tests/test_discussions.py
noirbizarre: Agreed :+1: : you should factorize.
Even better, you should write a `MetricsSettings` class for this case. See https://github.com/opendatateam/udata/blob/master/udata/tests/frontend/test_markdown.py#L16-L21 for an example (it's the pattern we use when we need to activate some settings)

l-vincent-l: This was done on purpose, I wanted to activate it as less as possible.

udata/tests/test_transfer.py
```diff
@@ -102,6 +102,9 @@ def test_request_transfer_to_same_organization(self):
 
 
 class TransferAcceptTest(TestCase, DBTestMixin):
+    def setUp(self):
+        self.app.config['USE_METRICS'] = True
```
noirbizarre: Don't forget the `super().setUp()`

PR #910, opened 2017-05-17
Fix js Model.save not updating

noirbizarre:
This PR fix some model `save()` method not performing update (and so breaking the ability for workflow to go back and forth on the steps editing this objects).

This is the very first step toward fully working workflows (cc @vinyll )

For the futur, it would be interesting to normalize the create/save/update methods on models.

### DIFF
--- a/js/models/dataset.js
+++ b/js/models/dataset.js
@@ -23,9 +23,11 @@ export default class Dataset extends Model {
      * Create or update the given dataset.
      */
     save(on_error) {
-        var ep = this.id ? 'datasets.update_dataset' : 'datasets.create_dataset';
+        if (this.id) {
+            return this.update(this, this.on_fetched, on_error);
+        }
         this.loading = true;
-        this.$api(ep, {payload: this}, this.on_fetched, on_error);
+        this.$api('datasets.create_dataset', {payload: this}, this.on_fetched, on_error);
     }
 
     update(data, on_success, on_error) {
--- a/js/models/organization.js
+++ b/js/models/organization.js
@@ -26,32 +26,31 @@ export default class Organization extends Model {
     }
 
     save() {
-        let endpoint = this.id ? 'organizations.update_organization' : 'organizations.create_organization';
-        this.$api(endpoint, {payload: this}, this.on_fetched);
+        if (this.id) {
+            this.update(this, this.on_fetched);
+        } else {
+            this.create();
+        }
     }
 
     create() {
-        this.$api('organizations.create_organization', {
-            payload: JSON.stringify(this.$data)
-        }, this.on_fetched);
+        this.$api('organizations.create_organization', {payload: this}, this.on_fetched);
     }
 
     role_for(obj) {
-        var user_id = obj.hasOwnProperty('id') ? obj.id : obj,
-            members = this.members.filter(function(member) {
-                return member.user.id === user_id;
-            });
+        const user_id = obj.hasOwnProperty('id') ? obj.id : obj;
+        const members = this.members.filter(member => member.user.id === user_id);
         return members.length ? members[0] : null;
     }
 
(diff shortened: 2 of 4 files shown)


js/models/dataset.js
```diff
@@ -23,9 +23,11 @@ export default class Dataset extends Model {
      * Create or update the given dataset.
      */
     save(on_error) {
-        var ep = this.id ? 'datasets.update_dataset' : 'datasets.create_dataset';
+        if (this.id) {
+            return this.update(this, this.on_fetched, on_error);
+        }
         this.loading = true;
-        this.$api(ep, {payload: this}, this.on_fetched, on_error);
+        this.$api('datasets.create_dataset', {payload: this}, this.on_fetched, on_error);
```
davidbgk: To be consistent, maybe introduce a `create` method?

js/models/organization.js
```diff
@@ -26,32 +26,33 @@ export default class Organization extends Model {
...
-            payload: JSON.stringify(this.$data)
-        }, this.on_fetched);
+        this.$api('organizations.create_organization', {payload: this}, this.on_fetched);
     }
 
     role_for(obj) {
-        var user_id = obj.hasOwnProperty('id') ? obj.id : obj,
-            members = this.members.filter(function(member) {
+        const user_id = obj.hasOwnProperty('id') ? obj.id : obj;
+        const members = this.members.filter(function(member) {
```
davidbgk: Arrow func?

js/models/reuse.js
```diff
@@ -22,16 +22,10 @@ export default class Reuse extends Model {
...
-            ep = 'reuses.update_reuse';
-            args = {payload: this, reuse: this.id,};
+            return this.update(this, this.on_fetched, on_error);
         }
-        else {
-            ep = 'reuses.create_reuse';
-            args = {payload: this,};
-        }
-        this.$api(ep, args, this.on_fetched, on_error);
+        this.$api('reuses.create_reuse', {payload: this}, this.on_fetched, on_error);
```
davidbgk: Same here.

js/models/organization.js
```diff
@@ -26,32 +26,33 @@ export default class Organization extends Model {
     }
 
     save() {
-        let endpoint = this.id ? 'organizations.update_organization' : 'organizations.create_organization';
-        this.$api(endpoint, {payload: this}, this.on_fetched);
+        if (this.id) {
+            this.update(this, this.on_fetched);
```
davidbgk: No `on_error` in these cases? (not consistent with datasets and reuses)

js/models/organization.js
```diff
@@ -26,32 +26,33 @@ export default class Organization extends Model {
...
-        return this.role_for(obj) != null;
+        return this.role_for(obj) !== null;
     }
 
     is_admin(obj) {
         if (obj.is_admin) return true;
-        var member = this.role_for(obj);
-        return member != null ? member.role === 'admin' : false;
+        const member = this.role_for(obj);
+        return member !== null ? member.role === 'admin' : false;
```
vinyll: Is this meaning `return member !== null && member.role === 'admin'`
?
In which case it would be easier to read.

js/models/dataset.js
noirbizarre: Yes, I agree, but
> For the futur, it would be interesting to normalize the create/save/update methods on models.

This PR is the minimal bugfix (no signature change, only fix the bug where it is), refactoring all the models and change it in every place where it's used wasn't in the scope. I think it needs a dedicated PR on the dev branch.

js/models/organization.js
noirbizarre: same as above

js/models/reuse.js
noirbizarre: same as above

Issue #760, opened 2017-02-15
Make all components to work without Vue.config.replace = false


noirbizarre:
## Detailed Description
Vue 0.12 had a different behavior when instanciating components, it wasn't replacing the root element.
When migrating to Vue 1.x, we had to toggle `Vue.config.replace = false` in order to maintain a compatibility with some component.
This setting setting is deprecated and then removed in Vue 2.0.

## Context
In order to stay close to standard Vue.js behavior, to allow upgrade and to simplify component styling (the component will be exactly as written, no custom surronding element), we need to drop `Vue.config.replace = false` and fix components for this behavior.

PR #944, opened 2017-05-29
Admin replace true

noirbizarre:
This PR provides easier Vue.js development/debugging:
  - Drop `Vue.config.replace = false`: compatible with Vue.js 1/2 and no more style guessing
    [#760](https://github.com/opendatateam/udata/pull/760) (and no more behavior differences between admin and front)
  - `name` on all components: no more `Anonymous Component` in Vue debugger
  - No more `Fragments`
  - More ES6 everywhere

Most of the fime, it's simply a surrounding `<div>` which is required.

Reviewers: sorry for the big PR

### DIFF
--- a/js/admin.routes.js
+++ b/js/admin.routes.js
@@ -11,48 +11,48 @@ const router = new VueRouter({
 
 router.map({
     '/': {
-        component: function(resolve) {
+        component(resolve) {
             require(['./views/home.vue'], resolve);
         }
     },
     '/me/': {
         name: 'me',
-        component: function(resolve) {
+        component(resolve) {
             require(['./views/me.vue'], resolve);
         }
     },
     '/me/edit/': {
         name: 'me-edit',
-        component: function(resolve) {
+        component(resolve) {
             require(['./views/me-edit.vue'], resolve);
         }
     },
     '/site/': {
         name: 'site',
-        component: function(resolve) {
+        component(resolve) {
             require(['./views/site.vue'], resolve);
         }
     },
     '/dataset/new/': {
-        component: function(resolve) {
+        component(resolve) {
             require(['./views/dataset-wizard.vue'], resolve);
         }
     },
     '/dataset/new/:oid/': {
-        component: function(resolve) {
+        component(resolve) {
             require(['./views/dataset-wizard.vue'], resolve);
         },
-        callback: function(view) {
+        callback(view) {
             if (!view.dataset.id) {
                 view.dataset.fetch(oid);
             }
         }
     },
     '/dataset/new/:oid/share': {
-        component: function(resolve) {
+        component(resolve) {
             require(['./views/dataset-wizard.vue'], resolve);
         },
-        callback: function(view) {
+        callback(view) {
             if (!view.dataset.id) {
                 view.dataset.fetch(oid);
             }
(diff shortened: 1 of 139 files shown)


js/admin.vue
```diff
@@ -1,9 +1,11 @@
 <template>
+<div>
```
thimy: Could the whole block be indented once more? (as well as in the other components where a root div has been added)

js/components/badges/modal.vue
```diff
@@ -63,31 +66,21 @@ export default {
...
     methods: {
         confirm() {
-            const to_add = this.selected.filter((badge) => {
-                        return this.initial.indexOf(badge) < 0;
-                    }),
-                to_remove = this.initial.filter((badge) => {
-                        return this.selected.indexOf(badge) < 0;
-                    });
-
+            const to_add = this.selected.filter(badge => this.initial.indexOf(badge) < 0);
```
noirbizarre: As much as possible, camel case in JS. As this PR is already a big one I tried to limit changes and so snake_case is kept where it is already.

js/components/dataset/details.vue
```diff
@@ -24,13 +25,14 @@
         <span v-for="tag in dataset.tags" class="label label-default">{{tag}}</span>
     </div>
     <div v-if="dataset.badges | length" class="label-list">
-        <strong>
+        <strong>c
```
noirbizarre: Thanks, I will remove that ASAP

js/admin.vue
noirbizarre: I tried not to indent anymore when it's possible to limit the diff size.
I can do it but it will produce a very very big diff.

noirbizarre:
Reviewers: do you want me to migrate to camel case everywhere and to perform full indentation in this PR ?
I tried to avoid that to limit the diff but I can do it if you prefer.

**Pros:**
  - it will be done once and for all
  - it's a start to set some coherent conventions

**Cons:**
  - it will produce a huge PR
  - changes from master will be a little bit harder to import

thimy:
I don't mind having a dedicated PR for this. :)

js/admin.routes.js
```diff
@@ -11,236 +11,236 @@ const router = new VueRouter({
...
                 name: 'dataset-resource',
-                component: function(resolve) {
+                component(resolve) {
                     require(['./components/dataset/resource/modal.vue'], resolve);
                 }
             },
             // '/resource/new/': {
             //     name: 'dataset-new-resource',
-            //     component: function(resolve) {
+            //     component(resolve) {
```
davidbgk: Remove completely?

js/components/datatable/row.vue
```diff
@@ -47,9 +46,9 @@ export default {
          *                             in the application scope when
          *                             the view is loaded
          */
-        load_cell: function(name) {
+        load_cell(name) {
             if (!this.$options.components.hasOwnProperty(name)) {
-                let options = require('./cells/' + name + '.vue');
+                const options = require('./cells/' + name + '.vue');
```
davidbgk: Use back ticks.

js/components/notifications/membership_request.vue
```diff
@@ -11,7 +11,7 @@ import BaseNotification from 'components/notifications/base';
 export default {
     mixins: [BaseNotification],
     methods: {
-        click: function() {
+        click() {
             this.$go('/organization/' + this.details.organization + '#membership-requests');
```
davidbgk: Add back ticks here too?

js/components/user/card-filter.vue
```diff
@@ -33,44 +36,42 @@
...
-                    page: '/user/' + user.slug + '/'
-                }});
-            });
+        users() {
+            return this.completions.map(user => new User({data: {
+                id: user.id,
+                avatar: user.avatar_url,
+                first_name: user.first_name,
+                last_name: user.last_name,
+                page: '/user/' + user.slug + '/'
```
davidbgk: Use back ticks :wink:

js/components/widgets/thumbnailer.vue
```diff
@@ -127,20 +125,20 @@ export default {
...
-        preview: function(coords) {
-            let bounds = this.Jcrop.getBounds(),
-                w = bounds[0],
-                h = bounds[1],
-                // Temp fix until https://github.com/vuejs/vue/issues/1697 is merged
-                // containers = this.$els.preview_containers;
-                containers = [...this.$el.querySelectorAll('.preview-container')];
+        preview(coords) {
+            const bounds = this.Jcrop.getBounds();
+            const w = bounds[0];
```
davidbgk: `const [w, h] = bounds;` should work here.

js/components/widgets/thumbnailer.vue
```diff
@@ -127,20 +125,20 @@ export default {
...
+            const h = bounds[1];
+            // Temp fix until https://github.com/vuejs/vue/issues/1697 is merged
+            // containers = this.$els.preview_containers;
+            const containers = [...this.$el.querySelectorAll('.preview-container')];
 
             containers.forEach(function(el) {
```
davidbgk: Arrow?

js/components/widgets/thumbnailer.vue
```diff
@@ -150,13 +148,13 @@ export default {
...
         },
-        setImage: function(src) {
+        setImage(src) {
             if (!src) return;
 
-            let that = this,
-                $pane = $(this.$els.crop_pane),
-                max_width = $pane.width(),
-                max_height = parseInt($pane.css('max-height').replace('px', ''));
+            const that = this;
```
davidbgk: Still required?

js/components/widgets/thumbnailer.vue
```diff
@@ -167,8 +165,8 @@ export default {
                     boxWidth: max_width,
                     boxHeight: max_height
                 }, function() {
```
davidbgk: Turn to arrow? (makes the `that` useless?)

js/components/widgets/thumbnailer.vue
```diff
@@ -167,8 +165,8 @@ export default {
...
-                    let bounds = this.getBounds(),
-                        size = Math.min(bounds[0], bounds[1]);
+                    const bounds = this.getBounds();
+                    const size = Math.min(bounds[0], bounds[1]);
```
davidbgk: `Math.min(...bounds)` ✨

davidbgk:
> Reviewers: do you want me to migrate to camel case everywhere and to perform full indentation in this PR ?

Given that the extra div looks non-significative to me and will maybe be removed on a future version 😝 I would let it as is.

On camel casing, I think the part where there is both `to_add` and `toAdd` within the same file should be fixed. Not sure if it's required everywhere for now.

js/components/widgets/thumbnailer.vue
noirbizarre: yeah ! :+1:

js/components/datatable/row.vue
noirbizarre: Not in this case: webpack 1.x dynamic import only works on ES5. I added a comment with references.

js/components/widgets/thumbnailer.vue
noirbizarre: Not in this case, the arrow shouldn't be used because the `jCrop` only expose the instance through the `this` scope and we need it (see the `that.jCrop = this`, this is the only way to get it or at least documented in the jCrop documentation)

js/components/widgets/thumbnailer.vue
noirbizarre: No scoping requirement and not one liner: an arrow function will only produce a little bit heavier code (ie. because of the extra `.bind(this)`)

noirbizarre:
Fixed the snake/camel case in `js/components/badges/modal.vue` and all others review's (fixable) issues.
Others have been explained as comment in the code.

js/components/widgets/thumbnailer.vue
davidbgk: I'm not sure we are at that level of performances versus consistency. But fair enough.

noirbizarre:
Yes, and it gives me a pretty clear visions of used standards and where they are not applied (mostly first written components).
I expect next PRs on admin/front JS to be thinner and more focused on features/bugfixes than style changes (with the exception of coding style PRs obviously :wink: )

js/components/widgets/thumbnailer.vue
noirbizarre: lol, I don't think the overhead is even noticeable. But this component is soon to be replaced ;)

PR #945, opened 2017-05-29
improve(admin): add an indicator on collapsible blocks

### LINKED ISSUE #915: Admin: visual feedback on collapsed items within the Quality block
capture d ecran 2017-05-24 a 11 36 49

E.g.: here the Description is collapsed but there is no visual feedback that it could be expanded to see details.

thimy:
capture d ecran 2017-05-29 a 11 28 14

Another way to do it would look like this: 
capture d ecran 2017-05-29 a 11 27 14
Plus and minus are a bit clearer, but they're a bit too big here because they're contained in a `h3`.
I believe `h3` is also too big though (bigger than the panel title), so it wouldn't be much of a problem if we diminish its size.

What do you prefer?

Fix #915

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,8 @@
   [#928](https://github.com/opendatateam/udata/issues/928)
     - Changed notification style to toast
     - Fix notifications that weren't displayed on form submission
+- Add a toggle indicator on dataset quality blocks that are collapsible
+  [#915](https://github.com/opendatateam/udata/issues/915)
 
 ## 1.0.11 (2017-05-25)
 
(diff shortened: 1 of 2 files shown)


noirbizarre:
I can't choose but:
- `+/-` is prettier
- chevrons are more consistent with other collapsible (ex: search panels).

To discuss: they were chevrons before but they were removed because the text is not always pertinent when the indicator is green.
**ex:** Questions/calls to actions are only relevant when the indicator is not green.

Maybe we should alter the panel content when the indicator is green (ie. simply remove the question ?)

davidbgk:
I would use chevrons because +/- can be interpreted as add/remove quantities in the given context.

odtvince:
I agree that there might be something to improve with the current convention for collapsible blocks.

The thing I like with the +/- proposition is that it stay light on the eye, but I am not totally sold to that solution for the following reasons:
* it can be mis-interpreted (as pointed by @davidbgk)
* it is only marginally more explicit than the current convention.

Therefore I would advocate to keep the chevrons until we find a better idea.

PR #948, opened 2017-05-30
improve(admin): fix datatable cell width

### LINKED ISSUE #922: Admin: titles are width-constrained on datasets block from organization pages
capture d ecran 2017-05-24 a 11 51 13

We should give more room to titles and less to other columns.

thimy:
Set a default width for dataset name and removed the `whitespace: nowrap` to prevent column to stretch regarding column title.

image

Fix #922

### DIFF
--- a/js/components/dataset/list.vue
+++ b/js/components/dataset/list.vue
@@ -21,7 +21,8 @@ export default {
                 key: 'title',
                 sort: 'title',
                 align: 'left',
-                type: 'text'
+                type: 'text',
+                width: 200
             }, {
                 label: this._('Creation'),
                 key: 'created_at',
--- a/js/components/datatable/table.vue
+++ b/js/components/datatable/table.vue
@@ -1,7 +1,15 @@
 <style lang="less">
 .datatable {
-    th {
-        white-space: nowrap;
+    thead tr th {
+        padding-right: 20px;
+        position: relative;
+        vertical-align: top;
+    }
+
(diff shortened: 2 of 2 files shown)


js/components/dataset/list.vue
```diff
@@ -21,7 +21,8 @@ export default {
                 key: 'title',
                 sort: 'title',
                 align: 'left',
-                type: 'text'
+                type: 'text',
+                width: 200
```
davidbgk: 1. What happens if we remove all fields’ `width`s? Doesn't it fit auto-magically?
2. What about using classes/vars for consistency across all admin tables?

thimy: 1. Removing all widths gives this: 
image
I feel better about not forcing sizes but I'm a bit bothered by the length of the third column, which was the initial issue.

2. Do you mean using classes for specific columns sizes? I agree this would be a better solution, this PR was meant as a quickfix but maybe we should rethink the datatable component then.

noirbizarre: I think we should bring back an older solution we had: having a grouped header `Update` with 2 nested columns: `Data` and `Metadata`.

noirbizarre: But I agree with @davidbgk : we shouldn't set a width on the column that should take all the empty space. Most of the time, titles are very long and span on 2 lines, with a fixed width on 200, they will span more.

PR #950, opened 2017-06-02
Improve/fix the init command

noirbizarre:
This PR improve the `udata init` command:
- migrate before trying to index
- search indexation/mapping reuse the `search init` command
- user is prompted for a superadmin user creation
- user is prompted for a default set of licenses to import
- user is prompted for some sample data creation
- does not prompt in a non-interactive environment

### DIFF
--- a/udata/commands/fixtures.py
+++ b/udata/commands/fixtures.py
@@ -1,49 +1,32 @@
 # -*- coding: utf-8 -*-
-
 from __future__ import unicode_literals
 
-from datetime import datetime
 import logging
 
 from udata.commands import manager
-from udata.core.dataset.factories import VisibleDatasetFactory, LicenseFactory
+from udata.core.dataset.factories import VisibleDatasetFactory
 from udata.core.discussions.factories import DiscussionFactory
-from udata.core.organization.factories import OrganizationFactory, TeamFactory
-from udata.core.reuse.factories import VisibleReuseFactory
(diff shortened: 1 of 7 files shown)


udata/commands/fixtures.py
```diff
@@ -1,49 +1,32 @@
...
-
 @manager.command
-def generate_fixtures():
+def generate_fixtures(datasets=5, reuses=1):
     '''Build sample fixture data (users, datasets and reuses).'''
-    user = AdminFactory(email='user@udata', password='password',
-                        confirmed_at=datetime.now())
-    log.info('Generated admin user "user@udata" with password "password".')
+    user = UserFactory()
+    log.info('Generated user "{user.email}".'.format(user=user))
```
abulte: No more password display? I guess this is because the "right way" is now to set your own password when prompted to create a super user. But what about a non-interactive install? Is it not interesting to display the password in this case (docker install)?

davidbgk: You can set your password afterwards but I don't know how simple it is with docker.

noirbizarre: Yes, exactly, for non interactive, you should use a dedicated command afterwards.
BTW, I think most of the commands need a coherence/consistence pass to:
- always have the same signature pattern (prompt vs args, switch/param for promptables...)
- ensure proper non-interactive environments handling
- always have the same display format

Issue #928, opened 2017-05-24
Admin: feedback missing on form validation


thimy:
## Detailed Description
When editing my own profile for example, saving changes redirects to profile page but we don't get any message about saving.

## Context
Users should get a feedback when modifying informations, the page behaviour is the same wether they save or cancel their changes.
We might need to check all form validation.

## Possible Implementation
Show a callout on saved change.

abulte:
Ok on `master` as of today.

<img width="296" alt="capture d ecran 2017-11-13 14 47 20" src="https://user-images.githubusercontent.com/119625/32728700-998ba426-c881-11e7-9c75-a36a98e4a3a4.png">

Issue #753, opened 2017-02-09
Establish some vue component naming convention


noirbizarre:
## Detailed Description
Currently there is no Vue component naming convention and some inconsistency. ex:

```
IssueList  < issues/list
DiscussionList < discussions/list
FollowerList < follow/list
…
DatasetCards < dataset/card-list
...
FormHorizontal < form/horizontal-form.vue
```

## Context
As a developper, I want a clear and consistent naming convention so:
- I can create new component with consisten naming
- I can find existing components easily
- Search and find in code is effective

## Possible Implementation
To be defined

abulte:
Migrating this to https://github.com/opendatateam/udata/issues/1312

PR #955, opened 2017-06-06
fix and improve(admin): notifications

thimy:
Changed notification style to toast 🍞
Fixed the notifications to be displayed on non-wizard 🎩 layout. (https://github.com/opendatateam/udata/issues/928)

image

### DIFF
--- a/js/components/dataset/resource/form.vue
+++ b/js/components/dataset/resource/form.vue
@@ -101,6 +101,7 @@ export default {
             type: Object,
             default() {return new Resource()}
         },
+        hideNotifications: false
     },
     data() {
         return {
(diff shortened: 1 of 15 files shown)


abulte:
@thimy does Vue 1 supports mixin? The `validate()` method seems very repetitive, maybe it can be mutualized in a mixin. My 2 cents, I obviously don't know the code and/or best practices for the project at all at this point ;-)

js/components/dataset/resource/form.vue
```diff
@@ -101,6 +101,7 @@ export default {
             type: Object,
             default() {return new Resource()}
         },
+        notifications: true
```
noirbizarre: This is the only one with positive variable (ie. `notifications` vs `hideNotifications`).
We should stay consistent and chose between one of the naming (personnaly I prefer `notifications` over `hideNotifications`)

thimy: Oops, forgot this one!

noirbizarre:
@abulte Yes, Vue 1.x support mixins.
I think the validation pattern is not the focus of this PR and should be handled in a separate PR (see the #526 EPIC, seems to fit in "jquery-validation (admin form validation)").
About conventions, we have some work to do like this #753 :wink:

thimy:
@abulte I'm not sure, not exactly acquainted to Vue yet, I'll look into this. 👍

thimy:
Notifications were indeed displayed in modals as well, fixed along with variable name.

PR #960, opened 2017-06-08
Escape HTML string for JS use in form description

### LINKED ISSUE #957: Login: "Forgot password" display bug
## Detailed Description
Fix the description of the password field.

capture d ecran 2017-06-08 12 20 47

## Your Environment
* Chrome 59
* macOS

abulte:
Fix #957 

This is probably overkill since from what I can see those descriptions are never displayed? Please advise ;-)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@
     - Fix notifications that weren't displayed on form submission
 - Add a toggle indicator on dataset quality blocks that are collapsible
   [#915](https://github.com/opendatateam/udata/issues/915)
+- Fix login form password help display [#957](https://github.com/opendatateam/udata/issues/957)
 
 ## 1.0.11 (2017-05-25)
 
--- a/udata/frontend/helpers.py
+++ b/udata/frontend/helpers.py
@@ -350,3 +350,9 @@ def to_json(data):
 def format_number(number):
     '''A locale aware formatter.'''
     return format_number_babel(number, locale=g.lang_code)
+
+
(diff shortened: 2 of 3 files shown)


abulte:
Possible other/further solutions:
- remove description display altogether for those forms (flask-security forms)
- display the description inline under the field (in this case we'll have a duplicate "forgot password" CTA on login)
- use `form-help` javascript tooltip like I've seen it on other parts of the site (admin)

I also find the duplicate fieldset / field label a little strange, maybe it's a good occasion to refactor those forms.

capture d ecran 2017-06-09 10 14 11

noirbizarre:
:+1: for the factoring, let's discuss this

davidbgk:
Escaping HTML to put it in data attributes of… HTML 🙃

Do we really need that help? Maybe `{% if f.description and '<' not in f.description %}` (your first proposal) is more pertinent here? On the screenshot it looks redundant to me.

abulte:
No I do not think we need that (escaping). It's over engineered and useless in the context. The filter could come in handy though ;-) The best way IMHO is to refactor those forms and deal with the descriptions at this moment - remove them or display them inline?

abulte:
PI they don't even display the field descriptions in the default `Flask-Security` templates: https://github.com/mattupstate/flask-security/blob/develop/flask_security/templates/security/_macros.html

davidbgk:
+1 for removing them.

abulte:
Replaced by #971.

PR #964, opened 2017-06-09
Fix #894 - badge emails and signals

### LINKED ISSUE #894: Badges notification mails are not sent (signals are not dispatched)
## Detailed Description
Badges notification mails are not sent because the badges signals ([`on_badge_added`](https://github.com/opendatateam/udata/blob/master/udata/core/badges/signals.py#L9) and [`on_badge_removed`](https://github.com/opendatateam/udata/blob/master/udata/core/badges/signals.py#L9) are never dispatched.

## Context
#885 and #887  highlighted the fact that badges metrics were not in use because signals are never dispatched.
Notifications mail are connected on the `on_badge_added` and so are never sent too.

## Possible Implementation
Dispatch the signals were they should be (on badges addition and removal)

## Your Environment
udata 1.0.x

abulte:
Fix #894

In addition to fixing email sending, the system is now more easily pluggable/extendable: it's easy to create a notification for a custom by subscribing to the `on_badge_added` and `on_badge_removed` signals.


udata/core/badges/tasks.py
```diff
@@ -23,20 +23,57 @@ def call_task(badge, **kwargs):
...
+    sender
+        The object that emitted the event.
+    the_badge
+        The `Badge` object awarded.
+        NB: `badge` is a reserved param name in Blinker.
+    '''
+    if the_badge.kind == PUBLIC_SERVICE and isinstance(sender, Organization):
+        recipients = [member.user for member in sender.members]
+        subject = _(
+            'Your organization "%(name)s" has been identified as public service',
```
davidbgk: The subject does not contain `a public service` as in emails body. Not sure which one is correct.

udata/core/badges/models.py
```diff
@@ -84,16 +86,20 @@ def add_badge(self, kind):
...
     def remove_badge(self, kind):
         '''Perform an atomic removal for a given badge'''
+        badge = self.get_badge(kind)
         self.update(__raw__={
             '$pull': {
                 'badges': {'kind': kind}
             }
         })
         self.reload()
+        on_badge_removed.send(self, the_badge=badge)
```
davidbgk: Is that signal in use somewhere?

udata/tests/api/test_organizations_api.py
```diff
@@ -682,22 +684,59 @@ def test_list(self):
...
+        self.assertListEqual([m.recipients[0] for m in mails], members_emails)
+
+    def test_create_same(self):
+        data = self.factory.attributes()
+        with self.api_user():
+            with self.assert_emit(on_badge_added):
+                self.post(
+                    url_for('api.organization_badges', org=self.organization),
+                    data)
+            with self.assert_not_emit(on_badge_added):
```
davidbgk: Might be handy to test that no email has been sent in that case.

udata/core/badges/models.py
abulte: Nope. Good for "plugability" though?

udata/tests/api/test_organizations_api.py
abulte: 👍

udata/core/badges/tasks.py
abulte: Probably `public service` (https://en.wikipedia.org/wiki/Public_service). Thanks.

abulte:
@noirbizarre is this OK now? Thanks.

udata/tests/api/test_organizations_api.py
```diff
@@ -682,22 +684,63 @@ def test_list(self):
     def test_create(self):
         data = self.factory.attributes()
         with self.api_user():
-            response = self.post(
-                url_for('api.organization_badges', org=self.organization),
-                data)
+            with self.assert_emit(on_badge_added):
```
noirbizarre: I think you can wrap this in a single context manager (if it pass the 80 chars constraint)

abulte: 👍

PR #967, opened 2017-06-12
Added License.guess

noirbizarre:
This PR adds a `License.guess()` helper to improve license matching while harvesting.
It performs many checks to find a license:
- exact match by url
- exact match normalized name
- exact match by opendefinition ID
- fallback on string matching with a Damerau-Levenshtein distance (case of typo)

### DIFF
--- a/udata/core/dataset/models.py
+++ b/udata/core/dataset/models.py
@@ -6,6 +6,7 @@ from datetime import datetime, timedelta
 from collections import OrderedDict
 
 from blinker import signal
+from stringdist import rdlevenshtein
 from flask import url_for
 from mongoengine.signals import pre_save, post_save
 from mongoengine.fields import DateTimeField
@@ -87,6 +88,11 @@ DEFAULT_CHECKSUM_TYPE = 'sha1'
 PIVOTAL_DATA = 'pivotal-data'
 CLOSED_FORMATS = ('pdf', 'doc', 'word', 'xls', 'excel')
 
+# Maximum acceptable Damerau-Levenshtein distance
+# used to guess license
+# (ie. number of allowed character changes)
+MAX_DISTANCE = 2
+
 
 class License(db.Document):
     # We need to declare id explicitly since we do not use the default
@@ -103,6 +109,28 @@ class License(db.Document):
 
     def __unicode__(self):
         return self.title
+
+    @classmethod
+    def guess(cls, text):
+        '''
+        Try to guess license from a string.
+
+        Try to exact match on identifier then slugified title
+        and fallback on edit distance ranking (after slugification)
+        '''
+        qs = cls.objects
+        text = text.strip().lower()  # Stored identifiers are lower case
+        slug = cls.slug.slugify(text)  # Use slug as it normalize string
+        license = qs(db.Q(id=text) | db.Q(slug=slug) | db.Q(url=text)).first()
+        if license is None:
+            # Try to single match with a low Damerau-Levenshtein distance
+            computed = ((l, rdlevenshtein(l.slug, slug)) for l in cls.objects)
+            candidates = [l for l, d in computed if d <= MAX_DISTANCE]
+            # If there is more that one match, we cannot determinate
+            # which one is closer to safely choose between candidates
+            if len(candidates) == 1:
+                license = candidates[0]
+        return license
 
 
 class DatasetQuerySet(db.OwnedQuerySet):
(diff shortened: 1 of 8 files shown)


udata/core/dataset/models.py
```diff
@@ -87,6 +88,10 @@
 PIVOTAL_DATA = 'pivotal-data'
 CLOSED_FORMATS = ('pdf', 'doc', 'word', 'xls', 'excel')
 
+# Maximum acceptable Damerau-Levenshtein distance
```
davidbgk: Isn't it related to the length of license codes‘ strings?

udata/core/dataset/models.py
```diff
@@ -104,6 +109,26 @@ class License(db.Document):
...
+        '''
+        qs = cls.objects
+        text = text.strip().lower()  # Stored identifiers are lower case
+        slug = cls.slug.slugify(text)  # Use slug as it normalize string
+        license = qs(db.Q(id=text) | db.Q(slug=slug) | db.Q(url=text)).first()
+        if license is None:
+            # Try to single match with a low Damerau-Levenshtein distance
+            computed = ((l, rdlevenshtein(l.slug, slug)) for l in cls.objects)
+            candidates = [l for l, d in computed if d <= MAX_DISTANCE]
+            if len(candidates) == 1:
```
davidbgk: What if length > 1, why do we take the first if not sorted?

udata/tests/dataset/test_dataset_model.py
```diff
@@ -235,3 +235,62 @@ def test_url_is_stripped(self):
...
+        license = LicenseFactory(title='xx yy')
+        found = License.guess('xx  yy')  # 2 spaces instead of 1
+        self.assertIsInstance(found, License)
+        self.assertEqual(license.id, found.id)
+
+    def test_match_by_title_with_mismatching_case(self):
+        license = LicenseFactory(title='x' * 6)
+        found = License.guess('xXxXxX')
+        self.assertIsInstance(found, License)
+        self.assertEqual(license.id, found.id)
```
davidbgk: Please add tests with real typos in licenses.

udata/core/dataset/models.py
noirbizarre: No, it's the number of operations on a single string to go to another.
It's not related to the length but I think you refer to elasticsearch which is doing progressive Levenshtein distance: do not allow typo on low string, allow 1 on medium string and 2 on large string. In our case the corpus is known, they all have title with a significant length to allow 2 typos and not to be close of another license name.

udata/tests/dataset/test_dataset_model.py
noirbizarre: What does it change? Test are written to isolate a single case the string is chosen to specificaly highlight this case (in this one, that case changes does not impact the match)

udata/core/dataset/models.py
noirbizarre: (Added anextra comment line)

udata/core/dataset/models.py
```diff
@@ -104,6 +110,28 @@ class License(db.Document):
...
+        qs = cls.objects
+        text = text.strip().lower()  # Stored identifiers are lower case
+        slug = cls.slug.slugify(text)  # Use slug as it normalize string
+        license = qs(db.Q(id=text) | db.Q(slug=slug) | db.Q(url=text)).first()
+        if license is None:
+            # Try to single match with a low Damerau-Levenshtein distance
+            computed = ((l, rdlevenshtein(l.slug, slug)) for l in cls.objects)
+            candidates = [l for l, d in computed if d <= MAX_DISTANCE]
+            # If there is more that one match, we cannot determinate
+            # which one is closer to safely chose between candidates
```
davidbgk: `choose`

davidbgk: `so we arbitrary take the first one`?

noirbizarre: No, we don't and let the license unknown.
We don't chose a license for the producer because if we chose something wrong (ie. a license that is more permissive that the original one), we are legaly responsible.
Example: there is some bug differences between cc-by-sa 2.0, 3.0 et 4.0 but if someone type cc-by-sa 5.0 (which doesn't exists), I don't want to choose arbitrary between the different versions.

This is a first implementation but we already discussed some possible improvements for some other existing case:
- store a list of known synonyms for each license (ie. License Etalab for License ouverte)
- have some rule for license version handling (ie. CC-BY-SA without version is latest 4.0 version)

davidbgk: Alright got it, my bad :)

noirbizarre:
Already done ;)

PR #968, opened 2017-06-12
Unlocalized route on I18nBlueprint

noirbizarre:
This PR allows to define some non-localized URLs (ie. with language prefix) on `I18nBlueprint`.

There are many place in the application where have defined 2 blueprints for a single set of views, a classic `Blueprint` and a localized `I18nBlueprint`.

This PR allows to perform all this on a single i18nBlueprint and also adds the reverse redirect for unlocalized URLs (ie. `/<lang>/unlocalized/endpoint` redirect to `/unlocalized/endpoint`).

### DIFF
--- a/udata/i18n.py
+++ b/udata/i18n.py
@@ -158,8 +158,11 @@ def init_app(app):
 
 
 def _add_language_code(endpoint, values):
-    if not (endpoint.endswith('.static') or endpoint.endswith('_redirect')):
-        values.setdefault('lang_code', g.get('lang_code', default_lang))
+    try:
+        if current_app.url_map.is_endpoint_expecting(endpoint, 'lang_code'):
+            values.setdefault('lang_code', g.get('lang_code', default_lang))
+    except KeyError:  # Endpoint does not exist
+        pass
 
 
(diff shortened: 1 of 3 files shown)


udata/i18n.py
```diff
@@ -158,8 +158,11 @@ def init_app(app):
 
 
 def _add_language_code(endpoint, values):
-    if not (endpoint.endswith('.static') or endpoint.endswith('_redirect')):
-        values.setdefault('lang_code', g.get('lang_code', default_lang))
+    try:
+        if current_app.url_map.is_endpoint_expecting(endpoint, 'lang_code'):
+            values.setdefault('lang_code', g.get('lang_code', default_lang))
+    except KeyError:  # Endpoint does not exists
```
davidbgk: No `s` at the end (but a point?)

davidbgk:
I wonder if it can be used for territories given that it doesn't look relevant to have `/en/` counterpart. It would allow to have `/territoires/` :wink:

noirbizarre:
I think in this case, it would be easier to have a raw `flask.Blueprint` in gouvfr for this case :+1:

PR #969, opened 2017-06-12
Allow resources to store extra metadata too

noirbizarre:
This PR allows resources to store extra metadata in the `extras` attribute like `Dataset.extras`.

Main use case: allows harvesters to store the remote resource identifier to improve the update behavior.


udata/core/dataset/api_fields.py
```diff
@@ -62,6 +62,7 @@
         attribute='modified', readonly=True,
         description='The resource last modification date'),
     'metrics': fields.Raw(description='The resource metrics', readonly=True),
+    'extras': fields.Raw(description='Extras attributes as key-value pairs'),
```
davidbgk: I think there is no need of `s` for `Extras` in description.

PR #976, opened 2017-06-14
Mention lzma dependency for macOS

oncletom:
Otherwise the following error would be shown:

```
  In file included from src/liblzma.c:1:
  src/liblzma.h:24:10: fatal error: 'lzma.h' file not found
  #include <lzma.h>
           ^
  1 error generated.
  error: command 'clang' failed with exit status 1

  ----------------------------------------
  Failed building wheel for pyliblzma
```

### DIFF
--- a/docs/local-dependencies.md
+++ b/docs/local-dependencies.md
@@ -5,6 +5,13 @@
 It is recommended to work within a virtualenv to ensure proper dependencies isolation.
 If you're not familiar with that concept, read [Python Virtual Environments - a Primer][].
 
+!!! note
+    On macOS, the library `lzma.h` has to be installed prior to running the next block of commands.
+
+    ```shell
+    $ brew install xz
+    ```
+
 Alright, now you can [install virtualenv][install-virtualenv] and then type these commands knowing what you are doing:
 
 ```shell


docs/local-dependencies.md
```diff
@@ -5,6 +5,13 @@
 It is recommended to work within a virtualenv to ensure proper dependencies isolation.
 If you're not familiar with that concept, read [Python Virtual Environments - a Primer][].
 
+!!! note
+    On macOS, the library `lzma.h` has to be installed prior to running the next block of commands.
+
+    ```shell
+    $ brew install xz
```
davidbgk: Maybe mention homebrew somewhere?

oncletom: @davidbgk even if it is already mentioned in the _System Dependencies_ page?

noirbizarre:
I think this is already done in the system dependencies section https://github.com/opendatateam/udata/blob/dev/docs/system-dependencies.md#user-content-osxhomebrew

oncletom:
@noirbizarre ah, I skipped that bit thinking the Docker setup was already enough. My bad! I guess we can close the PR then 👍

PR #980, opened 2017-06-15
Middleware URLs: doc and harmonization

### LINKED ISSUE #881: Specify login/password in udata.cfg when using ES, mongo and redis
## Detailed Description

I want to get `udata` running on Heroku.
Heroku gives me access to `ES`, `mongo` and `redis`, as long as I am able to connect to URLs such as:

```
mongodb://$LOGIN:$PASSWORD@$HOST:$PORT/$LOGIN
redis://$LOGIN:$PASSWORD@$HOST:$PORT/$LOGIN
https://$LOGIN:$PASSWORD@$HOST:$PORT/$LOGIN
http://$LOGIN:$PASSWORD@$HOST:$PORT/$LOGIN
```

Please could you make `udata.cfg` compatible to these URLs format?

abulte:
- document the use of RFC-1738 formatted URLs for middleware configurations (with authentication support)
- use `MONGODB_SETTINGS` with an URL instead of `MONGODB_*` settings

Fix #881

### DIFF
--- a/docs/adapting-settings.md
+++ b/docs/adapting-settings.md
@@ -134,14 +134,9 @@ ELASTICSEARCH_URL = 'http://<user>:<password>@<host>:<port>'
 
 ## Mongoengine/Flask-Mongoengine options
 
-### MONGODB_SETTINGS
+### MONGODB_HOST
 
-**default**:
-```python
-MONGODB_SETTINGS = {
-    'host': 'mongodb://localhost:27017/udata'
-}
-```
+**default**: `'mongodb://localhost:27017/udata'`
 
 The mongodb database used by udata.
 During tests, the test database will use the same name suffixed by `-test`
@@ -152,9 +147,7 @@ for more details.
 Authentication is also supported in the URL:
 
 ```python
-MONGODB_SETTINGS = {
-    'host': 'mongodb://<user>:<password>@<host>:<port>/<database>'
-}
+MONGODB_HOST = 'mongodb://<user>:<password>@<host>:<port>/<database>'
 ```
 
 ## Celery options
--- a/docs/adapting-settings.md
+++ b/docs/adapting-settings.md
@@ -129,7 +129,7 @@ ELASTICSEARCH_URL = 'elasticserver:9200'
 RFC-1738 formatted URLs are also supported:
 
 ```python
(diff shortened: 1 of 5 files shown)


docs/adapting-settings.md
```diff
@@ -126,30 +126,37 @@ The elasticsearch server url used for search indexing.
...
 
 See [the official Flask-MongoEngine documentation][flask-mongoengine-doc]
 for more details.
 
+Authentication is also supported in the URL:
+
+```python
+MONGODB_SETTINGS = {
+    'host': 'mongodb://<user>:<password>@<host>:<port>/<database>'
+}
```
noirbizarre: I think for a single parameter the `MONGODB_HOST` is cleaner than a single property in a dict

udata/settings.py
```diff
@@ -16,9 +16,9 @@ class Defaults(object):
     CONTACT_EMAIL = 'contact@example.org'
     TERRITORIES_EMAIL = 'territories@example.org'
 
-    MONGODB_HOST = 'localhost'
-    MONGODB_PORT = 27017
-    MONGODB_DB = 'udata'
+    MONGODB_SETTINGS = {
+        'host': 'mongodb://localhost:27017/udata'
+    }
```
noirbizarre: Use the `MONGODB_HOST` version

udata/models/__init__.py
```diff
@@ -101,8 +103,13 @@ def resolve_model(self, model):
 
 
 def init_app(app):
+    # use `{database_name}-test` database for testing
     if app.config['TESTING']:
-        app.config['MONGODB_DB'] = '{MONGODB_DB}-test'.format(**app.config)
+        parsed_url = urlparse(app.config['MONGODB_SETTINGS']['host'])
```
noirbizarre: Use the `MONGODB_HOST`

abulte:
@noirbizarre Yes, you're right. It's not the encouraged way of `Flask-Mongoengine` but it does feel cleaner. And it works, which I was not aware of ;-)

noirbizarre:
Yes, it's at the end of the configuration part: http://flask-mongoengine.readthedocs.io/en/latest/
I don't know why they prefer to user the dict form as default because to me it has many disavantages

PR #985, opened 2017-06-19
Fix frontend modal not being openable twice

noirbizarre:
This PR factorizes modal behavior and fix frontend modals not being openable twice.

### DIFF
--- a/js/mixins/modal.js
+++ b/js/mixins/modal.js
@@ -1,0 +1,24 @@
+import Vue from 'vue';
+
+/**
+ * A mixin holding the modal container Behavior.
+ * @type {Object}
+ */
+export default {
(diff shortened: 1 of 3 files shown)


js/mixins/modal.js
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * A mixin holding the modal container Behavior.
+ * @type {Object}
+ */
+export default {
+    methods: {
+        /**
+         * Insert a modal Vue in the application.
+         * @param  {Object} options     The modal component definition (options passed to Vue.extend())
+         * @param  {Object} data        Data to assign to modal properties
```
abulte: Should be `propsData`?

PR #990, opened 2017-06-22
Restore Read The Docs search

oncletom:
> It appears that the form action for the searchbox is being injected with this javascript: https://github.com/rtfd/readthedocs.org/blob/master/readthedocs/core/static-src/core/js/doc-embed/mkdocs.js

cf. rtfd/readthedocs.org#1088

refs docksal/docksal#120

### DIFF
--- a/docs/custom_theme/searchbox.html
+++ b/docs/custom_theme/searchbox.html
@@ -1,0 +1,5 @@
+<div role="search">
+  <form id="rtd-search-form-alt" class="wy-form" action="{{ base_url }}/search.html" method="get">
+    <input type="text" name="q" placeholder="Search docs" />
+  </form>
+</div>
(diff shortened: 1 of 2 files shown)


docs/custom_theme/searchbox.html
```diff
@@ -0,0 +1,5 @@
+<div role="search">
+  <form id ="rtd-search-form-alt" class="wy-form" action="{{ base_url }}/search.html" method="get">
```
noirbizarre: Extra space to remove after `id`

oncletom:
@noirbizarre it does work locally (still) — and hopefully it should work (again) on Read the Docs, as it worked on websites which deployed it. If it does not work with the `dev` branch on Read the Docs, we could revert the commit.

noirbizarre:
Dev branch is not deployed on readthedocs :/
But let's do it, it will when we'll merge dev on master (hopefully today :wink: )

oncletom:
And… it works! 🎉

PR #994, opened 2017-06-23
Redirects for old towns/counties/regions, fix #991

### LINKED ISSUE #991: Territories: redirections from old URL are not working properly
For instance http://demo.data.gouv.fr/fr/territory/county/13/ is not redirected.

Should redirect to http://demo.data.gouv.fr/fr/territories/departement/13@1860-07-01/Bouches-du-Rhone/

Probably related to the fact county is now departement.

davidbgk:


udata/features/territories/views.py
```diff
@@ -54,6 +54,33 @@ def redirect_town(code):
                             level='commune', code=code))
 
 
+@blueprint.route('/territory/town/<code>/', endpoint='town2')
```
noirbizarre: If the `endpoint` is not significant (ie. not called in other functions), just omit it. There is a sane default to the function/method name.

PR #996, opened 2017-06-23
improve(search): remove useless class

thimy:
Improves the searchbar's consistency.


noirbizarre:
Did you check that this does not have impact on lower resolutions (mobile, tablet in the chrome or firefox debugger) ?

thimy:
image
Does it look ok to you?

PR #998, opened 2017-06-23
improve(breadcrumb): fix alignment to container

thimy:
## Before
image

## After 
image

The breadcrumb should now be aligned with the global container and not be slightly on the left as it was before.


noirbizarre:
Did you test on lower resolutions (mobile, tablet in the chrome or firefox debugger) ?

PR #999, opened 2017-06-23
Search: Display figures inline in tabs

oncletom:
Also, advocate for a CSS class semantic which reflects it is not a breadcrumb but a search bar.

screen shot 2017-06-24 at 00 24 05

# Before / After

searchbar gif

### DIFF
--- a/udata/templates/search.html
+++ b/udata/templates/search.html
@@ -24,41 +24,31 @@
     {% if datasets %}
         <li class="active" data-tab="datasets">
             <a href @click.prevent="setTab('datasets')">
-                {{ _('Datasets')}}
-                <span class="badge">{{ datasets.total }}</span>
+                {{ ngettext('%(num)d dataset', '%(num)d datasets',datasets.total)}}
             </a>
         </li>
     {% endif %}
     {% if reuses %}
         <li{% if not datasets %} class="active"{% endif %} data-tab="reuses">
             <a href @click.prevent="setTab('reuses')">
-                {{ _('Reuses') }}
-                <span class="badge">{{ reuses.total }}</span>
+                {{ ngettext('%(num)d reuse', '%(num)d reuses', reuses.total) }}
             </a>
         </li>
     {% endif %}
     {% if organizations %}
         <li{% if not datasets and not reuses %} class="active"{% endif %} data-tab="organizations">
             <a href @click.prevent="setTab('organizations')">
-                {{ _('Organizations') }}
-                <span class="badge">{{ organizations.total }}</span>
+                {{ ngettext('%(num)d organization', '%(num)d organizations', organizations.total) }}
             </a>
         </li>
     {% endif %}
     {% if territories %}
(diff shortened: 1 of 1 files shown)


udata/templates/search.html
```diff
@@ -16,31 +16,29 @@
...
 {% block breadcrumb_bar %}
 <!-- Nav tabs -->
-<ul class="nav nav-tabs search-tabs">
+<ul class="nav nav-tabs search-tabs nav-tabs--search-tabs">
     {% if datasets %}
         <li class="active" data-tab="datasets">
             <a href @click.prevent="setTab('datasets')">
-                {{ _('Datasets')}}
-                <span class="badge">{{ datasets.total }}</span>
+                {{ _('%d Datasets' % datasets.total)}}
```
oncletom: Maybe there is a better way to pass translation arguments?

noirbizarre: Yes, you can (should) use `ngettext` which allows to have singular/plural support and write:
```html+jinja
{{ ngettext('%(num)d dataset', '%(num)d datasets', datasets.total) }}
```
See http://jinja.pocoo.org/docs/2.9/extensions/#newstyle-gettext for details

oncletom:
Thanks @noirbizarre, very helpful! I tried to simplify the tab display based on the territories count in e8929c4. Feel free to tell me if it is a good idea or not 🙂 

Also, I somewhat expected `N_` to be available as I saw it being aliased in `i18n.py`. It does not seem to be available in the jinja templates though. Is that expected?

udata/templates/search.html
```diff
@@ -16,48 +16,39 @@
...
-                    <span class="badge">{{ territories.0.name }}</span>
-                </a>
-            {% else %}
-                <a href @click.prevent="setTab('territories')">
-                    {{ _('Territories') }}
-                    <span class="badge">{{ total }}</span>
-                </a>
-            {% endif %}
+            <a href="{% if total==1 %}{{ territories.0.url }}{% endif %}" {% if total>1%}@click.prevent="setTab('territories'){% endif %}>
+                {{ ngettext('Territory: %(name)s', '%(num)d territories:', name=territories.0.name, total) }}
```
noirbizarre: Smart !!

udata/templates/search.html
```diff
@@ -16,48 +16,39 @@
...
-                    {{ _('Territory:') }}
-                    <span class="badge">{{ territories.0.name }}</span>
-                </a>
-            {% else %}
-                <a href @click.prevent="setTab('territories')">
-                    {{ _('Territories') }}
-                    <span class="badge">{{ total }}</span>
-                </a>
-            {% endif %}
+            <a href="{% if total==1 %}{{ territories.0.url }}{% endif %}" {% if total>1%}@click.prevent="setTab('territories'){% endif %}>
```
noirbizarre: Given we already are in a `{% if territories %}` (ie. total > 0), I think the '{% endif %} {% if total >1 %}` can be even more simplfied in a `{% else %}` ;)

oncletom:
@noirbizarre changes made! 🙂

PR #1003, opened 2017-06-26
Make placeholders presents only in theme (fix #707)

### LINKED ISSUE #707: Make placeholders presents only in theme
## Detailed Description
Placeholders are split between themes and core. It should be consistent and I propose to make it a themable element (so to put them in the themes).

## Context
When customizing uData , I want to make the placeholders consistent with my theme.
When coding on uData , I don't want to search and ask questions on how placeholders are handled.

## Possible Implementation
Make the placeholders a themable element. **Optionnaly** provide a fallback on core mecanism.

## Your Environment
Version 1.0

noirbizarre:
This PR ensures coherence on placeholders handling.
This is now fully a theme responsibility.


CHANGELOG.md
```diff
@@ -87,6 +87,8 @@
 - Integrating latest versions of GeoZones and GeoLogos for territories.
   Especially using history of towns, counties and regions from GeoHisto.
   [#499](https://github.com/opendatateam/udata/issues/499)
+- Placeholders are now entirely provided by themes
```
abulte: Maybe explicit that we're talking about *images* placeholders, I was confused, vs input fields placeholders.

PR #1004, opened 2017-06-26
Entrypoint-based harvesters

noirbizarre:
This PR makes harvester declaration entrypoint-based, like themes are.

This avoid using a global dictionnary to stora harvesters.
The documentation has been adapted (and a cookiecutter is coming [here](https://github.com/opendatateam/cookiecutter-udata-harvester)

### DIFF
--- a/docs/harvesting.md
+++ b/docs/harvesting.md
@@ -165,12 +165,10 @@ from __future__ import unicode_literals
 from udata.models import db, Resource
 from udata.utils import faker
 
-from . import BaseBackend, register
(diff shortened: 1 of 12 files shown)


docs/harvesting.md
```diff
@@ -214,6 +212,29 @@ class RandomBackend(BaseBackend):
 
 ```
 
+You need to properly expose the harvester as a `udata.harvesters` entrypoint in you `setup.py`:
```
abulte: you -> your

PR #1007, opened 2017-07-03
Fix avatar placeholders

### LINKED ISSUE #1006: Avatar placeholders not displayed
The avatar placeholders are not being displayed at various locations:
- [ ] header bar on front default theme for a user
- [ ] profile view in admin for a user
- [ ] profile view in admin for an organization

abulte:
Fix #1006

### DIFF
--- a/udata/frontend/helpers.py
+++ b/udata/frontend/helpers.py
@@ -131,26 +131,24 @@ def obfuscate(email):
 
 
 @front.app_template_filter()
-@front.app_template_global()
-def avatar_url(obj, size):
+@contextfilter
+def avatar_url(ctx, obj, size):
     if hasattr(obj, 'avatar') and obj.avatar:
         return obj.avatar(size)
     elif hasattr(obj, 'logo') and obj.logo:
         return obj.logo(size)
     else:
-        return placeholder(None, 'user')
+        return placeholder(ctx, None, name='user')
 
 
-@front.app_template_global()
-@front.app_template_filter()
-def owner_avatar_url(obj, size=32):
+def owner_avatar_url(ctx, obj, size=32):
     if hasattr(obj, 'organization') and obj.organization:
         return (obj.organization.logo(size)
                 if obj.organization.logo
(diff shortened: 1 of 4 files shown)


udata/frontend/helpers.py
```diff
@@ -132,25 +132,27 @@ def obfuscate(email):
 
 @front.app_template_filter()
 @front.app_template_global()
-def avatar_url(obj, size):
+@contextfilter
```
noirbizarre: I think we need to split in two in order to be able to use `@contextfilter` and `@contextfunction`, unless:
- both decorators works on a same function
- we try to stick to the global method or the filter everywhere

abulte: You mean create two distinct filters? Like a new `avatar_url_context`?

abulte:
Actually `owner_avatar_url` is not called as a filter anywhere, we might turn it into a function.

noirbizarre:
Yes, I just wanted to point the issue with the build: a context is required for placeholder (to access the current theme) and jinja expose 2 ways to pass the context:
- a `@contextfilter` decorator for filters which only pass context when used as a filter
- a `@contextfunction` decorator for function which only pass context when used as a function

Yes, we agree this needs a signature change, but the tricky part is not the signature change but how to feed the first context parameter.

*Ideal case 1*: we use these functions as both template filters and global functions but hopefully having both `@contextfilter` and `@contextfunction` on a same function is working in both case.
*Case 2*: we only use these functions as filters => we remove the `@front.app_template_global()` and keep the `@contextfilter` and the `@front.app_template_filter()`.
*Case 3*: we only use these functions as global functions => we remove the `@front.app_template_filter()` and keep the `@contextfunction` and the `@front.app_template_global()`.
*Worst case 4*: we use these functions as both template filters and global functions but having both `@contextfilter` and `@contextfunction` on a same function is not working => we either need to split these function in 2 separates ones or ensure we are only using template function or filter, not both.

abulte:
@noirbizarre it looks like case 2!

PR #1019, opened 2017-07-10
fix(search): populate Search field with user provided value

### LINKED ISSUE #1016: Homepage Search input is cleared out when user has already started typing something
## Detailed Description

While the homepage is still loading but well displayed, I start to type a search query. If the page load finishes in the meantime, it will discard anything the user would have typed in the meantime.

Behaviour is illustrated by the video below.

bug gif

## Context

When I type a query in the search field , I want to lose the content of the input field and be able to start a search when I am done, whatever the state of the component is.

## Possible Implementation

Not sure — make sure upgrading the Vue component does not affect the user experience.

## Your Environment

* Browser Name and version: Chrome evergreen
* Operating System and version (desktop or mobile): macOS 10.something

oncletom:
This pull request populates the _query_ field when the Vue component is initialised.

# The Good

It works.

# The Bad

I have not found any particular documentation for `$options.el` so it might break when upgrading Vue.js.

There are probably better ways to do:

- passing a `value="..."` property to the search field as a prop (could not figure out how to pass the paramter to the search macro)
- maybe a more official way to bind to the original input#search field

# The Ugly

The Vue component relies on the DOM of the original so it can break just by changing the HTML template.

# What's next?

Once this issue is fixed, I would like to remove the code duplication between the HTML view and the Vue component. Either by extending the form and providing only the autocomplete feature.

fix #1016

### DIFF
--- a/js/components/site-search.vue
+++ b/js/components/site-search.vue
@@ -63,11 +63,18 @@ export default {
         size: String,
         territoryId: String,
     },
+    created() {
+        const {query} = this.$location;
+
+        const originalField = this.$options.el.querySelector('#search');
+        const queryStringValue = decodeURIComponent(query.q.replace('+', ' '));
+
+        this.query = originalField.value || queryStringValue || '';
+    },
     data() {
-        const query = this.$location.query.q || '';
         return {
             current: -1,
-            query: decodeURIComponent(query.replace('+', ' ')),
+            query: null,
             show: false,
             cache: new Cache('site-search', sessionStorage),
             groups: [
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Display organization metrics in the organization page tab labels [#1022](https://github.com/opendatateam/udata/pull/1022)
 - Organization dashboard page has been merged into the main organization page [#1023](https://github.com/opendatateam/udata/pull/1023)
+- Fix an issue causing a loss of data input at the global search input level [#1019](https://github.com/opendatateam/udata/pull/1019)
 
 ## 1.1.0 (2017-07-05)
 


oncletom:
@abulte yeah, exactly. I kind of wanted to start the conversation with the most obvious and minimal change prior to jump in rewriting the component ;-)

I'll be away in user interviews with @thimy tomorrow — but should be online between 2pm and 3:30pm 🙂

abulte:
@oncletom :+1:

abulte:
@oncletom do you have a reliable way of testing this? Because I'm having trouble :-/ You may try the code below. I'm hoping the `null` won't erase the existing DOM value (as opposed to empty string), but I'm not sure.

```javascript
    created() {
      if (this.$location.query.q) {
        this.query = decodeURIComponent(this.$location.query.q.replace('+', ' '));
      }
    },
    data() {
        // const query = this.$location.query.q || '';
        return {
            current: -1,
            query: null,
            show: false,
            cache: new Cache('site-search', sessionStorage),
            groups: [
                group('datasets', this._('Datasets')),
                group('reuses', this._('Reuses')),
                group('organizations', this._('Organizations'), 'organization'),
                group('territory', this._('Territories'), 'territory'),
            ],
            minLength: 2,
            placeholders: placeholders,
        }
    },

```

oncletom:
@abulte I will give a go at it tomorrow (will be in the office) 👋 if you are around — I agree it would be nicer to compute the field value only once (`#created()`) instead  of at anytime (`#data()`), and prior to the user first interaction with the component.

js/components/site-search.vue
```diff
@@ -63,11 +63,18 @@ export default {
         size: String,
         territoryId: String,
     },
+    created() {
+        const {query} = this.$location;
+
+        const originalField = this.$options.el.querySelector('#search');
+        const queryStringValue = decodeURIComponent(query.q.replace('+', ' '));
+
+        this.query = originalField.value || queryStringValue || '';
```
oncletom: This does the trick. We either pick:

1. the value at runtime of the search field (altered by user or provided by a `input[value]` attribute;
2. otherwise we pick it up from the query string;
3. otherwise field value has not been setup

oncletom:
@abulte I have made the change 😃

oncletom:
Changelog updated 👍

PR #1022, opened 2017-07-12
feature(organization): display datasets/members/reuses counts in org pages

oncletom:
This is an incremental step to make the tabs a bit more informative as they were.
It also prepares an upcoming pull request which removes the right hand indicators.

image

### DIFF
--- a/udata/templates/organization/display.html
+++ b/udata/templates/organization/display.html
@@ -43,12 +43,12 @@
 {% block secondary_content %}
 {# Datasets and reuses tabs #}
 {% set dataset_tabs = (
-    ('datasets', _('Datasets'), datasets, {'organization': org.id|string}),
-    ('private-datasets', _('Private datasets'), private_datasets, {}),
+    ('datasets', ngettext('%(num)d dataset', '%(num)d datasets', datasets.total or 0), datasets, {'organization': org.id|string}),
+    ('private-datasets', ngettext('%(num)d private dataset', '%(num)d private datasets', private_datasets.total or 0), private_datasets, {}),
 ) %}
 {% set reuse_tabs = (
-    ('reuses', _('Reuses'), reuses, {'organization': org.id|string}),
-    ('private-reuses', _('Private reuses'), private_reuses, {}),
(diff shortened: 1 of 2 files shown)


udata/templates/organization/display.html
```diff
@@ -43,12 +43,12 @@
 {% block secondary_content %}
 {# Datasets and reuses tabs #}
 {% set dataset_tabs = (
-    ('datasets', _('Datasets'), datasets, {'organization': org.id|string}),
-    ('private-datasets', _('Private datasets'), private_datasets, {}),
+    ('datasets', ngettext('%(num)d dataset', '%(num)d datasets', datasets.total or 0), datasets, {'organization': org.id|string}),
+    ('private-datasets', ngettext('%(num)d private dataset', '%(num)d private datasets', private_datasets.total or 0), private_datasets, {}),
```
oncletom: Here, I discovered about the `private_datasets` concept. I was not if it was consistently populated so I assumed it had the same datastructure as `datasets`.

udata/templates/organization/display.html
```diff
@@ -140,7 +140,7 @@
     {% endfor %}
 
     {% if org.members %}
-    <tab id="members" header="{{ _('Members') }}">
+    <tab id="members" header="{{ ngettext('%(num)d member', '%(num)d members', org.members | length) }}">
```
oncletom: By reading the template, I assume we don't paginate the `members` therefore it contains the entire list of `org.members`.

The `display_base.html` refers to `org.metrics.members` but on a fresh install this number is not computed (maybe a bug?).

udata/templates/organization/display.html
abulte: I guess you can test by creating or editing a dataset and checking "private".

udata/templates/organization/display.html
davidbgk: AFAIK, metrics are computed via cron tasks using rabbitmq so if you want to test that part you probably need to tweak your local instance. Not straightforward 😞

oncletom: Cool, good to know! On my local instance I did not setup the workers but rabbitmq instance is up (I use `docker-compose up` to setup the app resource dependencies when I develop stuff).

oncletom:
Changelog added 👍

PR #1023, opened 2017-07-12
improvement(organization): move <org>/dashboard/ to <org>/#dashboard

oncletom:
A few things have been done:

- [x] removing _Dashboard_ link in org right hand column
- [x] removing obsolete test
- [x] removing `:org/dashboard/` view class (not sure if widgets were still in use)
- [x] redirecting `:org/dashboard/` to `:org/#dashboard`
- [ ] testing the dashboard tab with real data

# Before

image

# After 

image

This PR is a continuation of and thus based on #1022


udata/core/organization/views.py
```diff
@@ -97,77 +97,8 @@ def get_context(self):
...
-                    'type': 'line',
-                }
-            ]
-        })
-
-        context['metrics'] = widgets
-
-        return context
+def organization_dashboard(org):
+    return redirect('%s#dashboard' % url_for('.show', org=org))
```
abulte: we should add `code=301` to `redirect()`, by default it's a `302` http://flask.pocoo.org/docs/0.12/api/#flask.redirect

oncletom:
@abulte yes indeed, I forgot to add the code back. It is done now.

We can wait for @davidbgk to get another pair of eyes on this change.

udata/core/organization/views.py
```diff
@@ -97,77 +97,8 @@ def get_context(self):
...
-                    'type': 'line',
-                }
-            ]
-        })
-
-        context['metrics'] = widgets
-
-        return context
+def organization_dashboard(org):
+    return redirect('%s#dashboard' % url_for('.show', org=org), code=301)
```
davidbgk: Oh, I didn't know you can use `.show` here, we use to put the explicit prefix (`organizations.show` in that case) not sure if it's relevant but I would switch to it for consistency.

oncletom: Agreed for the consistency.

The dot-something was found here: http://flask.pocoo.org/docs/0.12/blueprints/#building-urls

Although it seems magical to me the route knows in which context the route is.

oncletom:
@davidbgk well, I was surprised to see the dashboard was actually quite empty without JavaScript (HTML stats are between [L25-32](https://github.com/opendatateam/udata/pull/1023/files#diff-07dc1f9e6bb15bf79b518b3d7d1284faL25), JS-based stats on [L34](https://github.com/opendatateam/udata/pull/1023/files#diff-07dc1f9e6bb15bf79b518b3d7d1284faL34)). All others metrics (reuses, datasets and followers) are metrics which are already displayed in the right hand side column of the org (underneath its logo) — and now, as a prefix of the org tab labels.

oncletom:
Changelog added (but merge #1022 first)

Issue #1025, opened 2017-07-12
Some geo coverage maps are not displayed correctly


abulte:
For example, the datasets for BAN and SIRENE do not show the France coverage on the leaflet widget.

https://next.data.gouv.fr/fr/datasets/base-sirene-des-entreprises-et-de-leurs-etablissements-siren-siret/

Version : `1.1.0` on `next.data.gouv.fr`

abulte:
@davidbgk can you look into this one too?

davidbgk:
I think it is related to https://github.com/opendatateam/udata/issues/1024#issuecomment-314769897

PR #1038, opened 2017-07-17
Fix territories migration for country/fr zone

davidbgk:
The command being idempotent, it had to be passed again with that fix but it will only migrate the forgotten zone.

Output:

6600 datasets and 3367 zones affected.
0 town zones, 0 county zones and 0 region zones updated.
2691 France zones, 0 DROM zones, 0 DROM-COM zones updated.
-> Done


abulte:
@davidbgk it should fix #1025, right?

davidbgk:
Indeed, I changed the commit message accordingly. Just when the CI has finished 😭

PR #1039, opened 2017-07-18
Improve purge and job command and document them

noirbizarre:
This PR:
- improve the `purge` command by allowing to purge only a given model
- fix the `job` command in case of keyword arguments
- document both commands

### DIFF
--- a/docs/administrative-tasks.md
+++ b/docs/administrative-tasks.md
@@ -26,3 +26,80 @@ You can also give a user administrative privileges with:
 ```shell
 $ udata user set_admin <email>
 ```
+
+## Purge data flagged as deleted
+
+When users delete some data in udata,
+it's only flagged as deleted and hidden in the frontend.
+This allows the administrative team to undelete data in case of error.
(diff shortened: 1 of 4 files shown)


docs/administrative-tasks.md
```diff
@@ -26,3 +26,80 @@ You can also give a user administrative privileges with:
 ```shell
 $ udata user set_admin <email>
 ```
+
+## Purge data flagged as deleted
+
+When user delete some data in udata,
```
davidbgk: userS or a user deleteS

docs/administrative-tasks.md
```diff
@@ -26,3 +26,80 @@ You can also give a user administrative privileges with:
...
+it's only flagged as deleted and hidden in the frontend.
+This allow the administrative team to undelete data in case of error.
```
davidbgk: allowS

noirbizarre:
This is configurable: purge jobs are manually defined so you can choose the frequency, every every minutes (does not make any sense), every hour, every day...

PR #1053, opened 2017-07-26
fix(search): a page without a search query string would break the initialisation of the search form

oncletom:
In continuity of #1019.

It should be shipped in `udata@1.1` to avoid the search to break.


abulte:
@oncletom #1019 has not been merged into `master`, no (urgent) need to push this in 1.1.

oncletom:
@abulte no problem, I removed the milestone.

PR #1057, opened 2017-07-26
fix(breadcrumb): display breadcrumb on all template inheriting from `base.html`

### LINKED ISSUE #1056: Breadcrumbs are not evenly displayed across pages
## Detailed Description

Several pages do not display the breadcrumb in spite of them being declared in their jinja/html templates:

* Datasets 
image
* Organizations
image

## Possible Implementation

Display the breadcrumb unless a good reasons prevails.

oncletom:
It also has the benefit of making clear what is the content type of a page.

# Dataset

image

# Organization

image

fix #1056

### DIFF
--- a/udata/templates/base.html
+++ b/udata/templates/base.html
@@ -1,4 +1,7 @@
 {% extends theme('raw.html') %}
+{% from theme('macros/breadcrumb.html') import breadcrumb with context %}
+{% set section_class = section_class|default('default') %}
+{% set toolbar_class = toolbar_class|default('') %}
 
 {% block body %}
     {% include theme("header.html") with context %}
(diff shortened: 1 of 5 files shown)


oncletom:
👍 good catch, I will give it another look on Monday.

oncletom:
@noirbizarre Likewise, I missed the follow-up on this issue. Will test on mobile this Tuesday (tomorrow) and ensure long titles don't mess up with the display.

A bit of context too, the intent of bringing back the breadcrumb was to:

1. remain consistent across the pages (it's weird to sometimes have a breadcrumb and sometimes not – especially on these heavily navigated contents)
1. signpost better where we are. It can be a bit confusing to know if were are on a user _or_ organisation page and also a dataset _or_ a reuse.

PR #1059, opened 2017-07-26
fix(user): create and associate timeline.less with timeline.vue

### LINKED ISSUE #1058: Timeline component on user profile page misses some styles
## Detailed Description

The `.timeline` component is not styled, probably because the related stylesheet is not embedded.

Plus the `static/user.css` file seems to be missing for some reason.

screen shot 2017-07-26 at 18 35 43

Can be seen on https://next.data.gouv.fr/fr/users/alexandre-bulte/

## Possible Implementation

Importing the `.timeline` stylesheet at the Vue component level.

oncletom:
# Before

screen shot 2017-07-26 at 18 35 43

# After

screen shot 2017-07-26 at 18 36 10

fixes #1058

### DIFF
--- a/js/components/activities/timeline.vue
+++ b/js/components/activities/timeline.vue
@@ -160,3 +160,30 @@ export default {
     }
 };
 </script>
+
+<style lang="less">
+@import "~less/udata/variables";
+
+@import "~bootstrap/less/mixins";
+@import "~admin-lte/build/less/mixins";
+@import "~admin-lte/build/less/timeline";
+
+.timeline {
+    li > .timeline-icon {
+        background: #f0f0f0;
+    }
+    li > .timeline-icon-more {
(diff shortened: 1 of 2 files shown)


oncletom:
I add @vinyll to the conversation — I did the change but I'm not sure I fully understood the trick, why the `static/user.css` is now generated whereas it was not the case prior to this change. Nor I could track which change broke the layout.

oncletom:
@noirbizarre thanks for pointing me in the right direction, I think it was the right decision 👍

What do you think of the changes?

I also added a screenshot of the `/en/dashboard/` route to reflect the changes (after having done a `inv clean` just to make sure fresh assets were built).

oncletom:
Change done 👍 

Way cleaner this way.

PR #1075, opened 2017-08-02
Gh 1068 handle countries

noirbizarre:
This PRs:
- ensure URL generation works for territories without validity
- removes special handling for country level (and so works when country is disabled)

### DIFF
--- a/udata/routing.py
+++ b/udata/routing.py
@@ -149,12 +149,15 @@ class TerritoryConverter(ModelConverter, PathConverter):
                 slug=territory.slug)
 
         code = getattr(obj, 'code', None)
-        validity = getattr(obj, 'validity', None)
         slug = getattr(obj, 'slug', None)
-        if code and validity and slug:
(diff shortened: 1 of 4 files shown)


udata/routing.py
```diff
@@ -149,12 +149,15 @@ def to_url(self, obj):
...
         slug = getattr(obj, 'slug', None)
-        if code and validity and slug:
+        validity = getattr(obj, 'validity', {})
+        if code and slug:
             return '{level_name}/{code}@{start_date}/{slug}'.format(
-                level_name=level_name, code=code, start_date=validity['start'],
-                slug=slug)
+                level_name=level_name,
+                code=code,
+                start_date=validity.get('start', 'latest'),
```
davidbgk: I don't get when `latest` is a pertinent fallback given that it redirects behind the scene?

noirbizarre: I think it need some clarification in another PR. Right now, this PR ensure it has the only supported behavior.

PR #1078, opened 2017-08-03
Harvest error handling improvement

noirbizarre:
This PRs improve harvest errors handling:
- Prevent multiple logging of a single exception
- Improve type detection error during DCAT backend initialization
- Ensure empty jobs always finish
- Don't store stacktrace any more in case of error (no relevant to the end user and already available while developping in the console or in Sentry in production)

### DIFF
--- a/udata/harvest/backends/base.py
+++ b/udata/harvest/backends/base.py
@@ -63,7 +63,7 @@ class BaseBackend(object):
 
     def harvest(self):
         '''Start the harvesting process'''
-        if self.perform_initialization():
+        if self.perform_initialization() is not None:
             self.process_items()
             self.finalize()
         return self.job
--- a/udata/harvest/backends/base.py
+++ b/udata/harvest/backends/base.py
@@ -84,8 +84,6 @@ class BaseBackend(object):
             if not self.dryrun:
                 self.job.save()
         except Exception as e:
-            log.error("Error in initialization : %s" % (str(e)))
-            log.exception(e)
             self.job.status = 'failed'
             error = HarvestError(message=str(e),
                                  details=traceback.format_exc())
--- a/udata/harvest/backends/base.py
+++ b/udata/harvest/backends/base.py
@@ -85,8 +85,7 @@ class BaseBackend(object):
                 self.job.save()
         except Exception as e:
             self.job.status = 'failed'
-            error = HarvestError(message=str(e),
-                                 details=traceback.format_exc())
+            error = HarvestError(message=str(e))
             self.job.errors.append(error)
             self.end()
             msg = 'Initialization failed for "{0.name}" ({0.backend})'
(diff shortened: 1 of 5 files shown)


udata/harvest/backends/base.py
```diff
@@ -84,11 +84,8 @@ def perform_initialization(self):
             if not self.dryrun:
                 self.job.save()
         except Exception as e:
-            log.error("Error in initialization : %s" % (str(e)))
-            log.exception(e)
             self.job.status = 'failed'
-            error = HarvestError(message=str(e),
-                                 details=traceback.format_exc())
```
abulte: Not a problem to loose the `details` here?

udata/harvest/backends/dcat.py
```diff
@@ -49,6 +51,18 @@ class DcatBackend(BaseBackend):
     def initialize(self):
         '''List all datasets for a given ...'''
         fmt = guess_format(self.source.url)
+        # if format can't be guessed from the url
+        # we fallback on the declared Content-Type
+        if not fmt:
+            response = requests.head(self.source.url)
+            mime_type = response.headers.get('Content-Type', '').split(';')[0]
```
davidbgk: `.split(';', 1)` ensure you're not splitting useless parts. Headers are so messy…

udata/harvest/tests/test_dcat_backend.py
```diff
@@ -172,3 +173,63 @@ def test_hydra_legacy_paged_collection_pagination(self):
...
+    def test_failure_on_initialize(self):
+        url = DCAT_URL_PATTERN.format(path='', domain=TEST_DOMAIN)
+        httpretty.register_uri(httpretty.GET, url, body='should fail')
+        org = OrganizationFactory()
+        source = HarvestSourceFactory(backend='dcat',
+                                      url=url,
+                                      organization=org)
+
+        actions.run(source.slug)
+
```
davidbgk: Why all these blank lines here and below?

udata/harvest/backends/base.py
noirbizarre: Not a problem because as stated in the description:
> (no relevant to the end user and already available while developping in the console or in Sentry in production)

This only remove the stacktrace in the admin interface

abulte: Ok. I have to say I was happy to find the traceback in the admin interface but if we have it in sentry I guess it's ok.

udata/harvest/tests/test_dcat_backend.py
noirbizarre: Like other tests, I try to respect Given/When/Then so I had a blank line between the initial state, the action and the different assertion blocks (there can be multiple assertions block so blank lines between assertions).
This is only for a better tests readability, to see the different logical steps

There is only one blank line before and one after, others before are not blank, this is the indentation

udata/harvest/backends/base.py
noirbizarre: Yes, as a developer, but regular user are scared and not helped by this stacktrace.
You have a far more detailed version in Sentry or in the shell while developping.

Idea of improvement in another PR: maybe get the Sentry ID

PR #1080, opened 2017-08-04
Migrate legacy spatial identifiers

noirbizarre:
This PRs handle unhandled spatial identifiers and level migration and fixes some cases:
- [x] Rename missing spatial levels
- [x] Fix `fr:district` which is `fr:arrondissement`
- [x] Handle all countries
- [x] Handle country groups
- [x] Handle EPCIs
- [x] Preserve granularity

### DIFF
--- a/udata/core/spatial/models.py
+++ b/udata/core/spatial/models.py
@@ -62,6 +62,7 @@ class GeoZone(db.Document):
     dbpedia = db.StringField()
     flag = db.ImageField(fs=logos)
     blazon = db.ImageField(fs=logos)
+    logo = db.ImageField(fs=logos)
 
     meta = {
         'indexes': [
--- a/udata/features/territories/commands.py
+++ b/udata/features/territories/commands.py
@@ -61,7 +61,7 @@ def migrate_zones_ids():
     drom_zone = GeoZone.objects(id='country-subset:fr:drom').first()
     dromcom_zone = GeoZone.objects(id='country-subset:fr:dromcom').first()
     france_zone = GeoZone.objects(id='country:fr').first()
-    for dataset in Dataset.objects.all():
+    for dataset in Dataset.objects(spatial_exists=True):
         if dataset.spatial and dataset.spatial.zones:
             counter_datasets += 1
             new_zones = []
--- a/udata/features/territories/commands.py
+++ b/udata/features/territories/commands.py
@@ -86,6 +86,13 @@ def migrate_zones_ids():
                         GeoZone
                         .objects(code=zone_id, level='fr:region')
                         .first())
+                elif kind == 'epci':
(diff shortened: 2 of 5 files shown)


udata/features/territories/commands.py
```diff
@@ -7,7 +7,10 @@
 import os
 import shutil
 import tarfile
+
+from collections import Counter
```
davidbgk: 😅

udata/features/territories/commands.py
```diff
@@ -50,70 +53,77 @@ def migrate_zones_ids():
...
+            # TODO: handle EPCIs
+            else:
+                new_zones.append(zone)
+        dataset.update(
+            spatial=SpatialCoverage(
+                granularity=dataset.spatial.granularity,
+                zones=[getattr(z, 'id', z) for z in new_zones if z]
+            )
+        )
+    log.info(Formatter().vformat('''Summary
```
davidbgk: I didn't know about `Formatter().vformat`, nice!

noirbizarre: Yes, I just found it! I needed to use the default value (because the `**` doesn't works when the value is not defined)

udata/features/territories/commands.py
noirbizarre: :wink:

udata/features/territories/commands.py
davidbgk: In which cases you don't have the counter defined?

noirbizarre: When you run the command twice

udata/features/territories/commands.py
```diff
@@ -50,70 +53,82 @@ def migrate_zones_ids():
...
+                    new_zones.append(
+                        GeoZone
+                        .objects(code=zone_id, level='fr:region')
+                        .first())
+                elif kind == 'epci':
+                    counter['epci'] += 1
+                    new_zones.append(
+                        GeoZone
+                        .objects(code=zone_id, level='fr:epci')
+                        .latest_valid())
```
davidbgk: Given the fast evolution of EPCIs, I'm not sure that using the latest by default is the more pertinent. Maybe better to use the update/create date to retrieve a more appropriated zone at the time of the submission?

noirbizarre: I agree but I think that if I go this path, I need to do the same for all zones (ie. choose town valid at dataset submission instead of town valid today)

udata/core/spatial/models.py
```diff
@@ -43,6 +43,9 @@ def valid_at(self, valid_date):
         return self(validity__end__gt=compare_string,
                     validity__start__lt=compare_string)
 
+    def latest_valid(self):
```
davidbgk: Still useful?

PR #1088, opened 2017-08-16
Place Territory page components on the grid

oncletom:
This pull request does several things:

- it makes sure the page is readable on small screens
- it gives more space to the _call to action_
- it relies solely on grid classes to align every element and please OCD people like me

# Desktop and large screens breakpoint

## Before

image

## After

image

# Tablet breakpoint

## Before

image

## After

image

# Small screens breakpoint

## Before

image

## After

image

fixes etalab/udata-gouvfr#184

### DIFF
--- a/udata/templates/embed-dataset.html
+++ b/udata/templates/embed-dataset.html
@@ -5,7 +5,7 @@
     }
     .dataset-card {
         border: 1px solid #EEE;
-        min-width: 300px;
+        min-width: 0;
         min-height: 170px;
         clear: both;
         margin: 1em;
--- a/less/udata/territories.less
+++ b/less/udata/territories.less
@@ -17,7 +17,9 @@
     .dataset-item--cta {
         @base-color: #7d8489;
         border: 1px solid @base-color;
-        padding: 0;
+        margin-bottom: @grid-gutter-width;
+        margin-top: @grid-gutter-width;
+        overflow: hidden;
 
         .format-label {
             width: 1em;
--- a/less/udata/territories.less
+++ b/less/udata/territories.less
@@ -3,15 +3,12 @@
         padding: 1em;
     }
     .dataset-list {
-        .col-sm-4 {
-            padding: 5px;
-        }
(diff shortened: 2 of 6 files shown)


udata/templates/embed-dataset.html
```diff
@@ -5,7 +5,7 @@
     }
     .dataset-card {
         border: 1px solid #EEE;
-        min-width: 300px;
+        min-width: 0;
```
davidbgk: Thanks for [the link](https://hackernoon.com/11-things-i-learned-reading-the-flexbox-spec-5f0c799c776b) in the commit message. 👍

oncletom: @davidbgk no prob – I just added some screenshots to pinpoint where the problem was (look at the _tablet breakpoint_).

davidbgk:
I'm good with alignments but beware of the result once real metadata and content is filled. Maybe load production data to check that? I'm particularly afraid of the footer bar which was the bottleneck when I integrated the block.

davidbgk:
> I'd be keen on reusing the existing Dataset layout for this page as it feels weird to refer to the same concept with 2 different layouts.

It's complicated :)

1. we do not link (yet?) to Datasets so it might be (even more) confusing
2. with territories, we reuse our embeds which is good in a way (eat your own dog food) but also bad because we're constrained especially on styling

So I'd say:

* short term: fix alignments and responsiveness (goal of that PR)
* long term: adapt Datasets' pages to be able to link to it and have a consistent block even for territories + remove custom embeds even if that was a good way to test/lazy load.

oncletom:
Understood — let's have a conversation about it when it's due 🙂

For now, we can either merge if everything is okay, deploy and see how the footer thingy goes.
Otherwise I will have to find a way to get a database dump and feed it to my mongodb instance.

davidbgk:
Let's do both, I think it's important that you can load a production database to test/detect anomalies but it should not block a bugfix.

PR #993, opened 2017-06-23
Documentation refactoring

noirbizarre:
This PR try to refactor documentation and introduce the production deployment parts.

PR #1089, opened 2017-08-16
Remove honcho and Procfiles

### LINKED ISSUE #977: I am unsure about what to do with the Procfile.sample
I am discovering the project by installing it for the first time.

1. System Dependencies install with Docker = 👍 
1. Local Dependencies install on macOS = 👍 

However I feel a bit lost when I am about to launch it when I am asked to modify the `Procfile`.  I am going to deconstruct my discomfort:

```
## Uncomment lines given your setup and workflow.
```

What is meant by _workflow_? The term is new at this stage.

```
## That first line actually launch the local web server,
## we prefer to do that manually to avoid mixing these
## logs with all other dependencies but YMMV
```

It seems a personal choice is expressed – I don't know if I should uncomment the line or not.
Plus I was not aware of the meaning of the _YMMV_ acronym so I got a bit side tracked looking after its meaning.

Decision: I will uncomment it but mostly because everything is commented out — I have to start somewhere.

```
## Launch Celery workers (required)
# worker: inv work
# beat: inv beat
```

They seem required although they are commented out. Should they be uncommented by default?

## Possible Implementation

To be honest, I would have rather had something ready to start and just have a choice between a few commands to run to avoid having to copy/modify a file:

- `honcho start` for a general use case
- `honcho start --procfile Procfile.docker` if using Docker
- possibly `honcho start --procfile Procfile.services` to launch system deps separately, as _they_ are really the only ones requiring an edit of the Procfile (if the services are not already launched on system boot)

## Your Environment

* Operating System and version (desktop or mobile): macOS 10.12 w/ homebrew + Docker

oncletom:
They are not visible in the documentation, which states nice ways to run the app on our machine.

fix #977


davidbgk:
Well, I use it because I'm Docker-allergic but I'm not developing anymore so… 🙄

oncletom:
@davidbgk the documentation mentions `invoke` only: http://udata.readthedocs.io/en/stable/development-environment/#running-the-project

I am happy to revert the changes and to document a lean way to run the project on our machines, with and without docker.

noirbizarre:
Sorry, I was sure to have removed the files in #993. Apparently not the case. Thanks for fixing that :+1:

.gitignore
```diff
@@ -64,7 +64,6 @@ coverage.xml
 # Force .gitignore files
 !.gitignore
 .noseids
-Procfile
```
noirbizarre: I think you can safely keep that line. Future procfile from developpers using it won't be commited.

docs/docker-deployment.md
```diff
@@ -1,13 +1,21 @@
...
+
+[Docker Compose][] makes it easy to spawn [system dependencies](system-dependencies.md) out of the box by running the following command:
+
+```shell
+docker-compose up
+```
 
 [Docker]: https://www.docker.com/
-[Docker image]: https://hub.docker.com/r/udata/udata/
+[Docker images]: https://hub.docker.com/r/udata/udata/
```
noirbizarre: If you change it here you need to change the reference above too.

docs/docker-deployment.md
```diff
@@ -1,13 +1,21 @@
-# Docker deployement
+# Docker deployment
 
 This documents the [Docker][] way to deploy uData.
 If you are intending to perform a classic deployment on a server, you should look at the [Installation section](installation.md).
 
-uData and its extensions are packaged as [Docker image][]
-You can go directly to [the dedicated repository][docker-repository] for an up to date docuementation
-and some sample configuration using [Docker Compose][].
+uData and its extensions are packaged and distributed as [Docker images][Docker image].
```
noirbizarre: The second bracket content is not required if the reference is the same as the link label. Here by the way, it's either you have `[Docker images][Docker images]` or `[Docker images][]`, either you change back the reference at the end of the file (see the next comment)

oncletom: Yeah, my bad, I hesitated to change the link reference for too long and mixed it up.
Will update accordingly.

oncletom:
@noirbizarre `.gitignore` amended and doc reference rectified. Good catch thanks 🙂

PR #1090, opened 2017-08-16
Maintain invoke 'underscore' behaviour and improve first time install documentation

oncletom:
For some reasons, commands are hyphenated now:

```
$ inv -l
Available tasks:

  assets-build    Install and compile assets
  assets-watch    Build assets on change
  beat            Run celery beat process
  clean           Cleanup all build artifacts
  cover           Run tests suite with coverage
  dist            Package for distribution
  doc             Build the documentation
  i18n            Extract translatable strings
  i18nc           Compile translations
  jstest          Run Karma tests suite
  qa              Run a quality report
  serve           Run a development server
  test            Run tests suite
  update          Perform a development update
  widgets-build   Compile and minify widgets
  widgets-watch   Build widgets on changes
  work            Run a development worker
```

The PR reflects the change in the documentation.

This is also the opportunity to make the text nicer around the first time install.

PS: PR can be rebased from `master` if the intent is to ship it with `1.1.1` or `1.1.2` (?).


davidbgk:
Not sure about the branch strategy ATM.

oncletom:
> force the old behavior back or update to the new behavior. You chose the 2nd one, it's fine to me (I prefer the `-` to the `_`).

Likewise.

I'm pondering the developer experience we offer in between _patch versions_ of `udata`. I was surprised to have to use a different command in between two code pulls. So thinking about it, I'd rather keep the underscores for now, and move to dashes for `udata@1.2`.

Maybe I'm overthinking and it's not such a big deal but who knows.

noirbizarre:
I think the real issue is that invoke requirement is not pinned. This is why the issue is already happening on master and fix should be submitted here.

Other alternative: pin latest version with `_` on master and pin latest version (so with `-`) on dev and submit the PR on dev

oncletom:
I have amended the commits and added an `invoke.yaml` — I was unsure how to configure that directly in the `tasks.py` file.

noirbizarre:
Why not keep the dashes ?
- We are the only known developpers right now (standard installation does not rely on `invoke`)
- We both agreed that `dashes` are cleaner than `underscore`
- I prefer to stay on the default `invoke` behavior than forcing a legacy one (it's postponing an issue to later)
- `invoke.yaml` is one more config file at the root (which is already crowded of)

oncletom:
Changes made 👍

PR #1092, opened 2017-08-17
Honor Markdown content in datasets description and dataset resources description

oncletom:
# Goal

Description is edited with markdown controls and is thus often used to describe the content of the resources. It's best to reflect the formatting intents of the writers to make them readable, and to make it easier to consume the exposed datasets.

# Security concerns

A few issues which could arise from this:

- malicious scripts
- malicious javascript actions in interactive attributes (`onclick`, etc.)
- text content is very long and requires an action from users to scroll

Content seem to be already sanitised (\o/) when saving.

# Caveats

I certainly have broken the tests — happy to change them if this PR is a direction where we feel good to go.

Long contents could be be clipped on the client side at some point:

- we preserve the integrity of the content in the JSON-LD data
- we preserve the usability of the informations, and keep it at hand

# Frontend

image

# Admin (rendered)

image

# Admin (editing)

image


oncletom:
So to recap, @noirbizarre has something in the pipe to make markdown render more even across the app — see #253.

We can either close this PR and properly address the overall issue _or_ merge it and properly address it later anyway.

PR #1094, opened 2017-08-22
Fix drag and drop

noirbizarre:
This PR fixes a lot of issue on file upload:
- [x] Update to latest fine uploader
- [x] fix upload endpoint detection (and use more explicit endpoint names)
- [x] restore missing dropzone highlight on resource form
- [x] fix filesize handling (when not defined, switch from legacy `File.fileSize` to `Blob.size`)
- [x] fix broken resource list inline upload layout:
  - [x] column numbers where mismatching (did not follow column addition)
  - [x] progress bar styles where not included
  - [x] mime type (way too long) was displayed instead of extension into format column
- [x] uploaded files where not replaced but duplicated
- [x] error handling on upload does not remove nor mark as errored the faulty files

### DIFF
--- a/js/components/dataset/resource/file-form.vue
+++ b/js/components/dataset/resource/file-form.vue
@@ -1,12 +1,18 @@
 <style lang="less">
+@import '~less/admin/variables';
+
 .resource-upload-dropzone {
-    border: 4px dashed #bbb;
+    border: 4px dashed @gray-lte;
     min-height: 150px;
     padding-top: 10px;
     padding-bottom: 10px;
 
     .lead {
         margin-bottom: 0;
+    }
+
+    .drop-active > & {
+        border: 4px dashed @green;
     }
 }
 </style>
--- a/js/components/dataset/resource/file-form.vue
+++ b/js/components/dataset/resource/file-form.vue
@@ -128,18 +128,17 @@ export default {
             if (typeof this.dataset !== 'undefined') {
                 params = {dataset: this.dataset.id};
             }
-            if (this.is_community) {
-                endpoint += 'community_';
-                params.community = this.resource.id;
-            } else {
-                endpoint += 'dataset_';
-            }
             if (this.resource.id) {
-                endpoint += 'resource';
-                params.rid = this.resource.id;
+                if (this.is_community) {
+                    params.community = this.resource.id;
+                } else {
+                    params.rid = this.resource.id;
+                }
             } else {
-                endpoint += 'resources';
+                endpoint += 'new_';
             }
+            endpoint += this.is_community ? 'community_' : 'dataset_';
+            endpoint += 'resource';
             return operations[endpoint].urlify(params);
         }
     },
--- a/js/components/dataset/resource/file-form.vue
+++ b/js/components/dataset/resource/file-form.vue
@@ -28,7 +28,7 @@
     </span>
     <div class="info-box-content">
         <span class="info-box-text">{{file.name}}</span>
-        <span class="info-box-number">{{file.filesize | filesize}}</span>
+        <span class="info-box-number">{{file.size | filesize}}</span>
         <div class="progress">
             <div class="progress-bar" :style="{width: progress+'%'}"></div>
         </div>
--- a/js/components/dataset/resource/file-form.vue
+++ b/js/components/dataset/resource/file-form.vue
@@ -130,7 +130,6 @@ export default {
         upload_endpoint() {
             const operations = API.datasets.operations;
             let params = {};
-            let endpoint = 'upload_';
             if (typeof this.dataset !== 'undefined') {
                 params = {dataset: this.dataset.id};
             }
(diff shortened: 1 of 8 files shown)


js/components/dataset/resource/file-form.vue
```diff
@@ -1,13 +1,19 @@
 <style lang="less">
+@import '~less/admin/variables';
+
 .resource-upload-dropzone {
-    border: 4px dashed #bbb;
+    border: 4px dashed @gray-lte;
```
oncletom: @noirbizarre I can see you hooked on the _variables_ thingy 🙂

js/components/dataset/resource/file-form.vue
```diff
@@ -22,7 +28,7 @@
     </span>
     <div class="info-box-content">
         <span class="info-box-text">{{file.name}}</span>
-        <span class="info-box-number">{{file.filesize | filesize}}</span>
+        <span class="info-box-number">{{file.size | filesize}}</span>
```
oncletom: I am pretty sure a linter would have caught this issue — @noirbizarre @vinyll do you know of anything to enforce data models for Vue.js?

js/components/dataset/resource/file-form.vue
```diff
@@ -128,18 +134,17 @@ export default {
...
+                if (this.is_community) {
+                    params.community = this.resource.id;
+                } else {
+                    params.rid = this.resource.id;
+                }
             } else {
-                endpoint += 'resources';
+                endpoint += 'new_';
             }
+            endpoint += this.is_community ? 'community_' : 'dataset_';
```
oncletom: Seems very error prone. Could we rather build a string like this?

```js
const route_new = this.resource.id ? '' : 'new_';
const route_namespace = this.is_community ? 'community_' : 'dataset_'; 

const endpoint = `upload_${route_new}${route_namespace}resource`;
```

Would make stuff more explicit, and probably simplify the above logic.

js/components/dataset/resource/list.vue
```diff
@@ -155,13 +161,17 @@ export default {
     },
     events: {
         'uploader:progress': function(id, uploaded, total) {
-            const el = this.$el.getElementById(`progress-${id}`);
+            const el = document.getElementById(`progress-${id}`);
```
oncletom: Aren't we breaking the rules of the Vue component? Right now it seems we are looking for something outside of the component itself.

js/components/dataset/resource/list.vue
```diff
@@ -155,13 +161,17 @@ export default {
...
             el.style.width = `${Math.round(uploaded * 100 / total)}%`;
         },
-        'uploader:complete': function(id, response) {
-            const file = this.$uploader.getFile(id);
-            this.files.$remove(this.files.indexOf(file));
+        'uploader:complete': function(id, response, file) {
+            this.files.splice(this.files.indexOf(file), 1);
```
oncletom: What is the intent of this line? Loos like we are modifying the files list in place but not sure why.

The previously used `$remove` invocation did not work for any specific reason?

js/components/dataset/resource/file-form.vue
noirbizarre: Yes, I always had, but there was no proper `@import ~` support in vue-loader initially so it was not possible to properly use them.
When I encounter a case were I can use them, I try to do/fix it

js/components/dataset/resource/file-form.vue
noirbizarre: Not sure that a linter can find that a variable name is removed from the spec on a non-typed variable. File(s) in this component (is)are provided by `fineuploader`. The fact its a `File` object is not indicated anywhere else than the fineuploader documentation :/

js/components/dataset/resource/list.vue
noirbizarre: No, we are looking for a component inside (defined ine the associated template). Sadly, `getElementById` is not defined on `Element`
In Vue.js 2.0, referencing raw `Element`s for direct access in a loop is supported. Sadly, it's only supported on component in Vue.js 1.x :/

js/components/dataset/resource/file-form.vue
noirbizarre: Yes we can (and I will), but that's only working for the string building part, not the parameters.

js/components/dataset/resource/list.vue
noirbizarre: Yes, we remove the file from the array, and for Vue.js to be able to detect the change, we need to use `splice()` (See https://v1.vuejs.org/guide/list.html#Array-Change-Detection)
The previous [`$remove()`](https://012.vuejs.org/guide/list.html#Augmented_Methods) was removed (:wink:) from Vue.js 1.0.

noirbizarre: My bad, the syntax is just not the same (and there is not a dedicated section about it): https://v1.vuejs.org/guide/list.html#Caveats
I wil change to `this.files.$remove(file)` and see if it's working :+1:

PR #1101, opened 2017-09-05
Added SPATIAL_SEARCH_EXCLUDE_LEVELS configuration

noirbizarre:
This PR makes the spatial zones levels to exclude from indexation configurable with `SPATIAL_SEARCH_EXCLUDE_LEVELS`

### DIFF
--- a/docs/adapting-settings.md
+++ b/docs/adapting-settings.md
@@ -92,6 +92,14 @@ The duration used for templates' cache, in minutes.
 
 This is the allowed resources extensions list that user can upload.
 
+## Spatial configuration
+
+### SPATIAL_SEARCH_EXCLUDE_LEVELS
+
+**default**: `tuple()`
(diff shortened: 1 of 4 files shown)


docs/adapting-settings.md
```diff
@@ -92,6 +92,14 @@ The duration used for templates' cache, in minutes.
 
 This is the allowed resources extensions list that user can upload.
 
+## Spatial configuration
+
+### SPATIAL_SEARCH_EXCLUDE_LEVELS
+
+**default**: `tuple()`
+
+List spatial levels that shoudn't be indexed (for time, performance and user experience).
```
oncletom: Is there an evident way to know the lists of spatial levels?

I have a very fuzzy idea of how to write a `tuple` in Python so maybe a small example would be ace.

PR #1108, opened 2017-09-06
Return 503 when all croquemort attempts failed

### LINKED ISSUE #1042: Datasets resources are marked as unreachable if the resources checker fails to work
## Detailed Description

Yesterday croque-mort was failing for some reasons which means resource checking were failing constantly. The side-effect is resources were marked as unreachable whereas they were reachable.

image

## Context

When I look at the resources when the resource checker is down, I want to witness no availability warning message, so I can download them.

## Possible Implementation

Error code should be checked more thorougly so as we can distinguish services error from resource checking status.

## Your Environment

* Operating System and version (desktop or mobile): N/A

abulte:
Fix #1042.

If we can not get the URL infos from croquemort after 10 attempts, it probably means that Croquemort is overloaded or unavailable rather than the destination URL not reachable. Thus, we raise a `503` that won't trigger any color coding on the front (stays gray).


oncletom:
@abulte good idea! Isn't 10 attempts a bit too many to assume croquemort is busy?

Another way of phrasing it is, how can we assume croquemort is too busy sooner? I kind of understand the `503` status is meant for that.

PR #1099, opened 2017-08-30
[WIP] Move croquemort features from udata to udata-croquemort

abulte:
Associated with https://github.com/opendatateam/udata-croquemort/pull/2.

Proof of concept, to be iterated and discussed. The coupling is ugly and some features are missing (i18n).

PR #1110, opened 2017-09-07
Move croquemort features from udata to udata-croquemort v2

abulte:
Replace #1099 

Depends on https://github.com/opendatateam/udata-croquemort/pull/2#issuecomment-327707710 — see there for more info.


abulte:
@noirbizarre I'd be happy to have your feedback at this point, especially about the `XXX` in the code. Once I know how to best organize the `CheckDatasetResource.get` logic, I should be able to try and test it.

abulte:
@noirbizarre proposed refactor (cf previous comment) in https://github.com/opendatateam/udata/pull/1110/commits/63f9fb86b37cce69badcd7c9692567fffc55b2a2 and tests added.

udata/linkchecker/backends.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals, absolute_import
+
+import logging
+import pkg_resources
+from datetime import datetime
+
+log = logging.getLogger(__name__)
+
+DEFAULT_LINKCHECKER = 'croquemort'
```
davidbgk: I think that croquemort should not be referenced here but comes from udata-croquemort's settings.

abulte: Maybe introduce a `DEFAULT_LINKCHECKER` configuration variable with `no_check` by default and has to be set to `croquemort` when you install `udata-croquemort`?

davidbgk: Or something like `LINK_CHECKER = 'path.to.DummyChecker'` by default, I kinda like the django-like Python path in settings.

abulte: I pushed something more coherent IMHO. `no_check` is now an entrypoint provided linkchecker in udata (there is your django-like path ;-)). `LINKCHECKING_DEFAULT_LINKCHECKER` is by default the key to this linkchecker. It will be replaced by `croquemort` if you install `udata-croquemort` and want it to be the default one (or whatever linkchecher you want).

setup.py
```diff
@@ -132,6 +132,9 @@ def pip(filename):
             'ods = udata.harvest.backends.ods:OdsHarvester',
             'ckan = udata.harvest.backends.ckan:CkanBackend',
             'dcat = udata.harvest.backends.dcat:DcatBackend',
+        ],
+        'udata.linkcheckers': [
+            'no_check = udata.linkchecker.backends:NoCheckLinchecker',
```
davidbgk: 👍

udata/core/dataset/api.py
```diff
@@ -516,26 +519,14 @@ def get(self):
...
-        if error or status >= 500:
-            # We keep 503 which means the URL checker is unreachable.
-            return error, status == 503 and status or 500
-        else:
-            return response, status
+    @api.doc('check_dataset_resource')
+    def get(self, dataset, rid):
+        '''Checks that a resource's URL exists and returns metadata.'''
+        resource = self.get_resource_or_404(dataset, rid)
+        result, status = check_resource(resource)
```
davidbgk: `return check_resource` ?

udata/core/dataset/models.py
```diff
@@ -366,21 +349,13 @@ def check_availability(self):
         Return a list of booleans.
         """
         # Only check remote resources.
+        # XXX is this enough? from the frontend we check every type
```
davidbgk: AFAIK, it is only in use within the admin to get global stats. To be checked.

udata/linkchecker/backends.py
```diff
@@ -0,0 +1,55 @@
...
+
+from flask import current_app
+
+log = logging.getLogger(__name__)
+
+
+ENTRYPOINT = 'udata.linkcheckers'
+
+
+class NoCheckLinchecker(object):
```
davidbgk: `NoCheckLinkChecker` (missing k)

udata/linkchecker/backends.py
```diff
@@ -0,0 +1,55 @@
...
+        return {
+            'check:status': 204,
+            'check:available': True,
+            'check:date': datetime.now()
+        }
+
+
+def get(name):
+    '''Get a linkchecker given its name or fallback on default'''
+    all_lc = get_all()
```
davidbgk: Not fond of shortened variables names but YMMV

udata/linkchecker/checker.py
```diff
@@ -0,0 +1,52 @@
...
+    if not result:
+        return {'error': 'No response from linkchecker'}, 503
+    elif result.get('check:error'):
+        return {'error': result['check:error']}, 500
+    elif not result.get('check:status'):
+        return {'error': 'No status in response from linkchecker'}, 503
+    # store the check result in the resource's extras
+    resource.extras.update(_get_check_keys(result))
+    resource.save()
+    return result, result.get('check:status')
```
davidbgk: Is that pertinent to keep the status in result dict?

udata/linkchecker/checker.py
```diff
@@ -0,0 +1,52 @@
...
+
+def check_resource(resource):
+    '''
+    Check a resource availability against a linkchecker backend
+
+    The linkchecker used can be configured on a resource basis by setting
+    the `resource.extras['check:checker']` attribute with a key that points
+    to a valid `udata.linkcheckers` entrypoint. If not set, it will
+    fallback on croquemort. If set to `no_check` it will assume the resource
+    is available via a dummy linkchecker.
```
davidbgk: Add a line about the fact that it returns a flask-like response (dict, status)

setup.py
davidbgk: (missing `k`)

udata/settings.py
```diff
@@ -164,13 +164,16 @@ class Defaults(object):
...
-    #     'url': 'http://localhost:8000',
-    #     'delay': 1,
-    #     'retry': 10,
-    # }
-    #
-    CROQUEMORT_IGNORE = []
+    LINKCHECKING_ENABLED = True
+    LINKCHECKING_IGNORE_DOMAINS = []
+    LINKCHECKING_CACHE_DURATION = 60 * 5  # in seconds
+    LINKCHECKING_DEFAULT_LINKCHECKER = 'no_check'
```
davidbgk: I don't get why we don't put the path here too? `udata.linkchecker.backends:NoCheckLinchecker`

udata/settings.py
```diff
@@ -184,6 +187,7 @@ class Defaults(object):
     #     'RESOURCE_DOWNLOAD': , # Demo = 5, Prod = ?
     #     'RESOURCE_REDIRECT': , # Demo = 6, Prod = ?
     # }
+    # TODO
```
davidbgk: TODO?

udata/linkchecker/checker.py
```diff
@@ -0,0 +1,52 @@
...
+
+def _get_check_keys(the_dict):
+    return {k: v for k, v in the_dict.iteritems() if k.startswith('check:')}
+
+
+def get_cache(resource):
+    '''Return a cached version from a resource's check if any and fresh'''
+    if resource.extras.get('check:date'):
+        cache_duration = current_app.config['LINKCHECKING_CACHE_DURATION']
+        if resource.extras['check:date'] <= datetime.now() \
```
noirbizarre: You might want to extract the expiration into an intermediary variable:
- prevent line break/split
- give meaning to the condition

udata/linkchecker/checker.py
```diff
@@ -0,0 +1,52 @@
...
+    fallback on croquemort. If set to `no_check` it will assume the resource
+    is available via a dummy linkchecker.
+    '''
+    cached_check = get_cache(resource)
+    if cached_check:
+        return cached_check
+    linkchecker_type = resource.extras.get('check:checker', 'croquemort')
+    LinkChecker = get_linkchecker(linkchecker_type)
+    if not LinkChecker:
+        return {'error': 'No linkchecker configured.'}, 503
```
noirbizarre: Maybe return `unkown` instead of an error. If it's optionnal, we might want not to consider the lack of linkchecker as an error.
Alternative: ensure there is always default one like the dummy one.

udata/linkchecker/checker.py
```diff
@@ -0,0 +1,52 @@
...
+    The linkchecker used can be configured on a resource basis by setting
+    the `resource.extras['check:checker']` attribute with a key that points
+    to a valid `udata.linkcheckers` entrypoint. If not set, it will
+    fallback on croquemort. If set to `no_check` it will assume the resource
+    is available via a dummy linkchecker.
+    '''
+    cached_check = get_cache(resource)
+    if cached_check:
+        return cached_check
+    linkchecker_type = resource.extras.get('check:checker', 'croquemort')
```
noirbizarre: The default value should be configurable, especially if the `croquemort` implementation is in an optionnal extension.
Other alternative: handle the fact that there may not be a default link checker (and provide an unkown status)

udata/linkchecker/checker.py
```diff
@@ -0,0 +1,52 @@
...
+    LinkChecker = get_linkchecker(linkchecker_type)
+    if not LinkChecker:
+        return {'error': 'No linkchecker configured.'}, 503
+    result = LinkChecker().check(resource)
+    if not result:
+        return {'error': 'No response from linkchecker'}, 503
+    elif result.get('check:error'):
+        return {'error': result['check:error']}, 500
+    elif not result.get('check:status'):
+        return {'error': 'No status in response from linkchecker'}, 503
```
noirbizarre: error or unkown ?

udata/settings.py
noirbizarre: This is the endpoint name

udata/settings.py
```diff
@@ -164,13 +164,16 @@ class Defaults(object):
...
-    #     'retry': 10,
-    # }
-    #
-    CROQUEMORT_IGNORE = []
+    LINKCHECKING_ENABLED = True
+    LINKCHECKING_IGNORE_DOMAINS = []
+    LINKCHECKING_CACHE_DURATION = 60 * 5  # in seconds
+    LINKCHECKING_DEFAULT_LINKCHECKER = 'no_check'
+
+    # `udata-croquemort` settings
```
noirbizarre: We can drop these lines I think

udata/templates/macros/metadata.html
```diff
@@ -46,8 +46,8 @@
 />
 {% endif %}
 
-<meta name="check-urls" content="{{ 'true' if config.CROQUEMORT else 'false' }}" />
-<meta name="check-urls-ignore" content="{{ config.CROQUEMORT_IGNORE|tojson|urlencode }}" />
+<meta name="check-urls" content="{{ 'true' if config.LINKCHECKING_ENABLED else 'false' }}" />
```
noirbizarre: Or `<meta name="check-urls" content="{{ config.LINKCHECKING_ENABLED|to_json }}" />`
Works for other booleans.

setup.py
abulte: 🙇

udata/linkchecker/backends.py
abulte: 🙇

udata/linkchecker/backends.py
abulte: 👍

udata/linkchecker/checker.py
abulte: I think it can be more straightforward on frontend.

udata/core/dataset/models.py
```diff
@@ -181,26 +177,13 @@ def closed_format(self):
...
+    def check_availability(self):
+        '''
+        Return the check status from extras if any.
 
-    @property
-    def is_available(self):
-        return self.check_availability(group=None)
+        Non checked resources are presumed available.
+        '''
+        return self.extras.get('check:available', True)
```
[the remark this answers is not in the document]
abulte: This is used by `dataset.quality.has_unavailable_resources`. I think we do not want to flag some datasets as "bad quality" just because we did not check their resources. (It was already done like that).

udata/linkchecker/backends.py
```diff
@@ -0,0 +1,55 @@
...
+    lc = all_lc.get(name)
+    if not lc:
+        default_lc = current_app.config.get(
+                                'LINKCHECKING_DEFAULT_LINKCHECKER')
```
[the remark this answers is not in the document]
abulte: I think it has to start by `LINKCHECKING_` to stay coherent with the other settings. And I used the `LINKCHECKING_` prefix and not `LINKCHECKER_` because `LINKCHECKER_ENABLED` does not really make sense since we can have multiple link checkers, `LINKCHECKING_ENABLED` is better IMHO.

udata/linkchecker/checker.py
abulte: You're right. And this is useless since I implemented to `no_check` fallback and `LINKCHECKING_DEFAULT_LINKCHECKER`.

udata/linkchecker/checker.py
abulte: There will always be the dummy one unless you configure a bad `LINKCHECKING_DEFAULT_LINKCHECKER`, and in this case it make sense to raise an error. 

NB: `503` is not really considered as an error on the frontend for now, nothing happens.

udata/linkchecker/checker.py
abulte: Same NB as above.

abulte: I'm probably gonna refactor this.

udata/core/dataset/models.py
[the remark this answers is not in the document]
noirbizarre: Nop, but I think we can safely set it to `unknown` (I saw this possibility in this PR). This is more accurate I think

js/front/dataset/index.js
```diff
@@ -160,21 +160,20 @@ new Vue({
...
+                        } else if (status >= 400 && status < 500) {
+                            el.classList.add('format-label-danger');
+                            el.setTooltip(this._('The resource cannot be found.'), true);
+                        } else if (status >= 500) {
+                            el.classList.add('format-label-warning');
+                            el.setTooltip(this._('An error occured on the remote server. This may be temporary.'), true);
                         }
-                    });
+                    })
+                    .catch(error => console.log('Something went wrong with the linkchecker', error));
```
davidbgk: Don't we want to put a `format-label-unchecked` or something like that in this case?

udata/core/dataset/models.py
```diff
@@ -181,26 +177,14 @@ def closed_format(self):
...
-        else:
-            return True  # We consider that API cases (types) are OK.
+    def check_availability(self):
+        '''
+        Return the check status from extras if any.
 
-    @property
-    def is_available(self):
-        return self.check_availability(group=None)
+        NB: `unknown` will evaluate to True in the aggregate checks using
```
davidbgk: Tricky

udata/core/user/models.py
```diff
@@ -135,11 +135,13 @@ def resources_availability(self):
                 *[org.check_availability() for org in self.organizations]
             )
         )
+        # Filter out the unknown
+        availabilities = [a for a in availabilities if type(a) is bool]
```
davidbgk: Why not using `isinstance`?

udata/settings.py
davidbgk: Are these settings documented somewhere?

udata/core/dataset/models.py
abulte: It kind of is :-/ I hesitated with `None` but the advantage of this solution is that we don't need to modify the aggregate checks (`all([])`).

udata/core/user/models.py
abulte: I like that when checking for a basic type because it reads as en English sentence.

abulte:
@davidbgk I can't reply on your comment on `udata/settings.py` (don't know why): nope they aren't documented, I have to do it 👍

js/front/dataset/index.js
abulte: Yes, could come in handy 👍

udata/core/user/models.py
noirbizarre: :+1:

abulte:
Rebased on `master`.

abulte:
@davidbgk is this OK for you?

js/front/dataset/index.js
```diff
@@ -160,20 +160,22 @@ new Vue({
...
-                                break;
-                            case 503:
-                                break;
-                            default:
-                                el.classList.add('format-label-danger');
-                                el.setTooltip(this._('The server cannot be found.'), true);
+                    this.$api.get(checkurl)
+                    .then((res) => {
+                        const status = res['check:status'];
+                        if (status >= 200 && status < 400) {
```
davidbgk: Thoughts for future: deal with redirection :)

udata/core/user/models.py
```diff
@@ -133,11 +133,13 @@ def resources_availability(self):
...
         if availabilities:
             # Trick will work because it's a sum() of booleans.
             return round(100. * sum(availabilities) / len(availabilities), 2)
         else:
-            return 0
+            return 100
```
davidbgk: Maybe explain why?

PR #1112, opened 2017-09-07
Deduplicate org members - fix #1111

### LINKED ISSUE #1111: Duplicate members in organizations
Cf https://github.com/etalab/data.gouv.fr/issues/1

Some organizations have duplicated members — probably because at some point the interface let us add duplicate members. We should probably add a migration to clean the DB.

abulte:
This adds a migration to ensure only one instance of a user exists in an organization, with the highest role assigned.

NB: the API is protected against double-adding a member, cf `test_organizations_api.py:MembershipAPITest.test_create_member_exists`.

Fix https://github.com/etalab/data.gouv.fr/issues/1 and #1111.

### DIFF
--- a/udata/migrations/2017-07-07-deduplicate-org-members.js
+++ b/udata/migrations/2017-07-07-deduplicate-org-members.js
@@ -1,0 +1,33 @@
+/*
+ * Ensure there's no duplicated members in each organization
+ */
+
+ function is_candidate() {
+     return this.members && this.members.length > 1;
+ }
+
+ var uniqueMembers;
+ var exist;
+ var nbOrgs = 0;
+
+ db.organization.find({'$where': is_candidate}).forEach(org => {
+     uniqueMembers = [];
+     org.members.forEach(member => {
+         if (exist = uniqueMembers.find(m => m.user.equals(member.user))) {
+             print(`Duplicate member ${member.user} for org ${org._id}`);
+             // make sure the higher role is assigned to the deduped user
+             if (member.role === 'admin') {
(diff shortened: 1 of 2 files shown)


udata/migrations/2017-07-07-deduplicate-org-members.js
```diff
@@ -0,0 +1,28 @@
...
+ * Ensure there's no duplicated members in each organization
+ */
+
+ function is_candidate() {
+     return this.members && this.members.length > 1;
+ }
+
+ var uniqueMembers;
+ var exist;
+ db.organization.find({'$where': is_candidate}).forEach(org => {
```
noirbizarre: Can be written `find({members: {$gt: 1}})` and I believe this is faster than using `$where`

udata/migrations/2017-07-07-deduplicate-org-members.js
```diff
@@ -0,0 +1,28 @@
...
+             if (member.role === 'admin') {
+                 exist.role = 'admin'
+             }
+         } else {
+             uniqueMembers.push(member);
+         }
+     })
+     if (uniqueMembers.length !== org.members.length) {
+         org.members = uniqueMembers;
+         db.organization.save(org);
```
noirbizarre: Maybe count and summarize the number of modified organizations

abulte:
@noirbizarre the `$where` was not working so I kept the current version. I added a summary line. All good for you?

noirbizarre:
:+1:

PR #1114, opened 2017-09-08
Harvest auto schedule

noirbizarre:
This PRs ensures that validated harvest sources are automatically scheduled (using the `HARVEST_DEFAULT_SCHEDULE` setting which is a cron expression) and that they can be rescheduled.

### DIFF
--- a/udata/harvest/actions.py
+++ b/udata/harvest/actions.py
@@ -101,6 +101,7 @@ def validate_source(ident, comment=None):
     if current_user.is_authenticated:
         source.validation.by = current_user._get_current_object()
     source.save()
+    schedule(ident, cron=current_app.config['HARVEST_DEFAULT_SCHEDULE'])
     launch(ident)
     return source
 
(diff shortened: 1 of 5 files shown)


udata/harvest/actions.py
```diff
@@ -153,28 +154,33 @@ def preview(ident):
     return backend.harvest()
 
 
-def schedule(ident, minute='*', hour='*',
+def schedule(ident, cron=None, minute='*', hour='*',
```
abulte: Do we still need the other karkgs? `minute`, `hour`...

noirbizarre: Yes, we need it, `udata harvest schedule` command line use it.

noirbizarre:
Right now in this PR, you can only change it from the commandline (before you could only set it, this PRs ensure it can be changed).
This PR is part of a bigger branch splitted in multiple PR. The next one to come on the subject expose the schedule information in the Harvest API payload and allows to set, change and unset it. An admin interface to set this is also provided in this coming PR.

PR #1117, opened 2017-09-09
Initial CircleCI 2.0 configuration

noirbizarre:
This PR try to enable CircleCI 2.0 and workflows.
The resuling build goes from around 15 minutes to between 7 and 8 minutes by using workflows (Python dependencies and tests run at the same time than assets testing and compilation)

CircleCI 2.0 does not allows to use tmpfs (to speed up the build) neither to use docker-compose with direct port access to reuse the existing `docker-compose.yml` file.

It requires #1118 and #1119 to make the build pass (these issues only appears with a fresh/clean dependencies cache)


noirbizarre:
Yes, I already saw that :unamused: 
Let's hope CircleCI improve that (tmpfs brings a lot of spped to the test)

noirbizarre:
BTY, this PR also fixes CircleCI cache issue

PR #1128, opened 2017-09-12
Hide forbidden actions from issue modal (fix etalab/data.gouv.fr#22)

noirbizarre:
This PR does 2 things:
- ensure that issue modal only display buttons that the user can use
- provide a default error handler for the admin part (right now, only handling 403)

### DIFF
--- a/js/admin.vue
+++ b/js/admin.vue
@@ -52,6 +52,17 @@ export default {
                 details: this._('The error identifier is {id}', {id: e.data.event_id}),
             });
         });
+    },
+    methods: {
+        handleApiError(error) {
+            const notif = {type: 'danger', icon: 'exclamation-circle'};
+            if (error.status === 403) {
+                notif.title = this._('Operation not permitted');
+                notif.details = this._('You are not allowed to perform this operation');
+                notif.icon = 'ban'
+            }
+            this.notifications.push(notif);
+        }
     }
 };
 </script>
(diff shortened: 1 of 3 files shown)


js/admin.vue
```diff
@@ -52,6 +52,17 @@ export default {
                 details: this._('The error identifier is {id}', {id: e.data.event_id}),
             });
         });
+    },
+    methods: {
+        handleApiError(error) {
+            const notif = {type: 'danger', icon: 'exclamation-circle'};
+            if (error.status === 403) {
+                notif.title = this._('Operation not permitted');
+                notif.details = this._('You are not allowed to perform this operation');
```
oncletom: I'm not sure to understand how it is displayed in context?

A person could get to see this message even though no button/action lead to it?

noirbizarre: No, this is a generic handler (handle only 403 but it's a start and it needs improvements) that **"can"** be used on any operation without the need for a specific error handler.
This one simply display the error in the notification zone (the toast).
But this is displayed only on error on actions triggered by the user (See: https://github.com/opendatateam/udata/pull/1128/files#diff-506081ce406eac251a8b297d56b75664R129)

Note: the current situation is that most the API operation don't have error handling. This allows future bugfixes/improvements... to take advantage of this helper and/or make it handle more generic cases.

PR #1133, opened 2017-09-14
Fix "Center the full picture" on user avatar upload

### LINKED ISSUE #1130: Could not convert string to float
Hello
we get the error could not convert string to float when upload image and try to center it.

https://www.data.gouv.fr/fr/

this error also appears on our server

see attached file

could_not_convert_string_to_float

abulte:
Fix #1130

### DIFF
--- a/js/components/widgets/image-picker.vue
+++ b/js/components/widgets/image-picker.vue
@@ -133,7 +133,10 @@ export default {
             if (this.HAS_FILE_API) {
                 const data = {};
                 try {
-                    data.bbox = this.$refs.thumbnailer.bbox;
+                    let bbox = this.$refs.thumbnailer.bbox;
+                    if (bbox) {
+                        data.bbox = bbox;
+                    }
                     this.upload(data);
                 } catch(e) {}
             } else {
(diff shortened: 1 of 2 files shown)


js/components/widgets/image-picker.vue
```diff
@@ -133,7 +133,10 @@ export default {
             if (this.HAS_FILE_API) {
                 const data = {};
                 try {
-                    data.bbox = this.$refs.thumbnailer.bbox;
+                    let bbox = this.$refs.thumbnailer.bbox;
```
noirbizarre: Shoumd be a `const` instead of a `let`

abulte: Yes :-/ Worth a new PR?

noirbizarre: Nop, this shouldn't have been merged: breaks on lower resolutions (non-retina) and mobile :/

abulte: Really? :-( How? I just tested on mobile (retina) and it looks ok.

noirbizarre: No sorry, wrong window ;)
The PR is welcomed for the `s/let/const/`

abulte: 😌

PR #1134, opened 2017-09-14
Ensure min-{height,width} for .user-face img

oncletom:
On a slow connection, user faces images might not be loaded for a while.
They appear with a 60px width but a 0px height.

This PR ensure they have a proper height even before we receive the first byte of the image.

# Before

image

# After

image

### DIFF
--- a/js/components/organization/members.vue
+++ b/js/components/organization/members.vue
@@ -38,6 +38,8 @@
         width: 76%;
         max-width: 60px;
         max-height: 60px;
+        min-height: 60px;
+        min-width: 60px;
     }
 
     strong {


oncletom:
@abulte I think we could — I suppose user images are square anyway? Just asking @noirbizarre in case there are some (not well-)known behaviours for user images.

oncletom:
You are right @noirbizarre, the min size can be smaller indeed.

screen shot 2017-09-18 at 12 54 19

That's something which did not show up just by resizing the window size to its minimum.

blayo:
@oncletom: which tool did you use to simulate a 320x568 on iphone 5 ?

abulte:
@blayo: looks like Chrome dev tools

oncletom:
Exactly :-) (Ctrl+Shift+M / Cmd+Shift+M in Chrome)

image

PR #1140, opened 2017-09-18
Spatial zones temporal resolution

noirbizarre:
This PR perform time resolution of spatial zones on:
- dataset submission (create or update)
- OEmbed API querying

This also adds some testing on spatial and GeoID querying:
- make the `GeoZone` model and factory handle predefined typed `validity.start` and `validity.end`
- make the `GeoZoneFactory` coherent by generating a real GeoID given its fake attributes
- test and fix `GeoZoneQuerySet.valid_at(validity)` (issue on exact match)
- add and test `GeoZoneQuerySet.latest()` and `GeoZoneQuerySet.resolve(geoid)` to factorize these reusable queries
- make `9999-12-31` a `END_OF_TIME` constant to make it more visible

Add a temporary Faker DateTime provider until joke2k/faker#596 is available for easier date testing

### DIFF
--- a/udata/core/spatial/models.py
+++ b/udata/core/spatial/models.py
@@ -12,10 +12,12 @@ from udata.i18n import lazy_gettext as _, gettext, get_locale
 from udata.models import db
 from udata.core.storages import logos
 
+from . import geoids
+
 
 __all__ = (
     'GeoLevel', 'GeoZone', 'SpatialCoverage', 'BASE_GRANULARITIES',
-    'spatial_granularities', 'parse_geoid',
+    'spatial_granularities',
 )
 
 
@@ -26,26 +28,6 @@ BASE_GRANULARITIES = [
 
 ADMIN_LEVEL_MIN = 1
 ADMIN_LEVEL_MAX = 110
-
-# Arbitrary date value used as validity.end
-# for zone not yet ended
-END_OF_TIME = date(9999, 12, 31)
-
-
-def parse_geoid(text):
-    '''
-    Parse a geoid from text
-    and return a tuple (level, code, validity)
-
-    GeoID, see https://github.com/etalab/geoids.
-    '''
-    if '@' in text:
-        spatial, validity = text.split('@')
-    else:
-        spatial = text
-        validity = 'latest'
-    level, code = spatial.rsplit(':', 1)
-    return level, code, validity
 
 
 class GeoLevel(db.Document):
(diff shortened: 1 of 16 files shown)


udata/core/spatial/models.py
```diff
@@ -27,6 +27,26 @@
 ADMIN_LEVEL_MIN = 1
 ADMIN_LEVEL_MAX = 110
 
+# Arbitrary date value used as validity.end
+# for zone not yet ended
+END_OF_TIME = date(9999, 12, 31)
```
davidbgk: I wonder if we should expose that date from geohisto and make it a dependency.

udata/features/oembed/api.py
```diff
@@ -45,18 +45,18 @@ def get(self):
...
+
                 try:
                     country, level, code, kind = item_id.split(':')
                 except ValueError:
                     return api.abort(400, 'Invalid territory ID.')
-                try:
-                    geozone = GeoZone.objects.get(
-                        id=':'.join([country, level, code]))
-                except GeoZone.DoesNotExist:
+                geoid = ':'.join((country, level, code))
```
davidbgk: Maybe we should have a counterpart to `parse_geoid`?

udata/core/spatial/models.py
```diff
@@ -27,6 +27,26 @@
...
+    and return a tuple (level, code, validity)
+
+    GeoID, see https://github.com/etalab/geoids.
+    '''
+    if '@' in text:
+        spatial, validity = text.split('@')
+    else:
+        spatial = text
+        validity = 'latest'
+    level, code = spatial.rsplit(':', 1)
```
davidbgk: We should test there is a `:` before

udata/core/spatial/models.py
```diff
@@ -39,9 +59,43 @@ class GeoLevel(db.Document):
...
+        return self(is_valid | no_validity)
+
+    def latest(self):
+        '''
+        Fetch the latest valid zone matching a QuerySet.
+
+        Ensuring the QuerySet unicity for (level, code)
+        is you responsibility.
+        '''
+        return self.order_by('-validity__end').first()
```
davidbgk: I think we should check validity for today. It would prevent from returning a zone that is not anymore active/valid.

noirbizarre: That's the expected behavior: I want the latest zone wether it exists or not anymore.
First implementation was using `today` but testing revealed that it wasn't working for ended zones (which we have).

udata/features/oembed/api.py
noirbizarre: There is already but it's not useable here because this is not using the same logic than anywhere else:
- `fr` prefix is hardcoded, `level` is not the same as in `udata.core.spatial`
- `code` is including the `@validity` part
This code section is already using the `GeoZone.objects.resolve()` helper one line below, but this part is not factorizable.
(BTW, I factorized a `udata.spatial.geoids` helper module)

udata/core/spatial/models.py
noirbizarre: Added

PR #1146, opened 2017-09-21
[WIP] UTF8 by the book

noirbizarre:
This PR tries to properly handle utf-8 on edge cases.

Detection:
- forces systeme default encoding to `ascii` during test
- makes the factories use unicode when using lorem generator
- feeds factories with more data (use traits)
- try to test facets rendering

Resolution:
- fixes the `__str__` and `__unicode__` handler on some models

### DIFF
--- a/udata/utils.py
+++ b/udata/utils.py
@@ -12,7 +12,9 @@ from datetime import date, datetime
 from calendar import monthrange
 from math import ceil
 from faker import Faker
+from faker.config import PROVIDERS
 from faker.providers import BaseProvider
+from faker.providers.lorem.la import Provider as LoremProvider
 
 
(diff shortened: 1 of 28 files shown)


udata/utils.py
```diff
@@ -230,3 +237,9 @@ class UDataProvider(BaseProvider):
     def unique_string(self, length=UUID_LENGTH):
         '''Generate a unique string'''
         return unique_string(length)
+
+
+@faker_provider  # Replace the default lorem provider with a unicode one
+class UnicodeLoremProvider(LoremProvider):
+    '''A Lorem provider that forces unicode in words'''
```
abulte: 👍

PR #1148, opened 2017-09-22
Test settings and tmpfs

noirbizarre:
This PR supersedes #902 and do the following:
- upgrades docker-compose to version 3 syntax
- provides Elasticsearch and Mongo test services with tmpfs storage
- provides optionnal test configuration for:
   - Elasticsearch: `ELASTICSEARCH_URL_TEST`
   - Mongo: `MONGODB_HOST_TEST`

This allows to run test locally with tmpfs middlewares which speed up the test (test duration almost divided by 2 on my station, from 9'45 to 5'10) by just adding this to my `udata.cfg`:
```python
MONGODB_HOST_TEST = 'mongodb://localhost:27018/udata'
ELASTICSEARCH_URL_TEST = 'localhost:9201'
```

**Note**: CircleCI does not support tmpfs right now, but the day he will, we will be able to speed up CircleCI tests too.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -29,11 +29,11 @@ services:
 
 
 
-        search:
-            image: udata/elasticsearch:2.4.5
-            volumes:
-              - ./data/search/data:/usr/share/elasticsearch/data
-            expose:
-              - "9200"
-            ports:
-              - "9200:9200"
+    search:
+        image: udata/elasticsearch:2.4.5
+        volumes:
+          - ./data/search/data:/usr/share/elasticsearch/data
+        expose:
+          - "9200"
+        ports:
+          - "9200:9200"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,27 +1,31 @@
-db:
-    image: mongo:3.2
-    command: mongod
-    volumes:
-      - ./data/db:/data/db
-    expose:
-      - "27017"
-    ports:
-    - "27017:27017"
+version: "3"
+services:
+    db:
+        image: mongo:3.2
+        command: mongod
+        volumes:
+          - ./data/db:/data/db
+        expose:
+          - "27017"
+        ports:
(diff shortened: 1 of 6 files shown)


docker-compose.yml
```diff
@@ -1,27 +1,40 @@
...
-broker:
-    image: redis
-    volumes:
-      - ./data/broker:/data
-    expose:
-      - "6379"
-    ports:
-    - "6379:6379"
+    db_test:
+        image: mongo:3.2.0
```
abulte: Why `3.2.0` here and `3.2` above?

docker-compose.yml
```diff
@@ -1,27 +1,40 @@
-db:
-    image: mongo:3.2
-    command: mongod
-    volumes:
-      - ./data/db:/data/db
-    expose:
-      - "27017"
-    ports:
-    - "27017:27017"
+version: "3"
```
abulte: Maybe use a different `docker-compose` file with tests services?

noirbizarre: This is too complex for most to handle:
- you need to switch each time you want to run tests (Most of the time I run my tests while the server running)
- this is not a production compose file
- this means we have to complexify the dev/test documentation
- one more file to the root :/

But it's open to debate. We can provide a second docker-compose file with these services to reduce memory usage for those who ain't running tests, but I think those who wants to test use the official docker image with one of the sample docker-compose.

abulte: This a just a matter of running `docker-compose -f docker-compose.yml -f docker-compose.tests.yml up` I think. But you're right about "those who wants to test use the official docker image with one of the sample docker-compose", so let's forget this.

noirbizarre: I changed it anyway, I find it clever to be able to run only whats needed :+1:

abulte: ✌️

PR #1168, opened 2017-09-26
Enconding fixes

noirbizarre:
Small fixes related to encoding (See #1146)

### DIFF
--- a/udata/models/document.py
+++ b/udata/models/document.py
@@ -39,9 +39,14 @@ class UDataDocument(Document):
         data[id_field] = getattr(self, id_field)
         return data
 
-
     def __str__(self):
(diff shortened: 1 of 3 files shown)


udata/models/document.py
```diff
@@ -39,10 +39,15 @@ def to_dict(self, exclude=None):
...
-
     def __str__(self):
         return self.__unicode__().encode('utf-8')
 
+    def __unicode__(self):
+        return '{classname}({id})'.format(
+            classname=self.__class__.__name__,
+            id=getattr(self, self._meta['id_field'])
+        )
+
```
abulte: :+1:

PR #1182, opened 2017-10-02
Fix URL validation for unicode support

noirbizarre:
This PR improves URL fields validation (both model and form level) by supporting more use cases:
- new tlds
- unicode domains
- unicode paths
- allows to accept only public URLs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Added a `udata info` command for diagnostic purpose [#1179](https://github.com/opendatateam/udata/pull/1179)
 - Add `.ttl` and `.n3` as supported file extensions [#1183](https://github.com/opendatateam/udata/pull/1183)
 - Improve logging for adhoc scripts [#1184](https://github.com/opendatateam/udata/pull/1184)
+- Improve URLs validation (support new tlds, unicode URLs...) [#1182](https://github.com/opendatateam/udata/pull/1182)
 
 ## 1.1.8 (2017-09-28)
 
--- a/udata/models/url_field.py
+++ b/udata/models/url_field.py
@@ -1,14 +1,85 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+import re
+
 from mongoengine.fields import URLField as MEURLField
+
+
+IP_MIDDLE_OCTET = r'(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5]))'
+IP_LAST_OCTET = r'(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))'
+
+URL_REGEX = re.compile(
+    r'^'
+    # scheme is validated separately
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Switch to [Crowdin](https://crowdin.com) to manage translations [#1171](https://github.com/opendatateam/udata/pull/1171)
 - **BREAKING** Switch to `Flask-Security`. `Flask-Security-Fork` should be uninstalled before installing the new requirements [#958](https://github.com/opendatateam/udata/pull/958)
 - Added a `udata info` command for diagnostic purpose [#1179](https://github.com/opendatateam/udata/pull/1179)
+- Improve URLs validation (support new tlds, unciode URLs...) [#1182](https://github.com/opendatateam/udata/pull/1182)
```
abulte: s/unciode/unicode/

udata/models/url_field.py
```diff
@@ -1,14 +1,80 @@
...
+
 class URLField(MEURLField):
     '''
     An URL field that automatically strips extra spaces
+    and support uncode domain and paths.
+
+    Public URL can be enforced with `public=True`
+
+    URL_REGEX has been extracted and adapted from:
+        https://github.com/kvesteri/validators/blob/master/validators/url.py
```
abulte: What is different versus the `validators.url` implementation? Public vs private?

udata/tests/forms/test_basic_fields.py
```diff
@@ -72,3 +72,17 @@ def test_with_invalid_url(self):
         form.validate()
         self.assertIn('url', form.errors)
         self.assertEqual(len(form.errors['url']), 1)
+
+    def test_with_unicode_url(self):
+        Fake, FakeForm = self.factory()
+
+        fake = Fake()
+        url = url = 'https://www.somewhère.com/with/accënts/'
```
abulte: ~~url =~~

udata/models/url_field.py
noirbizarre: Schema validation is handled separately (MongoEngine implementation).
Raw strings are used instead of unicode

noirbizarre:
I didn't activated the `public=True` on any model yet because I think it requires a little debate.
I think this is interesting to have `public=True` on reuses and resources URL and on org and user websites.
But, this is in the case of a public portal.

abulte:
@noirbizarre Yes I agree, it's debatable because some users may want to link to internal resources.

abulte:
@noirbizarre Maybe a config flag? `ALLOW_PRIVATE_RESOURCES`?

noirbizarre:
Yep, maybe in a separate PR. This one fix unicode and lands the fact that we can filter out private URL and another PR will provide the ability to configure this

udata/models/url_field.py
noirbizarre: I added it in the comment

PR #1193, opened 2017-10-03
Switch from user placeholders to identicon

noirbizarre:
This pull request replace user placeholders by [identicons](https://en.wikipedia.org/wiki/Identicon) (aka. a visual representation of a hash).

## Algorithm 
The user identifier is used as value to be hashed (not the email as it is not public and so not usable from external applications and also because it can change over the time).

## Customization
The avatar/identicon backend is customizable:
- you can switch implementations (using the `AVATAR_PROVIDER`)
- you can configure each implementation when possible
- you can create your own (using the `udata.avatars` python entrypoint)

Each parameter comes with a default value, overridable by theme and then local config

## Implementations

This PR comes with 3 different implementations.

### `internal` (default)

This implementation use [pyidenticon](http://pydenticon.readthedocs.io) to generate user avatars. This is udata which is generating the avatar and so there is no external request when displaying them.

Parameters are the follwing:

- `AVATAR_INTERNAL_SIZE` *(default: `7`)*: Number of blocks (the matrix size) used by the internal provider. (*Ex*: `7` will render avatars on a 7x7 matrix)
- `AVATAR_INTERNAL_FOREGROUND` *(default: `['rgb(45,79,255)', 'rgb(254,180,44)', 'rgb(226,121,234)', 'rgb(30,179,253)', 'rgb(232,77,65)', 'rgb(49,203,115)', 'rgb(141,69,170)']`)*: A list of foreground colors used by the internal provider to render the avatars
- `AVATAR_INTERNAL_BACKGROUND` *(default: `'rgb(224,224,224)'`)*: The background color used by the internal provider
- `AVATAR_INTERNAL_PADDING` *(default: `10`)*: The padding (in percent) used by the internal provider

screenshot-www data dev-2017-10-03-12-36-57-183
screenshot-www data dev-2017-10-03-14-03-16-141
screenshot-www data dev-2017-10-03-14-14-53-775

### `robohash`

A provider redirecting to [robohash](https://robohash.org) avatars.

Parameters are:
- `AVATAR_ROBOHASH_SKIN` *(default: `'set1'`): The skin (set) used by the robohash provider.
- `AVATAR_ROBOHASH_BACKGROUND` *(default: `'bg0'` (transparent background)): The background used by the robohash provider.

screenshot-www data dev-2017-10-03-12-40-39-639
screenshot-www data dev-2017-10-03-14-05-04-871
screenshot-www data dev-2017-10-03-14-05-41-966

### `adorable`

A provider redirecting to [Adorable avatars](http://avatars.adorable.io).
This provider is not customizable.

screenshot-www data dev-2017-10-03-12-37-45-975
screenshot-www data dev-2017-10-03-14-04-16-819
screenshot-www data dev-2017-10-03-14-06-58-564

## Fixes

This PR fixes:
- the avatar display in user cards (use same spacing as dataset and reuse cards)
- the spacing and sizing on the followers display on reuse page (see above screenshots)

### Naming

It's hard to find a good wording for variables and helpers. Sometimes identicon comes, sometime it's avatar. Any suggestion is welcome on this issue/subject...

### Packaging

This feature is located in the `udata.features.identicon` but the API is `/avatars/<identifier>/<size>`.
Given the previous point, it might be enhencable. Suggestions are welcome.

### Caching

This kind of feature requires cache, the question is wether it's the udata responsibility or the reverse proxy.


udata/features/identicon/backends.py
```diff
@@ -0,0 +1,131 @@
...
+from flask import redirect, send_file, current_app
+
+from udata import theme, entrypoints
+
+ADORABLE_AVATARS_URL = 'https://api.adorable.io/avatars/{size}/{identifier}.png'
+ROBOHASH_URL = 'https://robohash.org/{identifier}.png?size={size}x{size}&set={skin}&bgset={bg}'  # noqa
+
+
+# Default values overriden by theme default and local config
+DEFAULTS = {
```
abulte: Why aren't the defaults stored in `settings.py`?

udata/frontend/helpers.py
```diff
@@ -164,12 +164,12 @@ def owner_url(obj):
 @front.app_template_filter()
 @contextfilter
 def avatar(ctx, user, size, classes=''):
-    markup = '''
-        <a class="avatar {classes}" href="{url}" title="{title}">
-            <img src="{avatar_url}" class="avatar" alt="{title}"
-            width="{size}" height="{size}"/>
-        </a>
-    '''.format(
+    markup = ''.join((
```
abulte: I liked the multiline string better but YMMV.

udata/templates/user/card.html
```diff
@@ -2,8 +2,8 @@
 {% set user_url = url_for('users.show', user=user) %}
 <div class="card">
     <a class="card-logo" href="{{ user_url }}">
-        <img src="{{ user|avatar_url(48) }}" alt="{{ user.fullname }}"
-            width="48" height="48">
+        <img src="{{ user|avatar_url(60) }}" alt="{{ user.fullname }}"
+            width="60" height="60">
```
abulte: Shouldn't we deal with Retina (here and everywhere else)? i.e. request a 120px avatar.

udata/features/identicon/backends.py
noirbizarre: To let the theme bring custome parameters.

udata/frontend/helpers.py
noirbizarre: Yep me too but the extras spaces where the source of the visual breakage :/

udata/templates/user/card.html
noirbizarre: Yes, but I think it requires a PR on its own (maybe dealing with `srcset`)

oncletom:
Nice finding!

The `internal` works well in thread! In a grid, it feels very heavy and more like an 8bits space invader clone rather than data.gouv.fr :-( Using the main colors of data.gouv.fr palette would probably remove the 'out of place' / 'in my face' feeling they give.

The `robohash` feels out of place with the voice and tone of data.gouv.fr — could be good in other udata contexts though.

`adorable` is uneven and I feel uncomfortable seeing some faces on data.gouv.fr — like the angry and semi-drunk ones.

---

I have used http://webcolourdata.com to extract the colour palette of data.gouv.fr, @noirbizarre could you try with these `internal` foreground colours?

- `#5685C4`
- `#483D68`
- `#483D68`
- `#565D67`
- `#016D1B`
- `#125E9D`
- `#E8512A`

noirbizarre:
I tottaly agree but this PR is only a kind of framework:
- robohash and adorable backends are simple examples of what is easily doable
- theme can customize the colors (your proposal should go in the `gouvfr` theme)

noirbizarre:
I joined the followers screenshot because it's something fixed in this PR.

But given:
screenshot-www data dev-2017-10-03-15-54-45-588

Most of the time it will look like:
screenshot-www data dev-2017-10-03-15-58-58-127


Maybe we can fix the display for reuses having a lots of followers (in another pr):
- only display the first ones followed by a "and XX more" label with an expander
- add a separation between avatars
- ...??

abulte:
`internal` (`pyidenticon`) should be what's deployed, right? I'm not OK with `adorable` either ;-) But `pyidenticon` is OK for me. Maybe with a color PR on `udata-gouvfr` as suggested.

noirbizarre:
As stated before, the adorable avatars and the robohash avatars are simple examples (**1 line of code for adorable** and **2 lines for robohash**) of what's possible with this approach.
I can remove the lines but experience showed me that each time something is customizable (on udata or any other project), people ask for examples. This time I already packaged the example. As I'm also doing the support, this allows me to give a simple github URL with samples.

**depending on a third party**: it's configuration, It's not the default one, and there is no dependency at all (the 2 implementations are simple HTTP redirects, no third party libraries or anything like this).

oncletom:
OK that's the bit I did not get — I thought we only had a choice among 3 configurations. Now I understand `identicon` is a configurable system onto which we can plug any other avatar system, including our own.

PR #1200, opened 2017-10-05
Migration to remove void resources

abulte:
Remove resources without `title` or `url` from the DB. On `data.gouv.fr`, this matches:

```
ObjectId("53698e19a3a729239d203399") Fonds stratégique pour le développement de la presse (FSDP) - Aides attribuées
ObjectId("583e801488ee387693c65bb3") SYME05 - Bornes de recharge pour véhicule électrique
```

### DIFF
--- a/udata/migrations/2017-10-05-purge-void-resources.js
+++ b/udata/migrations/2017-10-05-purge-void-resources.js
@@ -1,0 +1,23 @@
+/*
+ * Remove resources with missing required fields (url or title)
+ */
+
+var nbRemoved = 0;
+
+const removedNoTitle = db.dataset.update(
+    {resources: {$gt: []}},
+    {$pull: {resources: {title: null}}},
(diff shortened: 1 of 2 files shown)


udata/migrations/2017-10-05-purge-void-resources.js
```diff
@@ -0,0 +1,23 @@
+/*
+ * Remove resources with missing required fields (url or title)
+ */
+
+var nbRemoved = 0;
+
+const removedNoTitle = db.dataset.update(
+    {resources: {$gt: []}},
```
noirbizarre: You can reduce the number of documents to lookup with `{resources.title: null}`.
If you want to perform it in a single pass:
```
{$or: [{'resources.title': null}, {'resources.url': null}]},
{$pull: {resources: {$or: [{title: null}, {url: null}]}}},
```

PR #1203, opened 2017-10-09
Improve licenses detection in harvesting

noirbizarre:
This PRs improve the `Licence.guess()` helper by allowing a list of string and a default fallback license.
It also expose the default license as `License.default()`
All harvesters now use the same algorithm.

### DIFF
--- a/udata/core/dataset/models.py
+++ b/udata/core/dataset/models.py
@@ -111,13 +111,30 @@ class License(db.Document):
         return self.title
 
     @classmethod
-    def guess(cls, text):
+    def guess(cls, *strings, **kwargs):
+        '''
+        Try to guess a license from a list of strings.
+
+        Accept a `default` keyword argument which will be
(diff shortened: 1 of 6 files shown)


udata/core/dataset/models.py
```diff
@@ -111,13 +111,27 @@ def __unicode__(self):
         return self.title
 
     @classmethod
-    def guess(cls, text):
+    def guess(cls, *strings, **kwargs):
+        '''
+        Try to guess a license from a list of strings.
```
abulte: Maybe document the `default` kwarg?

abulte: Or declare it explicitly in the method signature.

noirbizarre: It's not possible to declare explicitely a keyword argument with variable arguments in Python 2 (already tried :/).
I'll add it to the documentation.

PR #1205, opened 2017-10-09
Fix URLs and Hashes overflowing resource modals

noirbizarre:
This PR ensures resources URLs and hashes are not overflowing from the resource modal (they are specific cases where word break is unapplyable because for the CSS engines they are a single non-breakable word) so the fix is to simply allow line break in the middle of a word)

screenshot-www data dev-2017-10-09-16-57-56-147

Fix etalab/udata-gouvfr#207

(There might be a better long term fix by changing the way resources are displayed, this is only a quick fix with the current solution)


js/front/dataset/resource-modal.vue
```diff
@@ -60,3 +60,12 @@ export default {
     }
 };
 </script>
+
+<style lang="less">
+.resource-modal {
+    .dl-wide dd {
```
oncletom: Why is it duplicated from `dataset/resource/modal.vue`?

noirbizarre: Because they are 2 differents modals with the same issue.
They both are imported into different places (one in the front dataset page, the other in the admin dataset page) and so style isn't shared :/

PR #1206, opened 2017-10-10
Fix the scroll to a discussion sub-thread

abulte:
I still have a problem on my local env (when page finishes loading, scroll to top of page) but at least the JS error is fixed by this PR.

### DIFF
--- a/js/components/discussions/thread.vue
+++ b/js/components/discussions/thread.vue
@@ -136,7 +136,7 @@ export default {
             this.detailed = true;
             if (index) {
                 this.$nextTick(() => {
-                    this.$scrollTo(`#${discussionIdAttr}-${index}`);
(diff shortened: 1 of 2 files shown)


js/components/discussions/thread.vue
```diff
@@ -136,7 +136,7 @@ export default {
             this.detailed = true;
             if (index) {
                 this.$nextTick(() => {
-                    this.$scrollTo(`#${discussionIdAttr}-${index}`);
+                    this.$scrollTo(`#${this.discussionIdAttr}-${index}`);
```
oncletom: A further improvement would be to have an `index` parameter in the `computed.discussionIdAttr` to avoid adding it manually everywhere in the page.

PR #1208, opened 2017-10-12
Point out to udata in feathub

pblayo:
It'd be more straightforward to point out to the udata's entry in feathub rather than to http://feathub.com.
(it works both with and without being logged in from github)


abulte:
Oh my, I did not even know about this. Are we still using feathub? If not, shouldn't we remove it?

pblayo:
Just for information, the change requests are one year old :

image

noirbizarre:
:+1: for removal

pblayo:
I've submitted the pull request #1300 to remove the reference to feathub from the contributing section's documentation.

PR #1214, opened 2017-10-16
Fix datasets which temporal coverage was reversed

noirbizarre:
This PRs fix some datasets that weren't indexed because temporal coverage `start` and `end` where switched

### DIFF
--- a/udata/migrations/2017-10-16-fix-switched-temporal-coverage.js
+++ b/udata/migrations/2017-10-16-fix-switched-temporal-coverage.js
@@ -1,0 +1,17 @@
+/*
+ * Ensure all datasets have positive temporal coverage
+ */
+
+var modified = 0;
+
+db.dataset.aggregate([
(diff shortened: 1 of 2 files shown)


udata/migrations/2017-10-16-fix-switched-temporal-coverage.js
```diff
@@ -0,0 +1,17 @@
...
+     {$project: {_id: true, temporal_coverage: true, reversed: {$cmp: ['$temporal_coverage.start','$temporal_coverage.end']}}},
+     {$match: {reversed:  1}}
+]).forEach(function({_id, temporal_coverage}) {
+    modified += db.dataset.update({_id}, {$set: {temporal_coverage: {
+        start: temporal_coverage.end,
+        end: temporal_coverage.start
+    }}}).nModified;
+});
+
+print(`${modified} datasets where having reserved temporal coverage.`);
```
abulte: s/where having/had/

PR #1216, opened 2017-10-16
Visual feedback for closed discussion (fix etalab/data.gouv.fr#31)

noirbizarre:
This PR try to highlight that a discussion has been closed:
- change the sentence "discussion started/closed..."
- display a picto on closed discussion (picto choice is debatable)

screenshot-www data dev-2017-10-16-20-27-36-803


oncletom:
What is the use to notify users a discussion is closed? Is it an information which is important to read as part of the top-level list or as part of a given discussion/thread?

noirbizarre:
Related to https://github.com/etalab/data.gouv.fr/issues/31

You can respond on a closed discussion. Users think that it's a bug when they can't respond (and notify us about it)

noirbizarre:
Having only the status in bold implies having html markup in the translation string which I try to avoir as much as possible.
I'll try to find a better alternative solution

oncletom:
Thanks, I missed the original prompt for the PR 👍 

Should be enough and we can iterate over it later on.

PR #1217, opened 2017-10-16
[WIP] Added Dataset.acronym field

noirbizarre:
This PR adds a `acronym` field on Dataset (like on Organization).
- improve search (match on acronym on both completion and search)
- improve UI/UX: uniformize acronym display accross datasets


udata/core/dataset/api_fields.py
```diff
@@ -121,6 +122,7 @@
 dataset_fields = api.model('Dataset', {
     'id': fields.String(description='The dataset identifier', readonly=True),
     'title': fields.String(description='The dataset title', required=True),
+    'acronym': fields.String(description='The dataset short name'),
```
abulte: Short name maybe misleading, maybe keep `acronym`.

udata/core/dataset/forms.py
```diff
@@ -106,6 +106,8 @@ class DatasetForm(ModelForm):
     model_class = Dataset
 
     title = fields.StringField(_('Title'), [validators.required()])
+    acronym = fields.StringField(_('Acronym'),
+        description=_('An optionnal short title or acronym'))
```
abulte: Idem. Do we want to promote "short titles"? What could those be if not an acronym?

noirbizarre: Good question. I think I'll keep only acronym for the wording right now

abulte: 👌

udata/core/dataset/api_fields.py
```diff
@@ -87,6 +87,7 @@
 
 dataset_ref_fields = api.inherit('DatasetReference', base_reference, {
     'title': fields.String(description='The dataset title', readonly=True),
+    'acronym': fields.String(description='The dataset short name', readonly=True),
```
abulte: s/short name/acronym/ I think.

PR #1222, opened 2017-10-18
Feature/better autocomplete

noirbizarre:
Should fix etalab/data.gouv.fr#30

@oncletom @abulte


udata/core/dataset/search.py
```diff
@@ -168,7 +171,16 @@ def is_indexable(cls, dataset):
                 not dataset.private)
 
     @classmethod
+    def get_suggest_weight(cls, temporal_weight, spatial_weight, featured):
+        '''Compute the suggest part of the indexation payload'''
+        featured_weight = 1 if not featured else FEATURED_WEIGHT
+        # XXX temporal_weight seems too high
```
abulte: Remove me

PR #1224, opened 2017-10-19
Post improvements

noirbizarre:
This PR brings some improvements to post display and listing:
- allows to enable discussions on posts with `POST_DISCUSSIONS_ENABLED`
- adds some basic post navigation (previous and next post, link to post list)
- make the post list discoverable and style it
- import some post page style that should be in core from `udata-gouvfr`

screenshot-www data dev-2017-10-19-10-56-16-361

screenshot-www data dev-2017-10-19-10-57-07-775

screenshot-www data dev-2017-10-19-10-57-36-610

I think this is only a first pass, navigation and search needs some styling, there might be some things missing but it's a start

### DIFF
--- a/docs/adapting-settings.md
+++ b/docs/adapting-settings.md
@@ -373,12 +373,29 @@ The padding (in percent) used by the internal provider
 The skin (set) used by the robohash provider.
 See <https://robohash.org/> for more details.
 
-## AVATAR_ROBOHASH_BACKGROUND
+### AVATAR_ROBOHASH_BACKGROUND
 
 **default**: `'bg0'` (transparent background)
 
 The background used by the robohash provider.
 See <https://robohash.org/> for more details.
+
+## Posts configuration
+
+Theses settings allow you to customize the post feature.
+
+### POST_DISCUSSIONS_ENABLED
+
+**default** `False`
+
+Whether or not discussions should be enabled on posts
+
+### POST_DEFAULT_PAGINATION
+
+**default** `20`
+
+The default page size for post listing
+
 
 ## Example configuration file
 
--- a/udata/core/discussions/tasks.py
+++ b/udata/core/discussions/tasks.py
@@ -3,7 +3,9 @@ from __future__ import unicode_literals, absolute_import
 
 from udata import mail
 from udata.i18n import lazy_gettext as _
-from udata.models import Dataset, Reuse
+from udata.core.dataset.models import Dataset
+from udata.core.reuse.models import Reuse
+from udata.core.post.models import Post
 from udata.tasks import task, get_logger
 
 from .signals import (
@@ -26,7 +28,7 @@ def connect(signal):
 
 
 def owner_recipients(discussion):
-    if discussion.subject.organization:
+    if getattr(discussion.subject, 'organization', None):
         return [m.user for m in discussion.subject.organization.members]
     else:
         return [discussion.subject.owner]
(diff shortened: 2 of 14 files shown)


docs/adapting-settings.md
```diff
@@ -373,13 +373,30 @@ The padding (in percent) used by the internal provider
...
 
+## Posts configuration
+
+Theses settings allow you to customize the post feature.
+
+### POST_DISCUSSIONS_ENABLED
+
+**default** `False`
+
+Wether or not discussions should be enabled on posts
```
abulte: s/wether/whether/

Wether is a prime example of a word that will slip past the spell check. It is easily confused with two of its homonyms, whether and weather. Flying fingers find it easy to miss the single letter that separates them. Unless you’re a farmer, you might not even know that wether is either a:

male sheep or ram (the Oxford Dictionary of Etymology traces its roots to Old English, Old High German, Old Norse and Goth)
or a:
castrated ram or billy goat (according to A Word A Day).

😄

udata/core/discussions/tasks.py
```diff
@@ -3,7 +3,9 @@
 
 from udata import mail
 from udata.i18n import lazy_gettext as _
-from udata.models import Dataset, Reuse
+from udata.core.dataset.models import Dataset
```
abulte: why the switch from `udata.models`? Just curious.

udata/core/post/views.py
```diff
@@ -28,17 +28,32 @@ class ProtectedPostView(PostView):
...
+    newer = others(created_at__gt=post.created_at).order_by('created_at')
+    return theme.render('post/display.html',
+                        post=post,
+                        previous_post=older.first(),
+                        next_post=newer.first())
 
 
 @sitemap.register_generator
 def sitemap_urls():
+    yield 'posts.list_redirect', {}, None, "weekly", 0.6
```
abulte: Isn't the code below unreachable?

udata/frontend/views.py
```diff
@@ -65,6 +64,25 @@ def get_context(self):
         context[self.context_name] = self.get_queryset()
         return context
 
+    @property
+    def page(self):
+        try:
+            params_page = int(request.args.get('page', 1) or 1)
+            return max(params_page, 1)
+        except:
```
abulte: More precise exception catching possible?

udata/frontend/views.py
```diff
@@ -65,6 +64,25 @@ def get_context(self):
...
+            # Failsafe, if page cannot be parsed, we falback on first page
+            return 1
+
+    @property
+    def page_size(self):
+        try:
+            params_page_size = request.args.get('page_size',
+                                                self.default_page_size)
+            return int(params_page_size or self.default_page_size)
+        except:
```
abulte: Idem.

docs/adapting-settings.md
noirbizarre: :joy:

udata/core/discussions/tasks.py
noirbizarre: As much as possible and when I can think about it, I try to remove improt from the `udata.models` facade:
- it's hard to refactor and introspect
- it's one of the circular dependencies sources 
- it's one of the things that increase the test duration

udata/core/post/views.py
noirbizarre: I don't think so. After this `yield` statement, I believe the interperter is processing the for loop.
What do you think prevent it ?

abulte: Well, I always thought of `yield` as a kind of `return` but after a quick test I realize I was wrong ;-)

udata/core/discussions/tasks.py
abulte: 👍

oncletom: Looks like something which could benefit of an issue scheduled for `1.3` or something :-)

PR #1233, opened 2017-10-23
Optimize reindex

noirbizarre:
This PR brings a huge performance improvement on reindexing as well as improved errors handling and a more flexible and unique `index` command.

## A single command

Both commands (`search init` and `search reindex`) have been merged into a single `search index` command.
This command is more flexible (models as variable arguments, plural or singular, delete obselete index by default...) and more powerful (allows to reindex multiple models in a single pass).

Previous commands still exists but will show a deprecation warning when used.

```shell
udata search init -df
WARNING: reindex command will be removed in udata 1.4, use index command instead
-> Initiliazing index "udata-prod-2017-10-21-15-52"
```

According to [deprecation policy](http://udata.readthedocs.io/en/stable/versionning/#deprecation-policy) removal should be in udata 1.4 (An issue should be created).

### Full reindexing

Simply execute the command without arguments

```
# Index deletion with user confirmation prompt
udata search index
# No confirmation prompt on deletion
udata search index -f
# Previous index deleted (or unfinished one on error)
udata search index -k
```

### Partial reindex

Models are now simple arguments instead of the `-t` option from the previous `search reindex` command.

This means that to only reindex reuses and organizations you can do it in a single pass with:

```shell
udata search index reuse organization
```

instead of two passes before:

```shell
udata search reindex -t reuse
udata search reindex -t organization
```

The command also accept plural forms as it is a common error:

```shell
udata search index datasets
```

## Performances

Performance improvements comes from following [Tune for indexing speed guide](https://www.elastic.co/guide/en/elasticsearch/reference/master/tune-for-indexing-speed.html) and the [bulk indexing section from the Update settings documentation](https://www.elastic.co/guide/en/elasticsearch/reference/master/indices-update-settings.html#bulk).

**Consequences:**
- indexation is way faster
- resources consumption on the Elasticsearch side (both CPU and RAM) remains low (and way lower than before) during the whole reindexing

### Before

```shell
time udata search init -df
-> Initiliazing index "udata-prod-2017-10-21-15-59"
-> Indexing Dataset objects
-> Indexing GeoZone objects
-> Indexing Organization objects
-> Indexing Reuse objects
-> Indexing User objects
-> Creating alias "udata-prod" on index "udata-prod-2017-10-21-15-59"

real	23m28,467s
user	11m39,063s
sys	0m24,022s
```

### After

```shell
time udata search init
-> Initiliazing index "udata-prod-2017-10-21-13-41"
-> Indexing Dataset objects
-> Indexing GeoZone objects
-> Indexing Organization objects
-> Indexing Reuse objects
-> Indexing User objects
-> Creating alias "udata-prod" on index "udata-prod-2017-10-21-13-41"

real	6m20,919s
user	4m56,989s
sys	0m6,892s
```

## Error handling

This PR improves error handling on indexation. No more ugly stacktrace, only the errors details (which where not displayed before by the way).
Now, both commands also properly handle kill signals and keyboard interrupt.

In case of error, the unfinished index is now properly deleted and so avoid having a lots of unfinished indexes (consuming ES memory).

```shell
-> Initiliazing index "udata-prod-2017-10-21-15-52"
-> Indexing Dataset objects
^C
WARNING: Interrupted by signal
-> Removing index udata-prod-2017-10-21-15-52
```

For debugging purpose, you can keep the unfinished index with the `-k/--keep` parameter.

## Documentation

At last, the `search index` command is now documented in the "Administrative tasks" section.

### DIFF
--- a/docs/administrative-tasks.md
+++ b/docs/administrative-tasks.md
@@ -114,3 +114,41 @@ $ udata job run job-name -a arg1 arg2 -k key1=value key2=value
 **Note**: this is a low level command.
 Most of the time, you won't need it because there will be a dedicated command
 to perform the task you need.
+
+
+## Reindexing data
+
+Sometimes, you need to reindex data (in case of model breaking changes, workers defect...).
+You can use the `udata search index` command to do so.
+
+This command supports both full reindex without arguments and partial with model names as arguments:
+
+```shell
+# Reindex everything
+udata search index
+# Only reindex reuses and organizations
+udata search index reuses organizations
+```
+
+By default the command deletes the previous index in case of success or the new unfinished index in case of error but you can ask to keep indexes with the `-k/--keep` parameter
+
+```shell
+# Reindex everything but keep the old index
+udata search index -k
+```
+
+When used from an interactive terminal the command also prompt for deletion confirmation if an index with the same name already exists. This can be bypassed with the `-f/--force` parameter.
+
+```shell
+# Reindex everything and delete old index
+udata search index -f
+```
+
+It's possible to do a partial reindex by providing models (both singular and plural are supported) as arguments:
+
+```shell
+# Only reindex datasets and reuses (plural form)
+udata search index datasets reuses
+# Only reindex datasets and reuses (singular form)
+udata search index dataset reuse
+```
--- a/udata/search/commands.py
+++ b/udata/search/commands.py
@@ -2,7 +2,10 @@
 from __future__ import unicode_literals
 
 import logging
+import sys
+import signal
 
+from contextlib import contextmanager
 from datetime import datetime
 
 from flask_script import prompt_bool
@@ -10,6 +13,7 @@ from flask_script.commands import InvalidCommand
 
 from udata.commands import submanager
 from udata.search import es, adapter_catalog
+
 
 from elasticsearch.helpers import reindex as es_reindex, streaming_bulk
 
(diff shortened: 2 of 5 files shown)


docs/administrative-tasks.md
```diff
@@ -114,3 +114,41 @@ $ udata job run job-name -a arg1 arg2 -k key1=value key2=value
 **Note**: this is a low level command.
 Most of the time, you won't need it because there will be a dedicated command
 to perform the task you need.
+
+
+## Reindexing data
+
+Sometimes, you need to reindex data (in case model breaking changes, defect of workers...).
```
abulte: s/in case/in case of/

abulte: s/defect of worker/worker defect/

docs/administrative-tasks.md
```diff
@@ -114,3 +114,41 @@ $ udata job run job-name -a arg1 arg2 -k key1=value key2=value
...
+You can use the `udata search index command` to do so.
```
abulte: s/`udata search index command`/`udata search index` command/

docs/administrative-tasks.md
```diff
@@ -114,3 +114,41 @@ $ udata job run job-name -a arg1 arg2 -k key1=value key2=value
...
+
+This command both support full reindex without arguments or partial with model names as arguments:
```
abulte: *supports both* full reindex without arguments *and* partial

docs/administrative-tasks.md
```diff
@@ -114,3 +114,41 @@ $ udata job run job-name -a arg1 arg2 -k key1=value key2=value
...
+
+```shell
+# Reindex everything
+udata search index
+# Only reindex reuses and organizations
+udata search index reuses organizations
+```
+
+By default the command does delete the previous index in case of success or the new unfinished index in case of error but you can ask to keep indexes with the `-k/--keep` parameter
```
abulte: s/does delete/deletes/

docs/administrative-tasks.md
```diff
@@ -114,3 +114,41 @@ $ udata job run job-name -a arg1 arg2 -k key1=value key2=value
...
+```
+
+When used from an interactive terminal the command also prompt for deletion confirmation if an index with the same name already exists. This can be bypassed with the `-f/--force` parameter.
+
+```shell
+# Reindex everything and delete old index
+udata search index -f
+```
+
+It's possible to do a partial reindex by providing models (support both singular or plural) as arguments:
```
abulte: both singular and plural are supported

udata/search/commands.py
```diff
@@ -35,6 +40,21 @@ def iter_adapters():
...
+def iter_qs(qs, adapter):
+    '''Safely iterate over a DB QuerySet yielding ES documents'''
+    for obj in qs.no_dereference().timeout(False):
+        if adapter.is_indexable(obj):
+            doc = adapter.from_model(obj).to_dict(include_meta=True)
+            yield doc
+
+
+def iter_for_index(docs, index_name):
+    '''Iter over ES documents ensuring a given index'''
```
abulte: s/Iter/Iterate/

udata/search/commands.py
```diff
@@ -44,14 +64,41 @@ def index_model(index_name, adapter):
...
+    See: https://www.elastic.co/guide/en/elasticsearch/reference/master/indices-update-settings.html#bulk
+    '''  # noqa
+    es.indices.put_settings(index=index_name, body={
+        'index': {
+            'refresh_interval': '-1'
+        }
+    })
+
+
+def enable_refresh(index_name):
```
abulte: Maybe put `refresh_interval` as a parameter with a default value of `1s`? In case this needs to be changed/configured.

udata/search/commands.py
```diff
@@ -44,14 +64,41 @@ def index_model(index_name, adapter):
...
+    '''
+    Enable refresh after indexing and force merge
```
abulte: Enable refresh and force merge. Used after indexing. ?

udata/search/commands.py
```diff
@@ -76,63 +123,105 @@ def set_alias(index_name, delete=True):
...
+        if model not in doc_types_names:
+            log.error('Unknown model %s', model)
+            sys.exit(-1)
+
     log.info('Initiliazing index "{0}"'.format(index_name))
     if es.indices.exists(index_name):
-        if force or prompt_bool(('Index {0} will be deleted, are you sure ?'
-                                 .format(index_name))):
+        if IS_INTERACTIVE and not force:
+            msg = 'Index {0} will be deleted, are you sure ?'
```
abulte: s/ ?/?/

udata/search/commands.py
noirbizarre: Done :+1:

PR #1239, opened 2017-10-24
Cleanup resources checksum

abulte:
Some resources (5 on my local copy of the prod DB) have a `checksum` field with a `type` but without `value`. This results in errors such as https://sentry.data.gouv.fr/etalab/www-datagouvfr/issues/137603/ when saving a resource or dataset.

This migration removes the faulty `checksum` fields.

### DIFF
--- a/udata/migrations/2017-10-24-cleanup-resources-checksum.js
+++ b/udata/migrations/2017-10-24-cleanup-resources-checksum.js
@@ -1,0 +1,20 @@
+/*
+ * Remove checksum from resources where it has no value
+ */
+
+var nbCleaned = 0;
+
+db.dataset.find({
+    'resources.checksum': {$exists: true},
+    'resources.checksum.value': {$exists: false},
+}).forEach(dataset => {
+    dataset.resources.forEach(resource => {
+        if (resource.checksum && !resource.checksum.value) {
+            delete resource.checksum;
+            nbCleaned++;
+        }
+    });
+    db.dataset.save(dataset);
+});
+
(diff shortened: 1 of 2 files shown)


abulte:
NB: the faulty datasets are old (2015) and there are only two of them, so I do not think there is an underlying issue anymore.

udata/migrations/2017-10-24-cleanup-resources-checksum.js
```diff
@@ -0,0 +1,22 @@
+/*
+ * Remove checksum from resources where it has no value
+ */
+
+var nbCleaned = 0;
+
+db.dataset.find({
+    $and: [
```
noirbizarre: Not need for `$and`, this is the default operator, ie.
```javascript
{'resources.checksum': {$exists: true}, 'resources.checksum.value': {$exists: false}}
```

udata/migrations/2017-10-24-cleanup-resources-checksum.js
```diff
@@ -0,0 +1,22 @@
...
+        {'resources.checksum': {$exists: true}},
+        {'resources.checksum.value': {$exists: false}},
+    ]
+}).forEach(dataset => {
+    dataset.resources.forEach(resource => {
+        if (resource.checksum && !resource.checksum.value) {
+            delete resource.checksum;
+            nbCleaned++;
+        }
+        db.dataset.save(dataset);
```
noirbizarre: You save the dataset for each resource, do it only once by moving the `db.dataset.save()` outside the `dataset.resources.forEach()`

PR #1240, opened 2017-10-24
Trigger a link check on a resource from the admin

abulte:


js/components/dataset/resource/availability.vue
```diff
@@ -1,28 +1,58 @@
 <template>
-    <span v-if="availability === 'AVAILABLE'" class="badge bg-green">✓</span>
-    <span v-if="availability === 'NOT_AVAILABLE'" class="badge bg-red">×</span>
-    <span v-if="availability === 'UNKNOWN'" class="badge bg-gray">?</span>
+    <div :title="_('Click to trigger a new check')" v-if="!checking" @click.stop="refresh()">
```
davidbgk: Even if available?

js/components/dataset/resource/availability.vue
```diff
@@ -1,28 +1,58 @@
...
                     return 'UNKNOWN';
             }
+        },
+        refresh() {
+            this.checking = true;
+            API.datasets.check_dataset_resource({dataset: this.dataset.id, rid: this.resource.id}, res => {
+                this.availability = this.computeAvailability(res.obj['check:available']);
+                this.checking = false;
+            }, err => {
+                console.log('Something went wrong with the check', err);
```
davidbgk: Set the availability to unknown in that case?

js/components/dataset/resource/availability.vue
abulte: It's far fetched, but maybe we want to be able to flag immediately a non available resource.

js/components/dataset/resource/availability.vue
abulte: 👍

js/components/dataset/resource/availability.vue
davidbgk: I mean, I don't get the point of refreshing the availability of an available resource?

abulte: This feature is meant to refresh the check status of a resource when an admin knows the status wrong. Like I said it's far fetched, but maybe we know a resource is unavailable and we want the status to reflect that immediately.

Plus, not displaying the feature when the resource is available would add some complexity and I don't think it's worth it.

I see your point though ;-)

davidbgk: Ok, get it 👍

abulte:
TODO: this feature is almost useless as is, since it still relies on cached result from the link checker. This should force the retrieval from the link checker, whatever the age of the cache.

davidbgk:
About that TODO, I wonder if the cache should be done at the udata-croquemort level or at the croquemort level, both have pros and cons :)

abulte:

Issue #1085, opened 2017-08-10
Improve the 'Resource' card in a Dataset page


oncletom:
# Current situation

Dataset improvements are happening in #1072.

## Resource

![image](https://user-images.githubusercontent.com/138627/29175243-6a5335e8-7de0-11e7-8709-1a861a768bcf.png)

The details of a resource are:

| Name | Visible | Type |
| --- | --- | --- |
| checksum | no | `string` |
| created_at | no* | `date` |
| description | no* | `text` |
| filesize | no | `integer` |
| filetype | no* | `enum` |
| format | yes* | `string` |
| id | no | `string` |
| is_available | yes | `object` |
| last_modified | yes | `date` |
| latest | no* | `string` |
| metrics | no | `object` |
| mime | no* | `string` |
| published | no* | `date` |
| title | yes | `string` |
| url | no* | `string` |

* these informations are displayed in a popup, after interacting with a resource.

# Ideas Generation

## Idea 1

![shape 12](https://user-images.githubusercontent.com/138627/29175505-23f61e48-7de1-11e7-8e1e-8aff739b81e5.jpeg)

Highlighted features

- show **resource type** (tabular, structured, sheets, shapes)
- show **filesize** in 4 chars
- show **popularity**
- **copy to clipboard** (direct URL or Permalink)

## Idea 2

![shape 12 7](https://user-images.githubusercontent.com/138627/29175498-20f30e7c-7de1-11e7-9bca-8dbdd745d94b.jpeg)

Highlighted features

- show **download** options (full or **geo subset**)
- **copy to clipboard**
- show **filesize** in 4 chars
- show **resource type** (tabular, structured, sheets, shapes)
- show **date coverage**

## Idea 3

![shape 12 4](https://user-images.githubusercontent.com/138627/29175390-d844ad5c-7de0-11e7-8a24-b3de17f506e9.jpeg)

Highlighted features

- show **column count** (for tabular-like resource)
- optional **column documentation**
- show **filesize** in 4 chars
- show **popularity**

## Idea 4

![shape 12 2](https://user-images.githubusercontent.com/138627/29175551-3da35400-7de1-11e7-8a04-1fc19bf6e369.jpeg)

Highlighted features

- **aggregate historical resources** (based on pattern/string distance/whatever)
- optionally show **other resources** (the `...` button)

## Idea 5

[...]

PR #1252, opened 2017-10-31
Improve Cards consistency

oncletom:
I have made Cards look a bit more even and made a few changes:

- `.rollover` removal — this is an anti-pattern on mobile
- Card namespacing — bottom border color is defined by the namespace they live in
- grid alignment
- Cards are the same both in a page content block and in the Community Resources block
- Homepage and Topics carousels are replaced with rows of Cards

Some non-visual related changes:

- Datasets are displayed by row of 2 in an Organization page (was 3)
- Organisation datasets and reuses are sorted by time coverage freshness and popularity
- 'Resources count' has been added to the Dataset card
- `frequency` has been removed from the Dataset card

# What's missing?

- [x] What about the `daterange` helper strategy? I have probably oversimplified it by keeping only the years.
- [x] Shall we use a Dataset Card display in the search results as well?
- [x] Reuse Card misses the owner informations (avatar, name)
- [x] Certified Logo
- [x] Reuse Card misses the cover
- [ ] Accessibility check

# Issues spotted

- [x] `list-group-item` elements are messed up in search facets
- [x] cards are stacked when floating on a new line (solved with flexbox)
- [ ] _add a reuse in dataset modal_ cards layout is broken

# Screenshots

## Datasets in Search

image

## Various cards in Timeline

image

## Datasets in Org

image

## Users in Org

image

## Reuses in Org (`features=[]`)

image

## Reuses in Community Resources (`features=['preview']`)

image

## Dataset in Reuse

image

refs #1085


abulte:
@oncletom pics or it did not happen ;-) Seriously, I'd love to see before/after screenshots.

oncletom:
@abulte 👀 👆

js/components/dataset/filters.js
```diff
@@ -13,26 +13,11 @@ export default {
...
-            //     start = short_month(value.start)
-            //     if delta.days > 31:
-            //         end = short_month(value.end)
-            // else:
-            //     start = short_day(value.start)
-            //     if value.start != value.end:
-            //         end = short_day(value.end)
             return end_label
-                ? this._('{start} to {end}', {start:start_label, end:end_label})
+                ? this._('{start_label}-{end_label}', {start_label, end_label})
```
abulte: This probably does not need to be translated anymore?

udata/core/organization/views.py
```diff
@@ -76,8 +76,8 @@ def get_context(self):
         if self.organization.deleted and not can_view.can():
             abort(410)
 
-        datasets = Dataset.objects(organization=self.organization).visible()
-        reuses = Reuse.objects(organization=self.organization).visible()
+        datasets = Dataset.objects(organization=self.organization).order_by('-temporal_coverage.end', '-metrics.reuses', '-metrics.followers').visible()
+        reuses = Reuse.objects(organization=self.organization).order_by('-metrics.reuses', '-metrics.followers').visible()
```
abulte: There is a `metrics.reuses` on a `Reuse` object?

udata/core/organization/views.py
```diff
@@ -76,8 +76,8 @@ def get_context(self):
         if self.organization.deleted and not can_view.can():
             abort(410)
 
-        datasets = Dataset.objects(organization=self.organization).visible()
-        reuses = Reuse.objects(organization=self.organization).visible()
+        datasets = Dataset.objects(organization=self.organization).order_by('-temporal_coverage.end', '-metrics.reuses', '-metrics.followers').visible()
```
abulte: I think this goes over the 80 characters line limit (same below).

udata/frontend/helpers.py
```diff
@@ -279,19 +279,11 @@ def daterange(value):
...
-    else:
-        start = short_day(value.start)
-        if value.start != value.end:
-            end = short_day(value.end)
-    return _('%(start)s to %(end)s', start=start, end=end) if end else start
+    start = value.start.year
+    if delta.days > 365:
+        end = value.end.year
+
+    return _('%(start)s-%(end)s', start=start, end=end) if end else start
```
abulte: No translation needed?

udata/templates/dataset/card.html
```diff
@@ -3,11 +3,11 @@
 {% from theme('macros/certified.html') import badge_if_certified with context %}
 <div class="card dataset-card">
     {% if dataset.organization %}
-        <a class="card-logo" href="{{ dataset_url }}">
+        <div class="card-logo">
```
abulte: Why drop the link to the org? Confusing?

udata/templates/dataset/card.html
```diff
@@ -17,21 +17,20 @@
     <div class="card-body">
         <h4>
             <a href="{{ dataset_url }}" title="{{dataset.title}}">
-                {{ dataset.title|truncate(120) }}
+                {{ dataset.title }}
```
abulte: I trust you tested with very long titles?

udata/templates/reuse/card.html
```diff
@@ -1,34 +1,32 @@
 {% cache cache_duration, 'reuse-card', reuse.id|string, g.lang_code %}
-<div class="card reuse-card">
+{% set features = features or [] %}
+<div class="card reuse-card {% if 'preview' in features %}card--with-cover{% endif %}">
+    {% if 'preview' in features %}
```
abulte: What is `features.preview`?

udata/templates/dataset/card.html
abulte: Or you're truncating in CSS?

js/components/dataset/filters.js
oncletom: Yes, correct!

udata/core/organization/views.py
oncletom: Yes.

Not sure `metrics.reuses` exists but these elements are available for display.

udata/core/organization/views.py
oncletom: Yes you are right. Can I chain calls on multiple lines? If so 👍

udata/templates/dataset/card.html
oncletom: They are truncated in CSS indeed :-)

udata/templates/reuse/card.html
oncletom: Maybe the label is wrong — that's what makes the card to display the reuse cover (in the Community Reuses examples) vs. the Reuse in an org listing.

I'm happy to stick with one design only to mitigate complexity.

udata/core/organization/views.py
abulte: With a `\` before the break line, yes.

abulte: https://atom.io/packages/flake8

udata/templates/reuse/card.html
abulte: Ok, why not!

abulte: (why not for 2 designs, seems to make sense)

udata/templates/dataset/card.html
oncletom: I found that confusing indeed: the link targets the Dataset whereas the `alt` attribute and the image signpost the org.

oncletom:
Applied to Search results:

udata/templates/home.html
```diff
@@ -9,17 +9,11 @@
...
 {% block content %}
+{% if last_post %}
 <div id="home-carousel" class="carousel slide megarousel">
-  <!-- Indicators -->
-    <ol class="carousel-indicators">
-        <li data-target="#home-carousel" data-slide-to="0" class="active"></li>
-        <li data-target="#home-carousel" data-slide-to="1"></li>
-        <li data-target="#home-carousel" data-slide-to="2"></li>
-    </ol>
-
```
pblayo: Nice change!

pblayo:
It's nice to have a check list. I'm not sure what would be the best way to test accessibility (the last item) though.

udata/core/organization/views.py
noirbizarre: I believe this also require settings the proper index on dataset model

udata/core/organization/views.py
noirbizarre: I agree with @abulte : there is no `metrics.reuses` on the `Reuse` objects (only on datasets et organizations), but there is instead a `metrics.datasets`

udata/frontend/helpers.py
```diff
@@ -277,21 +277,16 @@ def daterange(value):
...
-    else:
-        start = short_day(value.start)
-        if value.start != value.end:
-            end = short_day(value.end)
-    return _('%(start)s to %(end)s', start=start, end=end) if end else start
+    start = format_date(value.start, date_format)
+    if delta.days > 365:
+        end = format_date(value.end, date_format)
+
+    return '{start!s}–{end!s}'.format(start=start, end=end) if end else start
```
noirbizarre: I think this is a little bit over simplified. This is open to discussion, but we had some cases where a little bit more details ware useful:
- for the elections, daily statistics were uploaded and the end of coverage allowed users to see than data have been updated for the current day
- still for the elections (and any other dataset which is not a time serie) displaying a year for a punctual event seems erroneus

Anyway, I find it easier to read with years only.

**Proposal**: keep the `daterange` filter as is and add a new `yearrange` filter. This will allows us to display more details when/where it's needed/pertinent

udata/templates/home.html
```diff
@@ -42,107 +36,44 @@ <h1 class="page-header">{{last_post.name}}
 
             </div>
         </div>
-        {% endif %}
-
-        <div class="item">
-            <div class="container">
 
-                <div class="row">
```
noirbizarre: These removals seem to be a major home page change. Can we have some screenshots showing before/after this change ?

pblayo:
We decided to split the PR between :
- cards
- search
- carousel

About cards, next things to be worked on :
- colors
- alignment

taniki:
here is the current state of the default template with @oncletom changes.

I wasn't theoretically fond to display card view in search results but thomas' works make it very appealling and consistent.

screenshot-2018-1-19 home - udata

screenshot-2018-1-19 search - udata

screenshot-2018-1-19 datasets - udata

abulte:
Closing this since we splitted it in multiple PRs and implemented a lot of it (at last). Thanks @oncletom!

PR #1259, opened 2017-11-07
Sidebar improvements: use list for tags and button for profile view

oncletom:
These are minor changes:

- the `+` link in owner/org sidebar panel is often mismatched as an "expand text" button otherwise.
- tags can benefit of using the `.label` class instead of `.btn` as well as being enclosed in an `<ul>` element

# Screenshots

## Reuse sidebar

image

## Dataset sidebar

image

### DIFF
--- a/less/udata/panels.less
+++ b/less/udata/panels.less
@@ -2,3 +2,14 @@
     background: rgba(255, 255, 255, 0.1);
     border: none;
 }
+
+.panel-body {
+    .tags {
+        .list-inline;
+            margin-left: 0;
+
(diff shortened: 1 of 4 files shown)


abulte:
@oncletom no need for style modification on default theme (like in https://github.com/opendatateam/udata/pull/1259)?

abulte:
Styled by default seems reasonable to me.

oncletom:
I cherry-picked the styles commit changes and removed it from etalab/udata-gouvfr#230 👍

PR #1260, opened 2017-11-07
Variable frequency link check strategy

abulte:
This PR introduces a variable caching strategy for check results. The base ideas are:
- check unavailable resources more often than available ones
- resources that have been available for a long time should be checked less and less frequently
- overall, trigger less checks

To achieve this:
- there is a new attribute `needCheck` on a resource JSON-LD in order to mutualize the logic between front and back
- there is a new extra `check:count-availability` that counts the successive checks with the same `check:available` value

The rules are (for now, this should be discussed here):
- available resources are checked at `LINKCHECKING_MIN_CACHE_DURATION *  check:count-availability` intervals
- unavailable resources are checked at `LINKCHECKING_MIN_CACHE_DURATION` intervals

Maybe we need a threshold of `check:count-availability` for unavailable resources above which they're considered "dead" and checked less frequently. Or maybe a linear progression of check frequency for unavailable resources and an exponential one for available resources.

NB: the cache verification is now done by the "clients" of the `linkchecker` module (i.e. front and command line tools for now) rather than the module itself.


udata/core/dataset/models.py
```diff
@@ -217,6 +217,27 @@ def check_availability(self):
         '''
         return self.extras.get('check:available', 'unknown')
 
+    def is_need_check(self):
```
noirbizarre: I think the `is_` prefix is weird and can be removed (I think this is a convention used on some project but as not in use here I think it's more readable without)

udata/settings.py
```diff
@@ -166,7 +166,7 @@ class Defaults(object):
 
     LINKCHECKING_ENABLED = True
     LINKCHECKING_IGNORE_DOMAINS = []
-    LINKCHECKING_CACHE_DURATION = 60 * 5  # in seconds
+    LINKCHECKING_MIN_CACHE_DURATION = 60 * 5  # in seconds
```
noirbizarre: Maybe it make more sense to express the duration in minutes (ie. checking multiple times by minutes does not seems realistic)

udata/core/dataset/models.py
```diff
@@ -217,6 +217,27 @@ def check_availability(self):
...
+
+        We check unavailable resources often. Available resources are checked
+        less and less frequently based on their historical availability.
+        '''
+        cache_duration = current_app.config['LINKCHECKING_MIN_CACHE_DURATION']
+        count_availability = self.extras.get('check:count-availability', 1)
+        is_available = self.check_availability()
+        if is_available == 'unknown':
+            return True
+        if is_available:
```
noirbizarre: `elif` (I know it produces the same result but it's just to mark the 3 states)

udata/linkchecker/tasks.py
```diff
@@ -53,5 +53,8 @@ def check_resources(self, number):
         resource_obj = get_by(dataset_obj.resources, 'id', rid)
         log.info('Checking resource %s (%s/%s)',
                  resource_id, idx + 1, nb_resources)
-        check_resource(resource_obj)
+        if resource_obj.is_need_check():
+            check_resource(resource_obj)
+        else:
+            log.info("\t-> Skipping this resource, cache is fresh enough.")
```
noirbizarre: Be careful with ad-hoc logging formating, it won't produce the same result everywhere.

udata/settings.py
abulte: I agree. But it won't be backward compatible. Do we have a doctrine on that?

udata/core/dataset/models.py
abulte: 👍

udata/core/dataset/models.py
abulte: 👍

udata/linkchecker/tasks.py
abulte: 👍

udata/settings.py
abulte: I changed the variable name anyway, so not really backward compatible from the beginning ;-) I'll make the change.

PR #1261, opened 2017-11-07
Check resource.format existence before computing closed_format

abulte:
This was causing the computing of `dataset.quality` to silently (sic) fail when one resource had no format declared.

### DIFF
--- a/udata/core/dataset/models.py
+++ b/udata/core/dataset/models.py
@@ -206,7 +206,7 @@ class ResourceMixin(object):
     @property
     def closed_format(self):
         """Return True if the specified format is in CLOSED_FORMATS."""
-        return self.format.lower() in CLOSED_FORMATS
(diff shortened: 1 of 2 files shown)


udata/core/dataset/models.py
```diff
@@ -206,7 +206,7 @@ def clean(self):
     @property
     def closed_format(self):
         """Return True if the specified format is in CLOSED_FORMATS."""
-        return self.format.lower() in CLOSED_FORMATS
+        return self.format and self.format.lower() in CLOSED_FORMATS
```
noirbizarre: So if format is not defined we consider that this is an open format :thinking: 
Maybe the opposite behavior makes more sense (ie. consider undefined format as closed format by default and so force user to define format if they don't want anymore the quality warning)

abulte:
@Axel indeed 👍 I'll tacle this properly in another PR, at least the hotfix is merged.

PR #1271, opened 2017-11-20
[doc] Explain that ./venv/bin is supposed to be in the path

pblayo:


pblayo:
Yes I'd say that adding `. venv/bin/activate` in the doc might help not to forget to do it...

PR #1272, opened 2017-11-24
Validate mongo config

noirbizarre:
This PR performs MongoDB parameters validation and test that settings are properly parsed.
This avoid silent errors.

Deprecation warning are now logged too.

### DIFF
--- a/udata/app.py
+++ b/udata/app.py
@@ -123,7 +123,23 @@ class UDataJsonEncoder(json.JSONEncoder):
         return super(UDataJsonEncoder, self).default(obj)
 
 
-def create_app(config='udata.settings.Defaults', override=None):
+WARNING_LOGGERS = 'elasticsearch', 'requests'
+
+
+def init_logging(app):
+    logging.captureWarnings(True)  # Display warnings
+    debug = app.debug or app.config.get('TESTING')
+    log_level = logging.DEBUG if debug else logging.WARNING
+    app.logger.setLevel(log_level)
+    for name in app.config['PLUGINS']:
+        logging.getLogger('udata_{0}'.format(name)).setLevel(log_level)
+    for logger in WARNING_LOGGERS:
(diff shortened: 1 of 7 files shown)


udata/models/__init__.py
```diff
@@ -102,15 +105,38 @@ def resolve_model(self, model):
...
+        config['MONGODB_HOST'] = config['MONGODB_HOST_TEST']
+    else:
+        parsed_url = urlparse(config['MONGODB_HOST'])
+        parsed_url = parsed_url._replace(path='%s-test' % parsed_url.path)
+        config['MONGODB_HOST'] = parsed_url.geturl()
+    validate_config(config)
+
+
 def init_app(app):
     # use `{database_name}-test` database for testing
```
abulte: Comment needs to be removed I think.

abulte: Or moved.

noirbizarre:
OK comments moved.

I also initialized logging earlier allowing to log deprecations.

udata/app.py
```diff
@@ -123,7 +123,23 @@ def default(self, obj):
...
+
+
+def init_logging(app):
+    logging.captureWarnings(True)  # Display warnings
+    debug = app.debug or app.config.get('TESTING')
+    log_level = logging.DEBUG if debug else logging.WARNING
+    app.logger.setLevel(log_level)
+    for name in app.config['PLUGINS']:
+        logging.getLogger('udata_{0}'.format(name)).setLevel(log_level)
+    for logger in WARNING_LOGGERS:
```
abulte: Maybe explain why we do that?

noirbizarre: :+1:

noirbizarre: Done (and renamed into `VERBOSE_LOGGERS` to be more explicit

abulte: 👌

PR #1273, opened 2017-11-26
Fix the doc about the cli option to specify licenses (fix #1270)

### LINKED ISSUE #1270: Optional argument of "udata licenses" is broken
The second argument of `udata licenses` is broken :

    $ udata licenses https://www.data.gouv.fr/api/1/datasets/licenses
    ...
    udata: error: too many arguments

It works without the second argument because udata fetches a default list of licenses :

    $ udata licenses
    -> Dropping existing licenses
    ...
    -> Done

pblayo:


pblayo:
Yes I agree that `filename` is not the best for an option that points to an url. I'm ok with `source`.

PR #1278, opened 2017-11-27
Add venv and .idea to .gitignore

pblayo:


pblayo:
@abulte : the reason why I'd add it to the local .gitignore is because I was surprised by an addition of .idea files through a `git commit -a`, without any previous `git add` related to these files (as @noirbizarre guessed). I think it'd avoid such a bad experience to another PyCharm user.

abulte:
Yes, that’s the point of `.gitignore`all right ;-) Still, I think that as a PyCharm user you should add it to a global `.gitignore` — regardless of this PR. No more surprise on any project.

PR #1283, opened 2017-11-29
display date of comments [FIX #1021]

### LINKED ISSUE #1021: Ajout date des réponses dans les discussions sur un dataset
Dans les discussions, on a juste une date pour l'ouverture d'un sujet, mais pas de date pour les échanges suivants... ça serait pas mal de l'avoir pour voir la réactivité dans les réponses.

taniki:


js/components/discussions/thread.vue
```diff
@@ -148,7 +151,11 @@ export default {
             } else {
                 this.$scrollTo(this);
             }
+        },
+        formatDate(val) {
+            return moment(val).format('LL')
```
abulte: Unfortunately as far as I'm concerned, we're not using StandardJS conventions, so I think we need a `;` here.

abulte:
Screenshots? 🙇

taniki:
@abulte 

image

taniki:
I have added some missing semicolons in the other part of the template. After some thought, this certainly could be a global filter.

taniki:
image

@abulte there is a legacy `.ellipsis` class on that part of the content. I am thinking about removing it before I might miss a point about why it what there first.

abulte:
It's not too bad as is.

js/components/discussions/thread.vue
```diff
@@ -97,10 +100,10 @@ export default {
             return `discussion-${this.discussion.id}`;
         },
         createdDate() {
-            return moment(this.discussion.created).format('LL')
+            return moment(this.discussion.created).format('LL');
```
pblayo: Was it intended to modify this line to add a semicolon while the function `formatDate` is added below with a very similar code:

```
return moment(val).format('LL');
```

Is it to avoid an indirection that `formatDate` isn't used here ?

pblayo: (it's minor, I marked it as "change requested" to make it visible because I'm not sure of the etiquette of this repository yet)

taniki: yep it was intended but only to comply with @abulte comment about not using StandardJS (which promote not using semi).

The best workaround for this case would be to implement a global filter but I am still a bit shy about code impact.

taniki:
Sorry I forgot to use the in-github squash option.

PR #1284, opened 2017-11-29
show message for emtpy dataset search [FIX #1044]

### LINKED ISSUE #1044: A search with unmatched results return a "datasets 1 to 0 of 0" label
## Detailed Description

As seen on https://www.data.gouv.fr/fr/topics/agriculture-et-alimentation/datasets?license=fr-lo&granularity=poi

image

## Context

When a theme search returns no results, I want to see a message stating there is no resealt, so I can try something else instead .

## Possible Implementation

N/C

## Your Environment

* Operating System and version (desktop or mobile): N/A

/via @gregoirenovel

taniki:
image

### DIFF
--- a/udata/templates/dataset/list.html
+++ b/udata/templates/dataset/list.html
@@ -15,11 +15,16 @@
 
 {% block breadcrumb %}
     <li class="active">
+    
         {{ _('Datasets') }}
-        <small>{{ _('%(start)s to %(end)s on %(total)s',
-            start=datasets.page_start,
-            end=datasets.page_end,
-            total=datasets.total) }}</small>
+
+        {% if datasets.total > 0 %}
+            <small>{{ _('%(start)s to %(end)s on %(total)s',
+                start=datasets.page_start,
+                end=datasets.page_end,
+                total=datasets.total) }}</small>
+		{% endif %}
+        
         {% include "dataset/search-labels.html" %}
     </li>
 {% endblock %}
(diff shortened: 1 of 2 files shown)


udata/templates/dataset/list.html
```diff
@@ -15,12 +15,16 @@
 
 {% block breadcrumb %}
     <li class="active">
-        {{ _('Datasets') }}
+			{% if datasets.total == 0 %}
```
noirbizarre: Even if there is no results, the breadcrum should be `Datasets`.

udata/templates/dataset/list.html
```diff
@@ -15,12 +15,16 @@
...
+				<small>{{ _('No dataset') }}</small>
+			{% else %}
+				{{ _('Datasets') }}
         <small>{{ _('%(start)s to %(end)s on %(total)s',
```
noirbizarre: But the paging details can be hidden instead

udata/templates/dataset/list.html
```diff
@@ -43,7 +47,15 @@
 
 {% block main_content %}
 <div class="row">
-    <div class="col-md-8 col-lg-9 smaller">
+
+		{% if not datasets %}
```
noirbizarre: Express the opposite, it should be clearer: `if datasets` ... `else` ... (positive condition and nominal case first)

taniki:
image

squashed and ready. it should also fix the messed up indenting

taniki:
related do #1044 

@noirbizarre 🆙

PR #1285, opened 2017-11-29
Make terms content an easily customizable markdown content

noirbizarre:
This PR externalize the terms handling.
udata now expect a markdown content into a remote or local file.
The target file is defined by the `SITE_TERMS` setting.
It default on an embedded markdown content which indicate to the administrator it needs to define terms.

**NB:** [this](https://github.com/opendatateam/udata/blob/master/udata/frontend/markdown.py#L19-L23) prevent `mailto:` links to render as links

### DIFF
--- a/docs/adapting-settings.md
+++ b/docs/adapting-settings.md
@@ -48,6 +48,13 @@ A secret key used as salt for cryptographic parts.
 
 The site identifier. It is used to attached some database configuration, metrics...
 
+### SITE_TERMS
+
+**default**: `'default'`
+
+The site terms in markdown. It can be either an URL to a markdown content or a local path.
+If this is an URL, the content is downloaded on the first terms page display and cached.
+
 ### PLUGINS
 
 **default**: `[]`
--- a/docs/adapting-settings.md
+++ b/docs/adapting-settings.md
@@ -48,11 +48,11 @@ A secret key used as salt for cryptographic parts.
 
 The site identifier. It is used to attached some database configuration, metrics...
 
-### SITE_TERMS
+### SITE_TERMS_LOCATION
 
-**default**: `'default'`
+**default**: `generic embedded terms`
 
-The site terms in markdown. It can be either an URL to a markdown content or a local path.
+The site terms in markdown. It can be either an URL or a local path to a markdown content.
 If this is an URL, the content is downloaded on the first terms page display and cached.
 
 ### PLUGINS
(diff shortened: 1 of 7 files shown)


udata/cgu.md
```diff
@@ -0,0 +1,6 @@
+# Generic terms and conditions
```
oncletom: `cgu.md` is based on a french naming (Conditions Générales d'Utilisation). Would not it be more consistent to name it `terms.md`?

Also, should we rather return a 404 page on the terms page if nothing has been specified? I'd be more comfortable to return nothing rather than a content which cannot be used as is.

Something like `udata doctor` could be used to list recommended changes to improve udata settings beyond its defaults.

oncletom:
Nice work :-)

Is there part of your work which can be reused in the future to handle more content via remote static pages/markdown resources?

docs/adapting-settings.md
```diff
@@ -48,6 +48,13 @@ A secret key used as salt for cryptographic parts.
 
 The site identifier. It is used to attached some database configuration, metrics...
 
+### SITE_TERMS
+
+**default**: `'default'`
+
+The site terms in markdown. It can be either an URL to a markdown content or a local path.
```
abulte: s/It can be either an URL to a markdown content or a local path./It can be either an URL or a local path to a markdown content./

abulte: s/terms/terms and conditions/

udata/cgu.md
abulte: 👍 for `terms.md`. Not sure about the `404`.

udata/cgu.md
```diff
@@ -0,0 +1,6 @@
...
+
+Be excellent to each other.
+
+You see this content because you don't have provided a proper `SITE_TERMS`
```
abulte: s/`SITE_TERMS`/`SITE_TERMS` setting/

docs/adapting-settings.md
```diff
@@ -48,6 +48,13 @@ A secret key used as salt for cryptographic parts.
...
+If this is an URL, the content is downloaded on the first terms page display and cached.
```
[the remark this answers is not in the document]
abulte: @oncletom https://github.com/opendatateam/udata/pull/1285/files#diff-6665fed81d09a73f3e985d1201025485R166

abulte:
[LGTM](http://www.lgtm.in/i/OL5qrh8fq)

PR #1289, opened 2017-12-04
Deeplink to resource - fix #1275

### LINKED ISSUE #1275: Deeplink to Community Resource
This won't fix the (probable) problem of the lack of visibility of community resources, but maybe we need a deep link to them. This would be extra useful for sharing them.

abulte:
This PR allows to make a deep link to a resource – community or standard.

A few points open to discussion:
- maybe we do not need the "link" icon in the resource box. The URL when the modal is opened may be enough
- there may be a cleaner solution (routes?) than hacking with `location.hash` – still, this seems to do the job

NB: the previous logic above https://github.com/opendatateam/udata/compare/master...abulte:gh-1275-community-resource-deeplink?expand=1#diff-50704b3aa6249195308d85bc02bc10acL70 should now be handled by `.stop` and `.prevent` modifiers on `@click` events.

### DIFF
--- a/js/front/dataset/index.js
+++ b/js/front/dataset/index.js
@@ -28,6 +28,9 @@ function parseUrl(url) {
     return a;
 }
 
+const RESOURCE_REGEX = /^#resource-([0-9a-f-]{36})$/;
+const RESOURCE_COMMUNITY_REGEX = /^#resource-community-([0-9a-f-]{36})$/;
+
 new Vue({
     mixins: [FrontMixin],
     components: {
@@ -43,6 +46,11 @@ new Vue({
         this.loadCoverageMap();
         this.checkResources();
         this.fetchReuses();
+        if (document.location.hash) {
+            this.$nextTick(() => { // Wait for data to be binded
+                this.openResourceFromHash(document.location.hash);
+            });
+        }
         log.debug('Dataset display page ready');
     },
     methods: {
@@ -64,13 +72,16 @@ new Vue({
         /**
          * Display a resource or a community ressource in a modal
          */
-        showResource(id, e, isCommunity) {
-            // Ensure edit button work
-            if ([e.target, e.target.parentNode].some(el => el.classList.contains('btn-edit'))) return;
-            e.preventDefault();
+        showResource(id, isCommunity) {
             const attr = isCommunity ? 'communityResources' : 'resources';
             const resource = this.dataset[attr].find(resource => resource['@id'] === id);
-            this.$modal(ResourceModal, {resource});
+            const communityPrefix = isCommunity ? '-community' : '';
+            location.hash = `resource${communityPrefix}-${id}`;
+            const modal = this.$modal(ResourceModal, {resource});
+            modal.$on('modal:closed', () => {
+                // prevent scrolling to top
+                location.hash = '_';
+            });
         },
 
         /**
@@ -233,6 +244,21 @@ new Vue({
                 this._('New tag suggestion to improve metadata'),
                 this._('Hello,\n\nI propose this new tag: ')
             );
-        }
+        },
+
+        /**
+         * Open resource modal if corresponding hash in URL.
+         * /!\ there is a similar function in <discussion-threads> (jumpToHash),
+         * jump may come from there too.
+         */
+        openResourceFromHash(hash) {
+            if (RESOURCE_REGEX.test(hash)) {
+                const [, id] = hash.match(RESOURCE_REGEX);
+                this.showResource(id, false);
+            } else if (RESOURCE_COMMUNITY_REGEX.test(hash)) {
+                const [, id] = hash.match(RESOURCE_COMMUNITY_REGEX);
+                this.showResource(id, true);
+            }
+        },
     }
 });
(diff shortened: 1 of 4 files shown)


js/front/dataset/index.js
```diff
@@ -64,13 +72,16 @@ new Vue({
         /**
          * Display a resource or a community ressource in a modal
          */
-        showResource(id, e, isCommunity) {
-            // Ensure edit button work
-            if ([e.target, e.target.parentNode].some(el => el.classList.contains('btn-edit'))) return;
```
noirbizarre: Why this has been removed ?

js/front/dataset/index.js
```diff
@@ -64,13 +72,16 @@ new Vue({
...
-            e.preventDefault();
+        showResource(id, isCommunity) {
             const attr = isCommunity ? 'communityResources' : 'resources';
             const resource = this.dataset[attr].find(resource => resource['@id'] === id);
-            this.$modal(ResourceModal, {resource});
+            const communityPrefix = isCommunity ? '-community' : '';
+            location.hash = `resource${communityPrefix}-${id}`;
+            const modal = this.$modal(ResourceModal, {resource});
+            modal.$on('modal:closed', () => {
+                // prevent scrolling to top
```
noirbizarre: Is this behavior common to all modals on close ?
Then maybe be we should issue a fix in a separate PR for this ?

js/front/dataset/index.js
```diff
@@ -28,6 +28,9 @@ function parseUrl(url) {
     return a;
 }
 
+const RESOURCE_REGEX = /^#resource-([0-9a-f-]{36})$/;
+const RESOURCE_COMMUNITY_REGEX = /^#resource-community-([0-9a-f-]{36})$/;
```
noirbizarre: I think both regexp might be combined into a single one:
 ```javascript
const RESOURCE_REGEX = /^#resource(?:-community)?-([0-9a-f-]{36})$/
```

File: js/front/dataset/index.js [hunk 4]
```diff
@@ -233,6 +244,21 @@ new Vue({
...
+        /**
+         * Open resource modal if corresponding hash in URL.
+         * /!\ there is a similar function in <discussion-threads> (jumpToHash),
+         * jump may come from there too.
+         */
+        openResourceFromHash(hash) {
+            if (RESOURCE_REGEX.test(hash)) {
+                const [, id] = hash.match(RESOURCE_REGEX);
+                this.showResource(id, false);
+            } else if (RESOURCE_COMMUNITY_REGEX.test(hash)) {
```
noirbizarre: If using a single regexp, you remove this `else if`

udata/templates/dataset/resource/list-item.html
```diff
@@ -1,12 +1,17 @@
 {% set resource_format = resource.format|trim|lower or 'data' %}
-<div id="resource-{{resource.id}}" class="list-group-item"
+<div id="resource{% if resource.from_community %}-community{% endif %}-{{resource.id}}" class="list-group-item"
```
noirbizarre: Hum, this distinction comes from this.
As ressources IDs (including community ressources) are unique and behavior is stricly the same, is this really necessary to have distinct ID pattern ?

udata/templates/dataset/resource/list-item.html
```diff
@@ -28,7 +33,7 @@ <h4 class="list-group-item-heading ellipsis">
             {% set edit_path = 'dataset/{id}/resource/{rid}' %}
         {% endif %}
         <div class="btn-group btn-group-xs tools">
-            <a class="btn btn-default btn-edit" v-tooltip title="{{ _('Edit') }}"
+            <a @click.stop class="btn btn-default btn-edit" v-tooltip title="{{ _('Edit') }}"
```
noirbizarre: OMG, is this working ? If so I understand the above removal !!

abulte: Yep, it should work ;-) At least it does on my machine.

udata/templates/dataset/resource/list-item.html
abulte: I need the distinction to use `showResource` that expects a `isCommunity` parameter, apparently in order to fetch the right resource in the JSON-LD.

The only alternative I can think of right now is to try to find the resource by id in both  `dataset[resources]` and `dataset[communityResources]`. I'm not sure it's a better alternative.

js/front/dataset/index.js
abulte: Nope, it's for this one only: since I set the `location.hash` with the current resource hash, I need to unset it when closing the modal. The `#_` is just a trick to avoid the top scroll that would be triggered with `#`.

abulte:
The link icon I'm talking about above:

capture d ecran 2017-12-05 10 22 28
capture d ecran 2017-12-05 10 22 36

js/front/dataset/index.js
noirbizarre: `const RESOURCE_REGEX = /^#resource(-community)?-([0-9a-f-]{36})$/` will permit to extract in a single pass the ID and the fact that's it's a community resource or not.

js/front/dataset/index.js
noirbizarre: With the single pass regex, you can write:
```javascript
const [, isCommunity, id] = hash.match(RESOURCE_REGEX);
this.showResource(id, Boolean(isCommunity));
```

PR #1293, opened 2017-12-07
Prevent reindex to fail on a given object but log error instead

noirbizarre:
Specific reindexation errors should not prevent the whole reindexation to finish.
This PRs ensure reindexation does not abort on specific objects errors and properly log errors so they can be fixed.


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Prevent `reindex` command to fail on a given object and log error instead. [#1293](https://github.com/opendatateam/udata/pull/1293)
```
abulte: s/to fail/from failing/

abulte:
Out of curiosity: what objects were failing and why?

noirbizarre:
That's the point, currently we can't know ;)

abulte:
Not even a traceback on the console?

noirbizarre:
Sadly, no. That's a common drawback with generators, if errors handling is not properly done, exception tracebacks are hidden :/

PR #1295, opened 2017-12-08
Fix broken post edit form page

taniki:


js/components/post/form.vue
```diff
@@ -1,16 +1,17 @@
...
 
 <script>
 import Post from 'models/post';
-import VerticalForm from 'components/form/vertical-form.vue';
 
 export default {
     name: 'post-form',
-    components: {VerticalForm},
+    components: {
+            vform: require('components/form/vertical-form.vue')
```
noirbizarre: As much as possible we avoid inline require

js/components/post/form.vue
```diff
@@ -1,16 +1,17 @@
 <template>
 <div>
-<vertical-form v-ref:form :fields="fields" :model="post"></vertical-form>
+    <vform v-ref:form :fields="fields" :model="post"></vform>
```
noirbizarre: We are doing the opposite, ie. transform `vfrom`/`hform` into  `vertical-form` `horizontal-form` for the following reason:
- this is more explicit
- web component spec prohibe single term markup (you need a dash in a web component markup)

taniki: argh. I was overzealous at getting a homogeneous code from the wrong example.

taniki:
@noirbizarre revert done

noirbizarre:
:ok_hand:

PR #1297, opened 2017-12-11
Import css dropdowns handling from udata-gouvfr

noirbizarre:
This PRs import CSS dropdowns behavior from gouvfr style (it's a ashred behavior) and fix hidden submenus on mobile.

screenshot-www data dev-2017-12-11-12-02-36-033

### DIFF
--- a/less/udata/dropdowns.less
+++ b/less/udata/dropdowns.less
@@ -1,0 +1,22 @@
+.dropdown {
+    // Desktop: display on hover
+    &.dropdown-hover:hover .dropdown-menu {
+        display: block;
+    }
+
+    // Clickable dropdowns
+    &.open .dropdown-menu {
+        display: inline;
+    }
+
+    // Mobile: display inline when menu is open
+    .navbar-collapse.in &.dropdown-hover .dropdown-menu {
+        display: inline;
+        width: 90%;
+        margin: 0 0 15px 15px;
+        position: relative;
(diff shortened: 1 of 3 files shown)


less/udata/dropdowns.less
```diff
@@ -0,0 +1,22 @@
...
+        display: block;
+    }
+
+    // Clickable dropdowns
+    &.open .dropdown-menu {
+        display: inline;
+    }
+
+    // Mobile: display inline when menu is open
+    .navbar-collapse.in &.dropdown-hover .dropdown-menu {
```
oncletom: How do we know this is for mobile? Just by reading the CSS selector, it is not obvious.

noirbizarre: `.navbar-collapse.in` is the bootstrap selector for the expanded menu on mobile only (`.navbar-collapse` is always visible on desktop)
This is not obvious (and I had to search), this is why I put the comment.

noirbizarre: BTW, why block the merge and request change ?

oncletom: @noirbizarre ok thanks. I was not sure if there was a more obvious way to signpost this. Something more obvious than a comment. I can live with it :-)

noirbizarre: Some mecanism in bootstrap are a little bit weird to understand :/

PR #1301, opened 2017-12-11
fix non-sorting posts behaviour in editorial view

taniki:
fix etalab/data.gouv.fr#41

This one was a deep dive in the full stack. I had to add the sort rule in the python. Not sure about the interaction with swagger.

There might be unnoticed bugs since only datasets and reuses since to acknowledge `sort` in their listing query.

### DIFF
--- a/js/components/post/list.vue
+++ b/js/components/post/list.vue
@@ -34,7 +34,7 @@ export default {
             },{
                 label: this._('Creation'),
                 key: 'created_at',
-                sort: 'created',
+                sort: 'created_at',
                 align: 'left',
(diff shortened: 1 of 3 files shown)


js/components/post/list.vue
```diff
@@ -34,7 +34,7 @@ export default {
             },{
                 label: this._('Creation'),
                 key: 'created_at',
-                sort: 'created',
+                sort: 'created_at',
```
pblayo: Discussed with @taniki : created needed to be changed into created_at so that the fix below could work.

PR #1303, opened 2017-12-12
Fix/community resource link icon - take 2

abulte:
Fix the fix in #1298 

Should be ok now, both on default and gouvfr themes.

capture d ecran 2017-12-12 11 21 23
capture d ecran 2017-12-12 11 22 49


abulte:
@noirbizarre not sure, I think they're useful for the discussions. I have to check again though.

PR #1322, opened 2017-12-18
fix absence of label for tabs on the org page (FIX #1320)

### LINKED ISSUE #1320: Label of tabs in the organization page are not displayed
As reported here etalab/data.gouv.fr#47

taniki:
Not sure it is the best solution but it seemed more efficient than forking/pr vue-strap. As far I know vue, `<slot>` here are a bit overkill since there is no way to access it simply from the `tabs` template.

image


taniki:
I proposed a PR to fix it in vue-strap 😎

noirbizarre:
:ok_hand:

PR #1331, opened 2017-12-21
Add a munin output to `udata worker status`

### LINKED ISSUE #1328: Add a munin output to `udata worker status`
Following #1324, it would be nice to have a Munin compatible output of this command.

Proposed Munin graphic representation:
- 1 chart per queue
- Foreach, a line with the number of a certain task type (ie `piwik-current-metrics`) queued at a given time

What's the output format Munin needs for that?

abulte:
Fix #1328

Usage:

```
$ udata worker status --munin -q default
piwik_current_metrics.value 10
update_metric.value 20
another_task.value 42
```

```
$ udata worker status --munin-config -q default
graph_title Waiting tasks for queue default
graph_vlabel Nb tasks
piwik_current_metrics.label piwik-current-metrics
update_metric.label update-metric
another_task.label another-task
# ... all possible tasks, even non queued ones #
```

Warnings at extensions init have been removed in order not to pollute the CLI output.

### DIFF
--- a/udata/commands/worker.py
+++ b/udata/commands/worker.py
@@ -35,16 +35,6 @@ def start():
     celery.start()
 
 
-def status_print_header(munin=False):
-    if not munin:
-        print('-' * 40)
-
-
-def status_print_title(queue, queue_length, munin=False):
-    if not munin:
-        print('Queue "%s": %s task(s)' % (queue, queue_length))
-
-
 def status_print_task(count, biggest_task_name, munin=False):
     if not munin:
         print('* %s : %s' % (count[0].ljust(biggest_task_name), count[1]))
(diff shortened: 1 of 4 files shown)


udata/commands/worker.py
```diff
@@ -21,25 +24,81 @@
...
 @m.command
 def start():
     """Start a worker"""
     celery.start()
 
 
-@m.command
-def status():
+def status_print_header(munin=False):
+    if not munin:
```
noirbizarre: I'm wondering if if might not be easier to read the if outside the call:
```python
if not munin:
    status_print_header()
```

udata/commands/worker.py
```diff
@@ -21,25 +24,81 @@
...
+        print('-' * 40)
+
+
+def status_print_title(queue, queue_length, munin=False):
+    if not munin:
```
noirbizarre: same here

noirbizarre:
I forgot: the documentation needs to be updated [here](https://github.com/opendatateam/udata/blob/master/docs/administrative-tasks.md)

PR #1332, opened 2017-12-22
Only show notification on success of update

### LINKED ISSUE #1266: Superfluous success notification when creating a dataset with an invalid field
When we input a wrong value in the dataset creation field, at least on the tags field, and the POST request fails we get a success notification which should not be there.

capture d ecran 2017-11-14 14 44 54

Related to #563

taniki:
FIX #1266 

There is still a remaining usability issue because the form does not highlight the faulty tag yet.

### DIFF
--- a/js/components/dataset/form.vue
+++ b/js/components/dataset/form.vue
@@ -78,17 +78,6 @@ export default {
         },
         validate() {
             const isValid = this.$refs.form.validate();
-
-            if (isValid & !this.hideNotifications) {
-                /*
-                this.$dispatch('notify', {
-                    autoclose: true,
-                    title: this._('Changes saved'),
-                    details: this._('Your dataset has been updated.')
-                });
-                */
-            }
-
             return isValid;
         },
         on_error(response) {
--- a/js/components/dataset/form.vue
+++ b/js/components/dataset/form.vue
@@ -80,11 +80,13 @@ export default {
             const isValid = this.$refs.form.validate();
 
             if (isValid & !this.hideNotifications) {
+                /*
                 this.$dispatch('notify', {
                     autoclose: true,
                     title: this._('Changes saved'),
                     details: this._('Your dataset has been updated.')
                 });
+                */
             }
 
             return isValid;
--- a/js/views/dataset-edit.vue
+++ b/js/views/dataset-edit.vue
@@ -30,12 +30,12 @@ export default {
     },
     methods: {
         save() {
-            let form = this.$refs.form;
+            const form = this.$refs.form;
             if (form.validate()) {
                 this.dataset.update(form.serialize(), (response) => {
-                    this.dataset.on_fetched(response); 
+                    this.dataset.on_fetched(response);
                     
-                    if(!form.hideNotifications){
+                    if (!form.hideNotifications) {
                         this.$dispatch('notify', {
                             autoclose: true,
                             title: this._('Changes saved'),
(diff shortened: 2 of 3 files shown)


js/components/dataset/form.vue
```diff
@@ -80,11 +80,13 @@ export default {
             const isValid = this.$refs.form.validate();
 
             if (isValid & !this.hideNotifications) {
+                /*
```
abulte: Remove commented part?

js/views/dataset-edit.vue
```diff
@@ -30,12 +30,24 @@ export default {
     },
     methods: {
         save() {
-            const form = this.$refs.form;
+            let form = this.$refs.form;
```
abulte: I think this should stay a `const`.

js/views/dataset-edit.vue
```diff
@@ -30,12 +30,24 @@ export default {
...
             if (form.validate()) {
                 this.dataset.update(form.serialize(), (response) => {
-                    this.dataset.on_fetched(response);
+                    this.dataset.on_fetched(response);
```
abulte: Indentation problem? I’m not sure why this line show up in the diff.

js/views/dataset-edit.vue
```diff
@@ -30,12 +30,24 @@ export default {
...
     methods: {
         save() {
-            const form = this.$refs.form;
+            let form = this.$refs.form;
             if (form.validate()) {
                 this.dataset.update(form.serialize(), (response) => {
-                    this.dataset.on_fetched(response);
+                    this.dataset.on_fetched(response); 
+                    
+                    if(!form.hideNotifications){
```
abulte: Spaces before and after `if (..) {` (this is what I deduce from existing code)

js/views/dataset-edit.vue
```diff
@@ -30,12 +30,24 @@ export default {
...
+                        this.$dispatch('notify', {
+                            autoclose: true,
+                            title: this._('Changes saved'),
+                            details: this._('Your dataset has been updated.')
+                        });
+                    }
+
                     this.$go({name: 'dataset', params: {oid: this.dataset.id}});
-                }, form.on_error);
+                }, (error)=>{
```
abulte: Spaces around arrow I think (and maybe drop the `()`)

js/views/dataset-edit.vue
```diff
@@ -30,12 +30,24 @@ export default {
...
+                    this.dataset.loading = false;
```
taniki: Not sure if this is the right place to do it

js/views/dataset-edit.vue
```diff
@@ -34,8 +34,20 @@ export default {
...
+                            autoclose: true,
+                            title: this._('Changes saved'),
+                            details: this._('Your dataset has been updated.')
+                        });
+                    }
+
                     this.$go({name: 'dataset', params: {oid: this.dataset.id}});
-                }, form.on_error);
+                }, error => {
+                    this.dataset.loading = false;
```
taniki: Not sure this is the right place for this

js/views/dataset-edit.vue
abulte: Sorry about « drop the () », I just realized we usually leave them (see response above).

js/views/dataset-edit.vue
abulte: What do you mean? It hides the loader, right? Looks good here.

PR #1336, opened 2017-12-29
Expose owner_avatar_url helper for themes

noirbizarre:
Allows theme to easily access the `owner_avatar_url` template filter


noirbizarre:
I didn't forget it, it just doesn't seems interesting to have this in the changelog, but I can add it.

Issue #1126, opened 2017-09-12
When URLs are saved, %20 are converted to plus signs


seiteta:
## Detailed Description
`%20` in URLs are converted to plus signs when they are saved.

## Exemple
<img width="602" alt="capture d ecran 2017-09-12 a 10 12 40" src="https://user-images.githubusercontent.com/5756228/30315111-2f9947da-97a3-11e7-9b81-e93cdad5c8ad.png">
<img width="644" alt="capture d ecran 2017-09-12 a 10 13 08" src="https://user-images.githubusercontent.com/5756228/30315118-34ee4a8c-97a3-11e7-9050-8f90a85599f4.png">

## Context
When I saved an URL as a ressource, I want it to stay the same, so I can share it on udata even if it contain `%20`.

## Your Environment
data.gouv.fr

pblayo:
According to @noirbizarre's investigation it seems to be a bug in swagger.js.

taniki:
This is also true for any basic field

seiteta:
Thanks 🙏

PR #1337, opened 2017-12-30
Fixes Swagger.js no setting jQuery.ajax contentType parameter

### LINKED ISSUE #1: Bug d'affichage des onglets de résultats de recherche
Ils viennent se superposer au _chemin de fer_ en français je crois.

capture d ecran 2014-09-17 a 22 02 53

Je corrigerais bien [en modifiant la CSS](https://github.com/etalab/udata/blob/13cb878f310add5e4709f9f3e7c99f434a0bcdd3/udata/static/less/udata/search.less#L192) mais je ne suis pas sûr du résultat attendu :)

noirbizarre:
This PR fixes a nasty bug with Swagger.js/jQuery.

## Context
Swagger.js only set the `Content-Type` HTTP header on request but does not give the `contentType` parameter to `jQuery.ajax`.

## Consequence
`jQuery` consider the request body as url encoded and reencode content. As a side effect, all `%20` characters are replaced by `+` signs.

## Resolution
This PR adds a hackish Swagger.js request interceptor which set the `jQuery.ajax` `contentType` parameter to the same value than the `Content-Type` header.
This fixes many side-effects:
- `%20` are not replaced by `+` signs anymore (#1126)
- default success and error handlers are taken in account (side effect of operation call signature change)
- any places where double encoding was occurring (I'm sure we missed some of them) are fixed too

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Fix topic creation [#1333](https://github.com/opendatateam/udata/pull/1333)
 - Add a `udata worker status` command to list pending tasks.[breaking] The `udata worker` command is replaced by `udata worker start`. [#1324](https://github.com/opendatateam/udata/pull/1324)
 - Prevent crawlers from indexing spammy datasets, reuses and organizations [#1334](https://github.com/opendatateam/udata/pull/1334) [#1335](https://github.com/opendatateam/udata/pull/1335)
+- Ensure Swagger.js properly set jQuery.ajax contentType parameter (and so data is properly serialized) [#1126](https://github.com/opendatateam/udata/issues/1126)
 
 ## 1.2.5 (2017-12-14)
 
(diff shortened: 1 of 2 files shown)


noirbizarre:
For reference this is a combination of:
- the default [`jQuery.ajax` `contentType` parameter](https://github.com/jquery/jquery/blob/3.2.1/src/ajax.js#L308)
- a change added 2 years ago: https://github.com/jquery/jquery/pull/2671
This swagger.js issue was invisible before jQuery 2.2.0

taniki:
Can it be also related to https://swagger.io/docs/specification/2-0/describing-request-body/ ?

`consume` parameter does not seem to be set to `application/json` anywhere. Maybe this is not implicit and swagger doesn't provide proper `content-type` to the js part.

No dev env here. will get a try and review this on tuesday!

noirbizarre:
The `consume` parameter properly set the `Content-Type` header, this is only the jQuery.ajax call which is broken.

Yes there might be something to fix in upstream swagger.js. Sadly, we are using Swagger.js 2.X and a fix on the 3.x won't be usable to us right now.

PR #1341, opened 2018-01-02
fix visual glitch by relying on expected usage

taniki:
### Before

screenshot from 2018-01-02 14-43-53

### After

screenshot from 2018-01-02 14-41-14


taniki:
a `.active .fa-angle-left` rule was already shipped by bootstrap

PR #1351, opened 2018-01-08
Sanitize mdstrip

abulte:
Jinja `do_striptags` result is not safe to piped into a `safe` filter, for example when we output `json_ld`.

Not totally sure about the impact of this on the rest of the site, but some nasty things happening on a reuse do not happen anymore and the reuse stil displays fine.

### DIFF
--- a/udata/frontend/markdown.py
+++ b/udata/frontend/markdown.py
@@ -7,7 +7,8 @@ import bleach
 import CommonMark
 from flask import current_app, Markup
 from werkzeug.local import LocalProxy
-from jinja2.filters import do_truncate, do_striptags
+from jinja2.filters import do_truncate
+from jinja2.utils import escape
 
 from udata.i18n import _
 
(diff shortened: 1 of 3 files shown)


noirbizarre:
Maybe simply use `bleach` like we do for markdown

abulte:
Hum, actually we should not *replace* `striptags` but *complement* it.

abulte:
New version with `bleach` and keeping the `striptags` feature. I'm unsure about https://github.com/opendatateam/udata/pull/1351/files#diff-8eed20078dbe50136ec367e07dcf4ed1L106

udata/frontend/markdown.py
```diff
@@ -100,6 +107,7 @@ def mdstrip(value, length=None, end='…'):
         value = value.split(EXCERPT_TOKEN, 1)[0]
     rendered = md(value)
     text = do_striptags(rendered)
+    text = bleach_clean(text)
```
noirbizarre: Seems to be working. I'm just wondering whether this should after or before the `do_striptags`

abulte: Someone `do_striptags` would reintroduce the injection. I tested because it seemed more logical but...

noirbizarre: Fine to me, I was just wondering the reasonning :+1:

PR #1370, opened 2018-01-15
Datepicker fix #1368

### LINKED ISSUE #1368: Datepicking gone wrong
## Detailed Description

When I create or edit a dataset, I have the following problem with the date picker on the "Temporal coverage" field: for a given year `Y`, the date picker switches from `November Y` to `December Y+1` and from `February Y` to `January Y-1`.

## Context

When create or edit a dataset , I want to be able to select all dates from a given year on the date picker.

## Possible Implementation

Fix it.

## Your Environment

* Browser Name and version: Safari, Chrome evergreen
* Operating System and version (desktop or mobile): Mac OS
* uData : data.gouv.fr `udata 1.2.8`

abulte:
This fixes #1368 but introduces a new problem for now: `currentMonthDays` is not recomputed when `current` changes.

I've been looking at this code for too long now, dropping this here for pairing / help.

I can not explain either why the original solution was not working (regarding the switch from November to December and from February to January).

**Edit: it now works with a hack around change detection.**

### DIFF
--- a/js/components/calendar.vue
+++ b/js/components/calendar.vue
@@ -253,10 +253,8 @@ export default {
     },
     watch: {
         selected(value) {
-            console.log('selected', JSON.parse(JSON.stringify(value)));
             if (!value) return;
             this.current = value.clone();
-            console.log('current', JSON.parse(JSON.stringify(this.current)));
         }
     }
 };
--- a/js/components/calendar.vue
+++ b/js/components/calendar.vue
@@ -80,8 +80,7 @@ export default {
     data() {
         const current = this.selected || moment();
         return {
-            currentMonth: current.month(),
-            currentYear: current.year(),
+            current: current,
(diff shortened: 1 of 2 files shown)


js/components/calendar.vue
```diff
@@ -220,7 +216,7 @@ export default {
         },
 
         isMonthDisabled(idx) {
-            const month = moment().month(idx).year(this.currentYear);
+            const month = moment().month(idx).year(this.current.year());
```
abulte: this should be `const month = this.current.clone().month(idx)`

abulte:
NB: I find the new code easier to read.

js/components/calendar.vue
noirbizarre: Agreed but non-significant here I think.

PR #1384, opened 2018-01-19
Use AND operator for multiple terms multi_match query

noirbizarre:
This PR ensure that queries multiple included terms use the `AND` operator.
It does not apply for negated terms as we want to ensure each negated term contribute to removal.

### DIFF
--- a/udata/tests/test_search.py
+++ b/udata/tests/test_search.py
@@ -556,7 +556,7 @@ class SearchQueryTest(SearchTestMixin, SearchTestCase):
     def test_query_with_multiple_including_and_excluding_terms(self):
         '''A query should detect negation on each term in query_string'''
         search_query = search.search_for(FakeSearch,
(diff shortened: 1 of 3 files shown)


udata/tests/test_search.py
```diff
@@ -541,6 +553,39 @@ def test_query_with_both_including_and_excluding_terms(self):
...
+                        'query': 'test value',
+                        'analyzer': search.i18n_analyzer._name,
+                        'type': 'cross_fields',
+                        'operator': 'and',
+                        'fields': ['title^2', 'description']
+                    }}
+                ],
+                'must_not': [
+                    {'multi_match': {
+                        'query': 'negated',
```
pblayo: What do you think about **first_negated** / **1st_negated** and **second_negated** / **2nd_negated** or **other_negated** as query's words?

PR #1386, opened 2018-01-19
Using cards in search results

taniki:
This also list using a faceted look and feel:
- list of datasets
- list of orgs
- list of reuses
- topics

extracted from #1252


abulte:

PR #1388, opened 2018-01-20
Ensure unicode objects are not casted to str but properly encoded

noirbizarre:
More fixes on unicode handling (Remove a double translation and ensure lazy_string are properly encoded in search facets)

Connects to #1381

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - `ValueError`s are not hidden anymore by the Bad Request error page, they are logged. [#1382](https://github.com/opendatateam/udata/pull/1382)
 - Spatial encoding fixes: prevent breaking unicode errors. [#1381](https://github.com/opendatateam/udata/pull/1381)
 - Ensure the multiple term search uses a `AND` operator [#1384](https://github.com/opendatateam/udata/pull/1384)
-- Facets encoding fixes: ensure lazy string are propery encoded. [#1388](https://github.com/opendatateam/udata/pull/1388)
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - `ValueError`s are not hidden anymore by the Bad Request error page, they are logged. [#1382](https://github.com/opendatateam/udata/pull/1382)
 - Spatial encoding fixes: prevent breaking unicode errors. [#1381](https://github.com/opendatateam/udata/pull/1381)
 - Ensure the multiple term search uses a `AND` operator [#1384](https://github.com/opendatateam/udata/pull/1384)
+- Facets encoding fixes: ensure lazy string are propery encoded. [#1388](https://github.com/opendatateam/udata/pull/1388)
```
abulte: s/string/strings/ or s/are/is/

udata/core/spatial/models.py
```diff
@@ -8,7 +8,7 @@
 from werkzeug import cached_property
 
 from udata.app import cache
-from udata.i18n import gettext as _, get_locale, language
+from udata.i18n import gettext as _, lazy_gettext as L_, get_locale, language
```
abulte: Is `L_` a known convention? I'd be more comfortable with `l_` since it's a function.

noirbizarre: Yes, it is, already used for [`ngettext`](https://github.com/opendatateam/udata/blob/master/udata/i18n.py#L109), [`pgettext`](https://github.com/opendatateam/udata/blob/master/udata/i18n.py#L114). 
I can make the same for `lazy_gettext`, ie. define it permanently into `i18n.py` instead of using `import .. as`

PR #1395, opened 2018-01-23
Prevent empty harvest run to fail

### LINKED ISSUE #1394: Harvester should not start a chord when there is no items to process
## Detailed Description
Some harvest jobs fail because there is no item to process.
Celery chords needs to have at least one item to process.

The following sentry error happen on such cases: https://sentry.data.gouv.fr/share/issue/9a3113bc1ab74f0ca0392a300f202b01/

## Possible Implementation
Prevent Celery from starting an empty chord when there is no item to process

## Your Environment
`udata==1.2.9`

noirbizarre:
Prevent empty harvest jobs to fail with Celery

Fix #1394

### DIFF
--- a/udata/harvest/tasks.py
+++ b/udata/harvest/tasks.py
@@ -18,12 +18,15 @@ def harvest(self, ident):
     source = HarvestSource.get(ident)
     Backend = backends.get(source.backend)
     backend = Backend(source)
-    if backend.perform_initialization():
+    items = backend.perform_initialization()
+    if items > 0:
         finalize = harvest_finalize.s(ident)
         items = [
             harvest_item.s(ident, item.remote_id) for item in backend.job.items
         ]
         chord(items)(finalize)
+    elif items == 0:
+        backend.finalize()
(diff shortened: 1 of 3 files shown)


udata/harvest/tasks.py
```diff
@@ -18,12 +18,15 @@ def harvest(self, ident):
     source = HarvestSource.get(ident)
     Backend = backends.get(source.backend)
     backend = Backend(source)
-    if backend.perform_initialization():
+    items = backend.perform_initialization()
+    if items > 0:
```
abulte: `if items` and `else`?

noirbizarre: No because failed bacth returns `None`

noirbizarre:
https://sentry.data.gouv.fr/etalab/www-datagouvfr/issues/123361/

noirbizarre:
https://sentry.data.gouv.fr/etalab/www-datagouvfr/issues/152712/

PR #1399, opened 2018-01-25
[wip] Put missing spinners on various parts

taniki:


taniki:
https://github.com/opendatateam/udata/blob/05e723b301c10664d2053f4342e31f4a50b71512/js/components/datatable/widget.vue#L16

Wondering if it is not more of a bug than missing stuffs. The line above should do the job right?

taniki:
Yep that's it. the condition is always true because the prop defaults `false`

PR #1410, opened 2018-01-30
Small fixes on search

noirbizarre:
This PR apply small fixes on search:
- facets right counters are vertically centered and ensure they do not overlap with text
- differentiate the "more result button": centered and no more right picto (which as the same as facet header)
- fix the missing separator between the initial facet result and the extra ones
- remove the hardcoded white background color, just reuse the body color

## Unexpanded

| Before | After |
|-----------|--------|
| screenshot-data xps-2018-01-29-21-43-31-554 | screenshot-data xps-2018-01-29-21-40-49-236 |

## Expanded

| Before | After |
|-----------|--------|
| screenshot-data xps-2018-01-29-21-43-57-495 | screenshot-data xps-2018-01-29-21-41-24-353 |

### DIFF
--- a/less/udata/search.less
+++ b/less/udata/search.less
@@ -5,7 +5,6 @@ ul.search-results {
 
     .search-result {
         position: relative;
-        background: white;
         border-bottom: 1px solid #F0F0F0;
         overflow: hidden;
         zoom: 1;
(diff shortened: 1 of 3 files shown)


less/udata/search.less
```diff
@@ -136,6 +135,34 @@ ul.search-results {
 .advanced-search-panel {
     .list-group {
         margin-bottom: 0;
+
+        .list-group-item {
+            // Fix the badge position on long facet labels
+            &:not(.more) {
+                @padding:15px; // Hard coded in bootstrap list-group compoent
```
taniki: small typo here `compo+n+ent`

PR #1424, opened 2018-02-06
I18n fix extraction

noirbizarre:
This PR fix plural and context extraction (the `setup.cfg` `keywords` parameters has undocumented format for `ngettext`, `pgettext` and `npgettext`).

This PR comes with a fixed `.pot` extraction so crowdin can match again plural and contexts

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,7 +4,7 @@ directory = udata/translations
 statistics = true
 
 [extract_messages]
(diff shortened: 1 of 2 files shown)


noirbizarre:
:joy:
The numbers tell `babel` how many parameters need to be extracted and which usage they have (`c` is for context`)

PR #1431, opened 2018-02-12
More entrypoints

noirbizarre:
This PR cleanups plugins magic import and switch the necessary ones to entrypoints. (`udata.views`, `udata.metrics`, `udata.models` and `udata.tasks`)
Their usage has been rationalized and all necessary functions grouped into the `udata.entrypoints` module.

This introduce the following changes:
- harvesters are only available if the plugin is enabled (avoid adding features/content with just installing a package)
- default `no_check` linkchecker is always available (not entrypoint anymore), others need to be activated in `PLUGINS`
- migrations are associated to `udata.models` entrypoint
- sentry can now detect all plugins and will report every versions
- plugin logging is now properly handled
- each enabled plugin can provide translations which are automatically loaded
- the `info` command is now splitted in two commands `info config` which allows to inspect the config and `info plugins` to inspect plugins
- the `metrics list` command has been added

This PR also adds an entrypoints documentation (ie. extending udata)

### DIFF
--- a/udata/entrypoints.py
+++ b/udata/entrypoints.py
@@ -7,8 +7,10 @@ import pkg_resources
 ENTRYPOINTS = {
     'udata.avatars': 'Avatar rendering backends',
     'udata.harvesters': 'Harvest backends',
-    'udata.i18n': 'Extra translations',
     'udata.linkcheckers': 'Link checker backends',
+    'udata.metrics': 'Extra metrics',
+    'udata.models': 'Models and migrations',
+    'udata.tasks': 'Tasks and jobs',
     'udata.themes': 'Themes',
     'udata.views': 'Extra views',
 }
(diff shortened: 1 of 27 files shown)


udata/entrypoints.py
```diff
@@ -3,11 +3,40 @@
 
 import pkg_resources
 
+# Here for documention purpose
```
abulte: s/documention/documentation/

udata/entrypoints.py
```diff
@@ -3,11 +3,40 @@
...
+    'udata.themes': 'Themes',
+    'udata.views': 'Extra views',
+}
+
+
+class EntrypointError(Exception):
+    pass
+
+
+def iter(name):
```
abulte: /!\ built-in function name

udata/sentry.py
```diff
@@ -32,7 +33,7 @@ def init_app(app):
                 register_signal, register_logger_signal
             )
             from raven.contrib.flask import Sentry
-        except:
+        except Exception:
```
abulte: Maybe `ImportError`?

noirbizarre:
Done :+1:

PR #1434, opened 2018-02-13
Switch to Authlib

noirbizarre:
This PR:
- [x] Switch to [Authlib](https://docs.authlib.org/en/latest/index.html) instead of the unmaintained [OAuthlib](https://oauthlib.readthedocs.io/en/latest/)/[Flask-OAuthlib](http://flask-oauthlib.readthedocs.io/en/latest/)
- [x] Ensure refresh tokens are working
- [x] Makes OAuth 2 implementation support all grantypes
- [x] Support token revocation
- [x] Makes `internal` client allowed to bypass authorization screen
- [x] Ports test to Pytest
- [x] Ensures tests cover all cases

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,6 +26,9 @@ norecursedirs = .git build .tox specs .cache udata/static udata/templates udata/
 python_files = test_*.py
 python_functions = test_*
 python_classes = *Test
+env =
+    # See: https://docs.authlib.org/en/latest/flask/oauth2.html
+    AUTHLIB_INSECURE_TRANSPORT=true
 
 [flake8]
 exclude =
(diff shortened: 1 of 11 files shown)


setup.cfg
```diff
@@ -26,6 +26,8 @@ norecursedirs = .git build .tox specs .cache udata/static udata/templates udata/
 python_files = test_*.py
 python_functions = test_*
 python_classes = *Test
+env =
+    AUTHLIB_INSECURE_TRANSPORT=true
```
abulte: What’s this?

udata/migrations/2018-02-12-authlib.js
```diff
@@ -0,0 +1,19 @@
+/*
+ * Remove checksum from resources where it has no value
+ */
```
abulte: Wrong comment I think.

setup.cfg
noirbizarre: Because HTTPS is part of OAuth specs and `authlib` enforce it unless this environment variable is explicitly set.
This allows tests (and only tests) to run OAuth flows in http (because the test serveur is `http://localhost`).

abulte: Ok, I missed the pytest section above.

PR #1448, opened 2018-02-19
Fix search API failing on unicode queries

noirbizarre:
This PR fix a bug preventing unicode search queries (any model) through the API.


noirbizarre:
No, most of the time, these are ascii IDs thaht are expected. `q` is the only field we are querying with free text (and so unicode).

PR #1453, opened 2018-02-20
Added frontend marker and theme unloading for testing

noirbizarre:
Easily loads frontend modules with the `@pytest.mark.fronted` and properly unload theme after test when it's not the default one (allows to properly handle theme initialization)

### DIFF
--- a/udata/tests/plugin.py
+++ b/udata/tests/plugin.py
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals
 
 import pytest
+import sys
 
 from contextlib import contextmanager
 from urlparse import urlparse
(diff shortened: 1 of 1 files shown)


udata/tests/plugin.py
```diff
@@ -61,13 +62,45 @@ def app(request):
     test_settings = get_settings(request)
     app = create_app(settings.Defaults, override=test_settings)
     app.test_client_class = TestClient
+    return app
+
+
+@pytest.fixture(autouse=True)
+def _load_frontend(request):
+    '''
+    Use `pytest.mark.frontend` specify that frontend/api should be loaded
```
abulte: s/Use/Using/ or s/specify/to specify/

udata/tests/plugin.py
```diff
@@ -61,13 +62,45 @@ def app(request):
...
+    '''
+    if 'app' not in request.fixturenames:
+        return
+
+    app = request.getfixturevalue('app')
+    marker = request.node.get_marker('frontend')
+    modules = set()
+
+    if marker and marker.args:
+        modules |= set(marker.args[0])
```
abulte: no need for `|=` here? `modules` will always be `set()`.

noirbizarre: True. Incoming one liner

PR #1462, opened 2018-02-26
Added link to password change into the admin

noirbizarre:
This PR simply add a link to password change into the admin. This is a short term solution.

Long term should be a refactoring of user/me admin page


js/views/me.vue
```diff
@@ -81,6 +85,9 @@ export default  {
         this._handler.remove();
     },
     methods: {
+        change_password() {
+            document.location = '/change';
```
abulte: This won't work if udata is served from a subfolder, does it?

noirbizarre: Yes but right now, it already can't be served in a subfolder because this pattern is already used in many places.

abulte: Ok :-)

Issue #1471, opened 2018-03-01
Elasticsearch analyser Internal server error.


micael-grilo:
Internal server error caused on search as "pt_analyzer" is missing

![image](https://user-images.githubusercontent.com/9539326/36855172-b9a53ffe-1d6a-11e8-9697-156ed61bc928.png)

## Detailed Description
uData is loooking for a specific Elasticsearch analyser depending on "DEFAULT_LANGUAGE" setting.
https://github.com/opendatateam/udata/blob/2c3885029cbf9401010b2832ca2999afaf6f342c/udata/search/__init__.py#L113-L115

In our case as our default language setting is "pt" this custom analyser doesn't exist, causing an AttributeError:
https://github.com/opendatateam/udata/blob/2c3885029cbf9401010b2832ca2999afaf6f342c/udata/search/analysis.py#L71 

## Possible Implementation

In our case we will need also a custom analyser named "pt_analyzer" because we will need special Elasticsearch filters as for example the "ASCII Folding Token filter" or "asciifolding", but it would be nice to set a default analyser when the i18n custom is missing.
https://github.com/opendatateam/udata/blob/2c3885029cbf9401010b2832ca2999afaf6f342c/udata/search/__init__.py#L113-L115

by changing line L115 to:
 `return getattr(analysis, '{0}_analyzer'.format(language)) or getattr(analysis, 'standard')`

Let me know if this change makes sense, and also if we could add a custom "pt_analyzer". 
Can we open a pull request? :)

Best!

noirbizarre:
Hi !

Thanks for the detailed issue. I understand perfectly the problem and the proposed changes make sense.

You should submit both pull requests:
- one to add a `pt_analyzer`
- one to handle the fallback on the `standard` analyzer

I'll be glad to help and answer your questions here or on gitter if you need (obviously I can't help on Portuguese specifics in `pt_analyzer`, but I can help on everything else)

noirbizarre:
Closed by #1482

PR #1482, opened 2018-03-06
Fallback to the standard if a specific i18n analyzer is missing

micael-grilo:
Issue: #1471 fall back to the standard analyzer if a specific i18n analyzer is missing


noirbizarre:
Thanks !

Closes #1471

PR #1484, opened 2018-03-07
High priority for sendmail tasks

abulte:


udata/settings.py
```diff
@@ -81,6 +85,10 @@ class Defaults(object):
             'queue': 'low',
             'routing_key': 'low.harvest',
         },
+        'udata_piwik.tasks.*': {
```
noirbizarre: You're right, we don't want some udata extension config there.
Another PR is coming for tasks declaration time default routing.

Only keep the `udata.*` in this PR

PR #1487, opened 2018-03-08
Tasks routing

noirbizarre:
This PR refactor the task routing: the default route can be defined at declaration time, on both `task` and `job`, by either using:
- `queue` and/or `routing_key` parameters
- `route` parameter from which `queue` and `routing_key` will be computed (`route=high.somewhere` gives `queue=high` and ` routing_key=high.somewhere`)

This also allows extension to define routing on their tasks.

Some route have been added to the existing tasks and jobs.

### DIFF
--- a/udata/search/__init__.py
+++ b/udata/search/__init__.py
@@ -15,7 +15,7 @@ from mongoengine.signals import post_save
 from speaklater import is_lazy_string
 from werkzeug.local import LocalProxy
 
-from udata.tasks import celery
+from udata.tasks import task
 
 
 from . import analysis
@@ -121,7 +121,7 @@ def get_i18n_analyzer():
 i18n_analyzer = LocalProxy(lambda: get_i18n_analyzer())
 
 
-@celery.task
+@task(route='high.search')
 def reindex(obj):
     model = obj.__class__
     adapter_class = adapter_catalog.get(model)
--- a/udata/tasks.py
+++ b/udata/tasks.py
@@ -18,6 +18,9 @@ class ContextTask(Task):
     abstract = True
     schedulable = None
     current_app = None
+    route = None
+    default_queue = None
+    default_routing_key = None
 
     def __call__(self, *args, **kwargs):
         with self.current_app.app_context():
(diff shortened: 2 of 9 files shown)


udata/search/__init__.py
```diff
@@ -121,7 +121,7 @@ def get_i18n_analyzer():
 i18n_analyzer = LocalProxy(lambda: get_i18n_analyzer())
 
 
-@celery.task
+@task(queue='high', routing_key='high.search')
```
abulte: why not `route='high.search'`? To stay consistent.

udata/tasks.py
```diff
@@ -26,6 +29,7 @@ def __call__(self, *args, **kwargs):
 
 class JobTask(ContextTask):
     abstract = True
+    default_queue = 'low'
```
abulte: Maybe `default`? Because, you know, `default` ;-)

udata/tasks.py
```diff
@@ -78,6 +108,19 @@ def log_test(self):
     self.log.error('This is a ERROR message')
 
 
+@job('test-low-queue', queue='low')
+@job('test-default-queue', queue='default')
+@job('test-high-queue', queue='high')
+def queue_test(self, raises=False):
```
abulte: can't be defined in test files?

udata/search/__init__.py
noirbizarre: Just forgot to remove the `queue`. I added the `route` after as a shortcut.

udata/tasks.py
noirbizarre: No, this is intentional, like `log-test` and `error-test`: this allow to manually run these jobs with `udata job run -d test-high-queue` and see if this is properly routing.
You can also run `udata job run -d test-high-queue raises=True` and you can have the time for a tasks to be processed because the exception will appear in Sentry with a timestamp.

udata/tasks.py
noirbizarre: This is `default` for tasks, jobs are meant to be long running tasks this is why I set them to `low` by default. But I can undo this, but this mean we have to declare `queue=low` manually for every job

abulte: Makes senses 👍

PR #1493, opened 2018-03-12
Upload cleanup

noirbizarre:
This PR:
- cleanup chunks after a successful upload
- allows fine uploader to retry failed chunks
- make sentry ignore upload progress
- takes advantage of Flask-FS 0.5 new `metadata` methods and so avoid the need for a local file copy (and so twice the upload size available in chunks)
- ensure tests storages are local and properly cleaned up

### DIFF
--- a/udata/core/storages/api.py
+++ b/udata/core/storages/api.py
@@ -45,6 +45,17 @@ class UploadStatus(Exception):
         self.error = error
 
 
+class UploadProgress(UploadStatus):
+    '''Raised on successful chunk uploaded'''
+    pass
+
+
+class UploadError(UploadStatus):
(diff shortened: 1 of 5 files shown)


udata/core/storages/api.py
```diff
@@ -45,6 +45,17 @@ def __init__(self, ok=True, error=None):
         self.error = error
 
 
+class UploadProgress(UploadStatus):
+    '''Raised on successful chunk uploaded'''
+    pass
+
+
+class UploadError(UploadStatus):
+    '''Raised any upload error'''
```
abulte: s/Raised/Raise or s/Raised/Raised by/

PR #1505, opened 2018-03-14
GeoZones/GeoIDs fixes

noirbizarre:
This PR ensure:
- `country-subset` (`country-subset:iso:name`) geoids are properly parsed (until a better ID is found in geozones for these zones)
- spatial zones properly load:
  - when not using `--drop`, perform update on existing zones
  - default values are handled
- Geozones with unicode label successfully (re)index

This PR also migrate existing geozones loaded with an incorrect validity (`validity: {}`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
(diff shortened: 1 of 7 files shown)


noirbizarre:
(The fix is only the first point but while I was working on it I saw other things to fix ;)

PR #1509, opened 2018-03-15
Lazy Mongo connexion

noirbizarre:
This PR delay the MongoDB conneection to the first real access.
This will:
- prevent warning on Celery start (and prevent non fork-safe usage in Celery)
- prevent the need to have MongoDB running for every commands, they will fail only on first real access if any


noirbizarre:
Yes, I was gonna forget ! (But handling Sentry issues by frequency and last seen reminded me fast enough!)

PR #1511, opened 2018-03-15
Support OAuth2 request-body credentials

noirbizarre:
This PR allows OAuth2 request-body credentials (ie. you can submit `client_id` and `client_secret` in the POST body when querying `/oauth/token`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Support request-body credential in OAuth2 [#1511](https://github.com/opendatateam/udata/pull/1511)
```
jdesboeufs: You can add something like `Fix a regression introduced in 1.3.0`.

PR #1530, opened 2018-03-26
Fixes search result vertical cut off (fix opendatalu/gouvlu#63)

noirbizarre:
See https://github.com/opendatalu/gouvlu/issues/63


noirbizarre:
Not in this case, `em` is relative to the parent font size :/

Issue #1466, opened 2018-02-28
Make API documentation country agnostic 


petzlux:
## Detailed Description
At the moment, the API documentation template examples are specific to data.gouv.fr . It would be nice if the content could be changed to remove these references, or at least make the examples as generic as possible. 

## Context
For deployments in other countries.

noirbizarre:
Closed by #1542

PR #1542, opened 2018-03-27
Bug #1466 - Make API documentation country agnostic

micael-grilo:
Hi, 

Please see in attach my contribution to Bug #1466.
I've tried to make the examples as generic as possible.

Best!

### DIFF
--- a/udata/templates/apidoc.html
+++ b/udata/templates/apidoc.html
@@ -80,7 +80,7 @@
     "page": 1,
     "page_size": 20,
     "total": 43,
-    "next_page": "http://{{config.SERVER_NAME}}/api/endpoint/?page=2",
+    "next_page": "https://{{config.SERVER_NAME}}/api/endpoint/?page=2",
     "previous_page": null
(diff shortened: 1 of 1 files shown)


noirbizarre:
Sorry, I forgot: for every PR with user noticeable changes we try to submit a CHANGELOG entry. Can you do so (in another PR because I merged this one to quickly) ?

micael-grilo:
@noirbizarre in fact with dynamic documentation (using a real organization) it would be better.
What you have in mind? Get the top or first 20 organizations and then choose one ramdomly to display as example?

Btw, the idea is to send an organization list (limit 20) as an parameter to the JINJA template in this line, then deal with the organization in the Jinja template, is it ok? Or do you suggest a better approach.
https://github.com/opendatateam/udata/blob/d02e371870d519dcf2c428427b5a183d5b87be82/udata/api/__init__.py#L270

I will create a new PR for the CHANGELOG.
Thanks

PR #1547, opened 2018-03-29
Resource modals: use API call

abulte:
This PR:
- introduces a new API endpoint to get a single resource
- uses this API endpoint to display a markdown formatted description on the the front resource modal
- uses the API endpoint to display check results (parsing is easier than w/ JSON-LD)
- uses the API endpoint to display back resource modal (instead of the whole dataset)

I kept both JSON-LD and API call results in the modal, because the JSON-LD stuff is loaded instantaneously (already present and needed in the markup of the dataset page). This can be discussed.

~I tried to use the new API endpoint on the back resource modal too (as of today we're loading the entire dataset when opening each modal) but there were some dependencies in the modal to the dataset object.~

capture d ecran 2018-03-29 11 51 32

### DIFF
--- a/js/components/dataset/resource/modal.vue
+++ b/js/components/dataset/resource/modal.vue
@@ -132,7 +132,7 @@ export default {
         return {
             edit: false,
             confirm: false,
-            dataset: new Dataset(),
+            dataset: this.$parent.$parent.dataset,
             resource: new Resource(),
             next_route: null
         };
@@ -167,10 +167,11 @@ export default {
                     params: parent.params
                 };
             }
-            this.dataset.fetch(this.$route.params.oid);
             if (this.$route.name.includes('community')) {
                 this.resource = new CommunityResource();
                 this.resource.fetch(this.$route.params.rid);
+            } else {
+                this.resource.fetch(this.$route.params.oid, this.$route.params.rid);
             }
         }
     },
--- a/js/front/dataset/resource-modal.vue
+++ b/js/front/dataset/resource-modal.vue
@@ -1,36 +1,36 @@
 <template>
 <modal class="resource-modal" v-ref:modal
-    :title="resource.name">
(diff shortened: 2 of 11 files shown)


js/components/dataset/resource/modal.vue
```diff
@@ -132,7 +132,7 @@ export default {
         return {
             edit: false,
             confirm: false,
-            dataset: new Dataset(),
+            dataset: this.$parent.$parent.dataset,
```
noirbizarre: Need a store ? :stuck_out_tongue_winking_eye:

js/front/dataset/resource-modal.vue
```diff
@@ -1,36 +1,36 @@
...
+    :title="resource.title">
 
     <div class="modal-body">
-        {{{ resource.description }}}
+        {{{ resource.description|markdown }}}
 
         <dl class="dl-horizontal dl-wide">
           <dt>{{ _('URL') }}</dt>
-          <dd><a :href="resource.contentUrl" @click="onClick">{{resource.contentUrl}}</a></dd>
+          <dd><a :href="resourceJsonLd.contentUrl" @click="onClick">{{resourceJsonLd.contentUrl}}</a></dd>
```
noirbizarre: Why not user `resource` everywhere ? This will allow to strip from the JSON-LD every non standard values we added

udata/tests/api/test_datasets_api.py
```diff
@@ -548,6 +548,19 @@ def setUp(self):
         self.login()
         self.dataset = DatasetFactory(owner=self.user)
 
+    def test_get(self):
+        '''It should fetch a resource from the API'''
+        with self.autoindex():
```
noirbizarre: I think the `autoindex` is useless: this API does not rely on search

PR #1567, opened 2018-04-09
Enforce a domain whitelist when resource.filetype is file

### LINKED ISSUE #988: Clean up filetypes for datasets submitted through API
Apart from inspire and geo-ide, we have a lot of resources submitted through the API which have the wrong file `type` attribute (`file` instead of `remote`). It should be fixed by a migration for the stock and avoided at the submission level.

Note: files from `static.dgfr` (very old version of dgfr) should be considered as `remote` too.

abulte:
Fix #988 

Migration in https://github.com/etalab/udata-gouvfr/pull/292

### DIFF
--- a/docs/adapting-settings.md
+++ b/docs/adapting-settings.md
@@ -93,6 +93,15 @@ The duration used for templates' cache, in minutes.
 
 This is the allowed resources extensions list that user can upload.
 
+### RESOURCES_FILETYPE_FILE_ALLOWED_DOMAINS
+
+**default**: `tuple()`
+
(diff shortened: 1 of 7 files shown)


docs/adapting-settings.md
```diff
@@ -93,6 +93,15 @@ The duration used for templates' cache, in minutes.
 
 This is the allowed resources extensions list that user can upload.
 
+### RESOURCES_FILETYPE_FILE_ALLOWED_DOMAINS
```
noirbizarre: What a long name, maybe something shorter ? `RESOURCES_FILE_ALLOWED_DOMAINS` ?

PR #1571, opened 2018-04-10
Feature/dynamic api doc

micael-grilo:
Hi, 

As suggested by @noirbizarre  on Pull #1542, this is a PR to create dynamic API documentation with real organizations.

L271 on udata/api/__init__.py can be changed to improve/tune organizations to display. ( As you know uData better than me, maybe there's a better param to get the top organizations.) :)

Best!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Current (in progress)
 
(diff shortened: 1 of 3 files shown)


noirbizarre:
As a first implementation, it's just fine. Let's iterate on this work.
I think having the same query without parameter won't change the displayed organizations as first query results are those with the maximum scoring, which include the number of datasets.

PR #1572, opened 2018-04-10
Upgrade to authlib 0.6

lepture:
Things you need to know:

1. client_type is removed, we use `token_endpoint_auth_method` to determine the authenticate method for client.
2. there should be a `revoked` status for token, a revoked token can not be re-used

### token.revoked

consider this:

1. Client Foo got user A's token: `abc`
2. user A revoked token: `abc`
3. user B issued a token to client Bar with token: `abc`
4. Client Foo can use `abc` to access user B's resource

Never delete a token.

### refresh token

The current implementation for refresh token is also dangerous, same reason as revoke token.


udata/api/oauth2.py
```diff
@@ -104,16 +105,12 @@ def check_redirect_uri(self, redirect_uri):
...
-            return self.confidential
-        elif client_type == 'public':
-            return not self.confidential
-        raise ValueError('Invalid client_type')
-
     def check_client_secret(self, client_secret):
         return self.secret == client_secret
 
+    def check_token_endpoint_auth_method(self, method):
+        return method in ('client_secret_post', 'client_secret_basic')
```
lepture: Implicit grant requires `none` method.

lepture:
Implicit grant requires a client without client_secret. I'm not sure how to fix the test.

noirbizarre:
I made a pull request to your pull request: https://github.com/lepture/udata/pull/1
It:
- removes last deprecation warnings
- allows the creation of public clients (without a `client_secret`)

This will allow to merge your pull request and from there we will be able to test it and adjust if needed

noirbizarre:
Thanks for the contribution !
We'll test it ASAP :ok_hand:

PR #1577, opened 2018-04-11
Fix org search sort

noirbizarre:
This PR fixes the broken organization "sort by creation date"


noirbizarre:
Don't pay attention

PR #1579, opened 2018-04-11
Display dataset recommendations from udata-recommendations

abulte:
Related to https://github.com/etalab/udata-gouvfr/pull/294 and https://github.com/opendatateam/udata-recommendations/pull/1/files

The `template_hook` thingy might not be ideal, RFC.

### DIFF
--- a/udata/frontend/__init__.py
+++ b/udata/frontend/__init__.py
@@ -21,11 +21,22 @@ log = logging.getLogger(__name__)
 front = I18nBlueprint('front', __name__)
 
 _footer_snippets = []
+_template_hooks = {}
 
 
 def footer_snippet(func):
     _footer_snippets.append(func)
     return func
+
+
+def template_hook(func):
+    _template_hooks[func.__name__] = func
+    return func
+
+
+@front.app_context_processor
+def inject_template_hooks():
+    return {'hook_%s' % k: v for (k, v) in _template_hooks.iteritems()}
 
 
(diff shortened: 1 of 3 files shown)


udata/models/extras_fields.py
```diff
@@ -66,3 +66,6 @@ def validate(self, value):
             types = ', '.join(t.__name__ for t in ALLOWED_TYPES)
             raise ValidationError('Value should be an instance of: {types}',
                                   types=types)
+        if isinstance(value, list):
```
noirbizarre: Maybe `isinstance(value, (list, tuple))`

udata/frontend/__init__.py
```diff
@@ -21,13 +21,24 @@
...
 
 
+def template_hook(func):
+    _template_hooks[func.__name__] = func
+    return func
+
+
+@front.app_context_processor
+def inject_template_hooks():
+    return {'hook_%s' % k: v for (k, v) in _template_hooks.iteritems()}
```
noirbizarre: Maybe prepare for Python 3, use `items()` (We won't have a lot of hooks, difference won't be noticeable)

udata/models/extras_fields.py
abulte: 👍

udata/frontend/__init__.py
abulte: 👍

PR #1583, opened 2018-04-12
Prepare for legacy PyPI closure and PyPI warehouse supporting markdown

noirbizarre:
This PR:
- update the `setup.py` file to support markdown description
- update PyPI links to `pypi.org` to ensure everything is ready for PyPI Legacy closure on 30 April 2018

See for reference:
- [The PyPI migration roadmap](https://wiki.python.org/psf/WarehouseRoadmap)
- [Markdown Descriptions on PyPI](https://dustingram.com/articles/2018/03/16/markdown-descriptions-on-pypi)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,27 +8,12 @@ import re
 from setuptools import setup, find_packages
 
 RE_REQUIREMENT = re.compile(r'^\s*-r\s*(?P<filename>.*)$')
-RE_MD_CODE_BLOCK = re.compile(r'```(?P<language>\w+)?\n(?P<lines>.+?)```', re.S)
-RE_CODE = re.compile(r'`(?P<code>[^`]+)`')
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -111,12 +58,14 @@ def pip(filename):
...
     description=__import__('udata').__description__,
     long_description=long_description,
+    long_description_content_type='text/markdown',
     url='https://github.com/opendatateam/udata',
     author='Opendata Team',
     author_email='opendatateam@data.gouv.fr',
     packages=find_packages(),
     include_package_data=True,
     install_requires=install_requires,
+    setup_requires=['setuptools>=38.6.0'],
```
abulte: 👍 Do this in `udata-gouvfr`?

noirbizarre: Yes, it's done. We need to do this for all packages

PR #1586, opened 2018-04-13
WIP: Factorize/unify uris validation

noirbizarre:
This PR factorize URls handling ensuring every part of the app use the same rules (models, forms, harvesters...).
The common helpers is fully tested, support utf8, rely on official IANA TLDs list by default and is configurable to work localy or on an intranet.

The following parameters are added:
### URLS_ALLOW_PRIVATE

**default**:  `False`

Whether or not to allow private URLs (private IPs...) submission

### URLS_ALLOW_LOCAL

**default**: `False`

Whether or not to allow local URLs (localhost...) submission.
When developping you might need to set this to `True`.

### URLS_ALLOWED_SCHEMES

**default**: `('http', 'https', 'ftp', 'ftps')`

List of allowed URL schemes.

### URLS_ALLOWED_TLDS

**default**: All IANA registered TLDS

List of allowed TLDs.
When using udata on an intranet, you might want to add you custom TLDs:

```python
from udata.uris import TLDS

URLS_ALLOWED_TLDS = TLDS + set(['custom', 'company'])

### CODE REVIEW

File: udata/forms/fields.py [hunk 1]
```diff
@@ -130,8 +129,8 @@ class URLField(EmptyNone, Field):
     def pre_validate(self, form):
         if self.data:
             try:
-                db.URLField().validate(self.data)
-            except db.ValidationError:
+                uris.validate(self.data)
```
abulte: why not use the logic from `udata/models/url_field.py` here?

udata/tests/test_model.py
```diff
@@ -261,14 +261,14 @@ def test_handle_unicode(self):
...
         url = 'http://10.10.0.2/path/'
-        URLTester(url=url).save()
+        PrivateURLTester(url=url).save()
         with self.assertRaises(ValidationError):
-            PublicURLTester(url=url).save()
+            URLTester(url=url).save()
 
 
-class DatetimedTest(DBTestMixin, TestCase):
+class Darivate(DBTestMixin, TestCase):
```
abulte: `Darivate`?

udata/tests/test_uris.py
```diff
@@ -0,0 +1,294 @@
...
+    'http://.www.foo.bar/',
+    'http://www.foo.bar./',
+    'http://.www.foo.bar./',
+    'http://[fffff:1450:4007:80e::2004]',
+    'http://[fffff:1450:4007:80e::2004]:8080',
+    'http://[fffff:1450:4007:80e::2004]:8080/index.html',
+    'http://[::]',
+    'http://[::]:8080',
+    'http://[::]:8080/index.html',
+]
```
abulte: Ok, I see what you mean when you said bulletproof ;-)

udata/uris.py
```diff
@@ -0,0 +1,121 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import re
+
+from flask import current_app
+from netaddr import IPAddress, AddrFormatError
+
+from udata.settings import Defaults
+# from tlds import tld_set as TLDS
```
abulte: Remove?

udata/uris.py
```diff
@@ -0,0 +1,121 @@
...
+    # resource path
+    r'(?:/\S*)?'
+    # query string
+    r'(?:\?\S*)?'
+    r'$',
+    re.UNICODE | re.IGNORECASE
+)
+
+# SCHEMES = Defaults.URLS_ALLOWED_SCHEMES
+# TLDS = Defaults.URLS_ALLOWED_TLDS
```
abulte: Remove?

udata/tests/test_model.py
noirbizarre: Weird, I do known from where that comes from

udata/tests/test_uris.py
noirbizarre: :stuck_out_tongue_winking_eye:

udata/tests/test_model.py
abulte: Probably a multi-select gone wrong ;-)

noirbizarre: Yes, it must be

udata/forms/fields.py
noirbizarre: This is the same logic, `udata/models/url_field.py` also use `uris.validate()` but to me this is more explicit to show the comme piece `uris.validate`

abulte: I liked the idea of decoupling a little bit more here: rely on `URLField` that relies on `uris` in turn. But not a big problem.

noirbizarre: If you prefer I can, but:
- I though it was easier to find the common dependency when being explicitly written
- `db.URLField.validate()` is an instance method relying on custom field configuration. I though about giving the same parameters here too, but this is not currently in use
- model and form validation are slightly different, and I wanted to be able highlight the differences while keeping a neutral behavior in `uris.validate`

abulte: LGTM

noirbizarre: :ok_hand:

PR #1616, opened 2018-04-24
Remove unwanted commas in default SECURITY_EMAIL_SUBJECT_* parameters

noirbizarre:
This PR removes unwanted commas in default `SECURITY_EMAIL_SUBJECT_*` (which makes it tuples instead of expected strings)


noirbizarre:
Don't worry, this does not seems to be critical :+1:

PR #1618, opened 2018-04-24
Make it easier to read sentry configuration

noirbizarre:
This PR expose the default Sentry configuration the same way than other modules: set default values in `udata.settings.Defaults`.
This avoid some extra `app.config.setdefault()` calls and make it easier to read the default configuration and the Sentry initialization.

The sentry configuration parameters are also documented.

### DIFF
--- a/docs/adapting-settings.md
+++ b/docs/adapting-settings.md
@@ -477,6 +477,43 @@ The default page size for post listing
 
 Max number of resources to display uncollapsed in dataset view.
 
+## Sentry configuration
+
+### SENTRY_DSN
+
+**default**: `None`
+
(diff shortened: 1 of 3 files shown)


docs/adapting-settings.md
```diff
@@ -477,6 +477,43 @@ The default page size for post listing
...
+
+The Sentry DSN associated to this udata instance.
+If defined, the Sentry support is automatically activated.
+
+### SENTRY_TAGS
+
+**default**: `{}`
+
+A key-value map of extra tags to pass as Sentry context.
+See: <https://docs.sentry.io/learn/context/>
```
abulte: Markdown link? (and below)

PR #1658, opened 2018-05-10
RESOURCES_FILE_ALLOWED_DOMAINS default value fix

abulte:
`tuple()` would break when doing things like this https://github.com/opendatateam/udata/blob/master/udata/core/dataset/forms.py#L44


docs/adapting-settings.md
```diff
@@ -95,7 +95,7 @@ This is the allowed resources extensions list that user can upload.
 
 ### RESOURCES_FILE_ALLOWED_DOMAINS
 
-**default**: `tuple()`
+**default**: `list()`
```
noirbizarre: Maybe `[]` instead of `list()`. `tuple()` is used because empty tuple is written `(, )` which isn't very readble

Issue #1637, opened 2018-05-02
Translate labels in registration form 


micael-grilo:
Apply i18n to "First name" and "Last name" labels in the registration form.

![image](https://user-images.githubusercontent.com/9539326/39521907-65f68088-4e08-11e8-94a2-faabaff00a80.png)

abulte:
We're (desperately) waiting for a new release of Flask-Security to fix this. https://github.com/mattupstate/flask-security/releases. Related to #1256.

We'll probably have to fork and release ourselves or pin a commit as a dependency.

micael-grilo:
These fields are an extension to the Flask-Security "RegisterForm",

https://github.com/opendatateam/udata/blob/acd849ab28b08c915ae3c2902796f78ebf6cfeb8/udata/auth/forms.py#L9-L13

The fields "First name" and "Last name" are already in the translation files, so applying i18n to these fields wouldn't solve it? (something like this... )
```python
from udata.i18n import lazy_gettext as _

class ExtendedRegisterForm(RegisterForm):
    first_name = fields.StringField(
        _('First name'), [validators.Required('First name is required')])
    last_name = fields.StringField(
        _('Last name'), [validators.Required('Last name is required')])
```

micael-grilo:
Fixed on v1.4.0

PR #1668, opened 2018-05-15
Apply i18n to First and Last name in registration form

micael-grilo:
Hi uData team, 

Please see my solution to #1637.

Best,


udata/auth/forms.py
```diff
@@ -4,10 +4,10 @@
 from flask_security.forms import RegisterForm
 from udata.forms import fields
 from udata.forms import validators
-
+from udata.i18n import lazy_gettext as _
```
abulte: Add a blank line (2 blank lines before class definition)

PR #1676, opened 2018-05-16
Fix the temporal coverage facet/field querystring parsing

### LINKED ISSUE #1673: Temporal Coverage filter doesnt work
This doesnt work https://data.public.lu/fr/datasets/?temporal_coverage=2015-01-01-2018-05-16

noirbizarre:
Fix #1673


udata/search/fields.py
```diff
@@ -79,6 +79,7 @@ def validate_parameter(self, value):
         return value
 
     def get_value_filter(self, value):
+        print('get_value_filter', value)
```
taniki: :see_no_evil:

PR #1690, opened 2018-05-28
Conditional community resources anchor

abulte:
As suggested by @cquest.

It was also suggested to add the number of community resources in the anchor link, but I feel it would be too much information. Please advise.

### DIFF
--- a/less/udata/resource.less
+++ b/less/udata/resource.less
@@ -1,10 +1,12 @@
 .resources-list {
-    h3 {
-        display: inline;
-    }
     .resources-list-header {
         margin-top: 20px;
         margin-bottom: 10px;
+        display: flex;
+        h3 {
+            margin-top: 0;
+            flex-grow: 1;
+        }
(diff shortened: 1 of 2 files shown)


abulte:
@noirbizarre right, I did not think about the "you can add one yourself" argument. So now I'm not so sure... @taniki WDYT?

abulte:
Since we don't explain how a community resource can be useful when there is none, I'd still lean toward hiding the link when there's none.

taniki:
I would recommend to separate section header and the link. Also indicate that's an interactive element with some wording. The repetition of "resource" is not very aesthetic too.

Conditional is fine. A counter will be good too to add density but then I will be weird to show it only if there is community resources.

screenshot-2018-5-30 base sirene des entreprises et de leurs etablissements siren siret - data gouv fr

abulte:
capture d ecran 2018-05-31 17 38 57

Issue #1688, opened 2018-05-25
close auto-completion after using a search box


taniki:
Currently, auto-completion stay open after navigation to search results

PR #1693, opened 2018-05-28
prevent autocomple to open on new page

taniki:
Currently if you use the search input, autocomplete will stay open after having loaded the new page because of a watcher on `this.query`.

### DIFF
--- a/js/components/site-search.vue
+++ b/js/components/site-search.vue
@@ -18,6 +18,7 @@
         <input name="q" type="search" class="form-control" autocomplete="off"
             :placeholder="placeholder || _('Search')"
             v-model="query" debounce="200"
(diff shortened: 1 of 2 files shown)


js/components/site-search.vue
```diff
@@ -18,6 +18,7 @@
         <input name="q" type="search" class="form-control" autocomplete="off"
             :placeholder="placeholder || _('Search')"
             v-model="query" debounce="200"
+            @keydown = "show = true"
```
abulte: spaces ;-)

taniki: :cowboy_hat_face:

taniki:
close #1688

PR #1698, opened 2018-05-29
Extras cleanup

noirbizarre:
This PR prepare the `ExtrasField`s refactoring (will be easier to read in multiple PR):
- port related tests to pytest syntax
- adds some missing form fields: `DateField` and `FloatField`
- fix the `assert_json_equal` by using the udata JSON serializer
- adds some missing `ExtrasField` model tests
- simplify the `ExtraField` form field signature by removing the need to pass the `extras` parameter

PR #1699, opened 2018-05-29
Extras parsing and improvements

noirbizarre:
This PR ensures registered extras attributes are properly parsed from JSON (or any forms).
It also remove the need for `db.Extra` as registered types can only be a subclass of Mongoengine `BaseField` or `EmbeddedDocument`.

`EmbeddedDocument` support might lack the advanced parsing part. (Should be in another PR when use cases comes)

Requires #1698 (which is included here)

### DIFF
--- a/udata/forms/fields.py
+++ b/udata/forms/fields.py
@@ -693,9 +693,12 @@ class ExtrasField(Field):
 
     @property
     def extras(self):
+        '''Getter to the model extras field'''
         return getattr(self._form.model_class, self.short_name)
 
     def parse(self, data):
+        '''Parse fields and store individual errors'''
+        self.field_errors = {}
         return dict(
             (k, self._parse_value(k, v)) for k, v in data.items()
         )
@@ -705,7 +708,10 @@ class ExtrasField(Field):
             return value
         expected = self.extras.registered[key]
         if expected in self.TYPES:
-            return field_parse(self.TYPES[expected], value)
+            try:
+                return field_parse(self.TYPES[expected], value)
+            except Exception as e:
+                self.field_errors[key] = getattr(e, 'message', str(e))
         else:
             return value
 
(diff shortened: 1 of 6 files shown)


udata/forms/fields.py
```diff
@@ -664,7 +664,24 @@ def pre_validate(self, form):
...
+def field_parse(cls, value, *args, **kwargs):
+    kwargs['_form'] = Form()
+    kwargs['_name'] = 'extra'
+    field = cls(*args, **kwargs)
+    field.process_formdata([value])
+    return field.data
+
+
 class ExtrasField(Field):
+    TYPES = {
```
abulte: Maybe `VALIDATED_PARSED_TYPES` or something? If I understand correctly, other types are supported but won't be parsed/validated.

udata/models/extras_fields.py
```diff
@@ -20,15 +19,26 @@ def __init__(self, **kwargs):
...
     def validate(self, value):
         super(ExtrasField, self).validate(value)
 
         errors = {}
         for key, value in value.items():
-            extra_cls = self.registered.get(key, DefaultExtra)
+            extra_cls = self.registered.get(key)
+
+            if not extra_cls:
+                if not isinstance(value, ALLOWED_TYPES):
```
abulte: Maybe use https://pythonhosted.org/six/#six.string_types instead of `basestring` in `ALLOWED_TYPES` while we're at it ;-)

udata/models/extras_fields.py
```diff
@@ -37,8 +47,8 @@ def validate(self, value):
                      else extra_cls(**value).validate())
                 else:
                     extra_cls().validate(value)
-            except ValidationError as e:
-                errors[key] = e.message
+            except Exception as e:
```
abulte: What was wrong w/ `ValidationError`? Seemed safer.

udata/models/extras_fields.py
abulte: Forget it, search and replace will be easy.

noirbizarre: Hum :thinking:
I tried to avoid using `six` until now, but maybe both `str` and `unicode`

udata/models/extras_fields.py
noirbizarre: Because there are other exceptions raised depending on the used field :/

udata/forms/fields.py
noirbizarre: Exactly, they are now allowed unless register, maybe ALLOWED_TYPES as a class attribute ?

abulte: `ALLOWED_TYPES` might be confusing, others are still allowed ;-) `SUPPORTED` is not wrong but is it clear enough?

noirbizarre: or `KNOWN_TYPES` ?

udata/models/extras_fields.py
abulte: Can we list them? I feel like this is the kind of code that can catch a totally unrelated exception from way back and make it difficult to debug... Like a unicode mayhem for example.

udata/forms/fields.py
abulte: 👍

udata/models/extras_fields.py
noirbizarre: Not exactly, the `extra_cls` is dynamic, listing exceptions means handling every case of every known and unkown fields.
But don't worry, the exception is serialized into the errors on the approriate key, they are only validation errors

abulte: I'm still worried, but OK ;-)

noirbizarre: OK, I'll try to immrove that in another PR

noirbizarre: OK, I narrowed the catched exeptions

PR #1717, opened 2018-06-05
Make "Main file" translatable

noirbizarre:
This PR adds amissing translatable string (so needs i18n extraction after merge)


noirbizarre:
:wink:

PR #1722, opened 2018-06-05
kombu==4.2.1

abulte:
Avoid this https://github.com/celery/celery/issues/4772 with `celery==4.1.1`


abulte:
Not needed, it should install 4.2.1 directly.

PR #1733, opened 2018-06-07
Fix community resource creation and display

### LINKED ISSUE #1732: Can not create community resource
Since 1.4.0 it's not possible to create a community resource anymore: the form/link goes to a standard resource creation.

abulte:
Fix #1732

### DIFF
--- a/js/front/dataset/index.js
+++ b/js/front/dataset/index.js
@@ -122,7 +122,11 @@ new Vue({
             location.hash = `resource${communityPrefix}-${id}`;
             const modal = this.$modal(
                 ResourceModal,
-                {datasetId: this.dataset['@id'], resource: this.resourceFromJsonLd(resourceJsonLd)}
+                {
+                    datasetId: this.dataset['@id'],
+                    resource: this.resourceFromJsonLd(resourceJsonLd),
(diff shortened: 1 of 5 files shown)


js/front/dataset/index.js
```diff
@@ -122,7 +122,11 @@ new Vue({
             location.hash = `resource${communityPrefix}-${id}`;
             const modal = this.$modal(
                 ResourceModal,
-                {datasetId: this.dataset['@id'], resource: this.resourceFromJsonLd(resourceJsonLd)}
+                {
+                    datasetId: this.dataset['@id'],
+                    resource: this.resourceFromJsonLd(resourceJsonLd),
+                    isCommunity: isCommunity,
```
noirbizarre: Isn't it possible to read this boolean from the `resource` paramter attributes ?

abulte: Probably yes 👍

abulte: Nope, nothing in there.

noirbizarre: OK, so LGTM
(maybe we should think about exposing this attribute in resources)

PR #1734, opened 2018-06-07
Delete dataset reference in community resource when deleted

abulte:
Fix [WWW-DATAGOUVFR-74AT](https://sentry.data.gouv.fr/etalab/www-datagouvfr/issues/615128/)

This crashes the community resources list API.

It might be a little strange to leave community resources hanging in the air after their dataset is deleted, but we probably don't want to remove them either.

### DIFF
--- a/udata/migrations/2018-06-07-community-resources-remove-references-to-deleted-datasets.js
+++ b/udata/migrations/2018-06-07-community-resources-remove-references-to-deleted-datasets.js
@@ -1,0 +1,18 @@
+/**
+ * Delete references to deleted datasets from CommunityResources
+ */
+
+var deleted = 0;
+
+db.community_resource.find().forEach(function(resource) {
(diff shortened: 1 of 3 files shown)


udata/migrations/2018-06-07-community-resources-remove-references-to-deleted-datasets.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Delete references to deleted datasets from CommunityResources
+ */
+
+var deleted = 0;
+
+db.community_resource.find().forEach(function(resource) {
+    var cursor = db.dataset.find({'_id' : resource.dataset});
```
noirbizarre: or
```javascript
const dataset = db.dataset.findOne({'_id' : resource.dataset});
if (dataset) {
  ...
```

PR #1745, opened 2018-06-12
Cleaner harvest validation error messages

noirbizarre:
This PR fixes and cleanup harvest messages:
- Validation is now tested
- Multiple validation errors are properly serialized with their field name
- Missing required field is now properly formatted

The validation error was always formatted like this:
```
Validation error:
- [field] expected basestring for dictionnary key: 42
- expected basestring for dictionnary key: 42
- expected basestring for dictionnary key: 42
- required key not provided: None
```
they will now be formatted like this:
```
Validation error:
- [field] expected basestring: 42
- [other] expected basestring: 43
- [nested.key] expected int: wrong
- [nested.missing] required key not provided
```


noirbizarre:
See https://sentry.data.gouv.fr/share/issue/2e974ecb67764af19db44e9babc18449/ for an example of bad formatted validation error

PR #1748, opened 2018-06-13
Can parse iso datetime in daterange

noirbizarre:
This PR improve daterange parsing to handle iso datetimes.

See https://sentry.data.gouv.fr/share/issue/bcc9e83697b34fed82f4bde84695d879/ for an example of harvested iso daterange failing.

Parser also properly handle `date` and `datetime` values.

### DIFF
--- a/udata/tests/test_utils.py
+++ b/udata/tests/test_utils.py
@@ -51,6 +51,28 @@ class DateRangeTest:
         assert daterange_end(None) is None
         assert daterange_end('') is None
 
+    def test_parse_daterange_start_date(self):
+        today = date.today()
+        assert daterange_start(today) == today
+
+    def test_parse_daterange_end_date(self):
+        today = date.today()
+        assert daterange_end(today) == today
+
+    def test_parse_daterange_start_datetime(self):
+        now = datetime.now()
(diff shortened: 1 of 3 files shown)


udata/tests/test_utils.py
```diff
@@ -51,6 +51,28 @@ def test_parse_daterange_end_empty(self):
...
+    def test_parse_daterange_end_datetime(self):
+        now = datetime.now()
+        assert daterange_end(now) == now.date()
+
+    def test_parse_daterange_start_full_iso(self):
+        assert daterange_start('1984-06-07T00:00:00+00:00') == date(1984, 6, 7)
+
+    def test_parse_daterange_end_full_iso(self):
+        assert daterange_end('1984-06-07T00:00:00+00:00') == date(1984, 6, 7)
+
```
abulte: no tests for `dashes <= 1`?

noirbizarre: They already exists (See https://github.com/noirbizarre/udata/blob/85dd63b1b529e78583caf2b65affaf42c249ceee/udata/tests/test_utils.py#L93-L97)

PR #1766, opened 2018-06-21
Migrate to python 3 🎉

abulte:


abulte:
@noirbizarre re https://github.com/opendatateam/udata/pull/1766/commits/75a1d5cd349a72cae1235cbf86cd4041a8c40de5: we need to figure out a smart way to call the right python binary. I started by using `python3.6` everywhere since it's what https://github.com/chriskuehl/python3.6-debian-stretch does by default, and thus it works 👌 on Circle. But on my local env, `python3.6` links to a system python, not the `venv` one, breaking all `inv` commands. I reverted to `python3` on all `inv` tasks except the one used by Circle (`pydist`). It's error prone though.

PR #1767, opened 2018-06-21
Skip gov.uk references when unreachable

noirbizarre:
This PR prevents test suite to fail when references.gov.uk pages are not reachable.

### DIFF
--- a/udata/tests/dataset/test_dataset_rdf.py
+++ b/udata/tests/dataset/test_dataset_rdf.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+import requests
 import pytest
 
 from datetime import date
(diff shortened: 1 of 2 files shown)


udata/tests/dataset/test_dataset_rdf.py
```diff
@@ -37,6 +38,14 @@
     (FREQ.continuous, 'continuous'),
 ]
 
+GOV_UK_REF = 'http://reference.data.gov.uk/id/year/2017'
+
+try:
+    requests.get(GOV_UK_REF, timeout=0.1)
```
abulte: `.head()` maybe

udata/tests/dataset/test_dataset_rdf.py
```diff
@@ -37,6 +38,14 @@
...
+    GOV_UK_REF_IS_UP = True
```
abulte: `else` :p

PR #1771, opened 2018-06-25
Slug redirections

noirbizarre:
This PR test, fix and enable slug redirections on any models with a permalink using the slug.

Slug (and their routing) where undertested (tested against `StringField` instead of `SlugField`) with false positive.

The following redirects will be enabled:
- UTF-8 slug redirects to the normalized one
- old slugs are redirected to the new one changes (until another object take the same slug)
- redirections are deleted on object deletion

### DIFF
--- a/udata/core/dataset/models.py
+++ b/udata/core/dataset/models.py
@@ -360,8 +360,8 @@ class Dataset(WithMetrics, BadgeMixin, db.Owned, db.Document):
                                   default=datetime.now, required=True)
     title = db.StringField(required=True)
     acronym = db.StringField(max_length=128)
-    slug = db.SlugField(max_length=255, required=True,
-                        populate_from='title', update=True)
+    slug = db.SlugField(max_length=255, required=True, populate_from='title',
+                        update=True, follow=True)
     description = db.StringField(required=True, default='')
     license = db.ReferenceField('License')
 
--- a/udata/models/slug_fields.py
+++ b/udata/models/slug_fields.py
@@ -6,6 +6,7 @@ import slugify
 
 from flask_mongoengine import Document
 from mongoengine.fields import StringField
+from mongoengine.signals import post_delete
 
 from .queryset import UDataQuerySet
 
@@ -28,6 +29,9 @@ class SlugField(StringField):
         self.separator = separator
         self.follow = follow
         super(SlugField, self).__init__(**kwargs)
+        if follow:
+            # Can't use sender=self.owner_document which is not yet defined
+            post_delete.connect(self.cleanup_on_delete)
 
     def __get__(self, instance, owner):
         if instance is not None:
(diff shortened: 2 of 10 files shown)


udata/core/dataset/models.py
```diff
@@ -360,8 +360,8 @@ class Dataset(WithMetrics, BadgeMixin, db.Owned, db.Document):
                                   default=datetime.now, required=True)
     title = db.StringField(required=True)
     acronym = db.StringField(max_length=128)
-    slug = db.SlugField(max_length=255, required=True,
-                        populate_from='title', update=True)
+    slug = db.SlugField(max_length=255, required=True, populate_from='title',
+                        update=True, follow=True)
```
abulte: What does `follow` do?

udata/models/slug_fields.py
```diff
@@ -83,19 +110,26 @@ def populate_slug(instance, field):
...
 
-    manual = (not previous and value or
-              field.db_field in instance._get_changed_fields())
+    # Field value has changed
+    changed = field.db_field in instance._get_changed_fields()
+    # Field initial value has been manually set
+    manual = not previous and value or changed
 
     if not manual and field.populate_from:
+        # value to slugify is extract from populate_from parameter
```
abulte: s/extract/extracted/

udata/models/slug_fields.py
```diff
@@ -130,14 +163,24 @@ def exists(s):
...
-
-        slug_follower.new_slug = slug
-        slug_follower.save()
+    if field.follow and old_slug != slug:
+        ns = instance.__class__.__name__
+        # Destroy redirections from this new slug
+        SlugFollow.objects(namespace=ns, old_slug=slug).delete()
+
+        if old_slug:
+            print('slugs', old_slug, slug)
```
abulte: remove

udata/routing.py
```diff
@@ -64,23 +71,47 @@ class ModelConverter(BaseConverter):
...
+
     def to_python(self, value):
         try:
-            obj = self.model.objects(slug=value).first()
-        except InvalidQueryError:  # If the model doesn't have a slug.
+            quoted = self.quote(value)
+            query = db.Q(slug=value) | db.Q(slug=quoted)
+            obj = self.model.objects(query).get()
+        except (InvalidQueryError, self.model.DoesNotExist):
+            # If the model doesn't have a slug or matching slug doesn't exists.
```
abulte: s/exists/exist/

udata/core/dataset/models.py
abulte: It's a custom trigger for the whole `SlugFollow` logic, right?

noirbizarre: Yes `follow=True` enable redirections following slug changes. I kept the parameter name as it was

PR #1777, opened 2018-06-28
Improve usability at new organization creation

micael-grilo:
This is a PR with a improvement to the search organization button on administration. The search results appear as you type and clicking on the buttons reloads the page unnecessarily causing some confusion to the users, so I've removed this page reloading feature.

It adds also a 'Find or create your organization' button so the user can click on it when no organization results have been found. 

Before:
image

After:
image

Best!

### DIFF
--- a/js/components/organization/card-filter.vue
+++ b/js/components/organization/card-filter.vue
@@ -13,9 +13,9 @@
                 <input type="text" name="search" class="form-control"
                     :placeholder="_('Search')" v-model="search_query" />
                 <div class="input-group-btn">
-                    <button type="submit" name="submit" class="btn btn-warning btn-flat">
+                    <div class="btn btn-warning btn-flat">
                         <span class="fa fa-search"></span>
-                    </button>
+                    </div>
                 </div>
             </div>
         </form>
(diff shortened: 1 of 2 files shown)


taniki:
Thanks for the PR! This sounds great. You can add a screenshot (before/after) for cosmetical improvements. It is easier for quick reviews.

micael-grilo:
Updated with screenshots! :)

js/components/organization/card-filter.vue
```diff
@@ -13,9 +13,9 @@
                 <input type="text" name="search" class="form-control"
                     :placeholder="_('Search')" v-model="search_query" />
                 <div class="input-group-btn">
-                    <button type="submit" name="submit" class="btn btn-warning btn-flat">
+                    <div class="btn btn-warning btn-flat">
```
abulte: Why did you remove the `button`?

micael-grilo: The button is just reloading the page and you lose all your search criteria when you click on it, it does nothing as the search happens when you are typing so no need for a submit button. ( Users get confused when they click on it, the page reloads and the search gets lost.)

PR #1797, opened 2018-07-18
Fix checksum display on resource modal

abulte:
Binding was not working on nested `checksum` object.


abulte:
poke @thomasmenant

PR #1804, opened 2018-07-26
Allows to serve static asserts on an external CDN domain

noirbizarre:
This PR allows to serve static assets from an external CDN defined by `CDN_DOMAIN`.

### DIFF
--- a/docs/adapting-settings.md
+++ b/docs/adapting-settings.md
@@ -446,6 +446,16 @@ This activates metrics, this is deactivated for tests
 
 uData use Flask-FS as storage abstraction.
 
+## Flask-CDN options
+
+See [Flask-CDN README](https://github.com/libwilliam/flask-cdn#flask-cdn-options) for detailed options.
+
+### CDN_DOMAIN
+
+**default**: `None`
+
+Set this to a domain name. If defined, udata will serve its static assets from this domain.
+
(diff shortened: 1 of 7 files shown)


docs/adapting-settings.md
```diff
@@ -446,6 +446,16 @@ This activates metrics, this is deactivated for tests
...
 
+## Flask-CDN options
+
+See [Flask-CDN README](https://github.com/libwilliam/flask-cdn#flask-cdn-options) for detailed options.
+
+### CDN_DOMAIN
+
+**default**: `None`
+
+Set this to a domain name. If defined, udata will serve it's static assets from this domain.
```
abulte: s/it's/its/

udata/frontend/helpers.py
```diff
@@ -35,6 +36,12 @@ def now():
     return datetime.now()
 
 
+def cdn_for(endpoint, **kwargs):
```
abulte: Not a fan of the name. `cdn_or_flask_url_for` ? Not great either :-/ but more descriptive.

noirbizarre: This is only meant to be used for CDN assets, not as a replacement for all flask `url_for`.
I kept the name but added a docstring to be more explicit.

PR #1716, opened 2018-06-04
Harvest config backend

noirbizarre:
This PR:
- allows harvest backends to register their allowed filters
- expose backends filters in the API
- ensure the `HarvestSource.config` object is properly writable (with validation)
- Starts testing the base backend class

Requires #1715  (included)

PR #1812, opened 2018-08-01
Harvest config admin

noirbizarre:
This PR follows #1716 starts exposing the harvester configuration (only filtering right now) in the admin form.

#### Without filtering support
screenshot-data xps-2018 08 01-15-45-12

#### With filtering support
screenshot-data xps-2018 08 01-15-45-42

### Backends specific filters and localised wordings

Each backend has the responsibility to expose its handled filters and provide translated labels (See #1716)

screenshot-data xps-2018 08 01-15-44-07

screenshot-data xps-2018 08 01-15-44-37

## Side-effects

This feature rely on fields change detection so form fields gain change detection in this PR

### DIFF
--- a/js/components/harvest/filter-field.vue
+++ b/js/components/harvest/filter-field.vue
@@ -1,0 +1,91 @@
+<template>
+<div class="input-group filter-group">
+    <select class="form-control filter-group__type" v-model="type">
+        <option v-for="t in TYPES" :value="t.value" :key="t.value">{{ t.label }}</option>
+    </select>
+    <select class="form-control filter-group__key" v-model="key">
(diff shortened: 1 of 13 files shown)


js/components/harvest/filter-field.vue
```diff
@@ -0,0 +1,91 @@
...
+    computed: {
+        hasData() {
+            return this.key || this.value;
+        },
+        isDefined() {
+            return this.key && this.value;
+        },
+        placeholder() {
+            if (!this.key || !this.choices) return;
+            return this.choices.find(c => c.key ==this.key).description;
```
abulte: missing space

PR #1815, opened 2018-08-02
Ensure OEmbed works with CDN and fixes some static serialization

noirbizarre:
This PR:
- ensures OEmbed (and legacy widget) compatibility with external CDN
- fixes some static URL serialization cases

By the way, every existing OEmbed/legacy widget integration still works because the assets are still reachable on the old URL.


noirbizarre:
Yes, I will :+1:

PR #1834, opened 2018-08-17
Fix missing dataset title on client-side card listing

noirbizarre:
This PR fixes the missing dataset title on editorial home datasets.

## Explanation
`full_title` is `Dataset` object computed property (it's `title` + `acronym` if present).
Everywhere but here this is a list of `Dataset` object instances which is given as parameter.

To fix this, this PR allows to cast js `List`objects with an optionnal `model` option in constructor.

## Before
screenshot-data xps-2018 08 17-19-01-59

## After
screenshot-data xps-2018 08 17-18-56-38

### DIFF
--- a/js/models/base.js
+++ b/js/models/base.js
@@ -206,6 +206,7 @@ export class List extends Base {
 
         this.items = this.$options.data || [];
         this.query = this.$options.query || {};
+        this.model = this.$options.model;
 
         this.sorted = null;
         this.reversed = false;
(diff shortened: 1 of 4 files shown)


js/models/base.js
```diff
@@ -268,7 +269,7 @@ export class List extends Base {
     }
 
     on_fetched(data) {
-        this.items = data.obj;
+        this.items = this.model ? data.obj.map(o => new this.model({data: o})) : data.obj;
```
taniki: this looks like some black magic code having a wider scope than the PR desc

noirbizarre: Details on how and why added in the description

PR #1857, opened 2018-08-27
Posts workflow rework

noirbizarre:
This pull-request rework the post publication workflow.

### Publication date and publish/unpublish API

The `private` boolean attribute has been replaced by a `published` datetime (migration provided).
The publication is now a separate action from the creation/edition (and so a separate API)

#### Unpublished state
screenshot-data xps-2018 08 25-19-54-10

screenshot-data xps-2018 08 28-11-55-18

#### Published state

screenshot-data xps-2018 08 25-19-52-21

screenshot-data xps-2018 08 27-17-35-27

### Publish action in form and save without reload

User can now save the post and still continue editing. It can also publish the post from the form.

#### Unpublished state
screenshot-data xps-2018 08 25-19-57-38

#### Published state
screenshot-data xps-2018 08 25-20-04-49

### Bigger textarea
Initial textarea size is 12 rows (to fit in the window). User can still grow the textarea or toggle fullscreen if more is required

screenshot-data xps-2018 08 27-14-55-59

### Other admin fixes

Fixes broken reuses and datasets cards 

## Post layout

This pull-request provide bug fixes and enhancements on the post UI

### Alignements fixes

Sidebar cards as well as pager are now properly aligned on the layout

screenshot-data xps-2018 08 25-19-38-16
screenshot-data xps-2018 08 25-19-13-44

### Full width content instead of empty sidebar

When there is no content in the sidebar, the post content will grow to full-width avoid a big blank on the right

### Pluralized sidebar sections

Sections titles are now pluralized. That means we have `Associated dataset`/`Associated datasets` and `Associated reuse`/`Associated reuses`.

| Singular | Plural |
|------------|----------|
| screenshot-data xps-2018 08 25-19-55-57 | screenshot-data xps-2018 08 25-19-51-04 |

Translations needs to be updated.

### Other changes

There is now a `draft` label on unpublished posts

screenshot-data xps-2018 08 25-20-02-58

The default theme now has a background color on pager to avoir a full blank page with floating buttons

screenshot-data xps-2018 08 27-15-06-33


noirbizarre:
> - not sure about the color of the two first modals. Both are a bit brutal

These are colors already used on other modals (warning and dangers). I choose to reuse existing classes and style, but this can be changed in another PR (modal colors/classes admin-wide)

> - it is also the same colors than the buttons

Not sure to understand but same here, no style change to existing conventions. All buttons are using the same classes. We can change that but admin-wide in another PR. This PR focus on the workflow.

> - buttons should be switched

I don't understand. Which one ? Switched from ? to ?

> - maybe keep yello buttons for cancel actions

This is already the case: the `btn-warning` class is only applyied to the `Cancel`/`Back` button

> - not sure about "unpublication" as an actual word

Yes it is, I checked for the official term before 👉 https://en.wiktionary.org/wiki/unpublication :wink:

noirbizarre:
**NB**: but I'm open to any other prefered alternative to `unpublish` but I wasn' able to find one

noirbizarre:
`Cancel`/`Confirm` buttons switched, screenshots updated

noirbizarre:
Updated modals header wording ('`Publish`/`Unpublish`).
PR description screenshots updated

noirbizarre:
Publish modal `Cancel` button switched to outlined. Screenshot updated

PR #1872, opened 2018-09-05
Update changelog for #1826

abulte:
My local environment was broken due to the absence of `manifest.json`. I had to run `inv assets-build` to fix this, both in `udata` and `udata-gouvfr`. So this PR:
- moves #1826 to a breaking change
- hints at `inv assets-build` to generate `manifest.json`


abulte:
@noirbizarre I modified the message to reflect that it's a dev command and that the theme part is optional. I still think we need to include this in the CHANGELOG to avoid people having their dev env broken like we did.

ThibaudDauce (approved):

PR #1874, opened 2018-09-06
Fixes assets.from_manifest and add raw parameter

noirbizarre:
This PR:
- fixes `assets.from_manifest` to handle both relative or absolute paths
- adds a `raw` parameter to retrieve only the value from manifest

### DIFF
--- a/udata/assets.py
+++ b/udata/assets.py
@@ -55,9 +55,15 @@ def exists_in_manifest(app, filename):
     return app in _manifests and filename in _manifests[app]
 
 
-def from_manifest(app, filename, **kwargs):
+def from_manifest(app, filename, raw=False, **kwargs):
     '''
-    Get the path to a static file for a given app entry of a given type
+    Get the path to a static file for a given app entry of a given type.
+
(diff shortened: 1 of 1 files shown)


udata/assets.py
```diff
@@ -55,9 +55,15 @@ def exists_in_manifest(app, filename):
     return app in _manifests and filename in _manifests[app]
 
 
-def from_manifest(app, filename, **kwargs):
+def from_manifest(app, filename, raw=False, **kwargs):
     '''
-    Get the path to a static file for a given app entry of a given type
+    Get the path to a static file for a given app entry of a given type.
+
+    :param str app: The to which is tied this manifest
```
abulte: Missing word?

udata/assets.py
```diff
@@ -55,9 +55,15 @@ def exists_in_manifest(app, filename):
...
+    :param str filename: the original filename (without hash)
+    :param bool raw: If True, doesn't add prefix to the manifest
```
abulte: s/If/if/

PR #1888, opened 2018-09-14
Harvest preview improvements

noirbizarre:
This pull requests fixes some bugs on harvest preview:
- ensure that the harvest source edit form preview is done against the current form content instead of the stored sourceconfig
- prevent some unicode errors in harvest error handling

### DIFF
--- a/js/components/harvest/form.vue
+++ b/js/components/harvest/form.vue
@@ -56,6 +56,11 @@ export default {
             if (field.field.id == 'backend') {
                 this.backendValue = value;
             }
+        },
+        'form:change': function(form) {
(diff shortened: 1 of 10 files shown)


js/components/harvest/form.vue
```diff
@@ -56,6 +56,11 @@ export default {
             if (field.field.id == 'backend') {
                 this.backendValue = value;
             }
+        },
+        'form:change': function(form) {
+            if (form.validate()) {
+                this.$dispatch('harvest:source:form:changed', this.serialize())
```
abulte: missing `;`

PR #1894, opened 2018-09-20
replace label and action of button when format of resource is 'url'

taniki:


taniki:
Hum. Can we then agree that all of code repository are link to webpages (and not to html file to download)?

noirbizarre:
Yes we can but don't worry I approved the PR.
I just wanted to give you some hints for later improvements.
Applying the same change to code repositories might be a another good PR.
Having a configurable list of format to open instead of download can be another.

PR #1898, opened 2018-09-27
Api specs fixes

noirbizarre:
This PR fixes a lot of `swagger.json` API specs bugs allowing it to validate and so being able to use swagger tools that weren't working (like the new version of Swagger UI).

Some fixes are on the `flask-restplus` side so it has been updated with it

### DIFF
--- a/udata/api/commands.py
+++ b/udata/api/commands.py
@@ -6,10 +6,11 @@ import os
 
 import click
 
-from flask import json
+from flask import json, current_app
+from flask_restplus import schemas
 
 from udata.api import api
-from udata.commands import cli
+from udata.commands import cli, success, exit_with_error
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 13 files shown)


noirbizarre:
I think we should. In this PR or another ?

noirbizarre:
Added validation testing. (`flask-restplus` now provide schemas and validation support)

udata/api/commands.py
```diff
@@ -49,3 +50,15 @@ def postman(filename, pretty, urlvars, swagger):
...
+
+
+@grp.command()
+def validate():
+    '''Validate the Swagger/OpenAPI specification with your config'''
+    with current_app.test_request_context():
+        schema = json.loads(json.dumps(api.__schema__))
+    try:
+        schemas.validate(schema)
+        success('API specificationsare valid')
```
abulte: s/specificationsare/specifications are/

PR #1901, opened 2018-09-28
Allow to restore assets after being deleted ( datasets, organizations and reuses ).

micael-grilo:
Context:
When I delete an asset ( dataset, organization, reuse), I want to be able to restore it.

Resume:
Assets can be accidentally deleted, and there's no way to easily restore them, neither through API or through Admin area. This PR adds:

- Ability to restore a asset through API:
-- Sending a POST with the deleted field Empty ("deleted:"null") will restore the asset.

-  Change Admin Area to accommodate this new feature.
-- In admin area you can now easily restore an asset that was deleted.
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Allows arguments and keyword arguments in the task `@connect` decorator [#1908](https://github.com/opendatateam/udata/pull/1908)
 
 ## 1.6.0 (2018-10-02)
 
--- a/udata/tasks.py
+++ b/udata/tasks.py
@@ -88,9 +88,9 @@ def get_logger(name):
     return logger
 
 
-def connect(signal):
+def connect(signal, *args, **kwargs):
     def wrapper(func):
-        t = task(func)
+        t = task(func, *args, **kwargs)
 
         def call_task(item, **kwargs):
             t.delay(item, **kwargs)


js/components/dataset/restore-modal.vue
```diff
@@ -0,0 +1,52 @@
...
+import Modal from 'components/modal.vue';
+
+export default {
+    components: {Modal},
+    props: {
+        dataset: Object,
+    },
+    methods: {
+        confirm() {
+            var restore_dataset =  this.dataset;
```
noirbizarre: Use `const` instead of `var` because we are trying to use ES6 as much as possible

js/components/organization/restore-modal.vue
```diff
@@ -0,0 +1,53 @@
...
+import Modal from 'components/modal.vue';
+
+export default {
+    components: {Modal},
+    props: {
+        organization: Object,
+    },
+    methods: {
+        confirm() {
+            var restore_organization = this.organization;
```
noirbizarre: Same here

js/components/reuse/restore-modal.vue
```diff
@@ -0,0 +1,50 @@
...
+import Modal from 'components/modal.vue';
+
+export default {
+    components: {Modal},
+    props: {
+        reuse: Object
+    },
+    methods: {
+        confirm() {
+            var restore_reuse = this.reuse;
```
noirbizarre: Same here

js/views/dataset.vue
```diff
@@ -203,6 +213,12 @@ export default {
                 {dataset: this.dataset}
             );
         },
+        confirm_restore() {
+            var m = this.$root.$modal(
```
noirbizarre: No need to store the reference `m`, it's not used later.

js/views/organization.vue
```diff
@@ -258,6 +274,8 @@ export default {
                     class: 'danger',
                     label: this._('Deleted')
                 }];
+            } else {
+                this.badges = [];
```
noirbizarre: I don't think this is related to this pull-request. Remove it or submit another one for this case

js/views/reuse.vue
```diff
@@ -189,6 +205,8 @@ export default {
                     class: 'danger',
                     label: this._('Deleted')
                 }];
+            } else {
+                this.badges = [];
```
noirbizarre: Same, not related

js/components/organization/restore-modal.vue
```diff
@@ -0,0 +1,53 @@
...
+            restore_organization.deleted = null;
+            API.organizations.update_organization(
+                {org: this.organization.id, payload: restore_organization},
+                (response) => {
+                    this.organization.fetch();
```
noirbizarre: To avoid an extra request, you can directly use the response payload to update the organization: `this.organization.on_fetched(reponse)`

js/components/dataset/restore-modal.vue
```diff
@@ -0,0 +1,52 @@
...
+            restore_dataset.deleted = null;
+            API.datasets.update_dataset({dataset: this.dataset.id, payload: restore_dataset},
+                (response) => {
+                    this.dataset.fetch();
```
noirbizarre: To avoid an extra request, you can directly use the response payload to update the dataset: `this.dataset.on_fetched(reponse)`

js/components/reuse/restore-modal.vue
```diff
@@ -0,0 +1,50 @@
...
+            restore_reuse.deleted = null;
+            API.reuses.update_reuse({reuse: this.reuse.id, payload: restore_reuse},
+                (response) => {
+                    this.reuse.fetch();
```
noirbizarre: To avoid an extra request, you can directly use the response payload to update the reuse: `this.reuse.on_fetched(reponse)`

js/views/organization.vue
micael-grilo: image
The purpose of this line is to clear the "Deleted" red badge.

noirbizarre: OK I understand :+1: 
But on some datasets/reuses/organization, there is more than one badge (more than only the deleted badge), you may need to do some filtering to only remove the deleted badge

PR #1909, opened 2018-10-04
Update translations

noirbizarre:
Updated translations from crowdin (manualy submit until crowding fixes its github integration)


bboulesteix:
PR liée à https://github.com/etalab/data.gouv.fr/issues/82.

PR #1918, opened 2018-10-15
Fix some external assets handling cases

noirbizarre:
This PR fixes some external assets cases:
- sometimes `external` keyword was ignored
- sometimes it didn't exists

This mostly fixes oembed cards.

### DIFF
--- a/udata/assets.py
+++ b/udata/assets.py
@@ -6,7 +6,7 @@ import json
 import os
 import pkg_resources
 
-from flask import current_app, url_for
+from flask import current_app, request, url_for
 from flask_cdn import url_for as cdn_url_for
 
 # Store manifests URLs with the following hierarchy
@@ -73,10 +73,15 @@ def from_manifest(app, filename, raw=False, **kwargs):
     path = _manifests[app][filename]
 
     if not raw and cfg.get('CDN_DOMAIN') and not cfg.get('CDN_DEBUG'):
-        prefix = 'https://' if cfg.get('CDN_HTTPS') else '//'
-        if not path.startswith('/'):
+        scheme = 'https' if cfg.get('CDN_HTTPS') else request.scheme
+        prefix = '{}://'.format(scheme)
+        if not path.startswith('/'):  # CDN_DOMAIN has no trailing slash
             path = '/' + path
         return ''.join((prefix, cfg['CDN_DOMAIN'], path))
+    elif not raw and kwargs.get('external', False):
+        if path.startswith('/'):  # request.host_url has a trailing slash
+            path = path[1:]
+        return ''.join((request.host_url, path))
     return path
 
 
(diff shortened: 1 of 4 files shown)


udata/assets.py
```diff
@@ -73,10 +73,15 @@ def from_manifest(app, filename, raw=False, **kwargs):
...
     if not raw and cfg.get('CDN_DOMAIN') and not cfg.get('CDN_DEBUG'):
-        prefix = 'https://' if cfg.get('CDN_HTTPS') else '//'
-        if not path.startswith('/'):
+        scheme = 'https' if cfg.get('CDN_HTTPS') else request.scheme
+        prefix = '{}://'.format(scheme)
+        if not path.startswith('/'):  # CDN_DOMAIN has no trailing slash
             path = '/' + path
         return ''.join((prefix, cfg['CDN_DOMAIN'], path))
+    elif not raw and kwargs.get('external', False):
+        if path.startswith('/'):  # request.host_url has a trailing slash
```
abulte: Same here.

udata/assets.py
```diff
@@ -73,10 +73,15 @@ def from_manifest(app, filename, raw=False, **kwargs):
     path = _manifests[app][filename]
 
     if not raw and cfg.get('CDN_DOMAIN') and not cfg.get('CDN_DEBUG'):
-        prefix = 'https://' if cfg.get('CDN_HTTPS') else '//'
```
abulte: Has the logic with `//` been removed from this PR? I think I remember we needed that in some case (like on embeds).

udata/assets.py
```diff
@@ -73,10 +73,15 @@ def from_manifest(app, filename, raw=False, **kwargs):
...
-        if not path.startswith('/'):
+        scheme = 'https' if cfg.get('CDN_HTTPS') else request.scheme
+        prefix = '{}://'.format(scheme)
+        if not path.startswith('/'):  # CDN_DOMAIN has no trailing slash
```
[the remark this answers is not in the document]
noirbizarre: no, you read it to fast:
the `startswith` is on `path` which is appended to `CDN_DOMAIN`

udata/assets.py
noirbizarre: Same here:
the `startswith` is on `path` which is appended to `request.host_url`

udata/assets.py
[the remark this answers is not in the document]
abulte: Right ;-) 🙈

udata/assets.py
noirbizarre: Yes, because it wasn't working with some protocols like `file://` (local testing, standalone HTML page).
Tested with embeds (this is while testing it I noticed this issue)

abulte: Don't you get mixed content warning/error in some cases?

noirbizarre: It shouldn't because `request.scheme` is in sync with the incoming request scheme.
I tested locally with:
- http without CDN
- https without CDN
- http with http CDN
- http with https CDN
- https with HTTPS CDN
Tested in siteassets and oembed

PR #1923, opened 2018-10-16
Harvest items can now match source.id before source.domain

abulte:
This should allow us to modify a Harvester url without recreating all the datasets as duplicates.

### DIFF
--- a/udata/harvest/backends/base.py
+++ b/udata/harvest/backends/base.py
@@ -246,10 +246,18 @@ class BaseBackend(object):
         after_harvest_job.send(self)
 
     def get_dataset(self, remote_id):
-        '''Get or create a dataset given its remote ID (and its source)'''
+        '''Get or create a dataset given its remote ID (and its source)
+        We first try to match `source_id` to be source domain independent
+        '''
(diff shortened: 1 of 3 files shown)


udata/harvest/backends/base.py
```diff
@@ -246,10 +246,18 @@ def end(self):
         after_harvest_job.send(self)
 
     def get_dataset(self, remote_id):
-        '''Get or create a dataset given its remote ID (and its source)'''
+        '''Get or create a dataset given its remote ID (and its source)
+        We first try to match `source_id` to be source domain independent
+        '''
+        dataset = Dataset.objects(__raw__={
+            'extras.harvest:source_id': self.source.id,
+            'extras.harvest:remote_id': remote_id,
```
noirbizarre: It should work, but you can do it in a in single pass:
```python
dataset = Dataset.objects(__raw__={
    'extras.harvest:remote_id': remote_id,
    '$or': [
        'extras.harvest:domain': self.source.domain,
        'extras.harvest:source_id': self.source.id,
    ]
}).first()

PR #1928, opened 2018-10-19
Fix enconding issue on Harvest Config validation

micael-grilo:
Fix encoding issues on harvest settings validation.

### DIFF
--- a/udata/harvest/forms.py
+++ b/udata/harvest/forms.py
@@ -40,7 +40,8 @@ class HarvestConfigField(fields.DictField):
                     msg = msg.format(f['key'], backend.name)
                     raise validators.ValidationError(msg)
 
-                f['value'] = f['value'].encode('utf-8') #Fix encoding error
+            if isinstance(f['value'], basestring):  # Not tested, this test may be improved
+                f['value'] = safe_unicode(f['value'])  # Fix encoding error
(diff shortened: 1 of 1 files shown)


udata/harvest/forms.py
```diff
@@ -39,6 +39,9 @@ def pre_validate(self, form):
                     msg = 'Unknown filter key "{0}" for "{1}" backend'
                     msg = msg.format(f['key'], backend.name)
                     raise validators.ValidationError(msg)
+
+                f['value'] = f['value'].encode('utf-8') #Fix encoding error
```
noirbizarre: You should check that the expected type is a string and make use of [`udata.utils.safe_unicode`](https://github.com/opendatateam/udata/blob/master/udata/utils.py#L277-L283):
```suggestion
            if isinstance(f['value'], basestring):  # Not tested, this test needs to be improved
                f['value'] = safe_unicode(f['value'])  # Fix encoding error
```

micael-grilo: "basestring", is the base for str and unicode, so should be fine!
I've opened this PR because we faced encoding issues when we used filters like this ( with special characters like í, ç, ã, etc.. ):
image

Tested with your solution, it's working as expected!

PR #1929, opened 2018-10-19
Tiny fix: display the list items properly

bzg:
Without the blank line, the list is not recognized as such.


bzg:
Super, merci !

PR #1979, opened 2019-01-05
export multiple models objects to CSV in a dataset

### LINKED ISSUE #1977: Les listes de jeux de données et de ressources ne sont pas fiables
Lorsque l'on tente de télécharger la liste des jeux de données `/datasets.csv`, le fichier a un nombre de ligne relativement aléatoire. Le serveur n'émet pas d'erreur visible pour signaler à l'usager que le fichier est incomplet.

Solution possible : ne pas faire du streaming mais en faire un jeu de données en bonne et dû forme.

Cela est également vrai pour la liste des ressources.

https://sentry.data.gouv.fr/etalab/www-datagouvfr/issues/648995/

abulte:
Fix #1977.

This introduces a job ~~with a `model` argument~~ that generates a CSV export of the model's objects from the DB. The file is then stored as a resource of a single dataset (created if not there). We could run jobs for each model we want to export every night. This reuses the `CSVAdapter` logic already available.

Things to be considered/discussed:
- ~~Should we keep the history of exports in the dataset? ATM this is what is done and it's probably valuable to have some history. But it can cause issues: a lot of resources on the page, privacy problems (especially for discussions)...~~ no history kept (on the frontend), resources are overridden
- ~~Should we have a single job that exports every (configurable) model or keep a job per model?~~ single job
- Should we optimize this a bit? ATM a temp CSV file is written in one chunk, maybe it would be more (memory) efficient to write it in chunks. I can export 100Mo in 4 minutes on my machine (resources) so it's not that slow for a job.

Things to be done:
- [ ] ~~Probably move the code in another file, this is not strictly a `datasets` matter~~ did not find a better place
- [x] Move some stuff into settings in order to be configurable (dataset slug, dataset defaults, allowed models)
- [x] Introduce a setting to choose which organization publishes the dataset
- [x] Tests


abulte:
Replaced by #2124

PR #1996, opened 2019-01-25
Fix harvest preview in edit form not taking configuration (features and filters)

noirbizarre:
This PR fix the lack of change events on harvest config form and so fix the missing configuration in the harvest edit form preview

### DIFF
--- a/js/views/harvester-edit.vue
+++ b/js/views/harvester-edit.vue
@@ -48,6 +48,13 @@ export default {
             }
         }
     },
+    ready() {
+        this.source.$on('updated', () => {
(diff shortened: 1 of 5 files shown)


js/views/harvester-edit.vue
```diff
@@ -48,6 +48,14 @@ export default {
             }
         }
     },
+    ready() {
+        this.source.$on('updated', () => {
+            this.$nextTick(() => {
+                console.log('updated', this.$refs.form.serialize())
```
abulte: remove?

PR #2000, opened 2019-01-28
Form fixes

noirbizarre:
This PR fixes some major bugs in forms:

- forms errors were mispositionned and not injected into vue.js forms
- server side error were silently ignored
- success notification was always dispatched even on on transport or server-side error
- some forms were using 2 differents models (Dataset and Resource) for the same thing
- some forms were never able to dispatch a success notification
- `hide-notification` was never `true`
- there was no deletion success notification

The fixes are:
- rely on the `updated` event to consider a form successfuly submitted and only then route to the next view or close modal and show the success notification
- ensure `jQuery.validate` properly inject errors into vue.js forms
- ensure server-side errors are properly injected into vue.js forms
- ensure only one object/model is managed by each form
- use the same notification logic for deletion
- dispatch a notification for non-field errors
- fix the error labels style

This PR is a quick and temporary fix but this need a major refactoring (Vuex, promise, async... whatever)


noirbizarre:
Yep, I tried to be stricter on separation of concerns between forms, views and models.
But, OMG, I'd like to have Vuex and async !!!!

PR #2012, opened 2019-01-31
Ensures all modals have the same buttons styles and orders

noirbizarre:
This PR ensures all modals have the same buttons styles and orders, same color code...

## Delete modals
screenshot-data xps-2019 01 31-15-15-28
screenshot-data xps-2019 01 31-15-14-31

## Edit modals
Cancel/Close button on the left, smaller, same color as the modal
Action buttons on the right, outlined, red and smaller if dangerous

screenshot-data xps-2019 01 31-15-39-36
screenshot-data xps-2019 01 31-15-15-57

### Before
screenshot-www data gouv fr-2019 01 31-16-41-34

### After
screenshot-data xps-2019 01 31-16-42-14

### Before
screenshot-data xps-2019 01 31-16-04-59

### After
screenshot-data xps-2019 01 31-16-10-22

### Before
screenshot-data xps-2019 01 31-15-48-13

### After
screenshot-data xps-2019 01 31-16-03-13

## Yes/No modals
- `No/Refuse` in red, on the left side
- `Yes/Accept` in green, on the right side

screenshot-data xps-2019 01 31-16-39-21

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Expose and import licenses `alternate_urls` and `alternate_titles` fields [#2006](https://github.com/opendatateam/udata/pull/2006)
 - Be consistent on search results wording and icons (Stars vs Followers) [#2013](https://github.com/opendatateam/udata/pull/2013)
 - Switch from a "full facet reset" to a "by term reset" approach in search facets [#2014](https://github.com/opendatateam/udata/pull/2014)
+- Ensures all modals have the same buttons styles and orders, same color code... [#2012](https://github.com/opendatateam/udata/pull/2012)
 
 ## 1.6.2 (2018-11-05)
 
--- a/js/components/badges/modal.vue
+++ b/js/components/badges/modal.vue
@@ -26,7 +26,7 @@
     </div>
 
     <footer class="modal-footer text-center">
-        <button v-if="confirm" type="button" class="btn btn-sm btn-info btn-flat pull-left"
+        <button v-if="confirm" type="button" class="btn btn-info btn-flat pull-left"
             @click="$refs.modal.close">
             {{ _('Cancel') }}
         </button>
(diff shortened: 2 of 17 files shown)


abulte:
Button placement consistency is awesome 👏 .

I wonder about some button colors though 🤔. The yes/no modal with white/red is very legible I think. It's not the case for the badges one. Maybe we can consider it a yes/no since there's only two choices?

noirbizarre:
I was wondering and I chosed to be consitent accross all modals.

Some of the possibilities I thinked about are:
- change the color of actions buttons from `btn-outlined` to something else (green for every actions, red for dangerous ?)
- drop the size changes between cancel/close buttons, danger buttons and actions buttons

abulte:
@tk we need your thoughts on this ;-)

taniki:
- drop the different button sizes. it bugged me a lot.
- I would switch the style of normal and off. the flat style looks like a button contrary to the simple outline but the whiteness make more contrast so maybe we should get a pale blue as outline.
- overall green/red is more legible but there is too much color already. In fact the white modal (discussions) is more clean/clear than blue ones. It also makes other colors more efficient.
- I won't use a button for "cancel" or "close" actions. More likely a simple link.
- I won't use "delete" button in "comment" or "edit" workflows. It should be in a different path in order to avoid mistakes, decision overload, path depths.

noirbizarre:
OK I hear the comments and I agree but this PR was only meant to align existing styles.

Can I implement the overall style change in another PR ? It needs to go over all modals to change them, and the last point in particular needs some major rewrite of some admin modals/edit/display views

taniki:
Yep mainly drop the different button sizes to have a complete alignement (: the overall consistency improvement provided by this PR is already a great thing (:

noirbizarre:
Thanks :+1: 

Done for the size change.

Issue #2021, opened 2019-02-03
A few typos in the documentation and UI strings


bzg:
There are a few typos in the documentation and UI strings.

PR #2022, opened 2019-02-03
Fixing a few typos in the documentation and UI/translation strings

bzg:
See  https://github.com/opendatateam/udata/issues/2021

This is mainly to get my feet wet and better understand the contribution process.

My understanding is that such a minor contribution still falls under the "bugfix" category and the PR against `master` is okay. Let me know if I should have made a PR against another branch.


abulte:
Welcome 💪 

About the branch : you're right, this is a bug fix and thus should made against `master`. However, we're in a very special state right now: the next version is supposed to be `udata 2.0` which lives in the `py3` branch, so maybe it would be be simpler to do it on that branch. But let's keep it on `master` for now: you'll experience the real workflow and we can backport the fix to `py3` w/o problem.

Some comments:
- you should create a branch on your fork before making a pull request, it's clearer that way -- we even have some suggestion about branch naming somewhere ;
- your PR branch should be rebased against `udata:master` before submitting the PR (the history is not clean on this PR because of that) ;
- you should add an entry to the CHANGELOG for each PR (below Current (in progress)) ;
- you should add `opendatateam/etalab` as a reviewer so that we can be notified.

Thank your for taking the time to make a contribution 🙏.

docs/adapting-settings.md
```diff
@@ -145,7 +145,7 @@ List of allowed URL schemes.
 **default**: All IANA registered TLDs
 
 List of allowed TLDs.
-When using udata on an intranet, you might want to add your own custom TLDs:
+When using uData on an intranet, you might want to add your own custom TLDs:
```
abulte: `udata` is now the canonical way :-)

bzg:
Thanks for the directions! I've submitted a new [PR](https://github.com/opendatateam/udata/pull/2023).

I don't see how to request a review, though. Shall I be part of `opendatateam` to request a review?

bzg:
Thanks a lot!

bzg:
Closing per https://github.com/opendatateam/udata/pull/2023#pullrequestreview-199507545

PR #2070, opened 2019-03-19
Fix reindex memory leaks

noirbizarre:
This PR fixes 2 memory leak on search indexing.

The memory profile have been sampled with the following configuration:

- **CPU:** Inter Core I7 7500U
- **RAM:** 16G
- **OS:** Archlinux
- **Datasets:** 10000

## Tooling

The profiling script is: https://gist.github.com/noirbizarre/7ea0b0526814131752e6902a4254df78

It requires:
 - [`memory-profiler`](https://pypi.org/project/memory-profiler/)
 - [`objgraph`](https://pypi.org/project/objgraph/)
 - [`matplotlib`](https://pypi.org/project/matplotlib/)
 - `graphviz`

Run it with:
```shell
mprof run -o output.dat profiling.py
```
Render graph with:
```shell
mprof plot output.dat -t "Title" -o output.png
```

### Memory profile from current master

profiling-master

**Max memory usage**: ~4G

## udata `SlugField`

The udata `SlugField` instances were keeping a reference to their parent document using `self.instance`. This reference was stored in a class instance which is instanciated once on document definition and so the garbage collector was never able to collect document instances using a slug field (as there was always a reference on it)

This has been fixed by refactoring the `SlugField` population into a signal handler and not keeping a static feference on the instance.

### Memory profile with the `SlugField` fix

profiling-fix

**Max memory usage**: ~3.6G

## MongoEngine querysets and `no_cache()`

MongoEngine default behavior is to cache any queryset just to avoid round trip on "count then query".
This behavior cause a massive memory leak while iterating over a queryset generator.

The fix is to add a `no_cache()` call before the loop.

This fix might be beneficial on other code parts.

### Memory profile with `no_cache()`

profiling-fix-no-cache

**Max memory usage**: ~300M

## MongoEngine 0.17.0

It doesn't fix the memory leak but the update was applied just to verify that it has no impact (See [this  PR](https://github.com/opendatateam/udata/pull/2063)).
As you can see, there is some meory drop and the hole reindexing take much more longer.
The update has been removed from the PR until the cause of these drops is found.

### Memory profile

profiling-fix-no-cache-0 17 0

**Max memory usage**: ~300M

### DIFF
--- a/udata/models/slug_fields.py
+++ b/udata/models/slug_fields.py
@@ -6,7 +6,7 @@ import slugify
 
 from flask_mongoengine import Document
 from mongoengine.fields import StringField
-from mongoengine.signals import post_delete
+from mongoengine.signals import pre_save, post_delete
 
 from .queryset import UDataQuerySet
 
@@ -18,6 +18,7 @@ class SlugField(StringField):
     A field that that produces a slug from the inputs and auto-
     increments the slug if the value already exists.
     '''
+    # Do not remove, this is required to trigger field population
     _auto_gen = True
 
     def __init__(self, populate_from=None, update=False, lower_case=True,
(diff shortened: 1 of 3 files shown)


udata/models/slug_fields.py
```diff
@@ -30,31 +31,26 @@ def __init__(self, populate_from=None, update=False, lower_case=True,
...
         super(SlugField, self).__init__(**kwargs)
-        if follow:
-            # Can't use sender=self.owner_document which is not yet defined
-            post_delete.connect(self.cleanup_on_delete)
 
     def __get__(self, instance, owner):
-        if instance is not None:
-            self.instance = instance
+        # mongoengine calls this after document initialization
+        # We register signals ahndler here to have a owner reference
```
abulte: s/ahndler/handlers/

Issue #2066, opened 2019-03-13
Discussions API with invalid id or class: improve error and return 404?


abulte:
When using the discussions API to create a new discussion, w/ a payload such as:

```
{
	"comment": "message",
	"subject": {
		"class": "Dataset",
		"id": "5937c5dec751df5a49c153a9"
	},
	"title": "title"
}
```

if the `id` is wrong (for example the id of a non existing dataset) or the `class` is wrong (non existing model), the API will reply with a `400` and this message:

```
{
  "errors": {
    "subject": [
      "Ce champ est requis."
    ]
  }
}
```

This is not very helpful for the caller, since there is indeed a `subject` field in the request.

Maybe we should reply instead with a `404` and a message such as:

```
{
  "errors": {
    "subject": [
      "Could not find an instance of the model <model> with the id <id>."
    ]
  }
}
```

NB: the `404` may not be strictly correct in this case, maybe sticking with `400` is OK and just improve the error message.

abulte:
After investigation, it seems that the `ValidationError` messages raised here https://github.com/opendatateam/udata/blob/master/udata/forms/fields.py#L384 are not bubbled up correctly into the API error.

abulte:
If we remove the `required` validator from https://github.com/opendatateam/udata/blob/master/udata/core/discussions/forms.py#L17, the correct error message will bubble up. `required` seems to take precedence even though it's not involved.

noirbizarre:
I've taken the `400` error path which seems more legit and standard (see #2075)
Was tricky because of `WTForm` which is hiding all error message when a `required` error is triggered.

abulte:
poke @l-vincent-l

l-vincent-l:
:clap: never thought it would be that hard!

noirbizarre:
Me neither !

PR #2075, opened 2019-03-21
Improve ModelField errors

noirbizarre:
This PR test and fixes some error cases on `ModelField` form field with `db.ReferenceField` and `db.GenericReferenceField`:
- Prevent 'required` error message to hide other ones
- Adds a lot of specific case more explicit error message
- Prevent some internal server errors

Fix https://github.com/opendatateam/udata/issues/2066

### DIFF
--- a/udata/forms/fields.py
+++ b/udata/forms/fields.py
@@ -382,63 +382,61 @@ class ModelFieldMixin(object):
 
 class ModelField(Field):
     def process(self, formdata, data=unset_value):
-        if not formdata:
-            return
-        # Process prefixed values as in FormField
-        newdata = {}
-        prefix = self.short_name + '-'
-        for key in formdata.keys():
-            if key.startswith(prefix):
-                value = formdata.pop(key)
-                newdata[key.replace(prefix, '')] = value
-        if newdata:
-            formdata.add(self.short_name, newdata)
+        if formdata:
+            # Process prefixed values as in FormField
+            newdata = {}
+            prefix = self.short_name + '-'
+            for key in formdata.keys():
+                if key.startswith(prefix):
+                    value = formdata.pop(key)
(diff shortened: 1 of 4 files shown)


udata/forms/fields.py
```diff
@@ -381,30 +381,64 @@ def process_formdata(self, valuelist):
...
-            except db.ValidationError:
-                message = _('{0} is not a valid identifier').format(specs['id'])
-                raise validators.ValidationError(message)
+                label = '{0}({1})'.format(model.__name__, oid)
+                msg = _('{0} does not exists').format(label)
+                raise validators.ValidationError(msg)
+
+    def pre_validate(self, form):
+        # If any error happen during process, we raise StopValidation here
+        # to prevent "DataRequired" validator from clearing errors
```
abulte: 🥇

udata/models/__init__.py
```diff
@@ -75,7 +75,7 @@ def resolve_model(self, model):
         try:
             return get_document(classname)
         except self.NotRegistered:
-            message = '{0} does not exists'.format(classname)
+            message = 'Model "{0}" does not exists'.format(classname)
```
abulte: s/exists/exist/

udata/forms/fields.py
noirbizarre: Yep, that was the tricky part to find !!! :sweat_smile:

noirbizarre:
It took a day to write `test_model_field.py` (test all error case for all combinations of "required/optionnal form field" + "db.GenericReferenceField/db.ReferenceField" + "all possible errors (at least known to me)".

But I just discovered some existing tests in `test_reference_field.py`. I need to merge them into `test_model_field.py` to make tests pass (the PR is missing the `choices` case)

noirbizarre:
Done, tests passed

PR #2087, opened 2019-03-26
[admin] delete a single comment in a discussion

### LINKED ISSUE #1965: Allow sub-discussion (comment) deletion
Today it's not possible to delete a comment inside a discussion, only a complete discussion. This is a problem for spam moderation.

We should allow the deletion of a single comment, at least for super-admins and probably also for organisation administrators.

abulte:
Not pretty (there's a `confirm` in there 🙈), but hey it's an admin-only feature and it works.

~~TODO: see what happens when the last comment in the discussion is removed, we should probably disable the feature.~~ The last comment on a thread cannot be deleted: API will raise `400` and feature is disabled in UI.

Fix #1965.

Capture d'écran 2019-03-26 18 56 20
Capture d'écran 2019-03-26 18 56 27

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
   - Use python based migrations instead of relying on mongo internal and deprecated `js_exec`
   - Handle rollback (optionnal)
   - Detailled history
+- Admins can delete a single comment in a discussion thread [#2087](https://github.com/opendatateam/udata/pull/2087)
 
 ### Breaking changes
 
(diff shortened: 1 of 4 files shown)


noirbizarre:
OMG `alert` usage !!! :trollface:

abulte:
Modal in a modal

noirbizarre:
:joy: 
You can play with `v-if` and `v-show` to switch from the modal in a confirmation modal. See: https://github.com/opendatateam/udata/blob/master/js/components/dataset/resource/modal.vue (This one is a little bit complex, but the logic is working)

abulte:
Yep, overengineered for now I think.

abulte:
Replaced by #2125

PR #2096, opened 2019-04-04
DCAT harvester refactoring

noirbizarre:
This PR refactor the DCAT harvesting to store only one graph for all datasets in a DCAT catalog.
The graph is stored in factorized way (ie. in JSON-LD with a `@context` to massively reduce the size.

These changes allows to process way more datasets in a single Job. As each harvesting job store all its items into a single MongoDB document, each job can store at most `16Mb` (See [MongoDB Limits](https://docs.mongodb.com/manual/reference/limits/))

This also prevent some data loss because properly slicing a graph is difficult in RDF when you don't known by advance all node properties. This will allow to parse more triplets for each dataset and resource.

These changes also allow processing of DCAT Dataset as Blank nodes instead of URIRef.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Refactor DCAT harvesting to store only one graph (and prevent MongoDB document size overflow) [#2096](https://github.com/opendatateam/udata/pull/2096)
 
 ## 1.6.6 (2019-03-27)
 
--- a/udata/harvest/tests/test_dcat_backend.py
+++ b/udata/harvest/tests/test_dcat_backend.py
@@ -94,6 +94,23 @@ class DcatBackendTest:
         assert dataset.tags == ['tag-1', 'tag-2']
         assert len(dataset.resources) == 1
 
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Refactor DCAT harvesting to store only one graph (and prevent MongoDB document size oveflow) [#2096](https://github.com/opendatateam/udata/pull/2096)
```
abulte: s/oveflow/overflow/

udata/harvest/tests/test_dcat_backend.py
```diff
@@ -94,6 +94,24 @@ def test_simple_flat(self, rmock):
...
 
+    def test_flat_with_blank_nodes(self, rmock):
+        filename = 'bnodes.jsonld'
+        url = mock_dcat(rmock, filename)
+        org = OrganizationFactory()
+        source = HarvestSourceFactory(backend='dcat',
+                                      url=url,
+                                      organization=org)
+
+        # Run the same havester twice
```
abulte: s/havester/harvester/

PR #2103, opened 2019-04-11
Fix munin zeros

noirbizarre:
This PR ensures that the `udata worker status --munin` always output tasks count even if count is zero.
Otherwise Munin sees it as "no data" (ie. vertical splitter in graph) and display count as `-nan` instead of `0`. It was also making Munin diagnosis difficult when a queue is empty.

This also ensure only tasks for monitored queue are listed in `udata worker status --munin-config` reducing the number of tasks by graph (extras tasks were always displayed as `-nan`. As a side-effect, caching tasks is not required anymore as it does not hit all the Celery workers anymore to list them.

A `udata worker tasks` command has been added to display all tasks and their queue/priority.

Some test jobs have been renamed to ensure they all have the same `test-` prefix and they are filtered out of munin commands.

The documentation link to the munin plugin has been updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Refactor DCAT harvesting to store only one graph (and prevent MongoDB document size overflow) [#2096](https://github.com/opendatateam/udata/pull/2096)
 - Expose sane defaults for `TRACKING_BLACKLIST` [#2098](https://github.com/opendatateam/udata/pull/2098)
 - Bubble up uploader errors [#2102](https://github.com/opendatateam/udata/pull/2102)
+- Ensure `udata worker status --munin` always outputs zero values so munin won't see it has a "no data" response [#2103](https://github.com/opendatateam/udata/pull/2103)
 
 ## 1.6.6 (2019-03-27)
 
--- a/docs/administrative-tasks.md
+++ b/docs/administrative-tasks.md
@@ -204,7 +204,7 @@ $ udata worker status --munin -q default
 $ udata worker status --munin-config -q default
 ```
 
-[munin-plugin]: https://github.com/etalab/munin-plugins/tree/master/udata-worker-status
+[munin-plugin]: https://github.com/etalab/munin-plugins/tree/master/udata
 
 
 ## Cache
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Refactor DCAT harvesting to store only one graph (and prevent MongoDB document size overflow) [#2096](https://github.com/opendatateam/udata/pull/2096)
 - Expose sane defaults for `TRACKING_BLACKLIST` [#2098](https://github.com/opendatateam/udata/pull/2098)
 - Bubble up uploader errors [#2102](https://github.com/opendatateam/udata/pull/2102)
+- Ensure `udata worker status --munin` always output zero values so munin won't see it has a "no data" response [#2103](https://github.com/opendatateam/udata/pull/2103)
```
abulte: s/output/outputs/

noirbizarre:
As stated [here](https://github.com/opendatateam/udata/pull/2103/files#diff-0e12c8806d6d77212205945f7510749fL29), we were using cache because we were using [Celery `inspect()`](https://github.com/opendatateam/udata/pull/2103/files#diff-0e12c8806d6d77212205945f7510749fL53) which is querying the state of every worker ([the `w` here](https://github.com/opendatateam/udata/pull/2103/files#diff-0e12c8806d6d77212205945f7510749fL55)).
It was working with the following limitations:
- the more we have workers, the more `inspect()` takes time (ie. `query duration * number of workers`)
- it only works when workers are started and for every queues
- it only works if the works if the workers are unfiltered (ie. when using one of `--queues/-Q`, `--exclude-queues/-X`, `--include/-I`)
- it only works if workers are reachables

If has been replaced by `celery.tasks` which doesn't query anything, these are only the tasks registered to celery in the code and so it don't depends on the workers at all (no network roundtrip, not state based)

But it's how I understand [this comment](https://github.com/opendatateam/udata/pull/2103/files#diff-0e12c8806d6d77212205945f7510749fL29). If cache is still required, I can put it back.

abulte:
@noirbizarre Ok thanks! Looks good.

PR #2104, opened 2019-04-11
Celery 4.3 - Redis 3.2

noirbizarre:
This PR upgrades to:
- Celery 4.3.0 (lots of fixes, Python 3.7 support, Redis-py 3+ support)
- Redis-py 3.2.1 (Python 3.7 support, improved connexion pool management...)
- Flask-caching (Redis-py 3+ support, cache only on success support, werkzeug 0.15 support)

Tested:
- [x] Celery basic task/job
- [x] Celery scheduled task/job
- [x] Celery worker & beat
- [x] sessions open/close
- [x] application cache
- [x] template cache

Includes #2069, #2090, #2092


abulte:
Shouldn't we include this in the 2.0 release instead?

abulte:
(because this seems like major updates, and maybe we don't want to iterate on 1.6.xxx to fix potential problems)

noirbizarre:
I wanted to wait for udata 2.0, but the latest Kombu version force Redis-py 3+ and so a fresh `udata` installation is broken. It only works when you already have a virtualenv with previous versions. :/
We can pin a Kombu version but I wasn't able to find one working (need to test Kombu version one by one).

I wiped my virtualenv so I don't have a working version :/ (Initialy this is why I submitted this PR, just to be able to work on `udata`)

Reading the changelogs, it seems to fix the workers being stuck on unresponsive or lost redis connexion :thinking:

noirbizarre:
I just tried again and downgrading worked (wasn't on a fresh install).
I will try to wipe again my virtualenv and if it's working, we can safely postpone this PR for udata 2.

PR #2125, opened 2019-05-03
[admin] delete a single comment in a discussion

### LINKED ISSUE #1965: Allow sub-discussion (comment) deletion

abulte:
Superseed #2087 

Not pretty (there's a `confirm` in there 🙈), but hey it's an admin-only feature and it works.

~~TODO: see what happens when the last comment in the discussion is removed, we should probably disable the feature.~~ The first comment on a thread cannot be deleted: API will raise `400` and feature is disabled in UI.

Fix #1965.

Capture d'écran 2019-03-26 18 56 20
Capture d'écran 2019-03-26 18 56 27

### DIFF
--- a/js/components/discussions/modal.vue
+++ b/js/components/discussions/modal.vue
@@ -51,7 +51,7 @@
                 v-for="(idx, message) in discussion.discussion">
                 <div class="direct-chat-info clearfix">
                     <span class="direct-chat-name pull-left">{{message.posted_by | display}}</span>
-                    <a v-if="$root.me.is_admin && canDeleteComment"  @click.prevent="confirm_delete_comment(idx)" href="" class="direct-chat-name direct-chat-delete">{{ _('Delete comment') }}</a>
+                    <a v-if="$root.me.is_admin && idx !== 0"  @click.prevent="confirm_delete_comment(idx)" href class="direct-chat-name direct-chat-delete">{{ _('Delete comment') }}</a>
                     <span class="direct-chat-timestamp pull-right">{{message.posted_on | dt}}</span>
                 </div>
                 <img class="direct-chat-img"  :alt="_('User Image')"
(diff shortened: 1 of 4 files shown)


js/components/discussions/modal.vue
```diff
@@ -42,9 +48,10 @@
         <h3>{{ discussion.title }}</h3>
         <div class="direct-chat-messages">
             <div class="direct-chat-msg"
-                v-for="message in discussion.discussion">
+                v-for="(idx, message) in discussion.discussion">
                 <div class="direct-chat-info clearfix">
                     <span class="direct-chat-name pull-left">{{message.posted_by | display}}</span>
+                    <a v-if="$root.me.is_admin && canDeleteComment"  @click.prevent="confirm_delete_comment(idx)" href="" class="direct-chat-name direct-chat-delete">{{ _('Delete comment') }}</a>
```
noirbizarre: `s/href=""/href/`

udata/core/discussions/api.py
```diff
@@ -133,6 +133,26 @@ def delete(self, id):
...
+    '''
+    @api.secure(admin_permission)
+    @api.doc('delete_discussion_comment')
+    @api.response(403, 'Not allowed to delete this comment')
+    def delete(self, id, cidx):
+        '''Delete a comment given its index'''
+        discussion = Discussion.objects.get_or_404(id=id)
+        if len(discussion.discussion) <= cidx:
+            api.abort(404, 'Comment does not exist')
+        elif len(discussion.discussion) == 1:
```
noirbizarre: Maybe add a condition to disallow deleting the first message (and so prevent to have an empty discussion heading followed by comments ie. admin should delete discussion instead of trying to delete the first message)

abulte: Done, I replaced the "last comment" condition by "first comment", makes more sense.

PR #2132, opened 2019-05-10
Revert "Harvest sources list API fix"

abulte:
Reverts opendatateam/udata#2131

The migration took 4 hours on next.data.gouv.fr until I lost the connection. It does not seem safe to release on production for now.


abulte:
Replaced by #2132

PR #2137, opened 2019-05-13
Extend debounce delay on autocomplete

abulte:
We see a lot of `suggest` API calls failing on the uWSGI side when we're under heavy load. This may help by sending less requests.


abulte:
Replaced by 2138

PR #2150, opened 2019-05-17
display remote source of harvested dataset

taniki:
work around to at least show a clickable link toward remote source of an harvested dataset. We could go deeper in information architecture later (:

image

cf https://github.com/etalab/data.gouv.fr/issues/127

### DIFF
--- a/udata/templates/dataset/display.html
+++ b/udata/templates/dataset/display.html
@@ -238,13 +238,7 @@
                             {% endif %}
 
                             {# Harvesting source #}
-                            {# ODS #}
-                            {% if dataset.extras['ods:url'] %}
-                                {% set source = dataset.extras['ods:url']  %}
-                            {# CKAN #}
-                            {% elif dataset.extras['remote_url'] %}
-                                {% set source = dataset.extras['remote_url']  %}
-                            {% endif %}
+                            {% set source = dataset.extras['ods:url'] or  dataset.extras['remote_url'] %}
 
                             {% if source %}
                             <li>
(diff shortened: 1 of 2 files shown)


udata/templates/dataset/display.html
```diff
@@ -237,6 +237,14 @@ <h3>{{ _('Informations') }}</h3>
                             </li>
                             {% endif %}
 
+                            {# Source ODS #}
```
abulte: More like "harvesting source"

udata/templates/dataset/display.html
```diff
@@ -237,6 +237,14 @@ <h3>{{ _('Informations') }}</h3>
...
+                            {% if dataset.extras['ods:url'] %}
```
abulte: Use `remote_url` for CKAN

udata/templates/dataset/display.html
```diff
@@ -238,18 +238,19 @@ <h3>{{ _('Informations') }}</h3>
...
                             {# Source #}
+
                             {# ODS #}
                             {% if dataset.extras['ods:url'] %}
-                            <li>
-                                <a href v-tooltip title="{{ _('Remote source') }}"><span class="fa fa-fw fa-external-link "></span></a>
-                                <a href="{{ dataset.extras['ods:url'] }}">{{ _('Remote source') }}</a>
-                            </li>
-
+                                {% set source = dataset.extras['ods:url']  %}
```
abulte: `{% set source = dataset.extras['ods:url'] or dataset.extras['remote_url'] %}` ? don't know if it works, maybe worth a try.

PR #2157, opened 2019-05-22
display user avatars at various places

taniki:
also use `.scalable` as full responsive images (i.e `width: 100%`)

next step might be to have a different placeholder for users and orgs

### DIFF
--- a/less/site.less
+++ b/less/site.less
@@ -221,7 +221,7 @@ hr {
 }
 
 img.scalable {
-    max-width: 100%;
+    width: 100%;
 }
 
 .clickable {
--- a/udata/templates/dataset/card.html
+++ b/udata/templates/dataset/card.html
@@ -8,9 +8,13 @@
                 width="70" height="70">
         </div>
         {{ badge_if_certified(dataset.organization) }}
+    {% elif dataset.owner %}
+    <div class="card-logo">
+        <img src="{{ dataset.owner.avatar(70, external=True) | placeholder('organization', external=True) }}" alt="">
+    </div>
     {% else %}
     <div class="card-logo">
-        <img src="{{ ''|placeholder('organization', external=True) }}" alt="">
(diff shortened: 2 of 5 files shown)


udata/templates/dataset/card.html
```diff
@@ -8,9 +8,13 @@
                 width="70" height="70">
         </div>
         {{ badge_if_certified(dataset.organization) }}
+    {% elif dataset.owner %}
+    <div class="card-logo">
+        <img src="{{ dataset.owner.avatar(70, external=True) | placeholder('organization', external=True) }}" alt="">
```
abulte: s/organization/user/ ?

udata/templates/dataset/search-result.html
```diff
@@ -9,6 +9,10 @@
                 width="70" height="70">
         </div>
         {{ badge_if_certified(dataset.organization) }}
+        {% elif dataset.owner %}
+        <div class="result-logo pull-left">
+            <img src="{{ dataset.owner.avatar(70)|placeholder('organization') }}" alt="">
```
abulte: s/organization/user/ ?

udata/templates/dataset/card.html
abulte: \+ `width="70" height="70"` ?

udata/templates/dataset/search-result.html
abulte: \+ `width="70" height="70"` ?

less/site.less
```diff
@@ -221,7 +221,7 @@ hr {
 }
 
 img.scalable {
-    max-width: 100%;
+    width: 100%;
```
abulte: seems risky 😨 this class is used 9 times in udata core alone

taniki: Move Fast and Break Things

abulte: poke @jdesboeufs :trollface:

PR #2158, opened 2019-05-23
Redirect unfiltered csv exports to dataset of datasets

abulte:
This redirects `/(datasets|resources|reuses|resources).csv` to the appropriate resource URL from the dataset of datasets. If the dataset of datasets feature is disabled, the export URLs behave as they used to (ie streaming export).

### DIFF
--- a/udata/core/site/views.py
+++ b/udata/core/site/views.py
@@ -89,7 +89,7 @@ def get_export_url(model):
 @blueprint.route('/datasets.csv', cors=True)
 def datasets_csv():
     params = multi_to_dict(request.args)
-    # redirect to EXPORT_CSV dataset if feature is enabled and no filter set
+    # redirect to EXPORT_CSV dataset if feature is enabled and no filter is set
     exported_models = current_app.config.get('EXPORT_CSV_MODELS', [])
     if not params and 'dataset' in exported_models:
         return get_export_url('dataset')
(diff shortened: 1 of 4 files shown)


udata/core/site/views.py
```diff
@@ -73,9 +73,30 @@ def map():
...
+            break
+    if not resource:
+        abort(404)
+    return resource.url
+
+
 @blueprint.route('/datasets.csv', cors=True)
 def datasets_csv():
     params = multi_to_dict(request.args)
+    # redirect to EXPORT_CSV dataset if feature is enabled and no filter set
```
taniki: "no filter set" ça ne veut pas dire qu'il y a n'y a pas de filtre. MAYBE "and no filter is set" ou "filter is not set"

udata/core/site/views.py
```diff
@@ -85,6 +106,14 @@ def datasets_csv():
 @blueprint.route('/resources.csv', cors=True)
 def resources_csv():
     params = multi_to_dict(request.args)
+    # redirect to EXPORT_CSV dataset if feature is enabled and no filter set
```
taniki: same

udata/core/site/views.py
```diff
@@ -93,6 +122,14 @@ def resources_csv():
 @blueprint.route('/organizations.csv', cors=True)
 def organizations_csv():
     params = multi_to_dict(request.args)
+    # redirect to EXPORT_CSV dataset if feature is enabled and no filter set
```
taniki: bis

taniki:
changelog

PR #2163, opened 2019-05-27
Allow .dxf upload

abulte:
Fix https://github.com/etalab/data.gouv.fr/issues/137


abulte:
replaced by https://github.com/opendatateam/udata/pull/2164

PR #2171, opened 2019-05-29
Replace featured reused by last reuses

taniki:
We should use search instead of last selected reuses if it is named "last reused". Otherwise we should rename the section. I would favor first option since its require less editorial work but we will need to sanitize search indexation.

CF https://github.com/etalab/udata-gouvfr/pull/408


taniki:
or should we rename the thing "Featured reuses" ?

taniki:
https://github.com/opendatateam/udata/blob/79d8ac45cf5ff745f1fe1f12bead3751ff815c97/udata/templates/topic/datasets.html#L13

this kind of line appears in several files but seems useless

udata/templates/topic/display.html
```diff
@@ -26,10 +26,10 @@
             {# Last reuses #}
             <div class="col-xs-12">
 
-                <h3>{{ _('Latest reuses') }}</h3>
+                <h3><a href="{{ url_for('topics.reuses', topic=topic) }}">{{ _('Latest reuses') }}</a></h3>
```
abulte: Isn't this a little confusing? This won't link to the _Latest reuses_ but to the _Featured reuses_ AFAIK.

Maybe:
- rename title to _Featured reuses_
- link explicitly to _Latest reuses_

abulte: Or remove all references to _Latest reuses_ but:
- this would mean deleting `topics.reuses` in the model
- it can be quite "dangerous" to automatically promote reuses for a topic (I think it has been (re)designed because there were some abuses)

I still like this solution 🙃

taniki: in this context `topic.reuses` is about the route "last reuses of this topic" and not to the `reuses` property/dict of the `topic` model.

taniki: (I won't have bothered rename the section and open a PR if there was an actual page/list about featured reuses)

taniki: Maybe check this : https://github.com/opendatateam/udata/blob/800d0ae88438fe4a103a9930e06862a3a0952cb0/udata/core/topic/views.py#L77

abulte: Ok for the link 🙈

udata/templates/topic/display.html
```diff
@@ -26,10 +26,10 @@
...
 
                 <ul class="card-list card-list--columned">
-                    {% for reuse in topic.reuses %}
+                    {% for reuse in recent_reuses %}
```
abulte: We still should delete `topics.reuse` in the model IMHO. Do it in this PR or open an issue for later 🤷‍♂

PR #2200, opened 2019-06-14
remove social media share button

taniki:
it is buggy and useless

FIX https://github.com/etalab/udata-gouvfr/issues/201


taniki:
@micael-grilo @petzlux have you an opinion on this feature?

PR #2204, opened 2019-06-18
update package.json

taniki:


taniki:
Pinning was introduced here : https://github.com/opendatateam/udata/commit/cccbce0e0251459062ac4e7ae16690a58291abc6

PR #2206, opened 2019-06-20
replace default search sort label

taniki:
Right now the default label for search sort is "---" this is not very explicit and might induce a believe of random or non-existent algorithm. "magic" is coming from kickstarter.com. We can use a more corporate copy text.

image

### DIFF
--- a/udata/templates/macros/search.html
+++ b/udata/templates/macros/search.html
@@ -18,7 +18,7 @@
 <div class="btn-group">
     <button type="button" class="btn btn-sorter btn-xs dropdown-toggle" data-toggle="dropdown">
         <span class="fa fa-fw fa-chevron-down hidden-sm"></span>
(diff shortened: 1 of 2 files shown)


udata/templates/macros/search.html
```diff
@@ -18,7 +18,7 @@
 <div class="btn-group">
     <button type="button" class="btn btn-sorter btn-xs dropdown-toggle" data-toggle="dropdown">
         <span class="fa fa-fw fa-chevron-down hidden-sm"></span>
-        {{ label|truncate(12, True) or '---' }}
+        {{ label|truncate(12, True) or _('science') }}
```
abulte: s/science/Science/ at least — I'm not sure about the wording either: Automatic? Default? Pertinency?

abulte: s/Pertinency/Relevance/ 😬

PR #2209, opened 2019-06-21
Prevent deleted harvesters from running until purged (fix #2185)

### LINKED ISSUE #2185: Cannot delete a harvester
When trying to delete an harvester, the `DELETE` request returns `204` but the harvester is not deleted, w/o error. It's also still scheduled.

noirbizarre:
This PR prevent deleted harvester from running until they are purged.

The purge have been named `purge-harvesters` to follow the convention.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 - Remove diff from js dependencies to fix CVE [#2204](https://github.com/opendatateam/udata/pull/2204)
 - Replace default sort label for better readability [#2206](https://github.com/opendatateam/udata/pull/2206)
 - Add a condition to up-to-dateness of a dataset [#2208](https://github.com/opendatateam/udata/pull/2208)
-- Prevent deleted harvester from runnning until purged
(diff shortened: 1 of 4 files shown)


noirbizarre:
I kept the current behavior (https://github.com/opendatateam/udata/blob/master/udata/harvest/models.py#L162): only nullify the source to keep the history. But I can change that, so WDT ?

noirbizarre:
Updated:
- fixed the missing "deleted" badge in the harvest admin view
- delete jobs when purging sources

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - Remove diff from js dependencies to fix CVE [#2204](https://github.com/opendatateam/udata/pull/2204)
 - Replace default sort label for better readability [#2206](https://github.com/opendatateam/udata/pull/2206)
 - Add a condition to up-to-dateness of a dataset [#2208](https://github.com/opendatateam/udata/pull/2208)
+- Prevent deleted harvester from runnning until purged
```
abulte: s/runnning/running/ + mention HarvestJob deletion

abulte: + mention deleted badge fix

PR #2210, opened 2019-06-26
Protect activity.atom against DoesNotExist

abulte:
Don't we just ❤️ MongoDB?

Protect against stuff like this https://sentry.data.gouv.fr/etalab/next-datagouvfr/issues/1847268/events/3319710/

Related to #2203

### DIFF
--- a/udata/core/site/views.py
+++ b/udata/core/site/views.py
@@ -5,6 +5,7 @@ import logging
 import requests
 
 from flask import request, json, redirect, url_for, current_app, abort
+from mongoengine.errors import DoesNotExist
 from werkzeug.contrib.atom import AtomFeed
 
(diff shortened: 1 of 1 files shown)


abulte:
> 👍 protect against this issue but the root cause should be fixed in there (for `User.DoesNotExists`): https://github.com/opendatateam/udata/blob/master/udata/core/user/models.py#L215-L242

Very theoretical. It should be handled at the dataset level https://github.com/opendatateam/udata/blob/800d0ae88438fe4a103a9930e06862a3a0952cb0/udata/core/dataset/tasks.py#L33 but it does in fact break https://sentry.data.gouv.fr/etalab/next-datagouvfr/issues/1847268/events/3319710/.

PR #2214, opened 2019-06-27
Harvest list api fix v2

noirbizarre:
This PR fixes the HarvestSources list API in case of a deleted harvested dataset.
It includes:
- reduced DB/API payload on listing (limit the data pulled by last_job attribute) and faster response
HarvestJob.items.dataset cleanup on dataset purge
- a cleanup migration for existing HarvestJobs

The migration duration depends on the harvest_jobs collection size and can take some time on big instances.

Replace #2131 and adds:
- a `HARVEST_JOBS_RETENTION_DAYS` parameter
- a `harvest-purge-jobs` job to cleanup harvest jobs older than retention policy
- the application configuration is now available in migrations as `config`
- the migration enforces `HARVEST_JOBS_RETENTION_DAYS` before the fix to reduce the amount of jobs to fix (makes way less iterations)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Ensure `filetype='remote'` is set when using the manual ressource form [#2236](https://github.com/opendatateam/udata/pull/2236)
 - Improve harvest sources listing (limit `last_job` fetched and serialized fields, reduce payload) [#2131](https://github.com/opendatateam/udata/pull/2131)
 - Ensure HarvestItems are cleaned up on dataset deletion [#2131](https://github.com/opendatateam/udata/pull/2131)
+- Added `config.HARVEST_JOBS_RETENTION_DAYS` and `harvest-purge-jobs` job to apply it
 
 ## 1.6.12 (2019-06-26)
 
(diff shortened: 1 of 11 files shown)


taniki:
#2228 should resolve UI problem related to harvester deletion

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
 - Display the datasets, reuses and harvesters deleted state on listing when possible [#2228](https://github.com/opendatateam/udata/pull/2228)
 - Fix queryless (no `q` text parameter) search results scoring (or lack of scoring) [#2231](https://github.com/opendatateam/udata/pull/2231)
 - Miscellaneous fixes on completers [#2215](https://github.com/opendatateam/udata/pull/2215)
+- Improve harvest sources listing (limit `last_job` fetched and serialized fields, reduce payload) [#2214](https://github.com/opendatateam/udata/pull/2214)
+- Ensure HarvestItems are cleaned up on dataset deletion [#2214](https://github.com/opendatateam/udata/pull/2214)
+- Added `config.HARVEST_JOBS_RETENTION_DAYS` and a `harvest-purge-jobs` job to apply it [#2214](https://github.com/opendatateam/udata/pull/2214)
```
abulte: Maybe add a reference to the migration and the fact that it can take a long time ⚠️

noirbizarre: Done :+1:

PR #2140, opened 2019-05-13
Geozones 2019

noirbizarre:
This PR:
- ensures that only the latest version of a spatial zone is indexed
- ensures that the `spatial migrate` commands works with any levels/any country
- fixes the spatial completion indexing
- fixes the search completion tokenizer `min_length` parameter handling
- fixes the `geoid.parse()` method to handle legacy identifiers (initialy `/` instead of `:` as separator)
- fixes the `zone.validity` constraints
- prepare for GeoZones 2019 (`wikidata` attribute)
- prompt for countries in `init` command

To update, simply execute:
```shell
udata spatial load --drop /path/or/url/to/geozones/archive.tar.xz
udata spatial migrate 
```
A minimalistic Geozones archive with only countries is  available on the [Geozones release page](https://github.com/etalab/geozones/releases): https://github.com/etalab/geozones/releases/download/2019.0/geozones-countries-2019-0-msgpack.tar.xz

PR #2223, opened 2019-07-03
Handles UTF-8 filenames in geologos archives

noirbizarre:
This PR fixes geologos archive loading (`udata spatial load_logos`) with utf-8 logo filenames on systems where `sys.getdefaultencoding()` returns `ascii`.


noirbizarre:
Merged in #2140

PR #2232, opened 2019-07-09
Pin werkzeug to 0.15.2

abulte:
Werkzeug >0.15.3 doubles our CI test/build time. 

We should test new releases after 0.15.4 to see if they fix this problem. In the meantime we can use 0.15.2.


abulte:
0.15.x is trouble. The build w/ 0.15.1 and 0.15.2 breaks w/o a traceback (looks like memory leak).

PR #2254, opened 2019-07-25
Improve oembed errors handling (no hidden error, CORS)

### LINKED ISSUE #2255: OEmbed errors are hidden
## Detailed Description

When an error occur while requesting an oembed widget, this error is hidden behind the default error handlers (ie. on 404, the error handler thinks `/api/1/oembed` does not exists and provide an error message telling so, see for example https://www.data.gouv.fr/api/1/oembed?url=https%3A%2F%2Fwww.data.gouv.fr%2Fdatasets%2F5862206588ee38254d3f4e5e%2F) and the response does not have CORS headers.

This means:
- a browser oembed consumer won't see the error because of the lack of CORS headers
- when reading the error, the consumer won't understand what happen

## Possible Implementation
Bypass the default error handlers and manually returns explicit errors with CORS (when it's possible)

noirbizarre:
This PR improve oembed errors handling and fixes #2255.

This one has been detected while working on https://github.com/etalab/datagouv-search-indicator: one of the dataset have been removed (INSEE legacy SIRENE dataset) and this prevent from requesting oembed for this dataset (or more precisely from seing it's a 404 and the dataset does not exists anymore)

**Note**: this fixes specific error case on oembed but in order to have CORS on all errors a fix is required on flask-restplus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - JS models load/save/update consistency (`loading` always `true` on query, always handle error, no more silent errors) [#2247](https://github.com/opendatateam/udata/pull/2247)
 - Ensures that date ranges are always positive (ie. `start` < `end`) [#2253](https://github.com/opendatateam/udata/pull/2253)
 - Enable completion on the "`MIME type`" resource form field (needs reindexing) [#2238](https://github.com/opendatateam/udata/pull/2238)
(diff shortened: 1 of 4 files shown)


bzg:
Does it fix an issue that is currently referenced somewhere? If not, can you just add this issue and reference it in the PR message?

noirbizarre:
Done :+1:

PR #2275, opened 2019-08-07
Configurable search

noirbizarre:
This PR extract as much search configuration as possible in settings allowing fine tuning search without repackaging/releasing udata.

### DIFF
--- a/docs/adapting-settings.md
+++ b/docs/adapting-settings.md
@@ -113,18 +113,6 @@ Define the resources preview mode. Can be one of:
 
 If you want to disable preview, set `PREVIEW_MODE` to `None`
 
-### SEARCH_AUTOCOMPLETE_ENABLED
-
-**default**: `True`
-
-Enables the search autocomplete on frontend if set to `True`, disables otherwise.
-
-### SEARCH_AUTOCOMPLETE_DEBOUNCE
-
-**default**: `200`
-
-The search autocomplete debounce delay on frontend, in milliseconds.
-
 ### ARCHIVE_COMMENT_USER_ID
 
 **default**: `None`
@@ -190,6 +178,94 @@ You can disable the feature by setting this to an empty list.
 **default**: `None`
 
 The id of a dataset that should be created before running the `export-csv` job and will hold the CSV exports.
+
+## Search configuration
+
+### SEARCH_AUTOCOMPLETE_ENABLED
+
+**default**: `True`
+
+Enables the search autocomplete on frontend if set to `True`, disables otherwise.
+
+### SEARCH_AUTOCOMPLETE_DEBOUNCE
+
+**default**: `200`
+
+The search autocomplete debounce delay on frontend, in milliseconds.
+
+### SEARCH_DATASET_MAX_TEMPORAL_WEIGHT
+
+**default**: `5`
+
+After this number of years, scoring is kept constant instead of increasing.
+
+### SEARCH_DATASET_FEATURED_WEIGHT
+
+**default**: `3`
+
+How much weight featured items get
+
+### SEARCH_DATASET_FIELDS
+
+**default**:
+```python
+(
+    'geozones.keys^9',
+    'geozones.name^9',
+    'acronym^7',
+    'title^6',
+    'tags.i18n^3',
+    'description',
+)
+```
+Overrides dataset search fields and their ponderation
+
+
+### SEARCH_REUSE_FIELDS
+
+**default**:
+```python
+(
+    'title^4',
+    'description^2',
+    'datasets.title',
+)
+```
+
+Overrides reuse search fields and their ponderation
+
+### SEARCH_ORGANIZATION_FIELDS
+
+**default**:
+```python
+(
+    'name^6',
+    'acronym^6',
+    'description',
+)
+```
+
+Overrides organization search fields and their ponderation
+
+### SEARCH_GEOZONE_FIELDS
+
+**default**: `None`
+
+Overrides geozone search fields and their ponderation
+
+### SEARCH_USER_FIELDS
+
+**default**:
+
+```python
+(
+    'last_name^6',
+    'first_name^5',
+    'about'
+)
+```
+
+Overrides user search fields and their ponderation
 
 ## Map widget configuration
 
--- a/docs/adapting-settings.md
+++ b/docs/adapting-settings.md
@@ -193,18 +193,6 @@ Enables the search autocomplete on frontend if set to `True`, disables otherwise
 
 The search autocomplete debounce delay on frontend, in milliseconds.
 
-### SEARCH_DATASET_MAX_TEMPORAL_WEIGHT
-
-**default**: `5`
-
-After this number of years, scoring is kept constant instead of increasing.
-
-### SEARCH_DATASET_FEATURED_WEIGHT
-
-**default**: `3`
-
-How much weight featured items get
-
 ### SEARCH_DATASET_FIELDS
 
 **default**:
@@ -220,6 +208,45 @@ How much weight featured items get
 ```
 Overrides dataset search fields and their ponderation
 
+### SEARCH_DATASET_MAX_TEMPORAL_WEIGHT
+
+**default**: `5`
+
+After this number of years, scoring is kept constant instead of increasing.
+
+**Warning** Index time parameter: reindeixing dataset is required for this parameter to be effective
+
+### SEARCH_DATASET_FEATURED_WEIGHT
+
+**default**: `3`
+
+How much weight featured items get
+
+**Warning** Index time parameter: reindeixing dataset is required for this parameter to be effective
+
+### SEARCH_DATASET_FEATURED_BOOST
+
+**default**: `1.5`
+
+Boost given to the featured datasets
+
+### SEARCH_DATASET_CERTIFIED_BOOST
+
+**default**: `1.2`
+
+Boost given to the datasets from certified organization
+
+### SEARCH_DATASET_REUSES_DECAY
+
+**default**: `0.1`
+
+Decay factor for reuses count on datasets
+
+### SEARCH_DATASET_FOLLOWERS_DECAY
+
+**default**: `0.1`
+
+Decay factor for followers count on datasets
 
 ### SEARCH_REUSE_FIELDS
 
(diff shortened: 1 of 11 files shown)


docs/adapting-settings.md
```diff
@@ -191,6 +179,157 @@ You can disable the feature by setting this to an empty list.
...
+(
+    'geozones.keys^9',
+    'geozones.name^9',
+    'acronym^7',
+    'title^6',
+    'tags.i18n^3',
+    'description',
+)
+```
+Overrides dataset search fields and their ponderation
```
abulte: +`.`

docs/adapting-settings.md
```diff
@@ -191,6 +179,157 @@ You can disable the feature by setting this to an empty list.
...
+
+### SEARCH_DATASET_MAX_TEMPORAL_WEIGHT
+
+**default**: `5`
+
+After this number of years, scoring is kept constant instead of increasing.
+
+**Warning** Index time parameter: reindeixing dataset is required for this parameter to be effective
```
abulte: +`.`
s/reindeixing/reindexing/

docs/adapting-settings.md
abulte: Idem for every paragraph/sentence.

docs/adapting-settings.md
```diff
@@ -191,6 +179,157 @@ You can disable the feature by setting this to an empty list.
...
+
+### SEARCH_DATASET_FEATURED_WEIGHT
+
+**default**: `3`
+
+How much weight featured items get
+
+**Warning** Index time parameter: reindeixing dataset is required for this parameter to be effective
```
abulte: s/reindeixing/reindexing/

noirbizarre:
the `fields` attribute got a special treatment (ie. default setting set to `None` and let the default in the in the class definition) because:
- it's easier to read and understand the adapter with the fields in it (it depends on internals, precisely on fields declared just above, and I find it easier to define the default having the definition on which it relies next to it)
- unlike other parameters, the `fields` processing is in the base class and removing it makes it totaly black magic (ie. no `fields` reference at all in the child class, no way to understand what happen unless you read the adapter base class code)

I can totally move it, but I think maintainance wil be a little bit harder then (and the search topic is not the easyiest to understand).

What do you prefer ? Totally extracted but harder to understand/maintain or special treatment ?

**NB**: For the deploy on check, already done there with and without this PR. When deploying on next, keep in mind that on next we don't have only this PR but the entire release and so the results may still change a little bit because of the previous search fixes/changes in it (but not so much for our tested case I think)

abulte:
> it's easier to read and understand the adapter with the fields in it (it depends on internals, precisely on fields declared just above, and I find it easier to define the default having the definition on which it relies next to it)

It's true as long as different settings are not set. When they are, you will be looking at code (field definition) that is not relevant and can be misleading.

> unlike other parameters, the fields processing is in the base class and removing it makes it totaly black magic (ie. no fields reference at all in the child class, no way to understand what happen unless you read the adapter base class code)

Yes, this is not very legible. But still, I'd rather have to dig through code than to easily find something that is not used (ie field definitions if different settings are set).

> NB: For the deploy on check, already done there a production DB, with and without this PR. When deploying on next, keep in mind that on next we don't have only this PR but the entire release and so the results may still change a little bit because of the previous search fixes/changes in it (but not so much for our tested case I think)

That latest on run for next on https://etalab.github.io/datagouv-search-indicator/#/next.data.gouv.fr/ is `2018/02/27`, I'd like to see a fresher one before this PR is deployed. Maybe you didn't commit/push/publish?

Yes, understood for the impact of the other PRs. We can compare current prod / next before this PR / next after this PR to have a fair idea of the impact I guess.

noirbizarre:
>> it's easier to read and understand the adapter with the fields in it (it depends on internals, precisely on fields declared just above, and I find it easier to define the default having the definition on which it relies next to it)

> It's true as long as different settings are not set. When they are, you will be looking at code (field definition) that is not relevant and can be misleading.

The whole purpose of this PR is being able to tune the search without touching the code. If you want to change the default, you are a developping and will need to repackage (kind of defeat the purpose of this) so I don't think this cover a real use case.

>> unlike other parameters, the fields processing is in the base class and removing it makes it totaly black magic (ie. no fields reference at all in the child class, no way to understand what happen unless you read the adapter base class code)

> Yes, this is not very legible. But still, I'd rather have to dig through code than to easily find something that is not used (ie field definitions if different settings are set).

The goal was to allow non-devs to play and find the right parameters so they can then become the new defaults, not to allow developpers to have personnal search parameters and code something only working on their setup. 

Anyway, It'done but I strongly believe there is no benefice in doing so, only a complexity overhead for developpers :/

>> NB: For the deploy on check, already done there a production DB, with and without this PR. When deploying on next, keep in mind that on next we don't have only this PR but the entire release and so the results may still change a little bit because of the previous search fixes/changes in it (but not so much for our tested case I think)

> That latest on run for next on https://etalab.github.io/datagouv-search-indicator/#/next.data.gouv.fr/ is 2018/02/27, I'd like to see a fresher one before this PR is deployed. Maybe you didn't commit/push/publish?

> Yes, understood for the impact of the other PRs. We can compare current prod / next before this PR / next after this PR to have a fair idea of the impact I guess.

I run it many time locally during the development to ensure there was no changes because I don't have a safe way to test just this PR on next and because the search indicator is the most (only?) efficient way to test the search during dev (it was meant to :wink: ).
It was planned from the beginning to do a before/after on next (and on prod to for fresher data and have history through release). I was a *NB* just to remind that it's not testing the PR but the release, which is why it's important to test locally during the dev and not rely only on next (which contains multiple changes) and prod (which data are always changing due to harvest, users interacting...) for search changes.

abulte:
> > > it's easier to read and understand the adapter with the fields in it (it depends on internals, precisely on fields declared just above, and I find it easier to define the default having the definition on which it relies next to it)
> 
> > It's true as long as different settings are not set. When they are, you will be looking at code (field definition) that is not relevant and can be misleading.
> 
> The whole purpose of this PR is being able to tune the search without touching the code. If you want to change the default, you are a developping and will need to repackage (kind of defeat the purpose of this) so I don't think this cover a real use case.
> 

Doesn't it cover the case where you read the code and go to change the parameters when you understood how search works? Doesn't it cover the case where you're working on an existing instance with potentially different settings?

Maybe you didn't get what I have in mind. I want to avoid someone stumbling on this code https://github.com/opendatateam/udata/blob/master/udata/core/dataset/search.py#L121 and think that is what is used in search, when the settings could be "silently" overridden in another file. Keep in mind that the person having modified the parameters can be different from the person reading the code: new developer on an existing project, team with more than one member...

> > > unlike other parameters, the fields processing is in the base class and removing it makes it totaly black magic (ie. no fields reference at all in the child class, no way to understand what happen unless you read the adapter base class code)
> 
> > Yes, this is not very legible. But still, I'd rather have to dig through code than to easily find something that is not used (ie field definitions if different settings are set).
> 
> The goal was to allow non-devs to play and find the right parameters so they can then become the new defaults, not to allow developpers to have personnal search parameters and code something only working on their setup.

I don't understand. That's not my goal at all. I don't even see how removing `cls.fields` would particularly "allow developpers to have personnal search parameters and code something only working on their setup.". Maybe I missed something.

> Anyway, It'done but I strongly believe there is no benefice in doing so, only a complexity overhead for developpers :/

I'm open to discussion, this is not an urgent or critical matter.

noirbizarre:
Sorry, the response appeared on merge submit :/
I respond here

> Doesn't it cover the case where you read the code and go to change the parameters when you understood how search works? Doesn't it cover the case where you're working on an existing instance with potentially different settings?

No its not, it should not happen (I hesitated to raise a warning when any `SEARCH_*_FIELDS` is set). 

The main use case was to allow datascientists (which expressed the need for this PR to be able to work and automate) to iterate quickly on parameters, in a docker environment by example, without having to touch the code, in order to find optimal settings to put as default. It's not meant to be production settings (doing so would mean that settings fixes/tunings will never go back upstream)

> Maybe you didn't get what I have in mind. I want to avoid someone stumbling on this code https://github.com/opendatateam/udata/blob/master/udata/core/dataset/search.py#L121 and think that is what is used in search, when the settings could be "silently" overridden in another file. Keep in mind that the person having modified the parameters can be different from the person reading the code: new developer on an existing project, team with more than one member...

That's exactly what I wanted to handle: without the `fields` attribute, you might have to read `elasticsearch-dsl` to understand where does that comes from. A new developper reading a search adapter code won't have the `fields` in it, and it will have to investigate to know this attribute even exists and is responsible for ponderation. In an IDE, he won't be able to use hyperlink navigation to find the behavior.

For me, there is 3 possibilities:
- you are a datascientist working on search parameters, you only need/have a packaged udata (or a docker) and your setting file where you just have set any `SEARCH_*_FIELDS` (any `SEARCH_*`). That was the need expressed twice (the 2 times we try without success to let datascientists find the best params)
- you are on a deployed instance and in this case, if you have defined any `SEARCH_*_FIELDS`, you have to know that upstream code changes might not be as effective on your instance than with the default settings (because devs won't have your settings when doing changes and test won't run with these too).
- you are developping on udata and in this case I hope you are working as much as possible with default values on search, otherwise your changes might not be effective outside your environment.

> I don't understand. That's not my goal at all. I don't even see how removing cls.fields would particularly "allow developpers to have personnal search parameters and code something only working on their setup.". Maybe I missed something.

That's not the removal of `cls.fields` which is permitting that, it's' what I'm understanding in:

> Yes, this is not very legible. But still, I'd rather have to dig through code than to easily find something that is not used (ie field definitions **if different settings are set**).

When developping (and so diging through the code), I believe you shouldn't rely on special settings (because if so, you are not testing the nominal case and your changes might only work in your environment where the proper settings apply)

> I don't understand. That's not my goal at all.

The PR goal was allowing to a non-dev to switch the workflow from:
- changing values in code (meaning finding value on code, even if you don't do Python or don't understand what's happening)
- packaging (which need a development environment, which was the main blocking point)
- running tests on an environment with significant data
- analyse
- repeat until it's OK
- submit default changes or give them to a dev which will do it

to:

- changing values in ~~code~~ settings
- ~~packaging~~
- running test
- analyse
- repeat until it's OK
- submit your changes to a dev which will do it

The `fields` attribute was kept to avoid workflow changes and complexity overhead for devs (ex: with settings only field, you have to know `FakeSearchAdapater` fields are expressed in `TestSettings.SEARCH_FAKE_FIELDS` instead of directly in `FakeSearchAdapter`, in the test or the fixture)

Without the  "let the datascientists do the tuning" part, I wouldn't have submitted this as it more complexity for us

PR #2277, opened 2019-08-08
Optimize CSV generation

abulte:
This introduces `no_cache` when generating CSV to lower memory footprint (https://github.com/MongoEngine/mongoengine/issues/365).

The `resources` (I think) export still eats up a lot of memory, this should be further looked into. This optimisation should at least allow the job to complete on our production machines.

It also looks like the memory is not completely released between the different models when doing a CSV export. Running them one after the other might help w/ memory pressure, thus the introduction a (optionnal) `model` argument.

NB: closing the csv file makes no difference memory wise but it's still a good idea I guess.

## before

out

## after

out_no_cache_all

### DIFF
--- a/udata/core/dataset/tasks.py
+++ b/udata/core/dataset/tasks.py
@@ -167,12 +167,16 @@ def export_csv_for_model(model, dataset):
 
 
 @job('export-csv')
-def export_csv(self):
+def export_csv(self, model=None):
     '''
     Generates a CSV export of all model objects as a resource of a dataset
     '''
     ALLOWED_MODELS = current_app.config.get('EXPORT_CSV_MODELS', [])
     DATASET_ID = current_app.config.get('EXPORT_CSV_DATASET_ID')
+
+    if model and model not in ALLOWED_MODELS:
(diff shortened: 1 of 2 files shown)


udata/core/dataset/tasks.py
```diff
@@ -157,6 +158,7 @@ def export_csv_for_model(model, dataset):
         csvfile.seek(0)
         # make a resource from this tmp file
         created, resource = store_resource(csvfile, model, dataset)
+        csvfile.close()
```
noirbizarre: Maybe use the context manager to avoid unclosed files on error or move this into the [`finally` block](https://github.com/opendatateam/udata/blob/1b371426cfe4a126327344775cd4047d9a2bf89a/udata/core/dataset/tasks.py#L167-L168)

noirbizarre:
The graph shows improvement but the RAM is not freed until the end of the operation, and not entirely freed (there's ~400mb not released). It might good to find what's preventing memory to be released in another PR.

**NB**: I bet on another MongoEngine memory leak :/

PR #2295, opened 2019-08-21
Remove user profile page(s)

abulte:
We discussed this change a few times, and always agreed it was not needed anymore. The profile page adds little value. It can be a problem wrt/ spam.


taniki:
When we addressed the list of defunct features, I remember having shared to have changed my mind about this. These pages useful for some lone gunners who are not part of an org.

abulte:
That is valid point. 

If we look at the stats for 2019, organization pages have `267 793` views, users `19 039` (x14 for orgs, 0,1% of total traffic for user pages ). Bounce rate is the same. 

If we look at the top 5:
Capture d'écran 2019-08-23 10 20 34

- `meilleurs-agents` should be an org
- `christian-quest-2` is an issue
- `team-reseaux-mobiles` should be an org
- ` immo-data` should be an org
- ` thomas-lourenco` is not an issue IMHO: someone that surfed on the DVF wave and only has three reuses

I guess Christian can create a `cquest.org` org if he wants to ;-)

I still think we can get rid of this. Alternative (regarding the spam problem): do not show the user's description on the profile page. If we decide to keep it, we should also think about making it more useful, but I don't think that's were data.gouv.fr is headed at (individual users vs orgs).

abulte:
@noirbizarre if we decide to go through w/ this, I need your advice on this: https://github.com/opendatateam/udata/pull/2295/files#diff-dfd4142ff996ccd8559b0801f2c22c26R18

taniki:
So, where is data.gouv.fr is headed at?

I am not thinking about retention rate here but how to promote activities of single users like our pal antoine augusti. #teammakeitmoreuseful

taniki:
ok with removing the description however

taniki:
Another concern is that it will be a major blow if there is no use profile page while having a failing admin/user dashboard

taniki:
OK I still think this is might be a hole in flow for power users (construction of reputation) but it is only minor. If one wants to show she is active she can make an handmade list of relevant contributions.

ThibaudDauce:

PR #2313, opened 2019-09-16
Add conditionnal template hooks for dataset sidebar

abulte:
Allows to conditionally hook into dataset sidebar template via an extension like https://github.com/abulte/udata-dataset-sidebar-example.


ThibaudDauce:

PR #2321, opened 2019-09-23
Fix the Python 3 build

noirbizarre:
This PR is an attempt to fix CircleCI build on Python official docker images:
- Latest version of venv/pip force an `ensurepip` call on venv creation but the `ensurepip` module is stripped from official docker images resulting in venv creation failure. This adds a `--without-pip` parameter on `venv` creation to prevent `ensurepip` call.
- the flake8 `max-line-length` was declared twice in `setup.cfg`. This removes the second declaration.


noirbizarre:
Docker images updated to include virtualenv

PR #2336, opened 2019-10-02
Upgrade mongo update

noirbizarre:
This PR:
- switches to MongoDB 3.6
- upgrade to latest `pymongo` and `mongoengine` versions
- fixes the removal of `EmbeddedDocument.save()` until refactoring

Open thought about MongoDB 4: it seems to improve performances and consistency (Secondary reads, distributed transaction, multi documents transactions...) without breaking changes for us (See https://docs.mongodb.com/manual/release-notes/4.2-compatibility/ and https://docs.mongodb.com/manual/release-notes/4.0-compatibility/)
Why not give a try to Mongo 4.2 upgrade ?

In case we stay on Mongo 3.6, don't use Master-Slave replication, it's removed into MongoDB 4 (this will prevent nightmare upgrade from 3.6 to 4.x)

### DIFF
--- a/udata/core/dataset/models.py
+++ b/udata/core/dataset/models.py
@@ -350,6 +350,11 @@ class Resource(ResourceMixin, WithMetrics, db.EmbeddedDocument):
     def dataset(self):
         return self._instance
 
+    def save(self, *args, **kwargs):
+        if not self.dataset:
+            raise RuntimeError('Impossible to save un orphan resource')
+        self.dataset.save(*args, **kwargs)
+
 
 class Dataset(WithMetrics, BadgeMixin, db.Owned, db.Document):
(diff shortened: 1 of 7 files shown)


udata/core/dataset/models.py
```diff
@@ -350,6 +350,11 @@ class Resource(ResourceMixin, WithMetrics, db.EmbeddedDocument):
     def dataset(self):
         return self._instance
 
+    def save(self, *args, **kwargs):
+        if not self.dataset:
+            raise RuntimeError('Impossible to save un orphan resource')
```
abulte:
```suggestion
            raise RuntimeError('Impossible to save an orphan resource')
```

PR #2341, opened 2019-10-09
Markdown: Proper GFM support and rendering consistency between front and back

noirbizarre:
This PR change the backend markdown implementation from `commonmark-py` to `mistune` which is faster and support GFM (ie. strikethrough, tables...).

This also fixes:
- the JSON-LD `description` field which should be a raw text field and was a mix of raw and html strings (without proper escapign which was a security issue) (See: https://w3c.github.io/json-ld-syntax/#restrictions-for-contents-of-json-ld-script-elements)
- client-side escaping errors
- the `mdstrip` which was rendering twice (and so was creating a small performance hit)
- the tags security filtering which was happening too early (and so let some tags pass)

As the GFM specs gives a list of unauthorized tags, the `MD_ALLOWED_TAGS` could be removed (not done here but could be)


udata/settings.py
```diff
@@ -149,28 +149,60 @@ class Defaults(object):
...
         'sup',
         'sub',
+        'table',
+        'td',
+        'th',
+        'tr',
+        'tbody',
+        'thead',
+        'tfooter',
+        # 'title',
```
abulte: Remove?

udata/frontend/helpers.py
```diff
@@ -401,3 +403,26 @@ def filesize(value):
...
+    elif isinstance(o,  (list, tuple)):
+        return [json_ld_script_preprocessor(v) for v in o]
+    elif isinstance(o, str):
+        return html.escape(o).replace('&#x27;', '&apos;')
+    else:
+        return o
+
+
+@front.app_template_filter()
+def embedded_json_ld(jsonld):
```
abulte: is this related to markdown somehow?

PR #2343, opened 2019-10-23
Remove embedded API doc

abulte:
This removes the embedded API doc. There is now a `API_DOC_EXTERNAL_LINK` setting that we simply link to. For data.gouv.fr it will be https://doc.data.gouv.fr/api/reference/.

Context: we do not need to maintain two different versions of the same doc. There was a CVE on `swagger-ui` npm package, so it's a good time for removing it.

Capture d'écran 2019-10-23 16 01 23


abulte:
I've waited a bit for the community to chime in since this is breaking. Almost a month w/o comment -> merge.

kojicz983:
@abulte we use to use this embedded API doc, do you have any suggestions on how to go about it after these changes?

abulte:
@kojicz983 we use a static page that displays the swagger doc. It should only require a few lines of code. https://github.com/etalab/doc.data.gouv.fr/blob/master/articles/_api/reference.md https://doc.data.gouv.fr/api/reference/. You may even be able to create a static page in your theme for this if you don't want another domain/server as we do and link to it via `API_DOC_EXTERNAL_LINK`.

kojicz983:
@abulte Thank you for your help!

Issue #2351, opened 2019-11-19
Changing another user's profile picture fails


quaxsze:
As an admin, trying to change another user's profile picture seems to fail.
The POST request to upload the image is a success but the GET request for profile retrieval does not return the profile picture.

![Capture d’écran 2019-11-19 à 15 42 05](https://user-images.githubusercontent.com/57526019/69156166-48980e00-0ae3-11ea-88be-38bd3583faa7.png)

quaxsze:
After some tests, it seems the endpoint for avatar uploading is not designed to make the difference between two users. It's uploads the avatar for the current authenticated user. Trying to change someone else's avatar results in actually changing our own avatar.
The endpoint must be rewriten.

PR #2354, opened 2019-11-22
fixed user's avatar change

quaxsze:
fix https://github.com/opendatateam/udata/issues/2351

### DIFF
--- a/udata/core/user/api.py
+++ b/udata/core/user/api.py
@@ -251,10 +251,10 @@ class UserListAPI(API):
         return user, 201
 
 
-@ns.route('/<user:user>/avatar', endpoint='given_user_avatar')
+@ns.route('/<user:user>/avatar', endpoint='user_avatar')
 class UserAvatarAPI(API):
-    @api.secure
-    @api.doc('given_user_avatar')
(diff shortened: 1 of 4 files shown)


udata/core/user/api.py
```diff
@@ -251,6 +251,19 @@ def post(self):
         return user, 201
 
 
+@ns.route('/<user:user>/avatar', endpoint='given_user_avatar')
```
abulte: `user_avatar` is fine I guess.

udata/core/user/api.py
```diff
@@ -251,6 +251,19 @@ def post(self):
...
+class UserAvatarAPI(API):
+    @api.secure
+    @api.doc('given_user_avatar')
+    @api.expect(image_parser)
+    @api.marshal_with(uploaded_image_fields)
+    def post(self, user):
+        '''Upload a new avatar for a given user'''
+        parse_uploaded_image(user.avatar)
+        user.save()
```
abulte: Missing permission check ;-) I think there's a decorator to allow a route only for admins.

quaxsze: Done

Issue #2300, opened 2019-08-30
Conflict between id and slug can cause major trouble


taniki:
If one use id of an existing dataset as a title of a new dataset, the URI will resolve to the new dataset. This can be a major issue if some users are using id instead of slugs for sake of stability.

Example:
- https://next.data.gouv.fr/fr/datasets/53699fe4a3a729239d206227/
- https://data.gouv.fr/fr/datasets/53699fe4a3a729239d206227/
- https://next.data.gouv.fr/fr/datasets/service-public-fr-annuaire-de-l-administration-base-de-donnees-locales/

abulte:
Probably a good idea to do this on organisations too.

abulte:
Cf (at least) https://github.com/opendatateam/udata/blob/fdec5338c0e4193b5ef6391304c45723b5e970e5/udata/core/organization/models.py#L93

taniki:
The same problem might also be risen in the API.

PR #2356, opened 2019-11-26
Resolve conflict between id and slug

quaxsze:
fix https://github.com/opendatateam/udata/issues/2300

### DIFF
--- a/udata/routing.py
+++ b/udata/routing.py
@@ -87,25 +87,24 @@ class ModelConverter(BaseConverter):
 
     def to_python(self, value):
         try:
-            obj = self.model.objects.get_or_404(id=value)
+            return self.model.objects.get_or_404(id=value)
         except NotFound:
-            obj = None
+            pass
         try:
-            if obj:
-                return obj
-            else:
-                quoted = self.quote(value)
-                query = db.Q(slug=value) | db.Q(slug=quoted)
-                obj = self.model.objects(query).get()
+            quoted = self.quote(value)
+            query = db.Q(slug=value) | db.Q(slug=quoted)
+            obj = self.model.objects(query).get()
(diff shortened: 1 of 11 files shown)


udata/tests/test_routing.py
```diff
@@ -83,6 +83,11 @@ def model_tester(model):
             assert isinstance(model, self.model)
             return 'ok'
 
+        @app.route('/wrong_model/<tester:model>')
+        def wrong_model_tester(model):
+            assert isinstance(model, self.model)
+            return str(model.id)
```
abulte: maybe use this `return` in the previous function, and remove this one. I don't see why we would break something else, `ok` is probably not used.

quaxsze: Done

udata/tests/test_routing.py
```diff
@@ -164,6 +164,18 @@ def test_redirect_to_normalized_utf8_slug(self, client):
         assert_redirects(client.get('/model/slüg'), url)
 
 
+class SlugAndIdTest(AsSlugMixin):
+    model = SlugTester
+    converter = SlugTesterConverter
+
+    def test_url_hijack(self, client):
```
abulte: Can you test the UUID-in-slug thing also?

quaxsze: Done

udata/routing.py
```diff
@@ -86,24 +86,25 @@ def quote(self, value):
             return url_quote(value)
 
     def to_python(self, value):
+        try:
```
abulte: Change the order slug/id/404 to id/slug/404 in the parent function documentation

udata/tests/test_routing.py
abulte: Add a documentation for this method

PR #2372, opened 2019-12-26
Fix responsive search view

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/151

### DIFF
--- a/less/udata/search.less
+++ b/less/udata/search.less
@@ -237,7 +237,7 @@ ul.search-results {
             @height: 84px;
             height: @height;
             .left-bar {
-                min-height: @height;
+                // min-height: @height;
             }
         }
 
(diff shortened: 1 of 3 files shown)


quaxsze:
This seems to happen in udata and not only in gouvfr.
It is due to the min-height of the component. It also happens on a Ipad screen.
Removing the min-height for both the small media and mid media solves the problem. Is it accurate or do you see any side effect?
I just commented the CSS for now.

less/udata/search.less
```diff
@@ -237,15 +237,15 @@ ul.search-results {
             @height: 84px;
             height: @height;
             .left-bar {
-                min-height: @height;
+                // min-height: @height;
```
abulte: remove commented lines ;-)

quaxsze: Done :)

Issue #2373, opened 2019-12-27
Discussions: non unique ids in DOM


abulte:
As reported by Chrome on https://next.data.gouv.fr/fr/datasets/service-public-fr-annuaire-de-l-administration-base-de-donnees-locales/

<img width="427" alt="Capture d'écran 2019-12-27 11 16 06" src="https://user-images.githubusercontent.com/119625/71513368-4b341300-289a-11ea-9358-ca58b4689756.png">

PR #2374, opened 2019-12-27
Fix non unique ids

quaxsze:
fix https://github.com/opendatateam/udata/issues/2373

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Fix admin messages and fix user roles selector default value [#2365](https://github.com/opendatateam/udata/pull/2365)
 - Fix new harvester's form tooltip showup [#2371](https://github.com/opendatateam/udata/pull/2371)
 - Fix responsive design of search results [#2372](https://github.com/opendatateam/udata/pull/2372)
-- Fix non-unique ids in datasets's comments [#2374](https://github.com/opendatateam/udata/pull/2374)
+- Fix non-unique ids in datasets' comments [#2374](https://github.com/opendatateam/udata/pull/2374)
 
 ## 1.6.18 (2019-12-13)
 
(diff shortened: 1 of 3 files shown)


js/components/discussions/thread-form.vue
```diff
@@ -11,47 +11,51 @@
 </template>
 
 <script>
-import log from 'logger';
+    import log from 'logger';
```
abulte: Can you revert the indentation as before? I can't review the changes ;-)

quaxsze: Done :)

js/components/discussions/thread-form.vue
```diff
@@ -22,10 +22,14 @@ export default {
...
           // duplicate POSTing.
           sending: false,
-          comment: ''
+          comment: '',
+          id: null
       }
   },
-  methods: {
+    ready: function() {
+      this.id = this._uid;
```
abulte: I did not know about `this._uid`, nice find. But have you seen this https://github.com/vuejs/vue/issues/5886#issuecomment-308625735?

I think using `this.discussionId` with a prefix like `new-comment-${this.discussionId}` should be unique enough (and you should be able to use it directly in `data`).

quaxsze: This might be simpler indeed

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Fix admin messages and fix user roles selector default value [#2365](https://github.com/opendatateam/udata/pull/2365)
 - Fix new harvester's form tooltip showup [#2371](https://github.com/opendatateam/udata/pull/2371)
 - Fix responsive design of search results [#2372](https://github.com/opendatateam/udata/pull/2372)
+- Fix non-unique ids in datasets's comments [#2374](https://github.com/opendatateam/udata/pull/2374)
```
abulte:
```suggestion
- Fix non-unique ids in datasets' comments [#2374](https://github.com/opendatateam/udata/pull/2374)
```

PR #2367, opened 2019-12-20
Adds the dataset acronym to the indexed fields

odtvince:

PR #2376, opened 2019-12-31
Adds the dataset acronym to the indexed fields

abulte:
Replaces #2367, courtesy of @odtvince


abulte:
This yields unconvincing results on next.data.gouv.fr : neither SIREN nor DVF queries yield more pertinent (first) result, in fact nothing seems to change but that was some quick testing.

AntoineAugusti:
Curious if you investigated a bit about this. Was it indexed eventually? Did you troubleshoot?

abulte:
Nothing more than last test above, this is very nebulous.

ThibaudDauce:

PR #2447, opened 2020-01-13
Fix territory routing for @latest

abulte:
When a URL like https://www.data.gouv.fr/fr/territories/commune/33063@latest/Bordeaux/ is requested, we now check if we have a GeoZone without `validity.end` before fetching the one with the latest `validity.end`. This seems to be necessary since the latest release of Geozone (?).

Fix https://github.com/etalab/data.gouv.fr/issues/159

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current (in progress)
 
 - New Crowdin translations [#2360](https://github.com/opendatateam/udata/pull/2360)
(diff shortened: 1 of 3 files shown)


abulte:
@davidbgk if you have any memory about this, I'd love your confirmation that this is a sane thing to do (or not).

davidbgk:
@abulte looks good to me 👍

PR #2451, opened 2020-01-21
Display link's text only to avoid KeyError

quaxsze:


udata/frontend/markdown.py
```diff
@@ -24,12 +24,16 @@
 
 
 def avoid_mailto_callback(attrs, new=False):
-    """Remove completely the link containing a `mailto`."""
+    """
+    Remove completely the link containing a `mailto` to avoid spam.
+    If case of a bad markdown formating for links, the href will not be found in attr and a KeyError will be raised.
```
abulte:
```suggestion
    In case of a bad markdown formatting for links, the href will not be found in attr and a KeyError will be raised.
```

PR #2458, opened 2020-03-09
fix gr./sp. in documentation

augusto-herrmann:
some improvements to English spelling and grammar in the documentation page about RDF


taniki:
Thanks ! What is bringing you around? :)

augusto-herrmann:
> Thanks! What is bringing you around? (:

I have been swimming in the open data ocean for more than a decade. That's also how long I've been involved with CKAN, having started its translation to Brazilian Portuguese in 2009. I came to uData to get to know it better.

[This forum post](https://teamopendata.org/t/un-dataverse-pour-le-monde-de-lentreprise/2225) intrigued me to find out how well uData handles data sharing in a controlled access environment. The parenthesis around "open" in the tag line

> Customizable and skinnable social platform dedicated to (open)data.

left me wondering if it might be suitable for data that is not open.

However, I did not find uData's documentation in the current state to be very inviting to newcomers or very descriptive. I did this PR as a small step trying to improve it.

taniki:
> left me wondering if it might be suitable for data that is not open.

uData is primarly targeting gov and agencies looking to enforce a wide open data policy. Most of the settings default to public sharing. I won't recommend this software for sensitive private data.

augusto-herrmann:
> >  left me wondering if it might be suitable for data that is not open.
> 
> uData is primarly targeting gov and agencies looking to enforce a wide open data policy. Most of the settings default to public sharing. I won't recommend this software for sensitive private data.

Perhaps you should consider, then, removing the parenthesis in the tag line saying (open)data? Because it definitely does give newcomers the impression that uData is suitable for closed as well as open data.

That issue aside, with uData working in the same space as CKAN (which has been around for years), it would be nice to have a comparison sheet or something in the documentation. The question to consider is: why should governments choose uData over CKAN for their open data portals?

A third question is: with uData being new, why did you choose to start it with Python 2.7, [which is no longer supported](https://pythonclock.org/) since January 1st, 2020, instead of Python 3? From [the documentation](https://udata.readthedocs.io/en/stable/system-dependencies/#python-requirements):

> uData requires Python 2.7 (installed by default on OSX and many Linux distributions)

There doesn't even seem to be [an issue for implementing Python 3 migration](https://github.com/opendatateam/udata/issues?q=is%3Aissue+Python+3) here in the repository. In comparison, CKAN is old and migrating it to Python 3 has been a priority for some time and [they're almost done with it](https://github.com/ckan/ckan/issues/5168). It would be nice to have another free and open source alternative to CKAN, but depending on an unsupported Python version definitely does not help the case for uData.

taniki:
udata isn't that new. It has been around since 6 years.

As a matter of fact, we just merged the new python 3 code base. It has been worked in pull requests and branches (:

augusto-herrmann:
> As a matter of fact, we just merged the new python 3 code base. It has been worked in pull requests and branches (:

Fantastic! Congratulations! :tada: :fireworks:

PR #2459, opened 2020-03-11
WIP: Metrics break down

quaxsze:


quaxsze:
`@api.route('/metrics/<id>', endpoint='metrics')` seems unused. After thinking it might be complicated to implement an endpoint finding any object by its ID in mongoDB. We can maybe remove this endpoint as it will be a feature of the future metrics service on InfluxDB.

quaxsze:
This is a first draft PR still WIP.
Get_metrics is for now a property. Any thoughts on which property/method is better is welcome.

udata/core/site/models.py
```diff
@@ -28,6 +30,112 @@ class Site(WithMetrics, db.Document):
 
     def __str__(self):
         return self.title or ''
+    
+    def count_users(self):
+        from udata.models import User
+        self.metrics["users"] = User.objects.count()
```
abulte: Filtering on activated users might be a good idea (`confirmed_at__ne=None`)

abulte: \+ `deleted=None`

abulte: (there might be a `visible()` method on User.objects too)

udata/core/site/models.py
```diff
@@ -28,6 +30,112 @@ class Site(WithMetrics, db.Document):
...
+    def count_reuses(self):
+        self.metrics["reuses"] = Reuse.objects.visible().count()
+        self.save()
+
+    def count_followers(self):
+        from udata.models import Follow
+        self.metrics["followers"] = Follow.objects(until=None).count()
+        self.save()
+    
+    def count_discussion(self):
```
abulte:
```suggestion
    def count_discussions(self):
```

udata/core/site/models.py
```diff
@@ -28,6 +30,112 @@ class Site(WithMetrics, db.Document):
...
+        self.save()
+
+    def count_max_org_datasets(self):
+        org = (Organization.objects(metrics__datasets__gt=0).visible()
+               .order_by('-metrics.datasets').first())
+        self.metrics["max_org_datasets"] = org.metrics['datasets'] if org else 0
+        self.save()
+    
+    @property
+    def get_max_metrics(self):
```
abulte: call it `max_metrics` or make it a function

udata/core/site/models.py
```diff
@@ -28,6 +30,112 @@ class Site(WithMetrics, db.Document):
...
+        return {
+            "max_dataset_followers": self.metrics.get("max_dataset_followers", 0),
```
abulte: nitpick: since those dict keys are the same than on `self.metrics`, maybe mutualise a list of keys.

udata/core/site/models.py
```diff
@@ -28,6 +30,112 @@ class Site(WithMetrics, db.Document):
...
+            "max_reuse_followers": self.metrics.get("max_reuse_followers", 0),
+            "max_org_followers": self.metrics.get("max_org_followers", 0),
+            "max_org_reuses": self.metrics.get("max_org_reuses", 0),
+            "max_org_datasets": self.metrics.get("max_org_datasets", 0)
+        }
+    
+    @property
+    def get_metrics(self):
+        metrics_dict = {
+            "datasets": self.metrics.get("datasets", 0),
```
abulte: nitpick: since those dict keys are the same than on self.metrics, maybe mutualise a list of keys.

udata/core/user/models.py
```diff
@@ -89,6 +90,13 @@ class User(WithMetrics, UserMixin, db.Document):
         'ordering': ['-created_at']
     }
 
+    __search_metrics__ = Object(properties={
```
abulte: nice syntax 👌

udata/core/user/models.py
```diff
@@ -236,9 +244,37 @@ def mark_as_deleted(self):
...
 
+    def count_datasets(self):
+        from udata.models import Dataset
+        self.metrics["datasets"] = Dataset.objects(owner=self).visible().count()
+        self.save()
 
-datastore = MongoEngineUserDatastore(db, User, Role)
+    def count_reuses(self):
+        from udata.models import Reuse
+        self.metrics["reuses"] = Reuse.objects(owner=self).count()
```
abulte: `visible()`?

udata/core/user/models.py
```diff
@@ -236,9 +244,37 @@ def mark_as_deleted(self):
...
+        self.save()
+    
+    def count_following(self):
+        from udata.models import Follow
+        self.metrics["following"] = Follow.objects.following(self).count()
+        self.save()
+    
+    @property
+    def get_metrics(self):
+        return {
```
abulte: nitpick: since those dict keys are the same than on self.metrics, maybe mutualise a list of keys.

udata/search/adapter.py
```diff
@@ -160,14 +159,5 @@ def as_request_parser(cls, paginate=True):
...
-    float: Float,
-}
-
-
 def metrics_mapping_for(cls):
-    props = {}
-    for name, metric in Metric.get_for(cls).items():
-        props[metric.name] = metrics_types[metric.value_type]()
-    return Object(properties=props)
+    pass
```
abulte: Remove fn?

udata/tests/site/test_site_api.py
```diff
@@ -13,32 +12,6 @@
...
-    name = 'fake-site-metric'
-    display_name = 'Fake site metric'
-    default = 0
-
-    def get_value(self):
-        return 2
-
-
-class MetricsAPITest(APITestCase):
-    def test_get_metrics_for_site(self):
```
abulte: we probably still need a test for that?

udata/tests/test_transfer.py
```diff
@@ -8,6 +8,8 @@
     transfer_request_notifications
 )
 from udata.models import Member
+from udata.core.user.metrics import update_owner_metrics
+from udata.core.organization.metrics import update_owner_metrics
```
abulte: I don't think you're supposed to include the same function name twice 😛 

Maybe the second one should be `update_organizaton_metrics`?

udata/core/organization/models.py
```diff
@@ -235,6 +243,49 @@ def json_ld(self):
...
+        return [
+            "datasets",
+            "members",
+            "reuses",
+            "followers",
+            "views",
+        ]
+
+    @property
+    def get_metrics(self):
```
abulte: use `get_metrics_keys()` here

udata/core/dataset/models.py
```diff
@@ -679,6 +685,50 @@ def json_ld(self):
...
+        return [
+            "discussions",
+            "issues",
+            "reuses",
+            "followers",
+            "views",
+        ]
+
+    @property
+    def get_metrics(self):
```
abulte: use `get_metrics_keys()` here

udata/core/reuse/models.py
```diff
@@ -163,6 +169,49 @@ def json_ld(self):
...
+        return [
+            "discussions",
+            "issues",
+            "datasets",
+            "followers",
+            "views",
+        ]
+
+    @property
+    def get_metrics(self):
```
abulte: use `get_metrics_keys()` here

abulte: well, not really possible here 🤔 this pattern (keys duplication) is broken. we should be able to find a better way 🌈

udata/core/organization/models.py
abulte: well, not really possible here 🤔 this pattern (keys duplication) is broken. we should be able to find a better way 🌈

abulte:
# Follow up review ⬆️ 

## question

> how do we display the number of private datasets for an org or a user (before and now)? I did not see any code related to that.

TBD, Stan is looking into it.


## general implementation

> computing all site-wide metrics at each modification of an object seems very risky performance wise. Do it through a batch? Can be cached/delayed somehow?

A scheduled task is probably a good idea (like once 1 hour).

> I did not see any caching anywhere — maybe it's not a bad idea, I just wan't to make we're not forgetting about that

Probably no need, except for direct computations of a few metrics (`len()`...) that deserve a `cached_property`.

## big picture

> what should do with the history (all Metrics currently stored, both in their own collection and on the objects)?

- we remove the metrics collection (**save a dump**)  when deploying this version (not a migration, let's not mess with people's DB) — we might use it someday to retrieve some metrics, or not
- about `obj.metrics`: add a `--drop` to the command that inits the metrics in this PR

> can the new metrics on the objects live alongside the old ones somehow? If not, how do we make sure they're not conflicting? Clean them up?

cf above

> should we already be sending some data to Influx?

Timestamped site-wide metrics could be nice to have in Influx. See how we organise things w/ `udata-piwik` already using Influx: maybe a `udata-metrics` extension that takes care of the connection to Influx (at least).

## do not forget

- `udata-gouvfr`
- `udata-piwik`

udata/core/followers/api.py
```diff
@@ -52,7 +52,7 @@ def get(self, id):
     @api.doc(description=NOTE)
     def post(self, id):
         '''Follow an object given its ID'''
-        model = self.model.objects.only('id').get_or_404(id=id)
+        model = self.model.objects.get_or_404(id=id)
```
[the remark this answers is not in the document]
quaxsze: Yes we need it to get the full object of the following

udata/core/dataset/search.py
```diff
@@ -33,11 +31,11 @@
 
 
 def max_reuses():
-    return max(current_site.metrics.get('max_dataset_reuses'), 10)
+    return max(current_site.get_max_metrics['max_dataset_reuses'], 10)
```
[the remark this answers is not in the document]
quaxsze: Done

udata/core/reuse/models.py
```diff
@@ -163,6 +169,49 @@ def json_ld(self):
...
+    def views_count(self):
+        return self.metrics['views']
+    
+    @property
+    def datasets_count(self):
+        return len(self.datasets)
+
+    def count_discussions(self):
+        from udata.models import Discussion
+        self.metrics["discussions"] = Discussion.objects(subject=self, closed=None).count()
```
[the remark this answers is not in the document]
quaxsze: It was there before. I guess we do not want to count closed discussions?

udata/core/site/models.py
quaxsze: The visible method checks if the user has any dataset or reuse to return True. I don't think it is relevant in this metric. We might want to know someday if there is a huge load of random users created suddenly.

udata/core/site/models.py
quaxsze: There's not exactly the same keys. It would means iterate on mutualised keys but filter to get discussions key for instance. Do you think of a clean way to do this? I think the code might just be less understandable. But it might also be the PTSD of the actual UData code being awfull to understand...

udata/core/site/models.py
quaxsze: cf above

udata/core/user/models.py
quaxsze: Done

udata/core/user/models.py
quaxsze: cf above

udata/core/site/models.py
```diff
@@ -26,8 +28,110 @@ class Site(WithMetrics, db.Document):
...
+        'max_org_reuses',
+        'max_org_datasets',
+        'datasets',
+        'discussions',
+        'followers',
+        'organizations',
+        'public-service',
+        'resources',
+        'reuses',
+        'users'
```
abulte: trailing comma

udata/core/user/models.py
```diff
@@ -237,9 +252,28 @@ def mark_as_deleted(self):
         Follow.objects(following=self).delete()
         mail.send(_('Account deletion'), copied_user, 'account_deleted')
 
+    def count_datasets(self):
+        from udata.models import Dataset
+        self.metrics["datasets"] = Dataset.objects(owner=self).visible().count()
```
abulte:
```suggestion
        self.metrics['datasets'] = Dataset.objects(owner=self).visible().count()
```

udata/core/user/models.py
```diff
@@ -237,9 +252,28 @@ def mark_as_deleted(self):
...
+        self.save()
 
-datastore = MongoEngineUserDatastore(db, User, Role)
+    def count_reuses(self):
+        from udata.models import Reuse
+        self.metrics["reuses"] = Reuse.objects(owner=self).visible().count()
```
abulte:
```suggestion
        self.metrics['reuses'] = Reuse.objects(owner=self).visible().count()
```

udata/core/user/models.py
```diff
@@ -237,9 +252,28 @@ def mark_as_deleted(self):
...
+        self.save()
 
+    def count_followers(self):
+        from udata.models import Follow
+        self.metrics["followers"] = Follow.objects.followers(self).count()
```
abulte:
```suggestion
        self.metrics['followers'] = Follow.objects.followers(self).count()
```

udata/core/user/models.py
```diff
@@ -237,9 +252,28 @@ def mark_as_deleted(self):
...
+        self.save()
+    
+    def count_following(self):
+        from udata.models import Follow
+        self.metrics["following"] = Follow.objects.following(self).count()
```
abulte:
```suggestion
        self.metrics['following'] = Follow.objects.following(self).count()
```

udata/core/reuse/models.py
```diff
@@ -68,6 +69,20 @@ def __str__(self):
 
     __badges__ = {}
 
+    __search_metrics__ = Object(properties={
+        'datasets': Integer(),
+        'followers': Integer(),
+        'views': Integer()
```
abulte, line 75:
```suggestion
        'views': Integer(),
```

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Current (in progress)
 
+- Full metrics refactoring [2459](https://github.com/opendatateam/udata/pull/2459)
```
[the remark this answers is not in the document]
abulte: Also document the use of the job for site metrics.

PR #2462, opened 2020-03-23
Add setting to allow images to be displayed in markdown

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/168


udata/frontend/markdown.py
```diff
@@ -86,9 +86,14 @@ def __call__(self, stream, source_tooltip=False, wrap=True):
         if source_tooltip:
             callbacks.append(source_tooltip_callback)
         
+        allowed_attr_md = current_app.config['MD_ALLOWED_ATTRIBUTES']
+        
+        if current_app.config['MD_ALLOWED_IMG']:
```
abulte: `MD_ALLOWED_ATTRIBUTES` with img in it is not enough?

quaxsze: No because bleach strips it of the `src` tag so the image does not display.

quaxsze: I included the title tag which is maybe useless

abulte: what about smtg like this?
```suggestion
        if 'img' in current_app.config['MD_ALLOWED_ATTRIBUTES']:
```

quaxsze: Good idea it seems cleaner but maybe less understandable from the config file. But i like the look of it 👍

quaxsze: The purpose here was to use a virgin variable to avoid modifying the config dict one.
if I use this:
`if 'img' in current_app.config['MD_ALLOWED_ATTRIBUTES']:`
I can't do something like this after:
`current_app.config['MD_ALLOWED_ATTRIBUTES']['img'].append('src')`
Because we end up with something like this after a few reloads:
`{'a': ['href', 'title', 'rel', 'data-tooltip'], 'abbr': ['title'], 'acronym': ['title'], 'img': ['src', 'src', 'src', 'src', 'src', 'src', 'src', 'src', 'src', 'src', 'src', 'src', 'src', 'src', 'src', 'src', 'src', 'src', 'src', 'src']}`

udata/settings.py
```diff
@@ -187,6 +187,7 @@ class Defaults(object):
         'a': ['href', 'title', 'rel', 'data-tooltip'],
         'abbr': ['title'],
         'acronym': ['title'],
+        'img': []
```
abulte: Why aren’t we using `src` and `title` here? It seems to work for the other attrs.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Current (in progress)
 
+- Add setting to allow images to be displayed in markdown [2462](https://github.com/opendatateam/udata/pull/2462)
```
abulte:
```suggestion
- Allow images to be displayed in markdown by default [2462](https://github.com/opendatateam/udata/pull/2462)
```

PR #2477, opened 2020-05-20
fix: modal content vertical overflow

revolunet:
When the resource modal content is verbose, we cant see all the content and cannot scroll down

Example : https://www.data.gouv.fr/fr/datasets/texte-provenant-des-pdfs-trouves-sur-data-gouv-fr/#resource-community-19755a30-3cb5-408d-9603-71fbbe284901

this fix limits height and make it scrollable


abulte:
Thanks @revolunet ;-)

Issue #2481, opened 2020-05-24
Handle CORS on error responses from API


abulte:
CORS is currently not supported on all errors responses (`40x`, `50x`) from the API.

This currently means:
- epic trouble when trying to use the API from a SPA: you never know why your request failed
- it's _impossible_ to implement chunked upload from a SPA, since `UploadStatus` is handled as an exception and thus does not support CORS — the process will fail after the chunk is uploaded

quaxsze:
CORS are now handled by Flask-CORS since #2510 
This seems to have fixed the issue.

PR #2485, opened 2020-05-26
WIP: added flask cors instead of restplus one

quaxsze:
fix https://github.com/opendatateam/udata/issues/2481

### DIFF
--- a/udata/api/__init__.py
+++ b/udata/api/__init__.py
@@ -8,7 +8,8 @@ from flask import (
     current_app, g, request, url_for, json, make_response, redirect, Blueprint
 )
 from flask_fs import UnauthorizedFileType
-from flask_restplus import Api, Resource, cors
+from flask_restplus import Api, Resource
+from flask_cors import CORS
 
 from udata import tracking, theme
 from udata.app import csrf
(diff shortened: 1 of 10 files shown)


udata/api/__init__.py
```diff
@@ -61,6 +62,8 @@
     'X-Forwarded-Proto',
 )
 
+co = CORS(allow_headers=PREFLIGHT_HEADERS)
```
abulte, line 65: avoid two letters variable names. `cors`?

udata/api/__init__.py
```diff
@@ -165,12 +168,7 @@ def page_parser(self):
 
 api = UDataApi(
     apiv1,
-    decorators=[csrf.exempt,
-                cors.crossdomain(origin='*',
-                                 credentials=True,
```
quaxsze, line 170: Credentials are set to true here.
Flask-cors has an option 'supports_credentials' but as the documentation states 'This option cannot be used in conjuction with a ‘*’ origin'.
Do we need to refine the origin parameter in order to use this option?

abulte: good question, let's see what works and doesn't work on demo as is.

abulte: (I don't see how we can refine the origin 🤔)

quaxsze:
Handled in https://github.com/opendatateam/udata/pull/2510

PR #2492, opened 2020-06-10
Add banner option in settings

JulienParis:
adding very little settings to allow banner display...

PR parallel to : https://github.com/etalab/udata-gouvfr/pull/480

- [x] to be cleaned from comments
- [ ] add banner html example ? 
- [ ] try banner on default theme ? 
- [ ] add test


JulienParis:
Just needed to inject the banner settings to the context, either on homepage or others to make the banner visible on every page...

JulienParis:
Capture d’écran 2020-06-11 à 08 10 39

JulienParis:
Capture d’écran 2020-06-11 à 08 10 49

udata/core/site/views.py
```diff
@@ -96,6 +96,8 @@ def home():
     }
     processor = theme.current.get_processor('home')
     context = processor(context)
+    context['banner_activated'] = current_app.config['BANNER_ACTIVATED']
```
abulte, line 99: I'm not sure you need this. Did you try to using global `config` context var? It should be available in the templates, cf https://github.com/opendatateam/udata/blob/master/udata/templates/apidoc.html#L5 for example.

udata/frontend/views.py
```diff
@@ -1,22 +1,28 @@
-from flask import request, redirect, abort, g, json
+import logging
+
+from flask import request, redirect, abort, g, json, current_app
 from flask.views import MethodView
 
 from udata import search, auth, theme
 from udata.utils import not_none_dict
 
+log = logging.getLogger(__name__)
```
abulte, line 9: remove (plus the import) since you're not using it.

udata/frontend/views.py
```diff
@@ -1,22 +1,28 @@
...
 
 class Templated(object):
     template_name = None
 
     def get_context(self):
-        return {}
+        basic_context = {}
+        return basic_context
```
abulte, line 16: Useless diff, revert to previous.

udata/core/site/views.py
JulienParis: changed that and [adapted the gouvfr theme for it](https://github.com/etalab/udata-gouvfr/pull/480/files#diff-60ca9eea7a34ed3b3dba3f52f6d46227), thx for the tip 👍

udata/settings.py
```diff
@@ -439,6 +439,9 @@ class Defaults(object):
 
     API_DOC_EXTERNAL_LINK = 'https://doc.data.gouv.fr/api/reference/'
 
+    # Global banner parameters
+    BANNER_ACTIVATED = True
```
abulte, line 443: set it to `False` by default

udata/settings.py
```diff
@@ -439,6 +439,9 @@ class Defaults(object):
...
+    BANNER_HTML_FILENAME = 'banner-2020-06-datactivist-survey.html'
```
abulte, line 444: set it to `''` by default

Issue #2489, opened 2020-06-10
System jobs not launched after udata>=2.0.0 upgrade


mayjolux:
Hi from Luxembourg!

We finally finished our migration from udata 1.6.20 to udata 2.0.0  and now even to 2.1.0.
However, I noticed that our Harvester Jobs (purging deleted datasets, organizations too) were not launching since I migrated from 1.6.20 to 2.0.0. If I define them in the System Tab of the admin panel, they show up as usual and get their default cron job planification of 0 0 * * *, but they don't get launched at that time.

If I go now on the machine and trigger the launch with the "udata harvest launch" method, the harvest is started without a problem, no error messages, everything just works fine. I don't know if I am missing something specific in the changes of udata 2.0.0.

In your breaking changes of udata 2.0.0 you say: "Deprecated celery tasks have been removed, please ensure all old-style tasks (pre 1.6.20) have been consumed before migrating" Since those harvest jobs are launched by celery workers I think it has maybe something to due with this? Maybe you have some additional information or encountered the same kind of problems during your migration?

In another small question, concerning the harvester creation and its planification, if I program the planification to 0 2 * * * for example in the UI of EDIT>Planification, the following is added to the cron expression b'0' b'2' b'\*' b'\*' b'*', which generates an error. I am not quite sure why this happens, do you have an idea?

Thanks for any help in this matter!

abulte:
Hi,

What is the output of `udata job scheduled`? You might also want to look at the state of your celery queues and see if the jobs are correctly consumed (`udata worker status` might help).

> In another small question, concerning the harvester creation and its planification, if I program the planification to 0 2 * * * for example in the UI of EDIT>Planification, the following is added to the cron expression b'0' b'2' b'*' b'*' b'*', which generates an error. I am not quite sure why this happens, do you have an idea?

Looks like a bug from the migration to python3. We're not using this feature lately, that's why we did not catch it. Can you fill a different issue for this?

mayjolux:
This is the output of the udata job scheduled:

b'Harvest Harvester IGSS': harvest(5ede2bd69e975616ca4ceab7) ↦ 0 10 * * *
b'Harvest Harvester STATEC': harvest(5ede0215ae9201709ddf956f) ↦ b'0' b'0' b'*' b'*' b'*'
b'Generate hourly Piwik metrics': piwik-current-metrics ↦ 0 * * * *
[...]

PR #2493, opened 2020-06-11
Add cast if bytes in haverster's cron schedule

quaxsze:
fix https://github.com/opendatateam/udata/issues/2489

### DIFF
--- a/udata/harvest/api.py
+++ b/udata/harvest/api.py
@@ -257,7 +257,6 @@ class ScheduleSourceAPI(API):
             data = request.json
         except Exception as e:
             data = request.data
-        if type(data) == bytes:
             data = data.decode('utf-8')
         return actions.schedule(ident, data)
 
--- a/udata/harvest/api.py
+++ b/udata/harvest/api.py
@@ -257,6 +257,8 @@ class ScheduleSourceAPI(API):
             data = request.json
         except Exception as e:
             data = request.data
+        if type(data) == bytes:
+            data = data.decode('utf-8')
         return actions.schedule(ident, data)
 
     @api.doc('unschedule_harvest_source')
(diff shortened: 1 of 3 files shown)


udata/harvest/api.py
```diff
@@ -257,6 +257,8 @@ def post(self, ident):
             data = request.json
         except Exception as e:
             data = request.data
+        if type(data) == bytes:
+            data = data.decode('utf-8')
```
abulte, line 261: when do you have bytes and when do we not? probably when it's not json? if it is, we should cast it in the above catch block.

if it's not, we should catch that at a higher level of abstraction I think, like in `actions.schedule` 

there is already a `str()` cast there BTW https://github.com/opendatateam/udata/blob/master/udata/harvest/actions.py#L212, maybe it's the one causing trouble, because:

```
>>> a = b'a'
>>> a
b'a'
>>> str(a)
"b'a'"
```

quaxsze: the str cast is there because the function signature accepts an hour, minute ect argument. Thoses arguments may be string or integers. This behavior can be found in the tests.
I will check if the bytes are retrieved only when it is not Json or not.

udata/harvest/api.py
```diff
@@ -257,6 +257,7 @@ def post(self, ident):
             data = request.json
         except Exception as e:
             data = request.data
```
abulte, line 259:
```suggestion
            data = request.data.decode('utf-8')
```

udata/harvest/api.py
```diff
@@ -257,6 +257,7 @@ def post(self, ident):
             data = request.json
         except Exception as e:
```
abulte, line 258: while we're at it, maybe catch only the right exception (`JSONDecodeError`?)

PR #2494, opened 2020-06-11
removed pyup file

quaxsze:


quaxsze:
I cannot remove the project from pyup. I do not have the rights for it.
At https://pyup.io/account/repos/github/opendatateam/udata/

quaxsze:
Do we want to remove it on every opendatateam project? If yes i will make the same PR for the other projects.

abulte:
This is nice (cleanup) but it won't change anything since update are disabled in the conf file ;-)

For the other repos we need to:
- work on a renovate config that works for us (this repo as a starting point)
- replace pyup conf w/ renovate config

PR #2486, opened 2020-06-02
Fix bleach domain parsing

JulienParis:
First PR in order to find a workaround about #2453 ...

- I had a little line in `.gitignore` to avoid committing some scripts in a `/scripts` folder ('cos I use some scripts to facilitate the reinstalls, the repo is [here](https://github.com/co-demos/udata-bash-scripts) ) ... but I scratched that ; 
- upgrade Bleach package : from `3.1.0` to `3.1.5` ; 
- a little switch in `markdown.py` when cleaning links, so to include the `mailto` scheme and adapt it in the `href` attribute... 
- added a little setting to allow mailto links or not if desired ... 
- as it's a first PR on uData for me I let (for this time only) all my logs/prints so you could see my debug process yourselves... I added the ìn progress` tag to the PR, and we could erase all logs from the PR once you've reviewed the code. WDYT ?

PR #2500, opened 2020-06-15
Fix internal links in markdown when not starting w/ slash

### LINKED ISSUE #2496: Some linkified urls are including the app server name
In a dataset's text, an url written in a markdown format is transformed into a `<a>` tag but its `href` attribute contains the app server name within the url... 

You can check an example of this behaviour here => [test Bleach dataset](https://www.data.gouv.fr/fr/datasets/dataset-test-1/)

## Detailed Description
 if you  write a text as : 
```
link 03 : [rte-france-03.com](rte-france-03.com)
```
you get on the client side something like : 
```
<p>link 03 : <a href="https://www.data.gouv.frrte-france-03.com">rte-france-03.com</a></p>
```
... which is wrong and not desired behaviour I guess : it shoud not contain `www.data.gouv.fr` before `rte-france-03.com` in the `href` attribute

## Possible Implementation
check PR #2486

abulte:
Fix #2496 

Close #2499 

Close #2486

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Fix internal links in markdown when not starting w/ slash [#2500](https://github.com/opendatateam/udata/pull/2500)
 
 ## 2.1.2 (2020-06-17)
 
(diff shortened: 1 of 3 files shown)


JulienParis:
I was just wondering if this needed to be mentionned in the change log ?

PR #2505, opened 2020-06-18
stats.app.etalab.studio

abulte:
Experimental integration of stats.app.etalab.studio (https://github.com/milesmcc/shynet).


abulte:
Bottom line: shynet is not powerful for our needs (not enough data / filtering capabilities).

Issue #2479, opened 2020-05-22
Implement PKCE via Authlib


abulte:
As noted in #2298, we're currently allowing implicit grant OAuth workflow for SPA apps. This workflow has been deprecated and replaced by PKCE. We should look into implementing PKCE on the server side via Authlib https://docs.authlib.org/en/stable/specs/rfc7636.html?highlight=#module-authlib.oauth2.rfc7636.

lepture:
Here is my change for adding PKCE in example-oauth2-server: https://github.com/authlib/example-oauth2-server/commit/a36b9e08270b12c82efe767f6792cab986c8a51e

PR #2510, opened 2020-07-06
Pkce implementation

quaxsze:
fix https://github.com/opendatateam/udata/issues/2479


udata/api/commands.py
```diff
@@ -59,3 +63,23 @@ def validate():
...
+
+    client = OAuth2Client.objects.create(
+        name='test-client',
+        owner=user,
+        redirect_uris=[uri]
+    )
+
+    click.echo(f'New OAuth client')
+    click.echo(f'Client ID {client.id}')
+    click.echo(f'Client secret {client.secret}')
```
JulienParis, line 85: just a suggestion : to echo the uri list the new client can use, so they could remember to add it to its app's oauth config ....

    ```click.echo(f'Client redirect_uris {client.redirect_uris}')```

udata/api/commands.py
```diff
@@ -59,3 +63,23 @@ def validate():
...
+def create_oauth_client(user_email, uri):
+    '''Creates an OAuth2Client instance in DB'''
+    user = User.objects(email=user_email).first()
+    if user is None:
+        exit_with_error('No matching user to email')
+
+    client = OAuth2Client.objects.create(
+        name='test-client',
+        owner=user,
+        redirect_uris=[uri]
```
JulienParis, line 80: suggestion to enable adding several uris in the cli instead of just one (but not sure about chosing `+`as main separator) : 
```        redirect_uris=uri.split('+')```

File: udata/api/commands.py [hunk 3]
```diff
@@ -59,3 +63,23 @@ def validate():
         success('API specifications are valid')
     except schemas.SchemaValidationError as e:
         exit_with_error('API specifications are not valid', e)
+
+
+@grp.command()
+@click.option('-u', '--user-email', help='User\'s email')
+@click.option('--uri', default='http://localhost:8080/login', help='Client\'s redirect uri')
```
JulienParis, line 70: suggestion about the uri declaration from the cli : allow declaring several urls by passing them as follow ? 

```@click.option('--uri', default='http://localhost:8080/login+http://localhost:8080', help='Client\'s redirect uri')```

Comment (JulienParis) [reply udata/api/commands.py:70]: or something like (cf [click docs](https://click.palletsprojects.com/en/7.x/options/#multiple-options):
```@click.option('--uri', default='http://localhost:8080/login', help='Client\'s redirect uri', multiple=True)```

**Comment by JulienParis** [comment]
just some notes I used to remember how to use the new oauth endpoints : 

### update udata libs 

```bash
cd ../udata
source venv/bin/activate
pip install --update authlib
pip install -e .
```

### create an oauth client

```bash
cd .../udata
source venv/bin/activate
udata api create-oauth-client -u my.udata.user.email@mail.com --uri=http://localhost:8080/login

# it should echo yout out your new client ID / client secret, something like
# New OAuth client
# Client ID 5f0blablablablabla1d
# Client secret eZvqvelbDOOblablablaVnX5nniOqGWFG7

# if your're developping locally
export AUTHLIB_INSECURE_TRANSPORT=1
```



```

File: udata/api/commands.py (hunk 3 above)
Comment (quaxsze) [reply udata/api/commands.py:70]: `default='http://localhost:8080/login+http://localhost:8080'` not sure what kind of syntax this is looks very JS to me 😄 
`default=['http://localhost:8080/login']` will not work click is not designed for this it would need to implement a custom argument type.
I will use `'http://localhost:8080/login http://localhost:8081 http://localhost:8082'` as it is easy to split a string by whitespace.
`default='http://localhost:8080/login+http://localhost:8080'` seems likewise but not sure what kind of syntax this is,  looks very JS to me 😄

Comment (quaxsze) [reply udata/api/commands.py:70]: Btw you already have a comment for this just above

**Comment by JulienParis** [comment]
note for later, I'm having this for now when posting on `/token`

```
Access to fetch at 'http://dev.local:7000/oauth/token' from origin 'http://localhost:8080' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

File: udata/api/commands.py (hunk 3 above)
Comment (abulte) [reply udata/api/commands.py:70]: https://click.palletsprojects.com/en/7.x/options/#multiple-options ?

Comment (abulte) [reply udata/api/commands.py:70]: (like Julien suggested actually)

File: udata/api/commands.py [hunk 4]
```diff
@@ -59,3 +63,24 @@ def validate():
...
+def create_oauth_client(user_email, uri):
+    '''Creates an OAuth2Client instance in DB'''
+    user = User.objects(email=user_email).first()
+    if user is None:
+        exit_with_error('No matching user to email')
+
+    client = OAuth2Client.objects.create(
+        name='test-client',
```
abulte, line 78: make this an option too

udata/api/oauth2.py
```diff
@@ -70,7 +72,7 @@ class OAuth2Client(ClientMixin, db.Datetimed, db.Document):
                           thumbnails=[150, 25])
 
     redirect_uris = db.ListField(db.StringField())
-    scopes = db.ListField(db.StringField(), default=['default'])
+    scope = db.StringField(default='')
```
abulte, line 75: Any particular reason for this change (aside that we don't use any scopes)?

udata/api/oauth2.py
```diff
@@ -79,6 +81,9 @@ class OAuth2Client(ClientMixin, db.Datetimed, db.Document):
         'collection': 'oauth2_client'
     }
 
+    def get_client_id(self):
```
abulte, line 84: `client_id` property below not good enough?

udata/api/oauth2.py
```diff
@@ -170,44 +154,81 @@ def __str__(self):
...
+        return self.scope
 
     def get_expires_in(self):
         return self.expires_in
 
     def get_expires_at(self):
         return (self.created_at - EPOCH).total_seconds() + self.expires_in
 
-    def is_refresh_token_expired(self):
+    def get_client_id(self):
```
abulte, line 165: same as above I guess

udata/api/oauth2.py
```diff
@@ -287,8 +317,6 @@ def authorize(*args, **kwargs):
         else:
             grant_user = None
         return oauth.create_authorization_response(grant_user=grant_user)
-    else:
-        abort(405)
```
abulte, line 291: Why remove this? Still seems useful, is someone sends a PUT or whatever

udata/api/oauth2.py
quaxsze: It's basically a copy of the way this is handled in authlib doc

udata/api/oauth2.py
quaxsze: We have to implement this method, the PKCE flow requires it. An execption `NotImplemented` is raised otherwise.

udata/api/oauth2.py
quaxsze: I do not see the use of it, as the `methods=['GET', 'POST']` in `@blueprint.route('/authorize', methods=['GET', 'POST'])` should handle it already.

udata/api/oauth2.py
abulte: use `get_client_id` in `client_id` then.

udata/api/oauth2.py
```diff
@@ -255,17 +282,17 @@ def request_invalid(self, request):
...
-        # TODO: return token.revoked
-        return False
+        return token.revoked
 
 
 @blueprint.route('/token', methods=['POST'], localize=False, endpoint='token')
 @csrf.exempt
 def access_token():
-    return oauth.create_token_response()
+    return oauth.create_token_response(request)
```
JulienParis, line 291: I think we should pre-chew the client's request before `create_token_response` so we could allow json data AND form data. Otherwise the flask implementation of Authlib will only accept form data and will ignore json data. That could be problematic to understand from a front point of view and counter-intuitive for third-party uses.

### proposal : 

```
from authlib.oauth2.rfc6749 import OAuth2Request
from authlib.integrations.flask_helpers import create_oauth_request

...

@blueprint.route('/token', methods=['POST'], localize=False, endpoint='token')
@csrf.exempt
def access_token():
    has_form = bool(request.form)
    req = create_oauth_request(request, OAuth2Request, use_json=not has_form)
    return oauth.create_token_response(request=req)
```

... that said if we want 3rd parties to send auth data only as form data (and not json) we could use `return oauth.create_token_response(request)` alone, but I think we should at least update the udata's docs and precise it in an "oauth2 / pkce" section...

### references / explanations from authlib : 

I tested various way before proposing this, because `create_oauth_request` was already integrated in the Flask's `AuthorizationServer` class, and it seemed weird indeed... But after exploring the authlib code I couldn't see where json data could be proceeded : 

- `AuthorizationServer` from Flask implementation does not contain/overwrite `create_token_response` : https://github.com/lepture/authlib/blob/master/authlib/integrations/flask_oauth2/authorization_server.py
- So the `create_token_response` function in Flask implementation's `AuthorizationServer` inherits from `rfc6749`'s `AuthorizationServer` and should pass on a request arg : https://github.com/lepture/authlib/blob/782a0fced780849418dc2a869528d10387e24b65/authlib/oauth2/rfc6749/authorization_server.py#L203
- the request is created from the class's own `create_oauth_request`, but never checks if json data is contained or not : https://github.com/lepture/authlib/blob/782a0fced780849418dc2a869528d10387e24b65/authlib/integrations/flask_oauth2/authorization_server.py#L110
- note that the json data from the client's request is only injected into an `OAuth2Request` instance if `use_json == true`, and none of the calls I've seen so far in the Flask implementation trigger this kwarg (but perhaps I've missed something ?) : https://github.com/lepture/authlib/blob/master/authlib/integrations/flask_helpers.py

quaxsze: As said in this [issue](https://github.com/lepture/authlib/issues/156), the RFC 6749 specifies the format 'application/x-www-form-urlencoded'

PR #2512, opened 2020-07-14
Admin: add schema field to resource

AntoineAugusti:
image

- Link data.gouv.fr to the schema catalog of schema.data.gouv.fr
- Add a schema field to the resource model
- Let producers select a schema from a list on resources

I left some comments/questions at various places.

## To discuss
- Configuration for the schema catalog URL
- The schema catalog of schema.data.gouv.fr only lists TableSchema schemas for now, should we change it? It's because we needed this endpoint for Validata, therefore it's made to include only TableSchema schemas. This can change.


js/models/schemas.js
```diff
@@ -0,0 +1,14 @@
+import {List} from 'models/base';
+
+
+export class Schemas extends List {
+    constructor(options) {
+        super(options);
+        this.$options.ns = 'datasets';
+        this.$options.fetch = 'schemas';
+        this.sorted = 'label';
```
AntoineAugusti, line 9: Sort schemas in alphabetical order for the frontend

udata/core/dataset/api.py
```diff
@@ -550,3 +554,13 @@ def get(self):
         '''List all resource types'''
         return [{'id': id, 'label': label}
                 for id, label in RESOURCE_TYPES.items()]
+
+
+@ns.route('/schemas/', endpoint='schemas')
+class SchemasAPI(API):
+    @api.doc('schemas')
+    @api.marshal_list_with(schema_fields)
+    @cache.cached(timeout=60*5)
```
AntoineAugusti, line 563: I wanted to add a configuration to specify a different TTL but I didn't manage to. It seems `current_app.config` isn't available here, not in the app context.

Any way to work around this? Is it needed?

udata/core/dataset/forms.py
```diff
@@ -79,6 +80,10 @@ class BaseResourceForm(ModelForm):
         _('Publication date'),
         description=_('The publication date of the resource'))
     extras = fields.ExtrasField()
+    schema = fields.StringField(
+        _('Schema'),
+        # choices=[schema['id'] for schema in ResourceSchemas.get()],
```
AntoineAugusti, line 85: Can't add this line here as `ResourceSchemas.get()` needs `current_app.config`. It isn't available there as it's not in the app context when called from here.

How to get around this?

udata/core/dataset/models.py
```diff
@@ -747,6 +750,23 @@ def from_community(self):
         return True
 
 
+class ResourceSchemas(object):
+    @staticmethod
+    def get():
```
AntoineAugusti, line 755: No caching for this method, caching is done at the API level

udata/settings.py
```diff
@@ -437,6 +437,10 @@ class Defaults(object):
     ARCHIVE_COMMENT_USER_ID = None
     ARCHIVE_COMMENT_TITLE = _('This dataset has been archived')
 
+    # Schemas parameters
+    ####################
+    SCHEMA_CATALOG_URL = None
```
AntoineAugusti, line 442: Is it a sensible default? Should I change this?

udata/tests/api/test_datasets_api.py
```diff
@@ -1325,3 +1326,27 @@ def test_resource_types_list(self):
         response = self.get(url_for('api.resource_types'))
         self.assert200(response)
         self.assertEqual(len(response.json), len(RESOURCE_TYPES))
+
+
+@pytest.mark.usefixtures('clean_db')
+class DatasetSchemasAPITest:
```
AntoineAugusti, line 1332: Wrote tests for my new endpoint using the `pytest` syntax. Other classes inherited from `APITestCase`, not compatible with `pytest`.

udata/tests/api/test_datasets_api.py
```diff
@@ -1325,3 +1326,27 @@ def test_resource_types_list(self):
...
+    modules = ['core.dataset']
```
AntoineAugusti, line 1333: I discovered adding this was key to get the routing of `api.get` working, can you explain why?

js/components/dataset/resource/form.vue
```diff
@@ -243,7 +244,23 @@ export default {
             }]
         },
         fields() {
-            return this.generic_fields.concat(this.file_fields);
+            return this.generic_fields.concat(this.schema_field).concat(this.file_fields);
```
AntoineAugusti, line 247: The schema field will not be added to the form if the schema catalog is empty / is not set.

udata/core/dataset/search.py
```diff
@@ -83,7 +83,8 @@ class Meta:
     resources = Object(properties={
```
AntoineAugusti, line 83: Commenting for the whole file.

I'm not sure to understand how this class works. I've tried to look at various fields and added the necessary lines. 
I don't understand if I need to add (in order of changes, top to bottom):
- a property
- a facet
- update the `serialize` method

udata/core/dataset/api.py
abulte: How about `SCHEMA_CACHE_DURATION = current_app.config.get()` at the module level and then use `SCHEMA_CACHE_DURATION`? I think it should work. https://github.com/opendatateam/udata/blob/master/udata/core/dataset/preview.py#L58

udata/core/dataset/forms.py
abulte: same as above, build the choices list at the module level.

udata/tests/api/test_datasets_api.py
abulte: It make the dataset module (and thus it's API) available for the test class. By default we try not to load too much stuff, in order to have faster tests.

udata/core/dataset/search.py
abulte: I'm not sure either :-/ let's keep this for a future PR.

udata/core/dataset/models.py
```diff
@@ -747,6 +750,23 @@ def from_community(self):
...
+        endpoint = current_app.config.get('SCHEMA_CATALOG_URL')
+        if endpoint is None:
+            return []
+
+        r = requests.get(endpoint)
+        r.raise_for_status()
+
+        result = []
+        for schema in r.json()['schemas']:
+            result.append({'id': schema['name'], 'label': schema['title']})
```
abulte, line 765: I guess a list comprehension would work here

udata/tests/api/test_datasets_api.py
```diff
@@ -1325,3 +1326,27 @@ def test_resource_types_list(self):
...
+
+    def test_dataset_schemas_api_list(self, app, api, rmock):
+        app.config['SCHEMA_CATALOG_URL'] = 'https://example.com/schemas'
```
abulte, line 1336: use `@pytest.mark.options` decorator for this

udata/tests/dataset/test_dataset_model.py
```diff
@@ -413,3 +415,18 @@ def test_multiple_strings(self):
         found = License.guess('should not match', license.id)
         assert isinstance(found, License)
         assert license.id == found.id
+
+
+class ResourceSchemasTest:
+    def test_resource_schemas_get(self, app, rmock):
+        app.config['SCHEMA_CATALOG_URL'] = 'https://example.com/schemas'
```
abulte, line 422: use decorator, same as above

udata/tests/dataset/test_dataset_model.py
```diff
@@ -413,3 +415,18 @@ def test_multiple_strings(self):
...
+        rmock.get('https://example.com/schemas', json={
+            'schemas': [{"name": "etalab/schema-irve", "title": "Schéma IRVE"}]
+        })
+
+        assert ResourceSchemas.get() == [{"id": "etalab/schema-irve", "label": "Schéma IRVE"}]
+
+    def test_resource_schemas_get_no_catalog_url(self, app):
+        app.config['SCHEMA_CATALOG_URL'] = None
```
abulte, line 430: use decorator

udata/tests/api/test_datasets_api.py
```diff
@@ -1325,3 +1326,27 @@ def test_resource_types_list(self):
...
+            'schemas': [{"name": "etalab/schema-irve", "title": "Schéma IRVE"}]
+        })
+
+        response = api.get(url_for('api.schemas'))
+
+        assert200(response)
+        assert response.json == [{"id": "etalab/schema-irve", "label": "Schéma IRVE"}]
+
+    def test_dataset_schemas_api_list_no_catalog_url(self, app, api):
+        app.config['SCHEMA_CATALOG_URL'] = None
```
abulte, line 1347: use decorator

udata/core/dataset/models.py
```diff
@@ -747,6 +750,23 @@ def from_community(self):
...
 
+class ResourceSchemas(object):
+    @staticmethod
+    def get():
+        endpoint = current_app.config.get('SCHEMA_CATALOG_URL')
+        if endpoint is None:
+            return []
+
+        r = requests.get(endpoint)
+        r.raise_for_status()
```
abulte, line 761: I think we need a smarter caching strategy and exception handling.

Something like that https://github.com/etalab/udata-gouvfr/pull/486/files#diff-1c9abf3241dab91b0e60c554f8bcd67aR72.

This would lead to removing the cache on the API and handle it here.

udata/core/dataset/models.py
```diff
@@ -747,6 +750,23 @@ def from_community(self):
         return True
 
 
+class ResourceSchemas(object):
```
abulte, line 753: our model classes are usually singular -> ResourceSchema

abulte:
> The schema catalog of schema.data.gouv.fr only lists TableSchema schemas for now, should we change it? It's because we needed this endpoint for Validata, therefore it's made to include only TableSchema schemas. This can change.

No need to change that for now. I don't think we'll do anything very interesting with other schemas for quite some time.

udata/core/dataset/forms.py
```diff
@@ -79,6 +80,12 @@ class BaseResourceForm(ModelForm):
         _('Publication date'),
         description=_('The publication date of the resource'))
     extras = fields.ExtrasField()
+    schema = fields.SelectField(
+        _('Schema'),
+        choices=[schema['id'] for schema in ResourceSchema.objects()],
```
AntoineAugusti, line 85: Do I need to explicitly add a `None` to the list? Tests pass without it.

udata/core/dataset/api.py
```diff
@@ -550,3 +551,18 @@ def get(self):
...
+    @api.doc('schemas')
+    @api.marshal_list_with(schema_fields)
+    def get(self):
+        '''List all available schemas'''
+        try:
+            # This method call is cached as it makes HTTP requests
+            return ResourceSchema.objects()
+        except LookupError:
+            abort(503, description='No schemas in cache and endpoint unavailable')
+        except ValueError:
```
AntoineAugusti, line 567: It was difficult to pick appropriate exceptions to raise, I'm open to suggestions

udata/core/dataset/models.py
```diff
@@ -747,6 +756,45 @@ def from_community(self):
         return True
 
 
+class ResourceSchema(object):
+    @staticmethod
+    @cache.memoize(timeout=SCHEMA_CACHE_DURATION)
+    def objects():
```
AntoineAugusti, line 762: Heavily inspired by the static pages' method you linked me to

AntoineAugusti:
Thanks for your guidance and your help @abulte!

I:
- tackled your PR comments
- reverted the commit introducing indexing - it'll be done later
- added a double caching mechanism to fetch schemas from the schema catalog
- cached at the model level instead of the API level
- enforced an enumeration for the `schema` property at the form level

udata/core/dataset/forms.py
abulte: No I don't think so.

udata/core/dataset/api.py
abulte: I wouldn't try to use base exceptions that are not made for that.

Either:
- use custom exceptions
- use the return value of the function (`[]` vs `None` for example) if it's enough

AntoineAugusti: I added custom exceptions 🔧

AntoineAugusti:
Added custom exceptions as you suggested @abulte. Thanks for your review.

udata/core/dataset/models.py
```diff
@@ -747,6 +759,45 @@ def from_community(self):
...
+            response = requests.get(endpoint, timeout=5)
+            # do not cache 404 and forward status code
+            if response.status_code == 404:
+                raise SchemasCatalogNotFoundException(f'Schemas catalog does not exist at {endpoint}')
+            response.raise_for_status()
+        except requests.exceptions.RequestException as e:
+            log.exception(f'Error while getting schema catalog from {endpoint}')
+            content = cache.get(cache_key)
+        else:
+            schemas = response.json()['schemas']
```
abulte, line 788:
```suggestion
            schemas = response.json().get('schemas', [])
```

AntoineAugusti:
I had to do an [extra commit](https://github.com/opendatateam/udata/pull/2512/commits/49ac4a4fa4163f053798ac286fe733a2c5a1211d) this morning to fix `dataset/forms.py`. When starting my local server, I had a stracktrace regarding the absence of the app context in this file, because of `choices=[schema['id'] for schema in ResourceSchema.objects()]` which needs the app context under the hood. I guess the app context is not available yet, hence we can't access the current config at this point.

I switched to a string field with a validator. The SelectField relies on a list of options, through the `choices` keyword. You can't leave it empty, that's why I switched to the String class.

This pattern is used elsewhere in the very same file
https://github.com/opendatateam/udata/blob/32b09032f3b60ef26ae105ccdbce4591dae7dc6a/udata/core/dataset/forms.py#L33-L48

I made sure the validator was working by editing the HTML
image

PR #2522, opened 2020-08-10
Remove theme, aka udata3

abulte:
This PR removes the default theme from udata, i.e. all HTML/CSS/JS assets related to frontend. The admin stays here (for now).

The theme will be totally contained in `udata-gouvfr`, cf https://github.com/etalab/udata-gouvfr/pull/492.


quaxsze:
Bare metal email preview after removing theme logic:
image

PR #2527, opened 2020-08-25
Show stringified extras in dataset modal

abulte:
Assuming `JSON.stringify` is safe enough since this can be a user provided value 🧐. 

## Before

Capture d’écran 2020-08-25 à 23 25 47

## After

Capture d’écran 2020-08-25 à 23 28 48

### DIFF
--- a/js/front/dataset/details-modal.vue
+++ b/js/front/dataset/details-modal.vue
@@ -35,7 +35,7 @@
                 <tbody>
                     <tr v-for="extra in dataset.extras">
                         <td>{{ extra.name }}</td>
-                        <td>{{ extra.value }}</td>
+                        <td>{{ stringify(extra.value) }}</td>
                     </tr>
                 </tbody>
             </table>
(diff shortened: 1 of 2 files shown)


abulte:
@AntoineAugusti (can't assign you a review)

js/front/dataset/details-modal.vue
```diff
@@ -57,6 +57,14 @@ export default {
     props: {
         dataset: Object
     },
-    components: {Modal}
+    components: {Modal},
+    methods: {
+        stringify(value) {
+            if (typeof value === 'string' || value instanceof String) {
+                return value;
+            }
```
AntoineAugusti, line 65: We can have floats and integers if I remember. Did you try testing if it's an object?

It seems it's not pretty to write the check https://stackoverflow.com/a/16608045

js/front/dataset/details-modal.vue
```diff
@@ -57,6 +57,14 @@ export default {
...
+            return JSON.stringify(value);
```
AntoineAugusti, line 66: From my understanding this + `{{ }}` is XSS safe 👮‍♀️

ThibaudDauce:
I think it's done :-)

Issue #2530, opened 2020-09-07
Removing a resource manually won't remove its file


abulte:
For example in `udata shell`:

```
# find a dataset with local files
dataset = ...
dataset.resources = []
dataset.save()
```

won't remove the resources' files.

abulte:
Solution: use the new method `remove_resource`.

```
dataset.remove_resource(resource)
dataset.save()
```

PR #2534, opened 2020-09-11
Add specific method for resource removal

quaxsze:
fix https://github.com/opendatateam/udata/issues/2530

### DIFF
--- a/udata/core/dataset/models.py
+++ b/udata/core/dataset/models.py
@@ -674,8 +674,6 @@ class Dataset(WithMetrics, BadgeMixin, db.Owned, db.Document):
             storages.resources.delete(resource.fs_filename)
 
         self.resources.remove(resource)
-        self.last_modified = datetime.now()
-        self.save()
 
     @property
     def community_resources(self):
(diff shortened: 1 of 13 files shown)


udata/core/dataset/models.py
```diff
@@ -667,6 +668,15 @@ def update_resource(self, resource):
         self.reload()
         post_save.send(self.__class__, document=self)
 
+    def remove_resource(self, resource):
+        # Deletes resource's file from file storage
+        if resource.fs_filename is not None:
+            storages.resources.delete(resource.fs_filename)
+
+        self.resources.remove(resource)
+        self.last_modified = datetime.now()
```
abulte, line 677: Keep `last_modified` out of this fn, like here https://github.com/opendatateam/udata/blob/master/udata/core/dataset/api.py#L267

When using the CLI we usually do not want to bump `last_modified`.

I'm not sure if we should keep the `save` here. It would probably be more foolproof. Maybe look at other examples.

abulte (commented):
Add tests 😬

udata/core/dataset/models.py
quaxsze: 👍

quaxsze: Regarding the tests the behavior and the storage is already tested in api tests. Do we want a specific model level test?

abulte: Maybe test the function directly, just to make sure.

abulte (approved):

Issue #2543, opened 2020-10-01
Community resource API: use dataset info when uploading one


abulte:
When using https://doc.data.gouv.fr/api/reference/#/datasets/upload_new_community_resource, eg `POST https://demo.data.gouv.fr/api/1/datasets/5f757d2d70762c2f015e466c/upload/community/`, the dataset id is not used when creating the `CommunityResource`. It is orphaned until we call `PUT https://demo.data.gouv.fr/api/1/datasets/community_resources/f55ddaff-755c-4ae4-b634-e02bb69d664b/` with `{dataset: <id>}`.

I think we could use the dataset id from the URL on the first call.

abulte:
Don't forget to test if the subsequent call (ie PUT with dataset id) still works, for compatibility's sake.

PR #2546, opened 2020-10-02
Add dataset into community resource at its creation

quaxsze:
fix https://github.com/opendatateam/udata/issues/2543


abulte (commented):
Write a test for the POST endpoint

abulte (approved):

Issue #2541, opened 2020-10-01
Resources updated from API: old file is not deleted


abulte:
1. upload a resource (file1) through the API
2. upload a new file (file2) on this resource
3. file1 is still on storage

Example: https://demo.data.gouv.fr/fr/admin/dataset/5f757d2d70762c2f015e466c/resource/c8554dd0-452f-41af-bb42-d39371385e0c

```
udata@demo-01:~$ ls fs/resources/from-api/*/*
fs/resources/from-api/20201001-085527/lamarche-data.gouv.xlsx  fs/resources/from-api/20201001-085634/donnees-hospitalieres-covid19-2020-05-06-19h00.csv
```

abulte:
Also true for CommunityResources:

https://demo.data.gouv.fr/fr/datasets/from-api/#resource-community-f55ddaff-755c-4ae4-b634-e02bb69d664b

```
udata@demo-01:~$ ls fs/resources/from-api/*/*.gpx
fs/resources/from-api/20201001-091713/ev3-blois-tours.gpx  fs/resources/from-api/20201001-095701/ev3-orleans-blois.gpx
```

PR #2548, opened 2020-10-05
Delete old file when updating resources and community resources

quaxsze:
fix https://github.com/opendatateam/udata/issues/2541


udata/core/dataset/api.py
```diff
@@ -310,6 +310,8 @@ def post(self, dataset, rid):
         ResourceEditPermission(dataset).test()
         resource = self.get_resource_or_404(dataset, rid)
         infos = self.handle_upload(dataset)
+        if resource.fs_filename is not None:
```
abulte, line 313: maybe we should put that as the very latest operation in the route, in case anything else fails we don't want to be in an inconsistent state (same for community resources).

abulte (approved):
Nice!

PR #2549, opened 2020-10-05
Update Basemaps URL and attributions

skgsergio:
# Summary
This PR updates basemaps URL to the new CDN URL and fixes the attributions to match current requirements.

## Full context
We've recently changed our CDN and we have a plan mid-long term (depends on how quick we manage to stop all the traffic there) to deprecate our `.global.ssl.fastly.net` domains.


ThibaudDauce:
Don't see these env in the current version of `udata`. Feel free to open an issue or a new PR if it's still a problem.

Issue #2560, opened 2020-10-27
DCAT catalog: encoding issue


abulte:
> Thanks for pointing us to the real DCAT-AP endpoint, we really appreciate it.
>
> Unfortunately, there is a unrecognized character in your data that prevent us from harvesting more then ~15.000 datasets. You can proof this by opening the following link in your browser. The browser will complain about wrong XML and point you to the right direction: https://www.data.gouv.fr/catalog.rdf?page=148&page_size=100
>
> the problematic character is hidden in the error message, but when you copy paste the string somewhere you can see it. Or you open the raw data of the page and search for the string. The line numbers are differently, though.

PR #2562, opened 2020-10-30
Escape XML's forbidden chars

quaxsze:
fix https://github.com/opendatateam/udata/issues/2560


quaxsze:
Replacement char is set to `?`. Any ideas of a better one?

Issue #2561, opened 2020-10-28
CLI user creation raises exception because of i18n


quaxsze:
When using `udata user create ` and submitting a password which does not match the requirements, this exception is raised:

`TypeError: sequence item 1: expected str instance, _LazyString found`

This happens because the error message of the password complexity checker is a translated lazy string.

A way of giving the CLI the right str should be found.

quaxsze:
Fixed by https://github.com/opendatateam/udata/pull/2567

PR #2566, opened 2020-11-03
Fix i18n in CLI for password complexity checker

quaxsze:
fix https://github.com/opendatateam/udata/issues/2561


abulte (commented):
I'm not sure this will work. AFAIK `lazy_gettext` is used to translate on the fly, based on the currently selected locale (eg user/request preference). This would break language switching in the form. Cf https://pypi.org/project/speaklater/.

quaxsze:
Seems to be fix by https://github.com/opendatateam/udata/pull/2567

PR #2568, opened 2020-11-06
WIP: Resource's url instead of permalink

quaxsze:


ThibaudDauce:
Is this still a change we want to make? @maudetes ? We still use `resource.latest` in the current version.

maudetes:
We prefer using the stable resource.latest for now :)

PR #2570, opened 2020-11-08
Fix error when retrieving GeoZones

mdamien:
The issue is that the script try to write to `/udata/fs/tmp/` but the `tmp` directory does not exist

Related issue: https://github.com/opendatateam/docker-udata/issues/197

I'm not sure it's the best place to ensure the `/tmp/` directory is there, maybe directly at the initialization phase in `udata/core/storages/__init__.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Escaping XML's forbidden characters [#2562](https://github.com/opendatateam/udata/pull/2562)
 - Ignore pattern feature for linkchecker [#2564](https://github.com/opendatateam/udata/pull/2564)
 - Fix TypeError when creating a superuser with an incorrect password [#2567](https://github.com/opendatateam/udata/pull/2567)
+- Fix error when retrieving GeoZones [#2570](https://github.com/opendatateam/udata/pull/2570)
 
 ## 2.4.0 (2020-10-16)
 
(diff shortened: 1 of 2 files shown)


quaxsze:
This solution might work but creates a specific behavior which differs from the rest of the storages.
Flask-fs uses `ensure_path` method on its `write()` and `save()` methods. The flask-fs way would be to download the file using the chunk storage and use its `save()` method.
Another simpler way would be to use python's temporary file feature instead of using a Flask-fs storage.

mdamien:
I could do:

```
        log.info('Downloading GeoZones bundle: %s', filename)
        with tmp.open(GEOZONE_FILENAME, 'w') as f:
            pass
        filename, _ = urlretrieve(filename, tmp.path(GEOZONE_FILENAME))
```

So it works with all backends

quaxsze:
It's actually even more hacky than the previous one :sweat_smile:
When I mentioned other storages, I meant that they download files using the chunk way and the `ensure_path` is implicitly called within the `save` function.
The solution is to either use a storage provided by Flask-FS and to proceed the way the librairy is designed for, or if this is over complicated for the wanted result, to access the system `tmp` by using python`s tempfile instead of using flask-fs.

mdamien:
@quaxsze ahah, this time I pushed something you will like I think, downloading the chunk way with `urlopen`

quaxsze:
Sorry my previous comment was not very clear.
When I mentioned the chunks way I was thinking of using chunk's storage like in the [handle_upload](https://github.com/opendatateam/udata/blob/267555b6ab7c2f375f619ccec2f9220e23934ab4/udata/core/storages/api.py#L129) function.
It might seem a bit over complicated that is why mentioned accessing the system tmp by using python`s tempfile instead of using flask-fs.

mdamien:
@quaxsze pushed a simple solution with `tempfile.NamedTemporaryFile`, let me know if it's what you're looking for

ThibaudDauce:
I don't think this bug exists anymore. Feel free to open an issue or a new PR if it's still the case.

Issue #2361, opened 2019-12-04
Posts returned by the API contain unpublished ones


davidbgk:
## Detailed Description

When I get Posts from the API, it contains unpublished posts, like `"content": "plop"`:
https://www.data.gouv.fr/api/1/posts/?page_size=10

## Context

We were investigating how to retrieve associated datasets to a post (not really relevant for the issue)

## Possible Implementation

See:
https://github.com/opendatateam/udata/blob/master/udata/core/post/api.py#L78-L79

PR #2578, opened 2020-11-23
Prevent unpublished posts to be retrieved

quaxsze:
Fix https://github.com/opendatateam/udata/issues/2361

### DIFF
--- a/udata/core/post/api.py
+++ b/udata/core/post/api.py
@@ -72,7 +72,7 @@ class PostsAPI(API):
     def get(self):
         '''List all posts'''
         args = parser.parse_args()
-        return (Post.objects.filter(published__ne=None).order_by(args['sort'])
+        return (Post.objects.published().order_by(args['sort'])
                             .paginate(args['page'], args['page_size']))
 
     @api.doc('create_post')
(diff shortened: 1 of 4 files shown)


udata/core/post/api.py
```diff
@@ -72,7 +72,7 @@ class PostsAPI(API):
     def get(self):
         '''List all posts'''
         args = parser.parse_args()
-        return (Post.objects.order_by(args['sort'])
+        return (Post.objects.filter(published__ne=None).order_by(args['sort'])
```
abulte, line 75: no `visible()` helper on Post model? Maybe it's time to add one :-)

quaxsze: actually there is one already :)

udata/core/post/factories.py
```diff
@@ -7,6 +7,15 @@
 from .models import Post
 
 
+class UnpublishedPostFactory(ModelFactory):
```
abulte, line 10: I'm not sure you need this, can't you use PostFactory with `published=None`?

abulte (approved):

udata/core/post/factories.py
quaxsze: Yes my bad

PR #2583, opened 2020-12-15
Add RDF for organizations

quaxsze:


quaxsze:
What is the use of `wants_rdf` like in the dataset's [view](https://github.com/opendatateam/udata/blob/9c269676b99abedeb3e138bf94ff2a7245ee9f3a/udata/core/dataset/views.py#L87) and of `rdf_links` ? Should we use them for organization's rdf view ?

udata/core/organization/views.py
```diff
@@ -93,6 +98,23 @@ def get_context(self):
...
 
+@blueprint.route('/<org:org>/rdf', localize=False)
+def rdf(org):
+    '''Root RDF endpoint with content negociation handling'''
+    format = RDF_EXTENSIONS[negociate_content()]
+    url = url_for('organizations.rdf_format', org=org.id, format=format)
+    return redirect(url)
+
+
+@blueprint.route('/<org:org>/rdf.<format>', localize=False)
```
abulte, line 109:
```suggestion
@blueprint.route('/<org:org>/catalog.<format>', localize=False)
```

udata/core/organization/views.py
```diff
@@ -93,6 +98,23 @@ def get_context(self):
...
+def rdf_format(org, format):
```
abulte, line 110: Use the logic from here https://github.com/opendatateam/udata/blob/f84280262dd78b85fb4793d8702c502e6c96a3d8/udata/core/site/views.py#L269 to output the datasets of the org.

udata/core/organization/views.py
```diff
@@ -93,6 +98,23 @@ def get_context(self):
...
+    if not EditOrganizationPermission(org).can() and org.deleted:
```
abulte, line 111: strange permission check for a read only view 🤔

udata/core/organization/views.py
```diff
@@ -93,6 +98,23 @@ def get_context(self):
         return context
 
 
+@blueprint.route('/<org:org>/rdf', localize=False)
```
abulte, line 101:
```suggestion
@blueprint.route('/<org:org>/catalog', localize=False)
```

abulte (approved):

abulte (changes requested):
oops previous review was not an approval :-)

udata/core/organization/rdf.py
```diff
@@ -28,4 +31,52 @@ def organization_to_rdf(org, graph=None):
...
+    org_catalog_url = url_for('organizations.rdf_catalog', org=org.id, _external=True)
+
+    catalog = graph.resource(URIRef(org_catalog_url))
+    catalog.set(RDF.type, DCAT.Catalog)
+    catalog.set(DCT.publisher, organization_to_rdf(org, graph))
+
+    for dataset in datasets:
+        catalog.add(DCAT.dataset, dataset_to_rdf(dataset, graph))
+    
+    if isinstance(datasets, Paginable):
```
abulte, line 49: This can't be factorized with the main catalog?

quaxsze: The dataset part is not exactly the same the main catalog add the owner or the organization to the dataset:
```
for dataset in datasets:
    rdf_dataset = dataset_to_rdf(dataset, graph)
    if dataset.owner:
        rdf_dataset.add(DCT.publisher, user_to_rdf(dataset.owner, graph))
    elif dataset.organization:
        rdf_dataset.add(DCT.publisher, organization_to_rdf(dataset.organization, graph))
    catalog.add(DCAT.dataset, rdf_dataset)
```

For the pagination part I thought about it but as the making of urls arguments depends on the model, for instance
```
first_url = url_for('organizations.rdf_catalog_format', org=org.id, page=1, **kwargs)
```
It might fall into something overly engineered and loss of the code readability to make it generic.
What do you think?

abulte (approved):
Maybe some doc to update? I don't know if the RDF/DCAT part is already documented somewhere.

PR #2586, opened 2021-01-12
Add resource's description size limit

quaxsze:


abulte (commented):
Maybe we need to factorize those limits (in https://github.com/opendatateam/udata/pull/2585/files too), and use two, maybe three distinct values. This would add a bit more coherence.

abulte (approved):

PR #2587, opened 2021-01-25
Fix organization's catalog view

quaxsze:


abulte (commented):
You should probably write a test to catch that previous error.

PR #2591, opened 2021-03-02
Admin wording and UX small enhancement

quaxsze:


abulte:
While we're at it, can we add an "Edit" button or link for admins that goes from the profile page https://demo.data.gouv.fr/fr/users/amalia-ouriachi-2/ to the admin page for this user? It would be great for moderation 💪

abulte:
Looks great on demo, thanks, I still have to do a code review.

js/components/user/profile.vue
```diff
@@ -20,13 +20,13 @@
...
 <div>
-<box :title="title" icon="user" boxclass="user-profile-widget">
-    <h3>{{user.fullname}}</h3>
+<box :title="user.fullname" icon="user" boxclass="user-profile-widget">
         <div class="profile-body">
             <image-button :src="user | avatar_url 100" :size="100" class="avatar-button"
                 :endpoint="endpoint" :editable="can_edit">
             </image-button>
         <div v-markdown="user.about"></div>
+        <h4>{{user.email}}</h4>
```
abulte, line 29: `v-if`?

udata/auth/helpers.py
```diff
@@ -0,0 +1,10 @@
+from flask import request
+from flask_security import current_user
+
+
+def request_is_admin() -> bool:
```
abulte, line 5: 😎

abulte, line 5:
```suggestion
def request_is_admin_or_self() -> bool:
```

abulte: `current_user_is_admin_or_self`?

js/components/user/profile.vue
quaxsze: Well as the user seeing this view is either an admin or himself, the api will always return the user's email.

abulte: Nope :-) you can access any user profile page in the admin as long as you're logged in (no need to be an admin or self).

udata/templates/user/base.html
```diff
@@ -28,6 +28,13 @@
 {% if current_user != user %}
 <div class="btn-group btn-group-xs">
     {{ follow_btn(user) }}
+    {% if sysadmin %}
+    <div class="btn btn-primary">
+        <a href="{{ url_for('admin.index', path='user/{id}/'.format(id=user.id)) }}" class="btn-primary">
+            {{ _('User\'s admin profile') }}
```
abulte, line 34:
```suggestion
            {{ _("User's admin profile") }}
```

abulte (approved):

taniki:
> While we're at it, can we add an "Edit" button or link for admins that goes from the profile page https://demo.data.gouv.fr/fr/users/amalia-ouriachi-2/ to the admin page for this user? It would be great for moderation 💪



Pour toutes les fois où j'ai cru devenir fou à pas trouver un usager dans l'admin

PR #2597, opened 2021-03-18
Pin twine

quaxsze:


requirements/circleci.pip
```diff
@@ -1,4 +1,4 @@
 -r test.pip
 invoke==1.3.0
 wheel
-twine
+twine==3.3.0
```
abulte, line 4: Add a comment to explain why we pinned

requirements/develop.pip
```diff
@@ -4,4 +4,4 @@ invoke==1.3.0
 -r report.pip
 Flask-DebugToolbar==0.10.1
 wheel
-twine
+twine==3.3.0
```
abulte, line 7: Add a comment to explain why we pinned

abulte (approved):

PR #2610, opened 2021-05-19
Add sqlite and ics format in allowed extensions

quaxsze:


abulte (commented):
Add `.db` too while we're at it.

abulte (approved):

ColinMaudry:
@abulte When does it get shipped to data.gouv.fr?

abulte:
Yesterday I think ;-)

PR #2619, opened 2021-06-04
Use udata's markdown configuration on SafeMarkup aswell

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/219

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Add a UdataCleaner class to use udata's markdown configuration on SafeMarkup aswell [#2619](https://github.com/opendatateam/udata/pull/2619)
+- Add a UdataCleaner class to use udata's markdown configuration on SafeMarkup as well [#2619](https://github.com/opendatateam/udata/pull/2619)
 
 ## 2.7.1 (2021-05-27)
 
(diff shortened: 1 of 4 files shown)


quaxsze:
We can rollback from `SafeMarkup` to `Markup` on avatars hooks, or we can allow `img` tags in the `SafeMarkup` class.
WDYT?

abulte:
Just rolling back won't cut it, `title=getattr(user, 'fullname', _('Anonymous user'))` is user provided, so it's open for XSS.

Maybe we should take a step back on those sanitization issues. We're piling up patches at the end of the process (ie rendering), when maybe we should work at the source (ie user input before storing in DB).

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Add a UdataCleaner class to use udata's markdown configuration on SafeMarkup aswell [#2619](https://github.com/opendatateam/udata/pull/2619)
```
abulte, line 5:
```suggestion
- Add a UdataCleaner class to use udata's markdown configuration on SafeMarkup as well [#2619](https://github.com/opendatateam/udata/pull/2619)
```

abulte (approved):

PR #2620, opened 2021-06-09
Migrate from raven to sentry-sdk (#2605)

### LINKED ISSUE #2605: [sentry] migrate from raven to sentry-sdk
Migrate from `raven` to `sentry-sdk`.

## Detailed Description
`raven` is the legacy (and deprecated) client for Sentry, it has been replaced by `sentry-sdk`.

## Context
Current `raven` version: `6.10.0`
Needed `sentry-sdk` version: `latest`

## Possible Implementation
https://docs.sentry.io/platforms/python/

maudetes:
Fix #2605.

Update sentry python lib from `raven` to `sentry-sdk`.

Question :
- How to  make sure that we log the same exact errors (ex: check Celery integration) ?
- Can we directly use a public sentry DSN ? The DSN with secret is not used and deprecated:
  > Deprecated DSN includes a secret which is no longer required by newer SDK versions.

### DIFF
--- a/udata/sentry.py
+++ b/udata/sentry.py
@@ -16,6 +16,7 @@ RE_DSN = re.compile(
     r'@(?P<domain>.+)/(?P<site_id>\d+)')
 
 SECRET_DSN_DEPRECATED_MSG = 'DSN with secret is deprecated, use a public DSN instead'
+ERROR_PARSE_DSN_MSG = 'Unable to parse Sentry DSN'
 
 # Controlled exceptions that Sentry should ignore
 IGNORED_EXCEPTIONS = HTTPException, PermissionDenied, UploadProgress
@@ -24,13 +25,14 @@ def public_dsn(dsn):
     '''Check if DSN is public or raise a warning and turn it into a public one'''
     m = RE_DSN.match(dsn)
     if not m:
-        log.error('Unable to parse Sentry DSN')
+        log.error(ERROR_PARSE_DSN_MSG)
+        raise ValueError(ERROR_PARSE_DSN_MSG)
 
     if not m["secret"]:
         return dsn
 
     log.warning(SECRET_DSN_DEPRECATED_MSG)
-    warnings.warn(SECRET_DSN_DEPRECATED_MSG, category=DeprecationWarning, stacklevel=2)
+    warnings.warn(SECRET_DSN_DEPRECATED_MSG, category=DeprecationWarning)
 
     public = '{scheme}://{client_id}@{domain}/{site_id}'.format(
         **m.groupdict())
--- a/udata/sentry.py
+++ b/udata/sentry.py
@@ -1,6 +1,7 @@
 import logging
 import pkg_resources
 import re
+import warnings
 
 from werkzeug.exceptions import HTTPException
 from udata import entrypoints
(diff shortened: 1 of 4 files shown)


abulte (commented):
Nice!

Re public DSN, I see two options (by order of preference):
1. keep it as is if it's compatible with both a public and private DSN, maybe with a deprecation warning if we detect a private one
2. check that we have a public DNS and throw an error if we have a private one (this will make sure nobody will leak a private DSN somehow someday _and_ force use to update our config)

Re tests: ideally this should be unit tested, define a Celery task that fails, see if a sentry mock is called.

jordanguedj:
I think we can get rid of `public_dsn` as we do have a public DSN for the project (it is currently configured using a deprecated one as we were waiting for the client to be upgraded) and Sentry explicitly deprecates DSNs including a secret.

abulte:
@jordanguedj agreed for us but maybe not for every udata instances in the world 😬

jordanguedj:
@abulte I agree that we should keep it backward compatible, there is no debate about this.
However we shouldn't force the use of deprecated DSNs (new DSNs aren't compatible with our current implementation).

abulte:
> However we shouldn't force the use of deprecated DSNs

Did I suggest that? 🤔 Not my intention :-) I'm suggesting in 1/ to show a deprecation warning for deprecated DSNs and in 2/ to not support them at all.

NB: just looked at the code, we're injecting the DSN into the template, so we should really make sure we manipulate a public DSN, either by refusing private DSNs or converting them to public ones as we do now.

abulte:
@jordanguedj sorry, I read and replied too fast, but if I sum up I think we agree:
- we do not want to force the use of deprecated (private) DSN (as the `public_dsn` currently do)
- we still want to "sanitize" the legacy/deprecated/private DSN to a public DSN (a refactored `public_dsn` would work) because we inject it in the templates — or refuse them altogether

udata/sentry.py
```diff
@@ -11,18 +12,26 @@
...
 # Controlled exceptions that Sentry should ignore
 IGNORED_EXCEPTIONS = HTTPException, PermissionDenied, UploadProgress
 
-
 def public_dsn(dsn):
-    '''Transform a standard Sentry DSN into a public one'''
+    '''Check if DSN is public or raise a warning and turn it into a public one'''
     m = RE_DSN.match(dsn)
     if not m:
         log.error('Unable to parse Sentry DSN')
```
abulte, line 27: not having a return or raise here is strange, don't you think (I know it's not your code)

udata/sentry.py
```diff
@@ -11,18 +12,26 @@
...
+
+    if not m["secret"]:
+        return dsn
+
+    log.warning(SECRET_DSN_DEPRECATED_MSG)
+    warnings.warn(SECRET_DSN_DEPRECATED_MSG, category=DeprecationWarning, stacklevel=2)
```
abulte, line 33: what does `stacklevel=2` do?

maudetes:
I fixed the public/private DSN question by showing a warning in case of secrets, and turn it into a public one anyway.

For the unit test, I suggest we add those in a different MR, I created the corresponding issue: #2621.
Independently, I manually checked that worker errors are caught by Sentry.

udata/sentry.py
maudetes: agreed, I'll change it!

udata/sentry.py
maudetes: It is an artifact, not relevant here. It unwinds to the previous caller for the stack trace. Here it would show `init_app` instead of `public_dsn` stack trace level. I'm removing it.

abulte (approved):

abulte:
@maudetes don't forget the CHANGELOG entry before merging 💡

jordanguedj (commented):
LGTM

PR #2624, opened 2021-06-17
Webhooks

abulte:
A naive approach at dispatching webhooks.

For now, only the dispatch logic is here, we still need to plug it to signals (dataset created...).

Also, some linting changes (enforce single quotes).

### DIFF
--- a/requirements/develop.pip
+++ b/requirements/develop.pip
@@ -1,4 +1,5 @@
 invoke==1.3.0
+flake8-quotes==3.2.0
 -r install.pip
 -r test.pip
 -r report.pip
--- a/requirements/install.pip
+++ b/requirements/install.pip
@@ -48,7 +48,7 @@ pytz==2019.3
 rdflib-jsonld==0.5.0
 redis==3.3.11
 rdflib==5.0.0
-requests==2.24.0
+requests==2.25.1
 simplejson==3.17.0
 StringDist==1.0.9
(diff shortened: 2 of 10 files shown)


abulte:
@quaxsze some hell broke loose in the tests because https://github.com/opendatateam/udata/blob/9bc0f731c23bb6fc45765bb312aabf94e697de3c/udata/core/user/views.py#L63 disappeared (errors were hidden before but my modifications made them pop), we should discuss this.

maudetes:
How did you pick the signals to plug in ? For example, don't we want to add the signals for resources addition and deletion (`Resource.on_added`, `Resource.on_deleted` that don't look connected at all today) ?

udata/features/webhooks/triggers.py
```diff
@@ -0,0 +1,79 @@
...
+def on_dataset_update(dataset):
+    updates, _ = dataset._delta()
+    if 'private' in updates and not dataset.private:
+        dispatch('datagouvfr.dataset.created', dataset.to_json())
+    else:
+        dispatch('datagouvfr.dataset.updated', dataset.to_json())
+
+
+@on_new_discussion.connect
+def on_new_discussion(discussion):
```
maudetes, line 32: Same question on discussion events, don't we want to add the `if not dataset.private` condition on dataset deletion ?

PR #2627, opened 2021-06-28
Import theme translations

maudetes:
Import any translations files from package, not only those located at `rootdir/translations/`.

It allows us to load theme translations, for example located at `rootdir/theme/gouvfr/translations/`.

### DIFF
--- a/udata/i18n.py
+++ b/udata/i18n.py
@@ -1,7 +1,7 @@
 import pkgutil
 
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


abulte (approved):
I guess we could switch to https://docs.python.org/3/library/pathlib.html since we're in python3, but if it ain't broke, don't fix it.

quaxsze (approved):

PR #2629, opened 2021-07-07
Migration for community resources schema

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/338

### DIFF
--- a/udata/migrations/2021-07-07-update-schema-for-community-resources.py
+++ b/udata/migrations/2021-07-07-update-schema-for-community-resources.py
@@ -1,0 +1,31 @@
+'''
+The purpose here is to update every community resource's metadata 'schema'
+with a new format (string to object)
+'''
+import logging
+
+from udata.models import CommunityResource
+
+log = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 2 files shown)


udata/migrations/2021-07-07-update-schema-for-community-resources.py
```diff
@@ -0,0 +1,31 @@
...
+def migrate(db):
+    log.info('Processing community resources.')
+
+    community_resources = CommunityResource.objects().no_cache().timeout(False)
+    for resource in community_resources:
+        save_res = False
+        if hasattr(resource, 'schema'):
+            schema = resource.schema
+            resource.schema = {}
+            if schema is not None and isinstance(schema, str):
```
maudetes, line 21: Could schema be another type? Maybe log a warning if it is the case?

quaxsze: The community resource model had a stringfield before and a dictfield now. So the schema cannot be another type than str

maudetes (approved):

PR #2631, opened 2021-07-13
Fix RDF endpoints Mime Type detection

quaxsze:


maudetes (approved):
To improve in a future iteration: https://github.com/etalab/data.gouv.fr/issues/374.

maudetes:
This dirty fix has been reverted (https://github.com/opendatateam/udata/commit/ff4b6175d6a690bac08c53f83441d8865a077db7) because it did not change the application/type that is still json by default.
XML contents are still dumped to json if no Accept header is specified.

PR #2633, opened 2021-07-16
Update sentry logging

maudetes:
Fix https://github.com/etalab/data.gouv.fr/issues/358 (show sentry event id).
Fix https://github.com/etalab/data.gouv.fr/issues/231 (update from raven to sentry).

`SENTRY_PUBLIC_DSN` was required to set meta tags accordingly (see https://github.com/etalab/data.gouv.fr/issues/352#issuecomment-876588788) and had been removed in #2620.
Sentry event id should now be displayed on error.

Left to discuss:
- Couldn't test that sentry event id showed on 400 or 500 due to DON'T PANIC werkzeug traceback.
- Not sure how to deal with this legacy eventListener [here](https://github.com/opendatateam/udata/blob/168214894eddf3cc879fe483569c3ca9fe3bd387/js/admin.vue#L58). Couldn't find a corresponding logic in the new sdk.
- Not sure what logger param was used for [here](https://github.com/opendatateam/udata/blob/168214894eddf3cc879fe483569c3ca9fe3bd387/js/raven.js#L6) and if we can find a corresponding configuration if still relevant.


quaxsze (approved):

maudetes:
Cherry-picked sentry event logging in https://github.com/opendatateam/udata/pull/2634.
Not sure we should update the sentry sdk as I can't find trace of support for vue 1 in the [documentation](https://docs.sentry.io/platforms/javascript/guides/vue/).

maudetes:
For some errors (not all of them), Sentry exception trace is missing with updated sdk.

image
VS
image

Couldn't find a solution to capture the stack trace correctly.

ThibaudDauce:
@maudetes Is this still something we want to do? :-)

PR #2636, opened 2021-07-26
Fix root api endpoint error

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/372

### DIFF
--- a/udata/api/__init__.py
+++ b/udata/api/__init__.py
@@ -228,7 +228,7 @@ def extract_name_from_path(path):
     base_path, query_string = path.split('?')
     infos = base_path.strip('/').split('/')[2:]  # Removes api/version.
     if base_path == '/api/1/':  # The API root endpoint redirects to swagger doc.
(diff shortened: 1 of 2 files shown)


maudetes:
This fix means we won't send signal on `/api/1/`, shouldn't we fix `extract_name_from_path` to support the root url instead ?

udata/api/__init__.py
```diff
@@ -227,6 +227,8 @@ def extract_name_from_path(path):
     """
     base_path, query_string = path.split('?')
     infos = base_path.strip('/').split('/')[2:]  # Removes api/version.
+    if base_path == '/api/1/':  # The API root endpoint redirects to swagger doc.
+        return safe_unicode('Api_doc')
```
maudetes, line 231: Can we change this strange half-snake :snake: half-upside-down-camel :camel: case? Or could we just use `apidoc` ? :p

maudetes (approved):

PR #2637, opened 2021-07-28
Return notfound() when router does not match DB territory object

quaxsze:


quaxsze:
Fixes one of https://github.com/etalab/data.gouv.fr/issues/339

maudetes (approved):

PR #2638, opened 2021-08-06
Add larger reuse thumbnail image

quaxsze:


quaxsze:
The images must be rerendered with the `udata images render` command, and the suggest index must be rebuild with `udata search index`

abulte (approved):

PR #2640, opened 2021-08-06
Use mongomock for testing

abulte:
Ref http://docs.mongoengine.org/guide/mongomock.html

This should speed up testing (twice as fast on my machine). 

:warning: remove `MONGODB_HOST_TEST` from your local `udata.cfg`.

Fix https://github.com/etalab/data.gouv.fr/issues/467

### DIFF
--- a/udata/settings.py
+++ b/udata/settings.py
@@ -494,6 +494,7 @@ class Testing(object):
         'title^2',
         'description',
     )
+    MONGODB_HOST_TEST = 'mongomock://localhost:27017/udata'
 
 
 class Debug(Defaults):
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,6 @@ jobs:
   python:
     docker:
       - image: udata/circleci:2-alpine
-      - image: circleci/mongo:3.6-stretch-ram
       - image: redis:alpine
       - image: udata/elasticsearch:2.4.5
     environment:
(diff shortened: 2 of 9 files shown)


abulte:
Unfortunately we still have two tests failing although I rewrote them and got rid of technical debt in the mean time :-/ Might be worth opening an issue on `mongomock`.

abulte:
Other refactoring idea: https://github.com/opendatateam/udata/blob/0758d23d3fff92d6f61dbb4f7666142a7e2bef0d/udata/core/reuse/api.py#L171 — followers also uses `GenericReferenceField` but thanks to the `model` abstraction, it always uses a full-fledged object to make the query, instead of an id (which is the problem with our two failing tests).

maudetes:
I think it is worth pushing this refactoring through to save time for testing, and additionally removing some technical debt.

udata/settings.py
```diff
@@ -494,6 +494,7 @@ class Testing(object):
         'title^2',
         'description',
     )
+    MONGODB_HOST_TEST = 'mongomock://localhost:27017/udata'
```
maudetes, line 497: Maybe update docs (`adapting-settings.md` and `testing-code.md`) with mongomock changes.

maudetes:
Issue created by @quaxsze with reproductible bug: https://github.com/mongomock/mongomock/issues/719

maudetes:
Closing for now, let's reopen it if the mongomock ticket gets resolved.

PR #2643, opened 2021-08-10
Activate plugins before creating app when testing

abulte:
Necessary for https://github.com/opendatateam/udata-recommendations/pull/183.

This duplicates what `pytest-flask` does but at a higher level (before app is created). I have mixed feelings about this but I can't think of a better idea and the logic is pretty sane.

EDIT: it's now scoped to plugins, good feelings now.

### DIFF
--- a/udata/tests/plugin.py
+++ b/udata/tests/plugin.py
@@ -101,12 +101,16 @@ def get_settings(request):
     if marker:
         return marker.args[0]
     _settings = getattr(request.cls, 'settings', settings.Testing)
-    # apply the options marker from pytest_flask as soon as app is created
+    # apply the options(plugins) marker from pytest_flask as soon as app is created
     # https://github.com/pytest-dev/pytest-flask/blob/a62ea18cb0fe89e3f3911192ab9ea4f9b12f8a16/pytest_flask/plugin.py#L126
     # this lets us have default settings for plugins applied while testing
+    plugins = []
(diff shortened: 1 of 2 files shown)


abulte:
Well, this is unfortunately too broad: options applied to a method will apply to (at least) all the methods of the class (see failing tests).

abulte:
Now scoped to the `plugins` options, which makes more sense.

udata/tests/plugin.py
```diff
@@ -101,7 +100,17 @@ def get_settings(request):
...
+    _settings = getattr(request.cls, 'settings', settings.Testing)
+    # apply the options(plugins) marker from pytest_flask as soon as app is created
+    # https://github.com/pytest-dev/pytest-flask/blob/a62ea18cb0fe89e3f3911192ab9ea4f9b12f8a16/pytest_flask/plugin.py#L126
+    # this lets us have default settings for plugins applied while testing
+    plugins = []
+    for options in request.node.iter_markers('options'):
+        option = options.kwargs.get('plugins', []) or options.kwargs.get('PLUGINS', [])
+        plugins += option
+    if plugins:
+        setattr(_settings, 'PLUGINS', plugins)
```
maudetes, line 112: We could set it even if `plugins` is empty ?

maudetes (approved):
The overall logic seems correct to me.

maudetes (approved):

PR #2647, opened 2021-08-18
Fix user delete command

abulte:
As per #2646

### DIFF
--- a/udata/core/user/commands.py
+++ b/udata/core/user/commands.py
@@ -67,7 +67,7 @@ def delete():
     if not user:
         exit_with_error('Invalid user')
     user.mark_as_deleted()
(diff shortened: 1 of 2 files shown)


udata/core/user/commands.py
```diff
@@ -66,7 +66,7 @@ def delete():
     user = User.objects(email=email).first()
     if not user:
         exit_with_error('Invalid user')
-    user.delete()
+    user.mark_as_deleted()
     success('User deleted successfully')
```
maudetes, line 70: Could you please change the message to something like 'User successfully marked as deleted' ? :pray:

maudetes (approved):
Don't forget changelog :)

PR #2652, opened 2021-08-20
Fix RDF output

abulte:
Bypass the unnecessary flask-restplus content negotiation and use Flask's.

TODO:
- [x] test

Fix https://github.com/etalab/data.gouv.fr/issues/374


maudetes (commented):
Yay! Seems to work nicely!

We need to do the same on other rdf endpoints (site/catalog and organization/catalog) :)

quaxsze (approved):

quaxsze (approved):

PR #2655, opened 2021-08-24
Update multiple dependencies

quaxsze:


quaxsze:
Too painful for the moment. We will be back on it later

PR #2659, opened 2021-09-09
Pin setuptools==58.0.0

maudetes:
Pinning setuptools in `setup_requires` doesn't seem to work (see https://stackoverflow.com/questions/58521386/should-setuptools-be-in-the-setup-requires-entry-of-setup-cfg-files).

Pinning in CI could probably work for now but it won't prevent any other env to use a newer `setuptools` version.

Any suggestion is welcome.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,8 +26,10 @@ jobs:
       - run:
           name: Install python dependencies
           command: |
-            virtualenv venv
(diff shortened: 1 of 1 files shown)


maudetes:
I tried to set build dependencies using https://www.python.org/dev/peps/pep-0518/ without success.

maudetes:
Not needed anymore due to #2660.

PR #2660, opened 2021-09-09
Update dependencies

maudetes:
# Updating dependencies

In order to fix https://github.com/etalab/data.gouv.fr/issues/512.

This updated deps were also tested in https://github.com/etalab/udata-front/pull/6.

## mongoengine

Changelog: https://github.com/MongoEngine/mongoengine/blob/master/docs/changelog.rst#changes-in-0200.

Some breaking changes, but it seems that few impact us directly.

Notes:
- `class_check` no more supported.

## feedparser

Changelog: https://github.com/kurtmckee/feedparser/blob/develop/CHANGELOG.rst

feedparser doesn't seem used on udata side. Maybe we should remove it?

## rdflib-jsonld

Using rdflib that has a jsonld plugin shipped already.
Documentation: https://github.com/RDFLib/rdflib-jsonld#using-the-plug-in-jsonld-serializerparser-with-rdflib

Notes:
- We used a hack to set the context to the url. Should we continue to do that?
- Somehow rdfs doesn't look supported?
- `DCT.term('format')` was replaced by `DCT.format`
- Some encoding changes

### DIFF
--- a/udata/tests/dataset/test_dataset_rdf.py
+++ b/udata/tests/dataset/test_dataset_rdf.py
@@ -803,7 +803,7 @@ class DatasetRdfViewsTest:
             response = client.get(url, headers={'Accept': 'application/ld+json'})
             assert200(response)
             assert response.content_type == 'application/ld+json'
-            assert response.json['@context']['@vocab'] == "http://www.w3.org/ns/dcat#"
+            assert response.json['@context']['@vocab'] == 'http://www.w3.org/ns/dcat#'
 
     @pytest.mark.parametrize('fmt,mime', [
         ('n3', 'text/n3'),
--- a/udata/tests/dataset/test_dataset_rdf.py
+++ b/udata/tests/dataset/test_dataset_rdf.py
@@ -140,7 +140,7 @@ class DatasetToRdfTest:
         assert r.value(DCAT.accessURL).identifier == URIRef(permalink)
         assert r.value(DCAT.bytesSize) == Literal(resource.filesize)
         assert r.value(DCAT.mediaType) == Literal(resource.mime)
-        assert r.value(DCT.term('format')) == Literal(resource.format)
+        assert r.value(DCT.format) == Literal(resource.format)
 
         checksum = r.value(SPDX.checksum)
         assert r.graph.value(checksum.identifier, RDF.type) == SPDX.Checksum
@@ -395,7 +395,7 @@ class RdfToDatasetTest:
         g.add((node, DCT.modified, Literal(modified)))
         g.add((node, DCAT.bytesSize, Literal(filesize)))
         g.add((node, DCAT.mediaType, Literal(mime)))
-        g.add((node, DCT.term('format'), Literal('CSV')))
+        g.add((node, DCT.format, Literal('CSV')))
 
         checksum = BNode()
         g.add((node, SPDX.checksum, checksum))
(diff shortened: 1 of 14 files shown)


abulte (commented):
First remark: yes, we should remove feedparser 🧹

udata/tests/dataset/test_dataset_rdf.py
```diff
@@ -803,16 +803,14 @@ def test_dataset_rdf_json_ld(self, client):
             response = client.get(url, headers={'Accept': 'application/ld+json'})
             assert200(response)
             assert response.content_type == 'application/ld+json'
-            context_url = url_for('api.site_jsonld_context', _external=True)
-            assert response.json['@context'] == context_url
+            assert response.json['@context']['@vocab'] == "http://www.w3.org/ns/dcat#"
```
abulte, line 806:
```suggestion
            assert response.json['@context']['@vocab'] == 'http://www.w3.org/ns/dcat#'
```

udata/tests/site/test_site_rdf.py
```diff
@@ -164,8 +164,7 @@ def test_catalog_rdf_json_ld(self, fmt, client):
         response = client.get(url, headers={'Accept': 'application/ld+json'})
         assert200(response)
         assert response.content_type == 'application/ld+json'
-        context_url = url_for('api.site_jsonld_context', _external=True)
-        assert response.json['@context'] == context_url
+        assert response.json['@context']['@vocab'] == "http://www.w3.org/ns/dcat#"
```
abulte, line 167:
```suggestion
        assert response.json['@context']['@vocab'] == 'http://www.w3.org/ns/dcat#'
```

abulte (commented):
Re Mongoengine: LGTM

Re `rdfs`: it's a breaking change we should document. It does not seem to be used in practice on our instance.

```
root@dataweb-05:~# grep '\.rdfs' /var/log/nginx/www.data.gouv.fr.access.log.1
root@dataweb-05:~# zgrep '\.rdfs' /var/log/nginx/www.data.gouv.fr.access.log.2.gz
root@dataweb-05:~# zgrep '\.rdfs' /var/log/nginx/www.data.gouv.fr.access.log.[2..10].gz
```

Re URL hack: If I understand correctly, since we don't hit the bug here https://github.com/RDFLib/rdflib-jsonld/issues/37 in the tests, we're fine?

maudetes:
> Re `rdfs`: it's a breaking change we should document. It does not seem to be used in practice on our instance.

Yep, I checked but I think rdfs wasn't documented at all actually: https://udata.readthedocs.io/en/latest/rdf/#content-negotiation.

> Re URL hack: If I understand correctly, since we don't hit the bug here [RDFLib/rdflib-jsonld#37](https://github.com/RDFLib/rdflib-jsonld/issues/37) in the tests, we're fine?

In my understanding, we took advantage of this "bug", and we used the `to_dict` as a hack to return an url as `@context` instead of the context dict itself. This hack isn't working anymore as is, and I changed the behaviour to return the context dict. I can take another look to go back to the prevous behaviour.

abulte:
> Yep, I checked but I think rdfs wasn't documented at all actually: https://udata.readthedocs.io/en/latest/rdf/#content-negotiation.

Okay, a mention in the changelog then !

> In my understanding, we took advantage of this "bug", and we used the to_dict as a hack to return an url as @context instead of the context dict itself. This hack isn't working anymore as is, and I changed the behaviour to return the context dict. I can take another look to go back to the prevous behaviour.

Well, your code looks cleaner, so LGTM!

abulte (approved):
CHANGELOG and good to go.

PR #2669, opened 2021-10-22
Switched list endpoints from ES to Mongo

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/586
fix https://github.com/etalab/data.gouv.fr/issues/587


maudetes (commented):
Thank you for this PR! :pray: I think it matches what had been discussed.

These are some global notes:
- [x] I think we can add some tests to make sure that we test on lists and search (for dataset, reuse and orga)?
- [x] Should we use a generic parser because there seems to be duplicate code between dataset, reuse and orga lists?
- [x] For now we can't sort on relevance or followers or ..., should we?
- [ ] I think we should also modify the call to the suggest route in admin. I created [this ticket](https://github.com/etalab/data.gouv.fr/issues/599).
- [ ] We should check the relevance of the mongo text search (does it work with lowercase, accents, etc.)? It didn't look satisfying on multiple words

udata/core/dataset/api.py
```diff
@@ -68,7 +67,19 @@
...
-search_parser = DatasetSearch.as_request_parser()
+
+dataset_parser = api.parser()
+dataset_parser.add_argument('q', type=str, location='args',
+                            help='The search query')
+dataset_parser.add_argument(
+    'sort', type=str, default='-created', location='args',
+    help='The sorting attribute')
+dataset_parser.add_argument(
+    'page', type=int, default=1, location='args', help='The page to fetch')
```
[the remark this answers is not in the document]
quaxsze: 0 causes 404, I guess what works with ES does not with mongo

udata/api/parsers.py
```diff
@@ -0,0 +1,36 @@
+from udata.api import api
+
+
+class ModelApiParser:
```
quaxsze, line 4: Draft of new parser to avoid code duplication. WDYT? @maudetes

udata/tests/api/test_datasets_api.py
```diff
@@ -69,13 +93,13 @@ def test_dataset_api_list_filtered_by_org(self):
             VisibleDatasetFactory()
             dataset_org = VisibleDatasetFactory(organization=org)
 
-        response = self.get(url_for('api.datasets'),
+        response = self.get(url_for('apiv2.dataset_search'),
```
quaxsze, line 96: Should we use Mongo here instead of ES? It is not really a search but a matching query isn't it? @maudetes

udata/core/dataset/api.py
```diff
@@ -59,16 +60,28 @@
...
 )
-from .search import DatasetSearch
 from .exceptions import (
     SchemasCatalogNotFoundException, SchemasCacheUnavailableException
 )
 from .rdf import dataset_to_rdf
 
+
+class DatasetApiParser(ModelApiParser):
+    sorts = {
```
maudetes, line 70: I think it's not exactly the same sorts parameters as in the `class DatasetSearch`. Shouldn't we be iso?

udata/core/organization/api.py
```diff
@@ -397,10 +411,16 @@ def put(self, org):
         return {'image': org.logo}
 
 
-dataset_parser = api.page_parser()
-dataset_parser.add_argument(
-    'sort', type=str, default='-created', location='args',
-    help='The sorting attribute')
+class DatasetApiParser(ModelApiParser):
```
maudetes, line 414: Could we use the DatasetApiParser defined in `dataset/api.py`?

udata/core/organization/api.py
```diff
@@ -397,10 +411,16 @@ def put(self, org):
...
+        'updated': 'last_modified',
+        'reuses': 'metrics.reuses',
+        'followers': 'metrics.followers',
+        'views': 'metrics.views',
+    }
+
+dataset_parser = DatasetApiParser()
 
 
 @ns.route('/<org:org>/datasets/', endpoint='org_datasets')
```
maudetes, line 426: `sort_mapping` is not needed anymore?

maudetes (commented):
Thank you again for this PR ! I made some comments but I don't think there is any major changes needed.

Additionnaly, my linter tells me that there are some long lines (the `search_text` lines). Do we have a limit set somewhere ?

abulte:
@maudetes re line length https://github.com/opendatateam/udata/blob/a5fba3e160abe3e1bd4ba8c36ff036153d179a22/setup.cfg#L43

udata/core/dataset/api.py
```diff
@@ -100,12 +113,15 @@
...
+    @api.expect(dataset_parser.parser)
     @api.marshal_with(dataset_page_fields)
     def get(self):
         '''List or search all datasets'''
-        search_parser.parse_args()
-        return search.query(Dataset, **multi_to_dict(request.args))
+        args = dataset_parser.parse()
+        datasets = Dataset.objects(archived=None, deleted=None, private=False)
+        if args['q']:
+            return datasets.search_text(args['q']).order_by('$text_score').paginate(args['page'], args['page_size'])
```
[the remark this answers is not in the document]
quaxsze: Sorting by text_score without a text query does not seem relevant no?
Anyway trying to sort without text search raises the following error 😬 
`*** pymongo.errors.OperationFailure: must have $meta projection for all $meta sort keys`

udata/tests/api/test_user_api.py
```diff
@@ -60,130 +60,53 @@ def test_unfollow(self):
...
-        '''It should suggest users without deduplicating homonyms'''
-        with self.autoindex():
-            UserFactory.create_batch(2, first_name='test', last_name='homonym')
+    def test_user_api_full_text_search_unicode(self):
+        '''It should find user with special characters'''
+        user = UserFactory(first_name='test', last_name='testé')
 
-        response = self.get(url_for('api.suggest_users'),
-                            qs={'q': 'homonym', 'size': '5'})
+        response = self.get(url_for('api.users', q='test testé'))
```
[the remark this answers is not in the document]
quaxsze: Damned

udata/core/dataset/api.py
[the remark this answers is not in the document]
maudetes: Ah no sorry, I meant with a text query.
An example logic could be:
```
if args['q']:
    search_datasets = datasets.search_text(args['q'])
    if args['sort']:
        return search_datasets.order_by(args['sort']).paginate [...]
    else
        return search_datasets.order_by('$text_score').paginate [...]
```
What do you think?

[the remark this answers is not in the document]
quaxsze: 👍

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 - Move template hook logic back to udata [#2671](https://github.com/opendatateam/udata/pull/2671)
 - Add dataset's acronym to catalog [#2675](https://github.com/opendatateam/udata/pull/2675)
 - Better URL-based License guess [#2672](https://github.com/opendatateam/udata/pull/2672)
+- :warning: API changes [#2669](https://github.com/opendatateam/udata/pull/2669):
+  - List endpoints for organizations, datasets, reuses and users are now querying MongoDB instead of ElasticSearch.
+  - Those endpoints use MongoDB full text search when `q`argument is used.
+  - A new API parser was implemented to replace the search one.
+  - The previous ElasticSearch endpoints were moved to APIv2 with folloing url pattern: `/{object}/search` (ex: `/datasets/search`).
```
maudetes, line 12:
```suggestion
- :warning: API changes [#2669](https://github.com/opendatateam/udata/pull/2669):
  - List endpoints for organizations, datasets, reuses and users are now querying MongoDB instead of ElasticSearch.
  - Those endpoints use MongoDB full text search when `q` argument is used.
  - A new API parser was implemented to replace the search one.
  - The previous ElasticSearch endpoints were moved to APIv2 with the following url pattern: `/{object}/search` (ex: `/datasets/search`).
```

udata/core/user/api_fields.py
```diff
@@ -91,6 +91,7 @@
 
 user_page_fields = api.model('UserPage', fields.pager(user_fields))
 
+
```
maudetes, line 94: remove this line

udata/core/dataset/api.py
maudetes: We should probably be ,iso for sorting to work in admin (sorting datasets by columns in /fr/admin/site/ for example).

udata/core/dataset/api.py
```diff
@@ -100,12 +116,21 @@
...
+        datasets = Dataset.objects(archived=None, deleted=None, private=False)
+        if args['q']:
+            search_datasets = datasets.search_text(args['q'])
+            if args['sort']:
+                return search_datasets.order_by(args['sort']).paginate(args['page'], args['page_size'])
+            else:
+                return search_datasets.order_by('$text_score').paginate(args['page'], args['page_size'])
+        if args['sort']:
+            return datasets.order_by(args['sort']).paginate(args['page'], args['page_size'])
+        return datasets.order_by(DEFAULT_SORTING).paginate(args['page'], args['page_size'])
```
maudetes, line 133: Could we simplify this code using something like:
```
sort = args['sort'] or ('$text_score' if args['q'] else None) or DEFAULT_SORTING
return datasets.order_by(sort).paginate(args['page'], args['page_size'])
```

udata/core/dataset/apiv2.py
```diff
@@ -124,6 +129,18 @@
...
+@ns.route('/search', endpoint='dataset_search')
+class DatasetSearchAPI(API):
+    '''Datasets collection search endpoint'''
+    @apiv2.doc('search_datasets')
+    @apiv2.expect(search_parser)
+    @apiv2.marshal_with(dataset_page_fields)
+    def get(self):
+        '''List or search all datasets'''
+        search_parser.parse_args()
+        return search.query(Dataset, **multi_to_dict(request.args))
```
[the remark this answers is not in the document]
maudetes: We use `search.query(OrganizationSearch, ...)` and `search.query(ReuseSearch, ...)` for example

udata/core/dataset/apiv2.py
```diff
@@ -124,6 +129,18 @@
 })
 
 
+@ns.route('/search', endpoint='dataset_search')
+class DatasetSearch(API):
```
maudetes, line 133: Since this class change, we have an issue on v2 orga and reuse search (that we should have seen with tests no?) because we define the OrganizationSearch search twice in apiv2! Could we rollback the class name changes?

udata/tests/api/test_user_api.py
```diff
@@ -60,130 +60,57 @@ def test_unfollow(self):
         self.assertEqual(Follow.objects.following(user).count(), 0)
         self.assertEqual(Follow.objects.followers(user).count(), 0)
 
-    def test_suggest_users_api_first_name(self):
```
maudetes, line 63: Could you add user suggest tests back since we're not removing the suggest in this PR? :pray:

maudetes (commented):
I add some comments on this PR but I don't think there are many blocking changes left after these.

We could improve the api parser later if we're not convinced by something better yet?

PR #2677, opened 2021-11-05
Realistic fixtures

quaxsze:
Fix https://github.com/etalab/data.gouv.fr/issues/527


udata/commands/fixtures.py
```diff
@@ -13,23 +14,85 @@
...
+            json_org = json_dataset.pop('organization')
+            json_resources = json_dataset.pop('resources')
+            for res in json_resources:
+                del res['latest']
+                del res['preview_url']
+                del res['last_modified']
+            json_fixture['resources'] = json_resources
+            json_fixture['dataset'] = json_dataset
+
+            json_org = requests.get(f"{data_source}{ORG_URL}/{json_org['id']}/").json()
```
maudetes, line 53: If we have two datasets from same organization, we would try to generate the organization twice later though?

maudetes (commented):
So nice these realistics datasets! :tada: 

Some general comments:
- [x] Could we add Discussion fixtures as well ?
- [ ] Some lines are longer than 100 chars
- [ ] Not sure this dataset-oriented approach will allow us to expand the fixtures with particular reuses, orga, posts etc. But I don't really have any better idea yet, we can start with this and see when we want to extend later
- [ ] For now, I have some territory-related errors on http://dev.local:7000/fr/datasets/. Could we fix this?

udata/settings.py
```diff
@@ -462,6 +462,8 @@ class Defaults(object):
                         'IssueAPI.post',
                         'SourcesAPI.post',
                         'FollowAPI.post']
+    
+    DATASET_SLUGS = []
```
maudetes, line 466: Could we name it more explicitely: `FIXTURE_DATASET_SLUGS` maybe?

And could you add a small entry in `adapting-settings.md` ?

udata/commands/fixtures.py
```diff
@@ -1,35 +1,135 @@
...
-        dataset = VisibleDatasetFactory(organization=organization)
-        DiscussionFactory(subject=dataset, user=user)
-        ReuseFactory.create_batch(reuses, datasets=[dataset], owner=user)
-
-    msg = 'Generated {datasets} dataset(s) with {reuses} reuse(s) each.'
-    log.info(msg.format(**locals()))
+@click.argument('data-source')
+def generate_fixtures_file(data_source):
+    '''Build sample fixture file based on datasets slugs list (users, datasets, reuses).'''
+    datasets_slugs = current_app.config['DATASET_SLUGS']
```
maudetes, line 37:
```suggestion
    datasets_slugs = current_app.config['FIXTURE_DATASET_SLUGS']
```

docker-compose.yml
```diff
@@ -18,11 +18,11 @@ services:
...
 
-    search:
-        image: udata/elasticsearch:2.4.5
-        volumes:
-          - ./data/search/data:/usr/share/elasticsearch/data
-        expose:
-          - "9200"
-        ports:
-          - "9200:9200"
+    # search:
```
maudetes, line 21: Woops?

maudetes (approved):
Outside of the `docker-compose.yml` comment that needs to be fixed, I think it's a good start :)

We can try to improve on the dataset-centric approach during this sprint if you have some time left or we can come back to it later?

PR #2678, opened 2021-11-08
Fix transfer ownership

maudetes:
Fix https://github.com/etalab/data.gouv.fr/issues/566

When transfering ownership, the issue comes from the [`pre-save` signal](https://github.com/opendatateam/udata/blob/a5fba3e160abe3e1bd4ba8c36ff036153d179a22/udata/models/owned.py#L38) called twice (on the first `save` call, then when recomputing metrics), without having the `changed_fields` cleared.

We suggest to fix the `pre-save` signal code instead of adding `pre_save` to `signal_kwargs['ignores']` on [`Reuse.count_datasets`](https://github.com/opendatateam/udata/blob/5b2dc91922e0309235ae5b128b11053870817389/udata/core/reuse/models.py#L192) for two reasons:
- the function isn't safe to be called twice in a row and should probably be in the first place (new signals could reintroduce this issue later on)
- to prevent any unexpected side effect that could follow this signal ignore (:spaghetti: spaghetti code)

**Done:**
- Reproduce the issue with a unit test
- Fix the issue


abulte (commented):
I'm unsure about this. 

On one hand, when re-implementing metrics, we used the other pattern: ignoring signals when recomputing metrics (obviously we forgot the `pre_save` signal). So wouldn't it be more coherent to reuse the same pattern?

On the other hand, what you said is true about your solution being more foolproof. But it also introduces _more_ business logic into signal handling and maybe we'd rather go towards having _less_ of those in them, and instead having the logic when sending the signal.

YMMV :-)

Sorry I don't have a definitive answer 😇

udata/models/owned.py
```diff
@@ -43,6 +43,9 @@ def owned_pre_save(sender, document, **kwargs):
     if not isinstance(document, Owned):
         return
     changed_fields = getattr(document, '_changed_fields', [])
+    if 'organization' in changed_fields and 'owner' in changed_fields:
+        # Ownership changes (org to owner or the other way around) have already been made
+        return
     if 'organization' in changed_fields:
```
abulte, line 49: While we're at it, if we go towards less business logic in signal handling, this whole `_previous_owner` seems overkill at first glance if it's only used here https://github.com/opendatateam/udata/blob/86173677392abedc532aabf1b78e842496c81f2e/udata/models/owned.py#L68. Why not send `on_owner_change` directly from here?

abulte: Forget it, it's probably because we want `post_save` :-) But still what a mess 🍝

maudetes:
I moved the ownership change logic from the `owned_pre_save` signal to the `clean` function (see https://github.com/opendatateam/udata/pull/2678/commits/6ce5261b5c69644b043c34de1efbafcb0cf6c814).
Would it be better this way?

quaxsze (approved):

PR #2681, opened 2021-11-17
Remove issues

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/469


js/views/organization.vue
```diff
@@ -23,11 +23,6 @@
         <reuse-list id="reuses-widget" class="col-xs-12" :reuses="reuses"></reuse-list>
     </div>
 
-    <div class="row">
-        <issue-list id="issues-widget" class="col-xs-12 col-md-6" :issues="issues"></issue-list>
-        <discussion-list id="discussions-widget" class="col-xs-12 col-md-6" :discussions="discussions"></discussion-list>
```
maudetes, line 28: Shouldn't we keep the discussion list?

js/views/site.vue
```diff
@@ -20,10 +20,6 @@
     <div class="row">
         <user-list id="users" class="col-xs-12" :users="users"></user-list>
     </div>
-    <div class="row">
-        <issue-list class="col-xs-12 col-md-6" :issues="issues"></issue-list>
-        <discussion-list class="col-xs-12 col-md-6" :discussions="discussions"></discussion-list>
```
maudetes, line 25: Same here, shouldn't we keep discussion list?

maudetes (commented):
Nice! :broom:

I think that we need to update udata-front as well (I see Issue imports, templates and DiscussionCsvAdapter at least).

js/views/reuse.vue
```diff
@@ -34,11 +34,6 @@
...
 
       <div class="row">
-        <issue-list
-          id="issues-widget"
-          class="col-xs-12 col-md-6"
-          :issues="issues"
-        ></issue-list>
         <discussion-list
           id="discussions-widget"
           class="col-xs-12 col-md-6"
```
maudetes, line 39:
```suggestion
          class="col-xs-12"
```

maudetes (approved):
It looks good to me!

I think that we shouldn't merge this before the udata-front corresponding PR though?

PR #2683, opened 2021-11-23
Fix Geom form error

quaxsze:
Fix https://github.com/etalab/data.gouv.fr/issues/619


udata/core/spatial/models.py
```diff
@@ -334,3 +334,11 @@ def top_label(self):
     def handled_zones(self):
         """Return only zones with a dedicated page."""
         return [zone for zone in self.zones if zone.handled_level]
+
+    def clean(self):
```
quaxsze, line 338: This ended up not being the cause of the error but seems important in a logical way. Do we keep it?

maudetes (approved):
As discussed, the user experience isn't great now because we can't see the geom in the admin front. The admin now raises a (not user-friendly) error when trying to set geom and zones both.

It should probably be improved soon (either by adding it in the admin front or refactoring the zones + geom fields)?

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
   - The command now uses a remote file (default) if a URL is provided or a local one if a path is provided.
   - This file can be generated by using the command `generate-fixtures-file`, which takes as an argument the URL of the source queried to retieve the data dumped to the file.
   - The command `generate-fixtures-file` uses a customizable list of datasets slugs to know which datasets to query.
+  - SpatialCoverage fixes [#2683](https://github.com/opendatateam/udata/pull/2683):
+    - The model cannot have `zones` and `geom` attributes set simultaneously anymore.
```
maudetes, line 13: I add a concern here.
Are we sure that we shouldn't set a geozone + a geometry at the same time?
We can have multiple geozones so why couldn't we have a geozone and an additional arbitrary geometry (ex a city and an arbitrary area around the city)?

PR #2689, opened 2021-12-07
Add reuse topic

maudetes:
Fix https://github.com/etalab/data.gouv.fr/issues/635
Fix https://github.com/etalab/data.gouv.fr/issues/637
Fix https://github.com/etalab/data.gouv.fr/issues/639
Fix https://github.com/etalab/data.gouv.fr/issues/634
Fix https://github.com/etalab/data.gouv.fr/issues/636

Add reuse topic simarly to what exists for reuse type. It's a StringField with a choice list between 12 topics.
In the admin form, a required select field is used.
Add indexation similarly, thus allowing to search with a filter on topic: http://dev.local:7000/api/1/reuses/?topic=work_and_training.
Finally, add a migration to set reuse topics to default `others` to prevent model validation error.


maudetes:
@quaxsze I didn't used the (deprecated) Topic model at all, let me know if I should have or if the StringField appears more logical.

quaxsze (approved):

quaxsze:
Nice ! 👏 
We should remove the legacy code from the previous topics feature.

maudetes:
> Nice ! clap We should remove the legacy code from the previous topics feature.

Should we prioritize [this ticket](https://github.com/etalab/data.gouv.fr/issues/400) for the next sprint? Or do you think we should remove the legacy topic code in this PR ?

quaxsze (approved):

PR #2690, opened 2021-12-08
Fix missing checksum type in admin form

maudetes:
Following #2688, it appeared that the resource checksum type was missing.

It wasn't a problem before because it wasn't send to the API as it was undefined, but with the previous PR it sends a `null` value, not allowed by the API.
Change the erroneous value `algo` by the correct `currentType`.


maudetes:
Not convinced about this fix, let's revert https://github.com/opendatateam/udata/pull/2688 for now to release last changes and improve this solution later.

maudetes:
Treated in https://github.com/opendatateam/udata/pull/2691 instead.

PR #2696, opened 2022-01-27
Migrate back suggest to APIV1

quaxsze:


maudetes (commented):
Thank you for this PR! I think we need to add some things as well in this PR.

- We need to migrate some suggests back: organization, reuse, spatial, tags.
- Some changes have been made to the admin that need to be reverted as well. Example: using apiv2 in base completer.

maudetes (approved):
I think we're almost good :)

We can remove apiv2 from admin, it isn't used anymore, right? We can add it back later if needed?

CHANGELOG.md
```diff
@@ -21,10 +21,7 @@
   - The previous ElasticSearch endpoints were moved to APIv2 with the following url pattern: `/{object}/search` (ex: `/datasets/search`).
 - :warning: Suggest changes [#2685](https://github.com/opendatateam/udata/pull/2685):
```
maudetes, line 22:
```suggestion
- :warning: Suggest changes [#2685](https://github.com/opendatateam/udata/pull/2685) and [#2696](https://github.com/opendatateam/udata/pull/2696):
```

maudetes (approved):

PR #2703, opened 2022-02-09
Add extras in kafka producer

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/522

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
   - The ModelAdapter, SearchQuery and SearchResult patterns were kept but heavily refactored.
   - Udata uses a Kafka producer to send documents to index to the search service.
   - Udata uses HTTP request to query the search service.
+- Add the `extras` field of Dataset, Reuse and Organization into their corresponding `ModelSearchAdapter` serialize method.
 
 ## 3.3.1 (2022-01-11)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
   - The ModelAdapter, SearchQuery and SearchResult patterns were kept but heavily refactored.
   - Udata uses a Kafka producer to send documents to index to the search service.
   - Udata uses HTTP request to query the search service.
-- Add the `extras` field of Dataset, Reuse and Organization into their corresponding `ModelSearchAdapter` serialize method.
+- Add the `extras` field of Dataset, Reuse and Organization into their corresponding `ModelSearchAdapter` serialize method. [#2703](https://github.com/opendatateam/udata/pull/2703)
 
 ## 3.3.1 (2022-01-11)
 
--- a/udata/core/dataset/search.py
+++ b/udata/core/dataset/search.py
@@ -1,3 +1,4 @@
+import datetime
 from udata.models import (
     Dataset, Organization, User, GeoZone, License
 )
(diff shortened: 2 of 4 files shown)


geoffreyaldebert:
can we add also `resources` properties from `dataset` ? It would be very useful 😇

udata/core/dataset/search.py
```diff
@@ -86,6 +86,18 @@ def serialize(cls, dataset):
...
 
+        serialized_resources = []
+        for resource in dataset.resources:
+            serialized_resources.append({
+                'id': resource.id,
+                'url': resource.url,
+                'format': resource.format,
+                'title': resource.title,
+                'schema': resource.schema,
+                'latest': resource.latest
```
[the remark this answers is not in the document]
geoffreyaldebert: I think you cover the essentials properties, thx.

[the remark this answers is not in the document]
geoffreyaldebert: filesize could be useful also if trustable. And beware, it's `filetype` not `file_type`

udata/core/dataset/search.py
```diff
@@ -82,9 +82,29 @@ def serialize(cls, dataset):
...
             'owner': str(owner.id) if owner else None,
-            'format': [r.format.lower() for r in dataset.resources if r.format]
+            'format': [r.format.lower() for r in dataset.resources if r.format],
+            'extras': dataset.extras
         }
 
+        serialized_resources = []
+        for resource in dataset.resources:
+            serialized_resources.append({
+                'id': resource.id,
```
maudetes, line 92:
```suggestion
                'id': str(resource.id),
```

udata/core/dataset/search.py
```diff
@@ -82,9 +82,29 @@ def serialize(cls, dataset):
...
+                'url': resource.url,
+                'format': resource.format,
+                'title': resource.title,
+                'schema': resource.schema,
```
maudetes, line 96: We'll need to read this information in the consumer. We could also put it at the dataset level, like we deed for format. What do you think?

maudetes (commented):
Thank you for this PR!

Some extras entries could be `datetime`, which can't be serialized.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
   - The ModelAdapter, SearchQuery and SearchResult patterns were kept but heavily refactored.
   - Udata uses a Kafka producer to send documents to index to the search service.
   - Udata uses HTTP request to query the search service.
+- Add the `extras` field of Dataset, Reuse and Organization into their corresponding `ModelSearchAdapter` serialize method. [#2703](https://github.com/opendatateam/udata/pull/2703)
```
maudetes, line 21: I think we'll need to refactor this Changelog after when merging SearchDev.

udata/core/dataset/search.py
```diff
@@ -82,9 +82,30 @@ def serialize(cls, dataset):
             'resources_count': len(dataset.resources),
             'organization': organization,
             'owner': str(owner.id) if owner else None,
-            'format': [r.format.lower() for r in dataset.resources if r.format]
+            'format': [r.format.lower() for r in dataset.resources if r.format],
+            'schema': [r.schema.get('name') for r in dataset.resources if r.schema],
+            'extras': dataset.extras
```
maudetes, line 87: Extras could be of datetime type, which can't be serialized.

maudetes (approved):
I think we're good :)

PR #2706, opened 2022-02-14
Redirect with 308 instead of 302

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/682


quaxsze:
Tested with a ressource upload to a renamed dataset.
```
    requests.post(
        'http://dev.local:7000/api/1/datasets/<renamed-dataset-slug>/resources/<resource-id>/upload/',
        files={"file": open(csv_path, "rb")},
        headers={"X-API-KEY": "..."},
    )
```
The method and the body are kept, and a identical POST request is made after redirection with code `308`

maudetes (approved):
Looks good, thank you!

Don't forget changelog

PR #2708, opened 2022-02-22
Normalize sort and serializing in search adapters

quaxsze:


maudetes (approved):
Yep, I think we're good

PR #2712, opened 2022-03-10
Move kafka producer to a more generic directory

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/769


udata/kafka/producer.py
```diff
@@ -0,0 +1,45 @@
...
+        return
+
+    if not index:
+        index = topic
+    value = {
+        'service': 'udata',
+        'data': document,
+        'meta': {
+            'message_type': message_type.value,
+            'index': index
```
maudetes, line 40: I wonder, do you think we should set this meta property in kwargs? Because `index` is not generic to all messages produced.

udata/kafka/producer.py
```diff
@@ -0,0 +1,45 @@
+import json
+from flask import current_app
+from kafka import KafkaProducer
+from udata.models import db, Dataset, Organization, Reuse
```
maudetes, line 4:
```suggestion
from udata.models import Dataset, Organization, Reuse
```

maudetes (commented):
Thank you! :pray:

How about renaming this folder `event` instead of `kafka`?

I think we should modify kafka imports in udata/search/commands.py as well.

udata/kafka/producer.py
quaxsze: True. `message_type` either as its for now:
```
    INDEX = 'index'
    REINDEX = 'reindex'
    UNINDEX = 'unindex'
```
But I guess more types will be added in the future?

maudetes: Hmm `message_type` is generic and can be left in `produce` function I guess.
I meant the `index` key, or additional arbitrary meta keys.

maudetes (approved):
I think we're good :)

PR #2713, opened 2022-03-11
Fix etalab/data.gouv.fr#659

ThomasG77:
Nécessite relecture des traductions en plus de l'aspect technique.

### DIFF
--- a/udata/translations/fr/LC_MESSAGES/udata.po
+++ b/udata/translations/fr/LC_MESSAGES/udata.po
@@ -169,7 +169,7 @@ msgstr "Le mainteneur officiel de la licence"
 
 #: udata/core/dataset/api_fields.py:25
 msgid "The license official URL"
-msgstr "L'URL officiel de la licence"
+msgstr "L'URL officielle de la licence"
 
 #: udata/core/dataset/api_fields.py:26
 msgid "Some arbitry flags"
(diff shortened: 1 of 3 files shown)


udata/translations/fr/LC_MESSAGES/udata.po
```diff
@@ -147,6 +147,284 @@ msgstr "a mis à jour un jeu de données"
...
+msgid "The resource title"
+msgstr "Le titre de la ressource"
+
+#: udata/core/dataset/api_fields.py:24
+msgid "The license official maintainer"
+msgstr "Le mainteneur officiel de la licence"
+
+#: udata/core/dataset/api_fields.py:25
+msgid "The license official URL"
+msgstr "L'URL officiel de la licence"
```
ThomasG77, line 172:
```suggestion
msgstr "L'URL officiel de la licence"
```

maudetes (commented):
Thank you for this PR! It seems to work!
If we add translations for dataset api fields, we could apply the same logic to all api fields defined. I don't think it is worth the trouble though.

I haven't checked the translations yet, but for information, we use crowdin to fill translations after extraction. See https://crowdin.com/project/udata. Since you've already made the translations here, you won't need to use crowdin though.

ThomasG77:
Crowdin not really different from Transifex. Not using it but would not be very different.

The PR fixes the issue because in the page, the dataset description fields are used in the form hovering.
So, as you stated, fixing all the api_fields in all entries would be the complete solution.
On short term, you may accept the PR and open a new card with low priority? We can always make the effort if required later as this PR is more or less the recipe for all API  endpoints

ThibaudDauce:
I think this issue will be fixed in the frontend for dataset creation in `udata-front`. Feel free to open a new issue or a new PR if it's still a problem (the issue https://github.com/etalab/data.gouv.fr/issues/659 was closed and the follow-up https://github.com/etalab/data.gouv.fr/issues/859 was marked as resolved)

PR #2715, opened 2022-03-14
Extend dcat properties support

maudetes:
Add support for frequency label, creation and modification date and use landing page property as remote url.

Fix https://github.com/etalab/data.gouv.fr/issues/767
Fix https://github.com/etalab/data.gouv.fr/issues/766
Fix https://github.com/etalab/data.gouv.fr/issues/770

### DIFF
--- a/udata/core/dataset/rdf.py
+++ b/udata/core/dataset/rdf.py
@@ -22,7 +22,7 @@ from udata.rdf import (
 from udata.utils import get_by, safe_unicode
 from udata.uris import endpoint_for
 
-from .models import Dataset, Resource, Checksum, License
+from .models import Dataset, Resource, Checksum, License, UPDATE_FREQUENCIES
 
 log = logging.getLogger(__name__)
 
@@ -299,6 +299,9 @@ def frequency_from_rdf(term):
             term = URIRef(uris.validate(term))
         except uris.ValidationError:
             pass
+    if isinstance(term, Literal):
+        if term.toPython() in UPDATE_FREQUENCIES:
+            return term.toPython()
     if isinstance(term, RdfResource):
         term = term.identifier
     if isinstance(term, URIRef):
--- a/udata/core/dataset/rdf.py
+++ b/udata/core/dataset/rdf.py
@@ -421,6 +421,14 @@ def dataset_from_rdf(graph, dataset=None, node=None):
     if isinstance(d.identifier, URIRef):
         dataset.extras['uri'] = d.identifier.toPython()
 
+    landing_page = url_from_rdf(d, DCAT.landingPage)
+    if landing_page:
+        try:
+            uris.validate(landing_page)
+            dataset.extras['remote_url'] = landing_page
+        except uris.ValidationError:
+            pass
+
     dataset.temporal_coverage = temporal_from_rdf(d.value(DCT.temporal))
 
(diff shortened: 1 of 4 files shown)


udata/core/dataset/rdf.py
```diff
@@ -400,6 +403,8 @@ def dataset_from_rdf(graph, dataset=None, node=None):
     dataset.title = rdf_value(d, DCT.title)
     dataset.description = sanitize_html(d.value(DCT.description))
     dataset.frequency = frequency_from_rdf(d.value(DCT.accrualPeriodicity))
+    dataset.created_at = rdf_value(d, DCT.issued, dataset.created_at)
+    dataset.last_modified = rdf_value(d, DCT.modified, dataset.last_modified)
```
abulte, line 407: A few thoughts/questions about those:
- no safeguard? Ie, is it a date, is it defined...
- more meta: do we still have a "machine" created and modified dates available on datasets if we override those? I don't think we can afford to loose those totally.

udata/harvest/tests/dcat/geonetwork.xml
```diff
@@ -41,10 +41,10 @@
           /
           2005-03-30T00:00:00+02:00</dct:temporal>
       <dct:issued xmlns:dct="http://purl.org/dc/terms/">2004-11-03</dct:issued>
-      <dct:updated xmlns:dct="http://purl.org/dc/terms/">2021-05-05</dct:updated>
+      <dct:modified xmlns:dct="http://purl.org/dc/terms/">2021-05-05</dct:modified>
```
abulte, line 44: Should we modify that test file? I got it from a "real" geonetwork dump. Also, shouldn't we support `dct:updated` too ?

Idea: create a new test file from the real stream of OREME.

maudetes: Yep to create a new test file from stream OREME. They also use `dct:updated` instead of `dct:modified` though.

Should we support terms that are not documented in wc3: https://w3c.github.io/dxwg/dcat/#Property:resource_update_date?

udata/core/dataset/rdf.py
maudetes: If undefined, the default value is used in  `rdf_value`. But indeed, there is no additional safeguard. A similar logic is used [for resources](https://github.com/opendatateam/udata/blob/7f8f9b8b6d6882f2fd51d48021a5b672bf9e7ab1/udata/core/dataset/rdf.py#L380).

I'm not clear with the different dates properties but it seemed to be the case for resources in DCAT and also in [udata-ods](https://github.com/opendatateam/udata-ods/blob/168ec81c097ffa06a362aaff977b79317329996b/udata_ods/harvesters.py#L158) and [udata-ckan](https://github.com/opendatateam/udata-ckan/blob/a7c5f4e67311152b066d697fc8899d5941b1f6d4/udata_ckan/harvesters.py#L155) as well. I went with the same logic, but indeed, we can discuss this in https://github.com/etalab/data.gouv.fr/issues/720

udata/harvest/tests/dcat/geonetwork.xml
maudetes: An extract from the OREME catalog to date was added as a text example.

abulte (approved):

PR #2718, opened 2022-03-25
Fix api v1 breaking change

maudetes:
When moving to api v2 for search architecture, existing list routes (`/api/1/datasets`) now use mongo list instead of a complex search query (available at `/api/2/datasets/search`). Existing filters of `/api/1/datasets` (`reuses` and `organizations` as well) have been dropped and only `sort` has been kept. See the [actual reference](https://doc.data.gouv.fr/api/reference/#/datasets/list_datasets).

This breaking change now impacts the reuse list on a dataset page in admin. Indeed, the `dataset` filter on the `api/1/reuses/?dataset=dataset_id` is not available. Ex: https://dev-02.data.gouv.fr/fr/admin/dataset/6124f0769c35f96b7d71feae/.

We need to discuss our strategy regarding this breaking change. The solutions could be:

1. let the breaking change and implement missing filter for admin only -> breaking change, small work
2. move `api/2/dataset/search` to `api/1/dataset` and remove mongo list -> small breaking change (not 100% iso), small work
3. re-implement **useful** filters on `/api/1/dataset` with the mongo list -> small breaking change, small work
4. re-implement **all** existing filters on `/api/1/dataset` with the mongo list -> no breaking change, considerable work
5. ... ?

We need to take into consideration:
- the impact of a breaking change -> what are the actual api usage ?
- code duplication
- workload
- logical responsibility of the code (difference between mongo list and elastic search)

This PR proposes a fix for this issue with solution 1.

### DIFF
--- a/udata/core/dataset/api.py
+++ b/udata/core/dataset/api.py
@@ -87,8 +87,8 @@ class DatasetApiParser(ModelApiParser):
         self.parser.add_argument('schema', type=str, location='args')
         self.parser.add_argument('schema_version', type=str, location='args')
 
-    @classmethod
-    def parse_filters(cls, datasets, args):
+    @staticmethod
+    def parse_filters(datasets, args):
         if args.get('q'):
             datasets = datasets.search_text(args['q'])
         if args.get('tag'):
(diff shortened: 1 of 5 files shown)


maudetes:
- [ ] Récupérer des stats sur l'utilisation actuelle

quaxsze (approved):

PR #2720, opened 2022-03-28
delete post save after add resource

quaxsze:


udata/core/dataset/models.py
```diff
@@ -697,7 +692,7 @@ def update_resource(self, resource):
         }
         self.update(**data)
         self.reload()
-        post_save.send(self.__class__, document=self)
+        self.on_resource_updated.send(document=self, resource_id=resource.id)
 
     def remove_resource(self, resource):
```
maudetes, line 697: Don't we want a `on_resource_deleted` signal as well? Or should we add it only when we need it?

maudetes (commented):
Thank you for this PR!

I think this fixes the issue of sending a post_save (and indexation) twice on adding a resource. However, when a user adds a resource in admin, he then has a step to modify the resource metadata, resulting in another signal being sent anyway.
We need to take this into consideration in the future when dealing with resources events.

udata/core/dataset/models.py
quaxsze: Maybe it's better to keep as few signals as possible?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Current (in progress)
 
+- Removed `post_save` signal emition within `add_resource` and `update_resource` methods. [#2720](https://github.com/opendatateam/udata/pull/2720)
```
maudetes, line 5:
```suggestion
- Removed `post_save` signal within `add_resource` and `update_resource` methods. [#2720](https://github.com/opendatateam/udata/pull/2720)
```

maudetes (approved):

PR #2728, opened 2022-04-12
Use mongoDB query instead of ES when no setting provided

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/804


quaxsze:
Looking for a possible other way of coding this.

quaxsze:
First draft of refactoring. WDYT @abulte @maudetes ?

quaxsze:
@maudetes

udata/search/result.py
```diff
@@ -48,14 +48,14 @@ def get_ids(self):
             return []
 
     def get_objects(self):
-        if not self._objects:
+        if not self.mongo_objects:
```
maudetes, line 51: Called when mongo_objects is None because no result has been found, resulting in `get_ids` getting called and crashing.

maudetes (commented):
Seems like a good compromise, to have working udata search without our search service architecture.

You need to add a `mongo_search` class method in reuses/search.py as well :)

udata/settings.py
```diff
@@ -29,7 +29,7 @@ class Defaults(object):
     MONGODB_CONNECT = False  # Lazy connexion for Fork-safe usage
 
     # Search service configuration
-    SEARCH_SERVICE_API_URL = 'http://127.0.0.1:5000/api/1/'
+    SEARCH_SERVICE_API_URL = ''
```
maudetes, line 32: Use `None` instead?

```suggestion
    SEARCH_SERVICE_API_URL = None
```

maudetes (approved):
:clap:

PR #2729, opened 2022-04-21
Improve extension detection

maudetes:
Fix https://github.com/etalab/data.gouv.fr/issues/632.

- [x] we can add a migration script to modify stock, should we?
- [x] is it a bad idea to use `current_app` in utils.py? -> I guess not
- [x] check ALLOWED_RESOURCES_EXTENSIONS actual overrides -> ALLOWED_RESOURCES_EXTENSIONS is extended but not overridden.

### DIFF
--- a/udata/core/storages/utils.py
+++ b/udata/core/storages/utils.py
@@ -51,10 +51,10 @@ def extension(filename):
     If it is in ALLOWED_RESOURCES_EXTENSIONS, we add it to the extension.
 
     Some examples of extension detection:
-    - test.unknown -> unkonwn.
-    - test.2022.zip -> zip.
-    - test.2022.csv.tar.gz -> csv.tar.gz.
-    - test.geojson.csv -> csv.
+    - test.unknown -> unknown
+    - test.2022.zip -> zip
+    - test.2022.csv.tar.gz -> csv.tar.gz
+    - test.geojson.csv -> csv
     '''
     filename = os.path.basename(filename)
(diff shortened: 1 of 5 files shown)


udata/settings.py
```diff
@@ -255,15 +255,19 @@ class Defaults(object):
     FS_IMAGES_OPTIMIZE = True
 
     # Default resources extensions whitelist
-    ALLOWED_RESOURCES_EXTENSIONS = [
+
+    ALLOWED_ARCHIVED_EXTENSIONS = [
```
abulte, line 259:
```suggestion
    ALLOWED_ARCHIVE_EXTENSIONS = [
```

abulte (approved):
Maybe explain the process with examples in the docstring of `extension`

udata/core/storages/utils.py
```diff
@@ -44,7 +44,18 @@ def mime(url):
...
 def extension(filename):
-    '''Properly extract the extension from filename'''
+    '''
+    Properly extract the extension from filename.
+    We keep the last extension except for archive extensions, where we check
+    previous extensions as well.
+    If it is in ALLOWED_RESOURCES_EXTENSIONS, we add it to the extension.
+
+    Some examples of extension detection:
+    - test.unknown -> unkonwn.
```
abulte, line 54:
```suggestion
    - test.unknown -> unknown.
```

abulte (approved):

PR #2732, opened 2022-05-04
Add endpoint to directly get specific resource by rid

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/824


udata/core/dataset/apiv2.py
```diff
@@ -217,3 +217,12 @@ def get(self, dataset):
...
             'total': len(res),
         }
+
+
+@ns.route('/resources/<uuid:rid>/', endpoint='resource')
+class ResourceAPI(API):
+    @apiv2.doc('get_resource')
+    @apiv2.marshal_with(resource_fields)
+    def get(self, rid):
+        return get_resource(rid)
```
maudetes, line 227: I think we should add `dataset id` in return value, for ulterior api calls.

udata/tests/apiv2/test_datasets.py
```diff
@@ -29,6 +29,18 @@ def test_get_dataset(self):
 
 class DatasetResourceAPIV2Test(APITestCase):
 
+    def test_get_specific(self):
+        '''Should fetch 1 page of resources from the API'''
```
maudetes, line 33:
```suggestion
        '''Should fetch serialized resource from the API based on rid'''
```

maudetes (commented):
Nice!

udata/core/dataset/apiv2.py
quaxsze: tell me WDYT of the new implem

udata/tests/apiv2/test_datasets.py
```diff
@@ -29,6 +29,19 @@ def test_get_dataset(self):
 
 class DatasetResourceAPIV2Test(APITestCase):
 
+    def test_get_specific(self):
```
maudetes, line 32: We could add a test for 404 and CommunityResource maybe?

maudetes:
On the WeakRef error, see the discussions [here](https://github.com/opendatateam/udata-ods/pull/218).
We don't have a clean solution to access the parent from an embedded document. See https://github.com/MongoEngine/mongoengine/issues/2492.

I suggest to fix the WeakRef error by marshaling manually the response, while `dataset` is in the scope.
I also added a fallback on `resource.dataset()` to return the parent `dataset` using a query if the WeakRef is failing.

maudetes (approved):
I think it's good for me, if you're okay with my suggestion :)

PR #2739, opened 2022-05-11
Fix test and size in suggests

quaxsze:
Fix tags, mime and format suggest size.


udata/tests/api/test_tags_api.py
```diff
@@ -31,7 +31,7 @@ def test_suggest_tags_api(self, api):
 
         for suggestion in response.json:
             assert 'text' in suggestion
-            assert suggestion['text'].startswith('test')
+            assert 'tes' in suggestion['text']
```
maudetes, line 34: `startswith` is used in other suggest test. The same issue could happen there as well?
Ex: test_suggest_datasets_api, test_suggest_organizations_api...

udata/core/tags/api.py
```diff
@@ -24,5 +24,5 @@ def get(self):
         '''Suggest tags'''
         args = parser.parse_args()
         q = normalize(args['q'])
```
maudetes, line 26: I don't think we should normalize tags in tag suggest. It returns unrelevant results by changing the query from `'co'` to `''`.

maudetes (commented):
Thank you for this fix!

udata/core/tags/api.py
quaxsze: This seems to be a wanted behaviour as the function returns a value of `''` if the query is inferior to `current_app.config['TAG_MIN_LENGTH']`
```
if len(value) < MIN_TAG_LENGTH:
    value = ''
```

maudetes: Agreed, but I don't think we want this normalization in the case of suggest.

maudetes (approved):
Thank you!

PR #2744, opened 2022-05-28
Fixed typo in commands in quickstart.md

restuccia:
The command is `docker compose` instead of `docker-compose`


maudetes (commented):
It actually depends on the compose version.
We should start updating the documentation (with a comment I guess for people still using the binary), but I suggest to change it everywhere at once in the doc.

I can do the change if you want :)

ThibaudDauce:
Did the full replacement in https://github.com/opendatateam/udata/pull/3009. Thanks!

PR #2749, opened 2022-06-16
Remove map related stuff

abulte:
This removes map related widgets, configuration...

In the admin, the map is replaced by textual informations.

When a `spatial.geom` is set:

Capture d’écran 2022-06-17 à 10 57 50

When `spatial.zones` are set:

Capture d’écran 2022-06-17 à 10 51 27

We're loosing a bit of information, especially when using a custom `geom` (683 visible datasets). I think the trade-off is worth it, especially since we're removing a dependency to an external service which I know nothing about (cartodb).

In the future, we might want to implement a new widget based on openmaptiles.geo.data.gouv.fr.

For reference, the "before" state:

Capture d’écran 2022-06-17 à 12 20 15

### DIFF
--- a/js/components/dataset/details.vue
+++ b/js/components/dataset/details.vue
@@ -19,17 +19,24 @@
     </div>
     <div v-if="dataset.spatial && dataset.spatial.granularity" class="label-list">
         <strong>
-            <span class="fa fa-clock-o fa-fw"></span>
+            <span class="fa fa-fw fa-bullseye"></span>
             {{ _('Territorial coverage granularity') }}:
         </strong>
         {{ dataset | granularity_label }}
     </div>
     <div v-if="spatialCoverage | length" class="label-list">
         <strong>
-            <span class="fa fa-fw fa-tags"></span>
+            <span class="fa fa-fw fa-bullseye"></span>
             {{ _('Territorial coverage') }}:
         </strong>
         <span v-for="zone in spatialCoverage" class="label label-default">{{zone}}</span>
+    </div>
+    <div v-if="dataset.tags | length" class="label-list">
+        <strong>
+            <span class="fa fa-fw fa-tags"></span>
+            {{ _('Tags') }}:
+        </strong>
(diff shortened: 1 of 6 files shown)


js/config.js
```diff
@@ -134,11 +134,6 @@ export const hidpi = (window.devicePixelRatio > 1 || (
     && window.matchMedia('(-webkit-min-device-pixel-ratio: 1.25),(min-resolution: 120dpi)').matches)
 );
 
-/**
- * Map configuration
- */
-export const map = _jsonMeta('map-config');
```
maudetes, line 140: You should also remove it [there](https://github.com/opendatateam/udata/blob/a8500d439f9c39d296a233183caaca48e4bbfa90/js/config.js#L187)

maudetes (commented):
`leaflet-map.vue`in admin depends on map configs and now fails

js/components/leaflet-map.vue
```diff
@@ -1,70 +0,0 @@
-<template>
```
maudetes, line 1: Could we remove leaflet package and related css entirely?

js/components/dataset/details.vue
```diff
@@ -17,12 +17,19 @@
...
+            <span class="fa fa-clock-o fa-fw"></span>
+            {{ _('Territorial coverage granularity') }}:
+        </strong>
+        {{ dataset | granularity_label }}
+    </div>
+    <div v-if="spatialCoverage | length" class="label-list">
         <strong>
             <span class="fa fa-fw fa-tags"></span>
-            {{ _('Tags') }}:
+            {{ _('Territorial coverage') }}:
```
maudetes, line 30: RIP to the tags? :p

maudetes (approved):
Seems good!

Don't forget the changelog entry.

js/components/dataset/details.vue
abulte: Yeah, tags are overrated :p

maudetes (approved):

PR #2750, opened 2022-06-23
Add protected extras

maudetes:
Originated from https://github.com/etalab/data.gouv.fr/issues/838, but has been treated using https://github.com/opendatateam/udata/pull/2762 instead for now.

The need for a protected extras still exist though.

# Questions and TODOs
- [ ] resource protected extras route needs dataset id -> but not always available
- [ ] how to register additionnal extras (used in plugins for example)
- [ ]  add protected_extras to serialize in dataset/events.py and search
- [ ] should we add protected_extras on discussions, users and orga ?
- [ ]  move all to Harvest (remote_url, dct:identifier, etc..) ?
- [ ] what about linkchecker extras ?
- [ ] what about csv-export:model ?


abulte (commented):
Congrats! Huge and tedious work :-)

I wonder if we could use a different pattern to access/write the protected extras. The dict pattern is not easy to read. Since we're using embedded documents, couldn't we use a XXXProtectClass(x=y) pattern? And/or maybe some kind of proxy or factory in order to have a more centralised view of who is writing what in the extras instead of hunting it down in the business logic.

I didn't think this through and I'm a bit foggy today 🌁, I'll let you know if I can make it clearer :-)

abulte:
Extra thought: since we're using HATEOAS, maybe we don't even need an embedded and can use a full-fledged Document? It it helps making the code more readable, obviously.

maudetes:
Done in https://github.com/opendatateam/udata/pull/2762

PR #2751, opened 2022-06-23
Clean up event code

maudetes:
Replace https://github.com/opendatateam/udata/pull/2745.

Close https://github.com/etalab/data.gouv.fr/issues/838.

Augmented data will be stored in extras discussed in https://github.com/etalab/data.gouv.fr/issues/728.

They will be moved to protected extras following https://github.com/opendatateam/udata/pull/2750.

---

Edit: Logic was moved to https://github.com/etalab/udata-event-consumer.
This PR only adds some clean-up on events.


udata/core/dataset/event.py
```diff
@@ -0,0 +1,58 @@
+import logging
+from uuid import UUID
+
+from .models import get_resource
+log = logging.getLogger(__name__)
+
+
+def consume_message_resource_analysed(key, value):
+    '''
+    Reads a message and update the resource extras with the payload values.
```
quaxsze, line 10: The three docstrings are the same could it be more explicits? The difference between the threee functions was not really straightforward to me 😬

quaxsze (approved):

PR #2752, opened 2022-07-04
Normalize score between 0 and 1

maudetes:
This PR is a suggestion to improve score normalization logic.

Normalization is based on max score.
Max score is not configurable anymore, but is set in Dataset model directly.

What do you think of this logic @quaxsze?


quaxsze:
Love the idea and the way it works 👏

quaxsze (approved):

PR #2753, opened 2022-07-05
Flask-restplus removal in apiv2

quaxsze:


quaxsze:
Upgraded `flask-security-too`. Migration is needed for the new mandatory `fs_uniquifier` attribute in the `User` model.

PR #2755, opened 2022-07-19
APIv2 restplus free

quaxsze:


maudetes (commented):
A very first batch of review

maudetes (commented):
We have an issue related on a hack to use a custom page size on search v2 endpoints.
It now fails with the new apiv2, could you take a look? https://github.com/etalab/data.gouv.fr/issues/880

udata/api/fields.py
```diff
@@ -90,3 +91,71 @@ def pager(page_fields):
             description='The previous page URL if exists')
     }
     return pager_fields
+
+
+class MaImageField(ma_fields.Field):
```
maudetes, line 96: Maybe we want to change the `Ma*Field` pattern, but don't have perfect ideas. How about `Marsh*Field`

udata/api/specs.py
```diff
@@ -0,0 +1,33 @@
...
+from udata.core.dataset.apiv2 import get_dataset_search, get_resources_paginated, get_specific_dataset_by_id, get_specific_resource_by_rid
+from udata.core.organization.apiv2 import get_organization_search
+from udata.core.reuse.apiv2 import get_reuse_search
+from udata.core.user.apiv2 import get_users_list, post_new_user
+
+
+def init_app(app):
+    app.config.update({
+        'APISPEC_SPEC': APISpec(
+            title="Udata",
```
maudetes, line 15:
```suggestion
            title="udata",
```

udata/api/specs.py
```diff
@@ -0,0 +1,33 @@
...
+            version="1.0.0",
+            openapi_version="3.0.2",
+            plugins=[MarshmallowPlugin()],
+        ),
+        'APISPEC_SWAGGER_URL': '/swagger/',
+        'APISPEC_SWAGGER_UI_URL': '/swagger-ui/',
```
maudetes, line 21: I think we don't need a swagger ui, if I understand correctly, could we deactivate it?

udata/api/specs.py
```diff
@@ -0,0 +1,33 @@
...
+
+def init_app(app):
+    app.config.update({
+        'APISPEC_SPEC': APISpec(
+            title="Udata",
+            version="1.0.0",
+            openapi_version="3.0.2",
+            plugins=[MarshmallowPlugin()],
+        ),
+        'APISPEC_SWAGGER_URL': '/swagger/',
```
maudetes, line 20: Should we set it to be on `api/2` instead?

udata/api/specs.py
```diff
@@ -0,0 +1,33 @@
...
+        'APISPEC_SWAGGER_UI_URL': '/swagger-ui/',
+    })
+
+    docs = FlaskApiSpec(app)
+    with app.app_context():
+        docs.register(get_dataset_search, endpoint='apiv2.dataset_search')
```
maudetes, line 26: I'm wondering if we could use a decorator instead, so that we don't forget to add this doc register?

udata/core/user/apiv2_schemas.py
```diff
@@ -0,0 +1,40 @@
...
+    id = fields.Str(dump_only=True)
+    first_name = fields.Str(dump_only=True)
+    last_name = fields.String(dump_only=True)
+    slug = fields.String(required=True)
+    about = fields.String()
+    page = MaURLFor(endpoint='users.show', mapper=lambda u: {'user': u}, fallback_endpoint='api.user', dump_only=True)
+    uri = MaURLFor(endpoint='api.user', mapper=lambda u: {'user': u}, dump_only=True)
+    avatar = MaImageField(dump_only=True)
+    avatar_thumbnail = MaImageField(dump_only=True, attribute='avatar', size=BIGGEST_AVATAR_SIZE)
+    email = fields.Function(lambda o: o.email if current_user_is_admin_or_self() else None, dump_only=True),
```
maudetes, line 32: email is not showing somehow

udata/api/specs.py
maudetes: Actual swagger url is `/api/1/swagger.json` for api v1

udata/api/specs.py
maudetes: We actually could have one to access it locally or on dev env.
I set it back, but it showed "No API definition provided.". I tried to look into it, but did not succeed. I think we can do without it for now but should keep it in mind.

maudetes (commented):
Nice! I find the code readable and it seems to work nicely. :clap: 

I have some comments, but mostly non blocking ones (except for the one on `secure`).

I wish we could have more tests on these endpoints, could we duplicate some tests that exists for api v1 maybe?

udata/api/fields.py
```diff
@@ -90,3 +91,67 @@ def pager(page_fields):
...
+            return
+        if self.size:
+            return value(self.size, external=True)
+        else:
+            return value.fs.url(value.filename, external=True)
+
+
+class MarshURLFor(ma_fields.Field):
+
+    _CHECK_ATTRIBUTE = False
```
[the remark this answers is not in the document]
quaxsze: I quote the doc:
```
Some fields, such as Method fields and Function fields, are not expected
to exist as attributes on the objects to serialize. Set this to False
for those fields
```

udata/api/__init__.py
```diff
@@ -302,14 +259,162 @@ def marshal_page_with(func):
...
+@apiv2_blueprint.route('/', endpoint='root')
+def apiv2_root():
+    '''
+    Useful to return the api v2 root url
+    '''
+    abort(404)
+
+
+@apiv2_blueprint.before_request
+def authentifyV2():
```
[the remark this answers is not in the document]
quaxsze: Before every request, as it is for APIv1

udata/core/user/apiv2.py
```diff
@@ -0,0 +1,59 @@
...
+        else:
+            return search_users.order_by('$text_score').paginate(kwargs['page'], kwargs['page_size'])
+    if kwargs.get('sort'):
+        return users.order_by(kwargs['sort']).paginate(kwargs['page'], kwargs['page_size'])
+    return users.order_by(DEFAULT_SORTING).paginate(kwargs['page'], kwargs['page_size'])
+
+
+@apiv2.route('/users/', endpoint='post_user', methods=['POST'])
+@csrf.exempt
+@UDataApiV2.secure
```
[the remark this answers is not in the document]
maudetes: My bad, it was my request app that remembered cookies of previous requests :p
It correctly returns a 401 :)

udata/tests/apiv2/test_datasets.py
```diff
@@ -81,7 +81,7 @@ def test_get_missing_param(self):
 
     def test_get_next_page(self):
         '''Should fetch 2 pages of resources from the API'''
-        resources = [ResourceFactory() for _ in range(80)]
+        resources = [ResourceFactory() for _ in range(40)]
```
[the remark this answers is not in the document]
abulte: 👍 probably because of https://github.com/opendatateam/udata/pull/2755/files#diff-a62c9bd3310470c9b7bfe01b30f76621a938f5d5be834d2b479dafd146b3d154R16 ?

udata/tests/apiv2/test_datasets.py
```diff
@@ -96,7 +96,6 @@ def test_get_next_page(self):
         response = self.get(data['next_page'])
         self.assert200(response)
         data = response.json
-        assert len(data['data']) == len(resources) - DEFAULT_PAGE_SIZE
```
abulte, line 99: This should still pass?

abulte (commented):
This is cool!

We need some tests on the new users v2 API I think

PR #2761, opened 2022-08-22
Fix image URLs for suggest endpoints

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/883

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Clean up event code [#2751](https://github.com/opendatateam/udata/pull/2751)
 - Replace mongo legacy image in CI [#2754](https://github.com/opendatateam/udata/pull/2754)
+- Fixes test `test_suggest_datasets_api` by modifying condition [#2759](https://github.com/opendatateam/udata/pull/2759)
 
 ## 4.1.1 (2022-07-08)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Clean up event code [#2751](https://github.com/opendatateam/udata/pull/2751)
 - Replace mongo legacy image in CI [#2754](https://github.com/opendatateam/udata/pull/2754)
 - Fixes test `test_suggest_datasets_api` by modifying condition [#2759](https://github.com/opendatateam/udata/pull/2759)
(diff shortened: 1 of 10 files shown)


udata/core/user/api.py
```diff
@@ -359,7 +359,7 @@ def get(self):
                 'id': user.id,
                 'first_name': user.first_name,
                 'last_name': user.last_name,
-                'avatar_url': user.avatar_url,
+                'avatar_url': user.avatar,
```
maudetes, line 362: Isn't it always `null`?

udata/core/dataset/api.py
```diff
@@ -615,7 +615,7 @@ def get(self):
                 'title': dataset.title,
                 'acronym': dataset.acronym,
                 'slug': dataset.slug,
-                'image_url': dataset.image_url,
+                'image_url': dataset.organization.logo.filename if dataset.organization else dataset.owner.avatar.filename,
```
maudetes, line 618: It should'nt be filename but the full url though.

maudetes (commented):
Do we want to return the `_thumbnail` image url on all?

udata/core/dataset/api.py
quaxsze: The marshalling will already build the URL it needs only the filename in order to make a valid url.
If passed the full url the result is something like
`https://static.data.gouv.fr/https://static.data.gouv.fr/avatars/8a/4e0caeb30843a08eca1a4a54e025c0-original.png`

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Clean up event code [#2751](https://github.com/opendatateam/udata/pull/2751)
 - Replace mongo legacy image in CI [#2754](https://github.com/opendatateam/udata/pull/2754)
 - Fixes test `test_suggest_datasets_api` by modifying condition [#2759](https://github.com/opendatateam/udata/pull/2759)
+- Fix image URLs for suggest endpoints [#2761](https://github.com/opendatateam/udata/pull/2761)
```
maudetes, line 12: Pay attention when merging with master, there has been a release since.

maudetes (approved):
I find the way the user avatar works weird (see my comment), but this is out of scope of this PR :)

Thank you for the fix!

PR #2769, opened 2022-10-03
Switch to pyproject.toml

quaxsze:
Fix https://github.com/etalab/data.gouv.fr/issues/939


maudetes (commented):
I may miss something but it seems that we're not specifying requirements in pyproject.toml (we had `install_requires`), thus having only `udata` being install and packaged instead of udata + its dependencies

pyproject.toml
```diff
@@ -0,0 +1,133 @@
...
+    "vine==1.3.0",
+    "voluptuous==0.11.7",
+    "werkzeug==0.16.1",
+    "wtforms-json==0.3.3",
+    "wtforms==2.2.1",
+    "xmltodict==0.12.0"
+]
+
+[project.optional-dependencies]
+dev = [
```
maudetes, line 95: We should probably document how to install these? Or should we install based on requirements/develop.pip ?

maudetes (commented):
Nice!

I have some issues on some deps when running `inv serve`. In particular, I saw mistakes on:
- MarkupSafe
- kombu[redis]

maudetes (commented):
I think we're all good on udata side! :clap:

ThibaudDauce:
Closing for now since the issue was closed

PR #2770, opened 2022-10-05
Switch to flask-rest-x and endpoints enhancement

quaxsze:
Fix https://github.com/etalab/data.gouv.fr/issues/940


maudetes (commented):
Oh yeah! Seems to work nicely, did you check the changelog of flask_restx for any side effect?

- [ ] Could you check for usage in plugins (ex: udata-front) as well
- [x] Don't forget changelog

maudetes (approved):

PR #2780, opened 2022-11-14
Fix harvest uri validation error

maudetes:
`uri` field in resource and dataset harvest metadata should both be `StringField` and not `URLField`.

Make test fail with an URI failing URL validation as distribution identifier and add fix.


abulte (approved):
🎉

PR #2781, opened 2022-11-14
Optimize DCAT harvesting on large multiple-paged catalogs

abulte:
This results in:
- successful preview on large multiple-paged DCAT catalogs
- fastest harvesting times on large multiple-paged DCAT catalogs

This is done through a different structure in `HarvestJob.data`. Previously we were storing the whole concatenated catalog in `HarvestJob.data.graph`, even when parsing multiple pages, and deserialising the whole catalog for each `HarvestItem`. Now we're storing a paged `HarvestJob.data.graphs` with one graph element for each page. Thus we only need to deserialise the right page when processing an `HarvestItem `.

This leads to 2x fastest overall speed for a 500 datasets catalog with five pages. I've also seen harvest process task being 3x fastest on larger catalogs.

This also introduces `HARVEST_MAX_ITEMS` setting, allowing to limit the number of harvested items per source on development platforms.

---

We could probably optimize more, for instance:
- store the relevant part of the graph for each `HarvestIem` before queuing the `harvest_job_item`, but I'm not sure how this would work DCAT-wise: could we have a small enough object? We also need to dig into `dataset_from_rdf` which probably won't be easy. We could also introduce bugs while manipulating the remote catalog, as it is we feed it straight to `dataset_from_rdf`
- create a `PageHarvestJob` with only the graph for the page stored in it, but this is heavy work since we would need to reconcile this with our current architecture and UI
- bulk process `HarvestItem` instead of one by one, but once again heavy work and impact on architecture and UI

For the preview part, we just stop iterating on pagination when we have enough objects. The drawback is that a faulty page somewhere in the catalog won't be catched when preview is ran, but I think we can live with it.

### DIFF
--- a/udata/harvest/tasks.py
+++ b/udata/harvest/tasks.py
@@ -22,10 +22,11 @@ def harvest(self, ident):
     if items > 0:
         finalize = harvest_job_finalize.s(backend.job.id)
         items = [
-            harvest_job_item.s(backend.job.id, item.remote_id)
+            (backend.job.id, item.remote_id)
             for item in backend.job.items
         ]
-        chord(items)(finalize)
+        chunks = harvest_job_item.chunks(items, 10)
+        chord(chunks.group())(finalize)
     elif items == 0:
         backend.finalize()
 
--- a/udata/harvest/tasks.py
+++ b/udata/harvest/tasks.py
@@ -25,8 +25,7 @@ def harvest(self, ident):
             (backend.job.id, item.remote_id)
             for item in backend.job.items
         ]
-        chunks = harvest_job_item.chunks(items, 10)
-        chord(chunks.group())(finalize)
+        chord(items)(finalize)
     elif items == 0:
         backend.finalize()
 
--- a/udata/harvest/tasks.py
+++ b/udata/harvest/tasks.py
@@ -22,7 +22,7 @@ def harvest(self, ident):
     if items > 0:
         finalize = harvest_job_finalize.s(backend.job.id)
         items = [
-            (backend.job.id, item.remote_id)
+            harvest_job_item.s(backend.job.id, item.remote_id)
             for item in backend.job.items
         ]
         chord(items)(finalize)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
   - Exhaustive list of changes [here](https://flask-security-too.readthedocs.io/en/stable/changelog.html#version-4-0-0).
 - Fix apiv2 swagger with harvest metadata and add apiv2 swagger tests [#2782](https://github.com/opendatateam/udata/pull/2782)
 - Add quality score to csv catalogs [#2785](https://github.com/opendatateam/udata/pull/2785)
-- Optimize DCAT harvesting on large multiple-paged catalogs [#2781](https://github.com/opendatateam/udata/pull/2781)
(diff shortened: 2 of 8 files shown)


maudetes:
For info, I came across [this commit](https://github.com/opendatateam/udata/commit/0dbb5e4cf0c51745fb4432034daf7b9f0313f601), that introduced the use of a single complete graph.

abulte:
@maudetes Interesting!

> This also prevent some data loss because properly slicing a graph is difficult in RDF when you don't known by advance all node properties. This will allow to parse more triplets for each dataset and resource.

maudetes (commented):
Nice! It seems to work way faster (I guess the largest the catalog, the best increase) !

What about the risk of having some information missing on the graph page?
Ex a dataset being defined on a page and its distribution on a different page.

abulte:
> What about the risk of having some information missing on the graph page?
> Ex a dataset being defined on a page and its distribution on a different page.

Ow, do you think it's possible? Would be poor design AMHA. I guess we could fallback on the complete graph if we don't find the complete node in the page, but looking at the code it won't be easy (`dataset_from_rdf` -> `resource_from_rdf` mayhem, and then what if the resource is just missing or invalid and we parse the whole catalog for nothing ?). Maybe we can let it fail for now and see what happens in real life.

maudetes:
>

maudetes:
> Ow, do you think it's possible? Would be poor design IMHO.

I've said that because some DCAT catalogs have distribution defined later in the graph.
Ex with this graph https://www.pigma.org/geonetwork/srv/fre/rdf.search, Datasets are defined first, then Distribution later on.
We could imagine having a pagination that cuts between node, ignoring the reference.

But indeed, let's see if it actually happens.
Are you aware of paginated DCAT that we could check? I'm thinking only of the geocatalogue

abulte:
> Ex with this graph https://www.pigma.org/geonetwork/srv/fre/rdf.search, Datasets are defined first, then Distribution later on.

It's the same on Geocatalogue, and it's truncated properly https://www.geocatalogue.fr/www/dcat/154e240559a401aa0c8b3f03dc5836dd/catalog_5.rdf.

> Are you aware of paginated DCAT that we could check? I'm thinking only of the geocatalogue

Looking 👀

abulte:
@maudetes So here's another paginated one: https://data.gov.ie/catalog.xml (anecdoticaly it fails because the whole catalog — 150k datasets can't be stored in mongo). The distribution is embedded in the dataset so no problem with this kind of pagination (if the catalog is not one of a medium-sized country 😅).

udata/harvest/tasks.py
```diff
@@ -22,10 +22,11 @@ def harvest(self, ident):
     if items > 0:
         finalize = harvest_job_finalize.s(backend.job.id)
         items = [
-            harvest_job_item.s(backend.job.id, item.remote_id)
+            (backend.job.id, item.remote_id)
             for item in backend.job.items
         ]
-        chord(items)(finalize)
+        chunks = harvest_job_item.chunks(items, 10)
```
maudetes, line 28: Did you see a difference with chunks?

maudetes (approved):
Seems to work smoothly

udata/harvest/tasks.py
abulte: Nope (should be tested on massive scale), it was a deadend, removed.

PR #2783, opened 2022-11-15
Kafka less

quaxsze:


udata/settings.py
```diff
@@ -419,6 +416,7 @@ class Defaults(object):
 
     FIXTURE_DATASET_SLUGS = []
     PUBLISH_ON_RESOURCE_EVENTS = False
+    RESOURCES_ANALYSER_URI = 'http://localhost:8000'
```
maudetes, line 419: I'm not 100% convinced by this name. It feels like we would call this to get an analysis, whereas our first intention is to notify it from a ressource event.

requirements/install.pip
```diff
@@ -149,8 +149,6 @@ jsonschema==3.2.0
     # via
     #   -r requirements/install.in
     #   flask-restx
-kafka-python==2.0.2
```
maudetes, line 152: Some kafka-python deps still live in doc.pip and report.pip?

udata/search/commands.py
```diff
@@ -62,32 +61,40 @@ def index_model(adapter, start, reindex=False, from_datetime=None):
...
-                kafka_uri=current_app.config.get('KAFKA_URI'),
-                topic=get_topic(message_type),
-                service='udata',
-                key_id=doc['id'],
-                document=doc,
-                meta={'message_type': message_type, 'index': index_name}
-            )
+            url = f"{current_app.config['SEARCH_SERVICE_API_URL']}/reindex"
+            try:
+                payload = {
```
maudetes, line 72: See my related comment on udata-search-service: https://github.com/opendatateam/udata-search-service/pull/35#discussion_r1034948159.
More generally, I find it weird to call reindex with an unindex message type, and an optional new index_name.

maudetes (commented):
Review in progress :)

udata/search/__init__.py
```diff
@@ -1,13 +1,11 @@
...
+import requests
 
 from flask import current_app
 from mongoengine.signals import post_save, post_delete
-from udata_event_service.producer import produce
 
-from udata.models import db, Dataset, Organization, Reuse
+from udata.models import db
 from udata.tasks import task, as_task_param
-from udata.event.values import KafkaMessageType
```
maudetes, line 9: Since we don't import from udata.event here anymore, udata.core.dataset.events is never imported when starting app, thus not connecting signals. It doesn't fail in test, since it is explicitly imported.

udata/settings.py
```diff
@@ -444,6 +442,7 @@ class Testing(object):
     URLS_ALLOWED_TLDS = tld_set | set(['test'])
     URLS_ALLOW_PRIVATE = False
     FS_IMAGES_OPTIMIZE = True
+    PUBLISH_ON_RESOURCE_EVENTS = False
```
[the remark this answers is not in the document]
quaxsze: Yes :)

udata/search/__init__.py
```diff
@@ -22,51 +21,47 @@ def reindex(classname, id):
...
-        message_type = f'{classname.lower()}.{action.value}'
-        produce(
-            kafka_uri=current_app.config.get('KAFKA_URI'),
-            topic=get_topic(message_type),
-            service='udata',
-            key_id=id,
-            meta={'message_type': message_type, 'index': classname.lower()}
-        )
+        url = f"{current_app.config['SEARCH_SERVICE_API_URL']}{adapter_class.search_url}/{str(id)}/unindex"
+        r = requests.delete(url)
```
maudetes, line 50: Does it actually makes sense that we're calling `DELETE` on an endpoint named `unindex`? :p

maudetes (commented):
Works well! Those are mostly cosmetic comments.

udata/search/__init__.py
quaxsze: Well If we refer to REST, it deletes a resource, so...

CHANGELOG.md
```diff
@@ -2,7 +2,10 @@
 
 ## Current (in progress)
 
-- Add dedicated extras endpoints on resources and datasets [#2779](https://github.com/opendatateam/udata/pull/2779)
+- :warning: Kafka removal [#2779](https://github.com/opendatateam/udata/pull/2779):
+  - Resources events publication uses now webhooks
+  - Search indexation (at runtime and with index command) are now made throught HTTP requests
+  - Add dedicated extras endpoints on resources and datasets
```
maudetes, line 8:
```suggestion
- :warning: Kafka removal [#2783](https://github.com/opendatateam/udata/pull/2783):
  - Resources events publication now uses webhooks
  - Search indexation (at runtime and with index command) are now made through HTTP requests
- Add dedicated extras endpoints on resources and datasets [#2779](https://github.com/opendatateam/udata/pull/2779)
```

maudetes (approved):
Small comments on changelog

PR #2779, opened 2022-10-25
Add objects endpoints to modify extras

quaxsze:
Fix https://github.com/etalab/data.gouv.fr/issues/983

PR #2784, opened 2022-11-15
Extra endpoints

quaxsze:
Fix https://github.com/etalab/data.gouv.fr/issues/983

Replaces https://github.com/opendatateam/udata/pull/2779


maudetes (approved):
Works nicely!

I still wonder about the expected body not showing in the Swagger definition due to this implementation. But it's non-blocking.

PR #2788, opened 2022-11-29
Add condition to avoid mail sending according to config var

quaxsze:


maudetes (approved):
Thank you for the *quick*fix :pray:

PR #2789, opened 2022-12-05
Add harvest.(created_at,modified_at,backend) to resources CSV export

abulte:


udata/core/dataset/csv.py
```diff
@@ -33,7 +33,11 @@ class DatasetCsvAdapter(csv.Adapter):
         ('tags', lambda o: ','.join(o.tags)),
         ('archived', lambda o: o.archived or False),
         ('resources_count', lambda o: len(o.resources)),
-        ('quality_score', lambda o: format(o.quality['score'], '.2f'))
+        ('quality_score', lambda o: format(o.quality['score'], '.2f')),
```
maudetes, line 36: Maybe leave quality_score at the end? Next to metrics.

udata/tests/dataset/test_csv_adapter.py
```diff
@@ -0,0 +1,64 @@
...
+        # dataset harvest dates should not be here
+        assert another_date.isoformat() not in d_row
+
+    def test_datasets_csv_adapter(self):
+        date_created = datetime(2022, 12, 31)
+        date_modified = date_created + timedelta(days=1)
+        dataset = DatasetFactory(
+            harvest={
+                'domain': 'example.com',
+                'backend': 'dcat',
```
maudetes, line 48: It's actually in capital case :p
```suggestion
                'backend': 'DCAT',
```

udata/tests/dataset/test_csv_adapter.py
```diff
@@ -0,0 +1,64 @@
...
+        DatasetFactory(resources=[ResourceFactory()])
+        adapter = DatasetCsvAdapter(Dataset.objects.all())
+        rows = list(adapter.rows())
+        d_row = [r for r in rows if str(dataset.id) in r][0]
+        # harvest.created_at
+        assert date_created.isoformat() in d_row
+        # harvest.modified_at
+        assert date_modified.isoformat() in d_row
+        # harvest.backend
+        assert 'dcat' in d_row
```
maudetes, line 62:
```suggestion
        assert 'DCAT' in d_row
```

udata/core/dataset/csv.py
```diff
@@ -74,6 +78,8 @@ class ResourcesCsvAdapter(csv.NestedAdapter):
         'created_at',
         'modified',
         ('downloads', lambda o: int(o.metrics.get('views', 0))),
+        ('harvest.created_at', lambda o: o.harvest and o.harvest.created_at),
```
maudetes, line 81: It would be nice to be able to say that a resource has been harvested or not, but I don't think we have a dedicated info at the resource level.

maudetes (commented):
Nice! Thanks for adding generic harvest infos, they seem useful.

An option could be to display a common value for created_at and modified_at that would be based on both model and harvest infos. It may be simpler for reusers but hide some logic (besides we could use the specific harvest info).

udata/tests/dataset/test_csv_adapter.py
abulte: since we don't care for the sake of this test, I made it clear :-)

udata/core/dataset/csv.py
abulte: Yep, but no we don't.

I tried to expose a `dataset.is_harvested` field based on `dataset.harvest.remote_id` presence. It works but it feels flaky. Decided not too :-)

Maybe we could add a dedicated property on `Dataset` with a well defined and contained logic, then expose it to the dataset and/or resources catalog. Feels like another PR though.

abulte:
> An option could be to display a common value for created_at and modified_at that would be based on both model and harvest infos. It may be simpler for reusers but hide some logic (besides we could use the specific harvest info).

I don't feel comfortable with that (yet). Catalog seems pretty low level to me, so not doing too much inference on the meaning of things before exporting seems like a good idea.

maudetes (approved):

PR #2792, opened 2022-12-19
Email change feature

quaxsze:
Fix https://github.com/etalab/data.gouv.fr/issues/992


udata/auth/views.py
```diff
@@ -19,6 +32,91 @@ def slash_url_suffix(url, suffix):
...
+    )
+
+
+def confirm_change_email_token_status(token):
+    expired, invalid, user, token_data = get_token_status(
+        token, 'confirm', 'CONFIRM_EMAIL', return_data=True)
+    new_email = None
+
+    if not invalid and user:
+        user_id, token_email_hash, new_email = token_data
```
maudetes, line 56:
```suggestion
        _, token_email_hash, new_email = token_data
```

udata/auth/views.py
```diff
@@ -19,6 +32,91 @@ def slash_url_suffix(url, suffix):
...
+    if invalid or expired:
+        return redirect(url_for('site.home'))
+
+    if user != current_user:
+        logout_user()
+        login_user(user)
+
+    if user.email == new_email:
+        msg = (
+            'Your change of email has already been confirmed.'
```
maudetes, line 87:
```suggestion
            'Your change of email has already been confirmed.',
```

udata/auth/views.py
```diff
@@ -19,6 +32,91 @@ def slash_url_suffix(url, suffix):
     return url.endswith('/') and ('%s/' % suffix) or ('/%s' % suffix)
 
 
+def send_change_email_confirmation_instructions(user, new_email):
```
maudetes, line 35: Do we want to add some type hints? :)

udata/auth/views.py
```diff
@@ -19,6 +32,91 @@ def slash_url_suffix(url, suffix):
...
+        login_user(user)
+
+    if user.email == new_email:
+        msg = (
+            'Your change of email has already been confirmed.'
+            'info')
+    else:
+        user.email = new_email
+        _datastore.put(user)
+        msg = (
```
maudetes, line 92: I'm wondering if we should we send a success email as well? (similar to what is done when you reset your password?)

maudetes (commented):
Good work, it seems to work smoothly! :)

Could you add some tests on this new feature? :pray: The case where `Your change of email has already been confirmed` was failing for example

udata/auth/views.py
```diff
@@ -19,6 +33,86 @@ def slash_url_suffix(url, suffix):
...
+        confirm_change_email_token_status(token))
+
+    if not user or invalid:
+        invalid = True
+        do_flash(*get_message('INVALID_CONFIRMATION_TOKEN'))
+    if expired:
+        send_change_email_confirmation_instructions(user, new_email)
+        do_flash(*(
+            (
+                'You did not confirm your change of email within {0}. '
```
maudetes, line 74: :warning: Some flash messages haven't been localized in this file

maudetes (approved):
Nice! :pray:

PR #2799, opened 2023-01-12
Fix date timezone and format for harvest previz

quaxsze:


udata/core/dataset/models.py
```diff
@@ -590,7 +590,14 @@ def last_update(self):
             for res in self.resources:
                 dates += [res.modified, res.published]
                 if res.harvest and res.harvest.modified_at:
-                    dates.append(res.harvest.modified_at)
+                    if isinstance(res.harvest.modified_at, date):
```
maudetes, line 593: Could we move this to a small helper? `last_update` now seems very complicated, when it isn't :p

udata/core/dataset/models.py
```diff
@@ -1,6 +1,6 @@
 import logging
 
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, date
```
maudetes, line 3:
```suggestion
from datetime import datetime, timedelta
```

udata/utils.py
```diff
@@ -139,6 +139,17 @@ def daterange_end(value):
         return result.replace(month=12, day=31)
 
 
+def date_or_aware_datetime_to_naive_datetime(given_date):
```
maudetes, line 142: To keep the `to_` pattern, we could maybename it like
```suggestion
def to_naive_datetime(given_date):
```

maudetes (approved):
Perfect :ok_hand:

PR #2800, opened 2023-01-13
Support csw dcat harvest

maudetes:
A proposed solution for https://github.com/etalab/data.gouv.fr/issues/913.

Use [DCAT export served by the CSW endpoint](https://github.com/geonetwork/core-geonetwork/wiki/DCAT-enhancements), available in GeoNetwork for harvesting.

This CSW backend is actually based on the DCAT backend with a different `parse_graph` function.


udata/harvest/backends/dcat.py
```diff
@@ -120,3 +123,62 @@ def process(self, item):
         dataset = self.get_dataset(item.remote_id)
         dataset = dataset_from_rdf(graph, dataset, node=node)
         return dataset
+
+
+class CswDcatBackend(DcatBackend):
+    display_name = 'CSW-DCAT'
+
+    def parse_graph(self, url, fmt):
+        body = '''<csw:GetRecords xmlns:csw="http://www.opengis.net/cat/csw/2.0.2"
```
maudetes, line 132: We could also consider using a lib to query a CSW endpoint, ex https://pypi.org/project/OWSLib.
Maybe we want to keep the dcat content raw instead?

udata/harvest/backends/dcat.py
```diff
@@ -120,3 +123,62 @@ def process(self, item):
...
+
+                for node in subgraph.subjects(RDF.type, DCAT.Dataset):
+                    id = subgraph.value(node, DCT.identifier)
+                    kwargs = {'nid': str(node), 'page': page}
+                    kwargs['type'] = 'uriref' if isinstance(node, URIRef) else 'blank'
+                    self.add_item(id, **kwargs)
+            graphs.append(graph)
+            page += 1
+
+            if int(search_results.attrib['nextRecord']) == 0 or \
```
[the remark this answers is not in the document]
fxprunayre: Sounds good.

See also https://github.com/geonetwork/core-geonetwork/blob/main/harvesters/src/main/java/org/fao/geonet/kernel/harvest/harvester/csw/Harvester.java#L338-L369

udata/harvest/backends/dcat.py
maudetes: Seems like in this example and others I've tried, `csw.records` is always empty with `outputschema='http://www.w3.org/ns/dcat#'` sadly

udata/harvest/backends/dcat.py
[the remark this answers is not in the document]
maudetes: Ah indeed, other edge cases to deal with! :pray:

udata/harvest/backends/dcat.py
ThomasG77: I've tested to run the script and not empty so not sure if there is a misunderstanding or it's related to the owslib version or something else.

ThomasG77: Wrong check on my side. It only works with default outputschema

maudetes: Since owslib doesn't seem to give us a satisfying result, let's keep with a raw query for now

udata/harvest/backends/dcat.py
```diff
@@ -120,3 +123,62 @@ def process(self, item):
...
+class CswDcatBackend(DcatBackend):
+    display_name = 'CSW-DCAT'
+
+    def parse_graph(self, url, fmt):
+        body = '''<csw:GetRecords xmlns:csw="http://www.opengis.net/cat/csw/2.0.2"
+                                  xmlns:gmd="http://www.isotc211.org/2005/gmd"
+                                  service="CSW" version="2.0.2" resultType="results"
+                                  startPosition="{start}" maxPosition="15"
+                                  outputSchema="http://www.w3.org/ns/dcat#">
+                    <csw:Query typeNames="gmd:MD_Metadata">
```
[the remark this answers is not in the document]
maudetes: I've added the sort property as mentioned in https://github.com/opendatateam/udata/pull/2800#discussion_r1129053500

udata/harvest/backends/dcat.py
```diff
@@ -120,3 +121,62 @@ def process(self, item):
...
+                            <Filter xmlns="http://www.opengis.net/ogc"><PropertyIsEqualTo>
+                                <PropertyName>documentStandard</PropertyName>
+                                <Literal>iso19139</Literal>
+                            </PropertyIsEqualTo></Filter>
+                        </csw:Constraint>
+                    </csw:Query>
+                </csw:GetRecords>'''
+        headers = {"Content-Type": "application/xml"}
+
+        content = requests.post(url, data=body.format(start=1), headers=headers).text
```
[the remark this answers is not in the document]
maudetes: Thank you for the inputs!
* We're keeping with the POST request for an improved user experience. We'll mention it in our harvesting doc (as well as an example of an expected endpoint to provide).
* I've removed `documentStandard`, we should probably add a mention about the need to expose a catalog with records that support mapping to DCAT in our documentation also!
* I've increased the page size to 200 and added the sort clause.

quaxsze (approved):

PR #2805, opened 2023-01-17
Handle None values in dataset and resource extras endpoints

abulte:
Fix https://github.com/etalab/data.gouv.fr/issues/1001

- If key does not exist: ignore
- If key exists: delete

### DIFF
--- a/udata/tests/apiv2/test_datasets.py
+++ b/udata/tests/apiv2/test_datasets.py
@@ -199,7 +199,10 @@ class DatasetExtrasAPITest(APITestCase):
         assert data['test::extra'] == 'test-value'
 
     def test_update_dataset_extras(self):
-        self.dataset.extras = {'test::extra': 'test-value', 'test::extra-second': 'test-value-second'}
+        self.dataset.extras = {
+            'test::extra': 'test-value',
+            'test::extra-second': 'test-value-second',
+        }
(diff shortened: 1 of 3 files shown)


quaxsze (approved):

udata/tests/apiv2/test_datasets.py
```diff
@@ -199,22 +199,30 @@ def test_get_dataset_extras(self):
...
         data = ['test::extra-second', 'another::key']
         response = self.put(url_for('apiv2.dataset_extras', dataset=self.dataset), data)
         self.assert400(response)
         assert response.json['message'] == 'Wrong payload format, dict expected'
 
-        data = {'test::extra-second': 'test-value-changed', 'another::key': 'another-value'}
+        data = {
+            'test::extra-second': 'test-value-changed',
+            'another::key': 'another-value',
+            'test::none': None,
```
maudetes, line 216: I don't think we test that it unsets a key here, right?

maudetes (commented):
Thank you!
I've made some minor comments :)

udata/tests/apiv2/test_datasets.py
abulte: True! I added a test.

maudetes (approved):
Nice! :clap:

PR #2807, opened 2023-01-19
Remove useless published date in resource model

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/1033

### DIFF
--- a/udata/core/dataset/rdf.py
+++ b/udata/core/dataset/rdf.py
@@ -256,6 +256,7 @@ def temporal_from_resource(resource):
     '''
     Parse a temporal coverage from a RDF class/resource ie. either:
     - a `dct:PeriodOfTime` with schema.org `startDate` and `endDate` properties
+    - a `dct:PeriodOfTime` with DCAT `startDate` and `endDate` properties
     - an inline gov.uk Time Interval value
     - an URI reference to a gov.uk Time Interval ontology
       http://reference.data.gov.uk/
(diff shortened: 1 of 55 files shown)


udata/core/dataset/rdf.py
```diff
@@ -133,7 +133,6 @@ def resource_to_rdf(resource, dataset=None, graph=None):
     r.add(DCT.description, Literal(resource.description))
     r.add(DCAT.downloadURL, URIRef(resource.url))
     r.add(DCAT.accessURL, URIRef(permalink))
-    r.add(DCT.issued, Literal(resource.published))
```
maudetes, line 136: Maybe keep this line but using `created_at`. Maybe treat it in https://github.com/opendatateam/udata/pull/2810, since we could want to use harvest info as well on the different dates exposed in RDF.

udata/tests/dataset/test_dataset_model.py
```diff
@@ -89,13 +89,6 @@ def test_update_resource_missing_checksum_type(self):
         with pytest.raises(db.ValidationError):
             dataset.update_resource(resource)
 
-    def test_last_update_with_resource(self):
```
maudetes, line 92: I think we should keep this test, but use resource.created_at instead ;)

We could add a case with an harvest date being the max

maudetes (commented):
We need to find a solution on the `published` field disappearing in the model

quaxsze:
`Published` attribute left in the model because of the complexity and the difficulty of the removal/migration. All code logic related to the setting og getting of the attribute was removed

maudetes (approved):
I think we're all clear! :clap:

PR #2808, opened 2023-01-19
Fix only SHA1 checksum is accepted when uploading resources

tboye:
Depending on the storage type and settings, it might happen that resource metadata will provide a non-SHA1 checksum. At the moment it causes an exception. This MR fixes that and makes resource uploader accept all other supported checksum types.

In addition, a similar inconsistency is fixed within CSV export implementation.

### DIFF
--- a/udata/core/dataset/api.py
+++ b/udata/core/dataset/api.py
@@ -29,6 +29,7 @@ from udata.auth import admin_permission
 from udata.api import api, API, errors
 from udata.api.parsers import ModelApiParser
 from udata.core import storages
+from udata.core.dataset.models import CHECKSUM_TYPES
 from udata.core.storages.api import handle_upload, upload_parser
 from udata.core.badges import api as badges_api
 from udata.core.followers.api import FollowAPI
@@ -362,7 +363,10 @@ class UploadMixin(object):
         if 'html' in infos['mime']:
             api.abort(415, 'Incorrect file content type: HTML')
         infos['title'] = os.path.basename(infos['filename'])
-        infos['checksum'] = Checksum(type='sha1', value=infos.pop('sha1'))
+        for checksum_type in CHECKSUM_TYPES:
+            if checksum_type in infos:
+                infos['checksum'] = Checksum(type=checksum_type, value=infos.pop(checksum_type))
+                break
         infos['filesize'] = infos.pop('size')
         del infos['filename']
         return infos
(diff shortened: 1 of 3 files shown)


udata/core/dataset/api.py
```diff
@@ -362,7 +363,10 @@ def handle_upload(self, dataset):
         if 'html' in infos['mime']:
             api.abort(415, 'Incorrect file content type: HTML')
         infos['title'] = os.path.basename(infos['filename'])
-        infos['checksum'] = Checksum(type='sha1', value=infos.pop('sha1'))
+        for checksum_type in CHECKSUM_TYPES:
+            if checksum_type in infos:
+                infos['checksum'] = Checksum(type=checksum_type, value=infos.pop(checksum_type))
+                break
```
maudetes, line 369: `infos['checksum']` may never be set if none of the checksum_type matches and it wouldn't raise an error here.
Using `next()` would raise an error directly instead, what do you think?

```suggestion
        checksum_type = next(algo for algo in CHECKSUM_TYPES if algo in infos)
        infos['checksum'] = Checksum(type=checksum_type, value=infos.pop(checksum_type))
```

maudetes (commented):
Thank you a lot for this contribution! :pray:
I'm making some suggestions here.

I'm also wondering if we could add some relevant tests, but not sure we have fixtures to test a storage returning a checksum other than `sha1` yet.

udata/core/dataset/api.py
tboye: I was actually hesitating whether we want to raise an exception here or wait for the caller to complain, but after all I agree it's best to do it here. Also, looks way more elegant with `next()`.

tboye:
@maudetes Regarding testing, from what I see there's indeed no infrastructure for that at the moment. Do you have any suggestions on how to address that?

maudetes (approved):
Thank you for the update!

Regarding the tests, I think a StorageMock could be created in `test_storages`.
But regarding the new code lines, it may be overkill and I'm okay with merging it without additional tests (`sha1` case is already tested).

maudetes:
> Just to make sure, do changes in `udata/core/dataset/tasks.py` also make sense? It was implemented in a bit strange way so I'm wondering if there was some reasoning behind it.

I also found the logic weird, but I think it was erroneous, taking a look at the rest of the logic.
All tests pass on the csv export functionality, and it runs correctly locally with a `sha1` storage (ex: running `udata job run export-csv organization`).

maudetes:
Well done for this work! I can merge if you update the branch with master :blush:

PR #2815, opened 2023-02-08
Rename dataset and resource internal dates for convention

quaxsze:


udata/core/dataset/api.py
```diff
@@ -360,6 +360,7 @@ def handle_upload(self, dataset):
         prefix = '/'.join((dataset.slug,
                            datetime.now().strftime('%Y%m%d-%H%M%S')))
         infos = handle_upload(storages.resources, prefix)
+        print('INFOS :', infos)
```
maudetes, line 363:
```suggestion
```

udata/core/dataset/events.py
```diff
@@ -23,7 +23,7 @@ def serialize_resource_for_event(resource):
         'checksum_type': resource.checksum.type if resource.checksum else None,
         'checksum_value': resource.checksum.value if resource.checksum else None,
         'created_at': to_iso_datetime(resource.created_at),
-        'modified': to_iso_datetime(resource.modified)
+        'modified': to_iso_datetime(resource.last_modified)
```
maudetes, line 26: Do we want to modify it to `last_modified` (on the reception side too)?

udata/core/dataset/models.py
```diff
@@ -290,11 +290,19 @@ class ResourceMixin(object):
...
 
-    created_at = db.DateTimeField(default=datetime.now, required=True)
-    modified = db.DateTimeField(default=datetime.now, required=True)
+    created_at_internal = db.DateTimeField(default=datetime.now, required=True)
+    last_modified_internal = db.DateTimeField(default=datetime.now, required=True)
     published = db.DateTimeField()  # DEPRECATED BUT LEFT FOR BACKWARDS COMPATIBILITY
     deleted = db.DateTimeField()
 
+    @property
+    def created_at(self):
```
maudetes, line 299: Maybe you could add minimal tests on those computed properties? :)

udata/core/dataset/models.py
```diff
@@ -290,11 +290,19 @@ class ResourceMixin(object):
     harvest = db.EmbeddedDocumentField(HarvestResourceMetadata)
     schema = db.DictField()
 
-    created_at = db.DateTimeField(default=datetime.now, required=True)
-    modified = db.DateTimeField(default=datetime.now, required=True)
+    created_at_internal = db.DateTimeField(default=datetime.now, required=True)
```
maudetes, line 293: Maybe fixtures.py should be updated as well to match the new behaviour?

udata/core/dataset/models.py
```diff
@@ -578,6 +586,14 @@ def check_availability(self):
             return []
         return [resource.check_availability() for resource in remote_resources]
 
+    @property
+    def created_at(self):
+        return self.harvest.created_at if self.harvest and self.harvest.created_at else self.created_at_internal
+
+    @property
+    def last_modified(self):
+        return max([self.last_modified_internal, to_naive_datetime(self.harvest.modified_at)]) if self.harvest and self.harvest.modified_at else self.last_modified_internal
```
maudetes, line 595: Some lines are really long, we should target 100 chars max on new code :)

udata/core/dataset/models.py
```diff
@@ -587,7 +603,7 @@ def last_update(self):
         if self.resources:
             dates = []
             for res in self.resources:
-                dates.append(res.modified)
+                dates.append(res.last_modified)
                 if res.harvest and res.harvest.modified_at:
                     dates.append(to_naive_datetime(res.harvest.modified_at))
```
maudetes, line 608: I think we can refactor this to use `res.last_modified` that takes the max for us :)

udata/core/dataset/models.py
```diff
@@ -290,11 +290,19 @@ class ResourceMixin(object):
...
-    created_at = db.DateTimeField(default=datetime.now, required=True)
-    modified = db.DateTimeField(default=datetime.now, required=True)
+    created_at_internal = db.DateTimeField(default=datetime.now, required=True)
+    last_modified_internal = db.DateTimeField(default=datetime.now, required=True)
     published = db.DateTimeField()  # DEPRECATED BUT LEFT FOR BACKWARDS COMPATIBILITY
     deleted = db.DateTimeField()
 
+    @property
+    def created_at(self):
+        return self.harvest.created_at if self.harvest and self.harvest.created_at else self.created_at_internal
```
maudetes, line 300: I'm scared we may have issues with naive datetimes, maybe use the helper used in last_update?

udata/core/dataset/models.py
```diff
@@ -290,11 +290,19 @@ class ResourceMixin(object):
...
+
+    @property
+    def last_modified(self):
+        return max([self.last_modified_internal, to_naive_datetime(self.harvest.modified_at)]) if self.harvest and self.harvest.modified_at else self.last_modified_internal
```
maudetes, line 304: If you wanna, we can update `harvest.modified_at` -> `harvest.last_modified` for coherence, even if you like the first better (optional)?

maudetes (commented):
Oh nice! :partying_face:
I think I like this option better, because computation is more transparent and the properties haven't changed for API or jinja templates. What do you think?
I haven't tested the migration or the code yet, I need to prepare a db to scramble. :p

Don't forget changelog ;)

udata/core/dataset/events.py
quaxsze: well `resource.modififed` won't exist anymore so we have to either modify it to `last_modified` or `last_modified_internal`

udata/core/dataset/models.py
quaxsze: the problem occured by comparing a naive date and an aware one. Here we are not doing any comparison.

udata/core/dataset/events.py
maudetes: I meant the left-hand part, `'modified'` in `resource_dict`

maudetes (commented):
Thank you for the update!

I have minor comments, except for the migration that didn't seem to work on my local db sadly.

udata/core/dataset/events.py
quaxsze: Well I guess we will have to modify it on the search service then ;)

udata/migrations/2023-02-08-rename-internal-dates.py
```diff
@@ -15,13 +16,23 @@ def migrate(db):
...
-        'resource.created_at': 'resource.created_at_internal',
-        'resource.modified': 'resource.last_modified_internal'}})
+        'last_modified': 'last_modified_internal'}})
 
     community_resource_collection = db.community_resource
     community_resource_collection.update_many({}, {'$rename': {
         'created_at': 'created_at_internal',
-        'modified': 'last_modified_internal'}})
+        'modified': 'last_modified_internal',
+        'harvest.modified_at': 'harvest.last_modified'}})
```
quaxsze, line 25: Is that needed? Community ressources are never harvested right? So I guess not any will have the field set?

udata/migrations/2023-02-08-rename-internal-dates.py
```diff
@@ -15,13 +16,23 @@ def migrate(db):
...
+
+    datasets = Dataset.objects(resources__0__exists=True)
```
quaxsze, line 27: `As mentioned in the documentation there is no way to directly rename fields within arrays with a single command. Your only option is to iterate over your collection documents, read them and update each with $unset old/$set new operations.`
According to this I did a loop for resources :/

udata/migrations/2023-02-08-rename-internal-dates.py
maudetes: Indeed, we don't have community resource with an harvest field :)

udata/migrations/2023-02-08-rename-internal-dates.py
```diff
@@ -0,0 +1,46 @@
...
+        'last_modified': 'last_modified_internal'}})
+    log.info(f'{result.modified_count} Datasets processed.')
+
+    community_resource_collection = db.community_resource
+    result = community_resource_collection.update_many({}, {'$rename': {
+        'created_at': 'created_at_internal',
+        'modified': 'last_modified_internal'}})
+    log.info(f'{result.modified_count} Community Resources processed.')
+
+    result = dataset_collection.update_many({}, [{
```
maudetes, line 29: second argument can't be an array, can it?

I have the following error at this step when running `udata db migrate`:
```
pymongo.errors.OperationFailure: BSON field 'update.updates.u' is the wrong type 'array', expected type 'object', full error: {'ok': 0.0, 'errmsg': "BSON field 'update.updates.u' is the wrong type 'array', expected type 'object'", 'code': 14, 'codeName': 'TypeMismatch'}
  ```

File: udata/core/dataset/models.py [hunk 12]
```diff
@@ -290,11 +290,18 @@ class ResourceMixin(object):
     harvest = db.EmbeddedDocumentField(HarvestResourceMetadata)
     schema = db.DictField()
 
-    created_at = db.DateTimeField(default=datetime.now, required=True)
-    modified = db.DateTimeField(default=datetime.now, required=True)
-    published = db.DateTimeField()  # DEPRECATED BUT LEFT FOR BACKWARDS COMPATIBILITY
```
maudetes, line 295: You can add it in the changelog :)

udata/core/dataset/models.py
```diff
@@ -481,16 +488,16 @@ def __str__(self):
     meta = {
         'indexes': [
             '$title',
-            'created_at',
-            'last_modified',
+            'created_at_internal',
```
maudetes, line 491: Did we test that sort on `created_at` still work with mongo search?

udata/core/dataset/events.py
maudetes: Could you prepare a PR there please? :pray:

udata/migrations/2023-02-08-rename-internal-dates.py
```diff
@@ -0,0 +1,46 @@
...
+'''
+import logging
+
+from mongoengine.connection import get_db
+from udata.models import Dataset
+
+log = logging.getLogger(__name__)
+
+
+def migrate(db):
```
maudetes, line 12: It seems to work nicely! :muscle:
Did you try renaming harvest field (on dataset and resource) to `last_modified` for total coherence? :p

udata/tests/api/test_datasets_api.py
```diff
@@ -96,6 +96,28 @@ def test_dataset_api_sorting(self):
...
+        user = self.login()
+        first = VisibleDatasetFactory(title="first created dataset")
+        second = VisibleDatasetFactory(title="second created dataset")
+        response = self.get(url_for('api.datasets', sort='created'))
+        self.assert200(response)
+        self.assertEqual(response.json['data'][0]['id'], str(first.id))
+
+        response = self.get(url_for('api.datasets', sort='-created'))
+        self.assert200(response)
+        self.assertEqual(response.json['data'][1]['id'], str(first.id))
```
maudetes, line 109: I would find it more understandable the other way around (as you did with last_modified)
```suggestion
        self.assertEqual(response.json['data'][0]['id'], str(second.id))
```

maudetes (commented):
I have a question on harvest field, but else it seems to work like a charm!
Congratulations!

maudetes (approved):

PR #2819, opened 2023-02-28
Update mongoDB from 3.6 to 6.0.4

quaxsze:
Fix https://github.com/etalab/data.gouv.fr/issues/1056


maudetes (approved):
Nice! I'm turning into a draft so that we can test it on our infra before merging :)

Don't forget to add a changelog entry

maudetes (approved):
:rocket:

PR #2823, opened 2023-03-06
Fix inv i18n

nicolaskempf57:
Fix `inv i18n` errors due some extensions now integrated to jinja2 v3 core.


maudetes (approved):
:pray: If you wanna, you can add the link to the PR in the existing flask v2 entry in the changelog :)

babel.cfg
```diff
@@ -3,11 +3,11 @@
 [jinja2: templates/**.html]
```
maudetes, line 3: Don't we mind the `jinja2` keyword though? :p

maudetes: We went from Jinja2==2.10.3 to 3.1.2, thus I was joking about it not being Jinja2 anymore but Jinja3 :p
Please don't mind me

PR #2824, opened 2023-03-07
Clean translations doc for plugins

maudetes:
Move doc in a dedicated section on `adding-translations.md`, and point to it in relevant parts.


docs/harvesting.md
```diff
@@ -240,3 +240,4 @@ You may take a look at the [existing backends][backends-repository] to see exiti
 [DCAT]: https://www.w3.org/TR/vocab-dcat/
 [backends-repository]: https://github.com/opendatateam/udata/tree/master/udata/harvest/backends
 [gov-uk-references]: http://reference.data.gov.uk/
+[cookiecutter-template]: https://github.com/opendatateam/cookiecutter-udata-harvester
```
nicolaskempf57, line 243: why is the link here and not with the ones in `adding-translations.md` ?

maudetes: It's actually not related to translations. It's another cookie cutter template that was mentioned [here](https://github.com/opendatateam/udata/blob/5342e20082eeaae1272665478671f5f8fe598865/docs/harvesting.md?plain=1#L229) but the link was missing.

PR #2825, opened 2023-03-07
Resource schema can now have an URL

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/1052


quaxsze:
Changelog will be modified after coming release.

js/components/dataset/resource/form.vue
```diff
@@ -252,7 +252,7 @@ export default {
                 const values = [{id: '', label: ''}].concat(schemas.data);
                 return [{
                     id: 'schema.name',
-                    label: this._('Schema'),
+                    label: this.resource.schema.url ? this._('Schema (Url present)') : this._('Schema'),
```
maudetes, line 255: I think having this message with an appropriate wording could do the trick.
How about something more explicit: `Schema (warning: setting it here would override the current schema url value)`?

We should validate with @geoffreyaldebert though.

udata/tests/api/test_datasets_api.py
```diff
@@ -574,17 +574,31 @@ def test_dataset_api_unfeature(self):
         dataset.reload()
         self.assertFalse(dataset.featured)
 
-    def test_dataset_api_unfeature_already(self):
```
maudetes, line 577: What did this test do to you to get smashed? :p

udata/tests/api/test_datasets_api.py
```diff
@@ -574,17 +574,31 @@ def test_dataset_api_unfeature(self):
...
-        '''It shouldn't do anything to unfeature a not featured dataset'''
-        self.login(AdminFactory())
-        dataset = DatasetFactory(featured=False)
+    def test_dataset_new_resource_with_schema(self):
+        '''Should filters datasets results based on query filters'''
```
maudetes, line 578: Woops

udata/tests/api/test_datasets_api.py
```diff
@@ -574,17 +574,31 @@ def test_dataset_api_unfeature(self):
...
+        dataset = DatasetFactory(owner=user)
+        data = dataset.to_dict()
+        resource_data = ResourceFactory.as_dict()
 
-        response = self.delete(url_for('api.dataset_featured',
-                                       dataset=dataset))
-        self.assert200(response)
+        resource_data['schema'] = {'name': 'my-schema', 'version': '1.0.0', 'url': 'http://example.com'}
+        data['resources'].append(resource_data)
+        response = self.put(url_for('api.dataset', dataset=dataset), data)
```
maudetes, line 586: You could directly use `api.resources` and use `resource_data` only :)

udata/core/dataset/models.py
```diff
@@ -299,6 +299,8 @@ def clean(self):
         super(ResourceMixin, self).clean()
         if not self.urlhash or 'url' in self._get_changed_fields():
             self.urlhash = hash_url(self.url)
+        if not bool('name' in self.schema) ^ bool('url' in self.schema):
```
maudetes, line 302: We're not saying anything about version. Having a version and an url doesn't make sense. Not sure we want to make it any more complex though

js/components/dataset/resource/form.vue
quaxsze: The message is too long and the space in the ui is short. We have no more than a few words

maudetes (approved):

PR #2828, opened 2023-03-15
Adding support for ssim format

restuccia:
Added the IATA ssim format to the allowed ones, cf. https://github.com/etalab/data.gouv.fr/issues/1066


maudetes (approved):
Thank you!

PR #2829, opened 2023-03-16
Update renovate config

maudetes:
Should we use renovate on top of dependabot?

Later on, may be useful to take a look at https://docs.renovatebot.com/modules/manager/pip-compile/.


maudetes:
@quaxsze any opinion on the renovate vs dependabot?

PR #2840, opened 2023-04-06
Do not fail on RecipientsRefused in sendmail

maudetes:
Fix https://github.com/etalab/data.gouv.fr/issues/1074

See underlying logic that sends the `SMTPRecipientsRefused` error (`The server rejected ALL recipients (no mail was sent)`): https://github.com/python/cpython/blob/3.11/Lib/smtplib.py#L808`
Add minimal test case on the send mail functionality, with refused recipients case.


jordanguedj (approved):
Thanks!

PR #2841, opened 2023-04-07
Fix SelectField validation failure

maudetes:
Following https://github.com/opendatateam/udata/pull/2826 and wtforms upgrade, we now have an error on SelectField Validation : https://errors.data.gouv.fr/organizations/sentry/issues/1667/?alert_rule_id=3&alert_timestamp=1680857196717&environment=production&project=3.
Calling the `PUT api/1/datasets/<dataset-id>` leads to this error for example.

The underlying SelectField validation behavior has changed in wtforms.
The code has changed from [2.2.1](https://github.com/wtforms/wtforms/blob/7152845801e88ab6a9c7d719b73d41bef6464b5d/wtforms/fields/core.py#L470) to [3.0.1](https://github.com/wtforms/wtforms/blob/94a5c268cd914798172ab6b6e85375c9447e3b70/src/wtforms/fields/choices.py#L134), leading to `iter_choices` being called when it wasn't before, where the issue occurs.

It seems the the lazystring created by `_(label)` in `localized_choices` is broken, which is why it fails with `unhashable type: '_LazyString'`.

This fix is a workaround. It checks if the label is lazystring already, preventing applying lazy_gettext twice, leading to a broken lazystring:
```
>>> from udata.i18n import lazy_gettext as _
>>> _('dataset')
l'jeu de données'
>>> _(_('dataset'))
<_LazyString broken>
```
___
However, two questions are left open before merging:
- [x] why doesn't it fail in tests when the exact same form validation should be applied? We have an issue if we have different behavior between tests and run
    - Because double translations do not fail in English, which is DEFAULT_LANGUAGE. We have turned it to `fr` in tests, to make sure to apply translations logic all around
- [x] I've seen lazy_gettext being applied twice without issues when debugging, why does it fail in this case?
    - If translation language is English, double translations do not fail (translation of dataset is dataset, and again and again?)

### DIFF
--- a/udata/forms/fields.py
+++ b/udata/forms/fields.py
@@ -5,7 +5,8 @@ from dateutil.parser import parse
 from flask import url_for
 from flask_mongoengine.wtf import fields as mefields
 from flask_storage.mongo import ImageReference
-from wtforms import Form as WTForm, Field as WTField, validators, fields, SubmitField
+from speaklater import is_lazy_string
+from wtforms import Form as WTForm, Field as WTField, validators, fields, SubmitField  # noqa
 from wtforms.utils import unset_value
 from wtforms.widgets import TextInput
 from wtforms_json import flatten_json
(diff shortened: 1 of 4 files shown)


udata/forms/fields.py
```diff
@@ -128,14 +129,6 @@ def process_formdata(self, valuelist):
             except ValueError:
                 self.data = None
                 raise validators.ValidationError(_('Invalid date/time input'))
-            except TypeError:
```
maudetes, line 131: Not needed anymore. Fixed in dateutil 2.3: https://bugs.launchpad.net/dateutil/+bug/1247643.

quaxsze (approved):

PR #2845, opened 2023-04-21
Compile translations for testing

maudetes:
Follow https://github.com/opendatateam/udata/pull/2841, where tests now use `fr` as default language.

Tests running in CI didn't have most translations since they hadn't been compiled, diverging from tests run locally with compiled translations.

Make sure to have compiled translations to use in tests, locally or in CI.

### DIFF
--- a/tasks/__init__.py
+++ b/tasks/__init__.py
@@ -54,6 +54,83 @@ def update(ctx, migrate=False):
 
 
 @task
+def i18n(ctx, update=False):
+    '''Extract translatable strings'''
+    header('Extract translatable strings')
+
+    info('Extract Python strings')
+    with ctx.cd(ROOT):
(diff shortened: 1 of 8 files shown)


tasks/__init__.py
```diff
@@ -54,6 +54,83 @@ def update(ctx, migrate=False):
...
+@task
+def i18nc(ctx):
+    '''Compile translations'''
+    header('Compiling translations')
+    with ctx.cd(ROOT):
+        ctx.run('python setup.py compile_catalog')
+
+
+@task(i18nc)
 def test(ctx, fast=False, report=False, verbose=False, ci=False):
```
maudetes, line 134: I've moved code around, but the only modification is `@task` -> `@task(i18nc)` as decorator the `test` task.

quaxsze (approved):

PR #2847, opened 2023-04-25
Sort by last update

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/1035


udata/core/dataset/search.py
```diff
@@ -84,6 +85,7 @@ def serialize(cls, dataset):
             'badges': [badge.kind for badge in dataset.badges],
             'frequency': dataset.frequency,
             'created_at': to_iso_datetime(dataset.created_at),
+            'last_update': to_iso_datetime(dataset.last_update),
```
maudetes, line 88: Made me realize, we don't test search serialization at all?

PR #2853, opened 2023-05-15
Use utcnow to handle utc datetimes all around

maudetes:
Fix https://github.com/etalab/transport-site/issues/3152

Since https://github.com/opendatateam/udata/pull/2810, we're now returning timezone aware datetimes.
We're currently expecting utc times stored in mongo. But `datetime.now()`, use as default values all around generates date in the current timezone.

This PR replaces it by `datetime.utcnow()`, to make sure to store utc datetimes in mongo.
This fix would only apply to future objects being created/modified.

Similar PRs have been opened in other udata plugins:
* https://github.com/etalab/udata-front/pull/251
* https://github.com/opendatateam/udata-search-service/pull/42
* https://github.com/opendatateam/udata-ckan/pull/246

For existing objects, we could apply a migration, but it may not be trivial, since some dates could have been supplied by the user, or when the servers used a different timezone?
My guess is it would not add much user value. Indeed, the bug is currently mostly visible on recent contents, not on days or years-old contents.
:arrow_right: I would be willing to have your opinion on this @AntoineAugusti :pray:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
     - Upgrade flask-mongoengine 0.9.5 -> 1.0.0, now returning a ValidationError on get_or_404 on invalid id
     - Upgrade mongoengine 0.26.0 -> 0.27.0
 -  Prevent raising unecessary error in index command [#2851](https://github.com/opendatateam/udata/pull/2851)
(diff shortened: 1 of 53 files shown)


quaxsze:
👏

quaxsze (approved):

AntoineAugusti:
@maudetes Seems acceptable to me to not perform a migration for old data 👌

PR #2820, opened 2023-03-02
[WIP] Feat/upgrade celery

maudetes:
Fix https://github.com/etalab/data.gouv.fr/issues/1062

It's a draft to test that dependencies are compatible when upgrading Celery, following https://github.com/etalab/data.gouv.fr/issues/1043.

Changelogs haven't been checked yet for deps upgraded here.
See also https://docs.celeryq.dev/en/stable/whatsnew-5.2.html#upgrading-from-celery-4-x.

PR #2859, opened 2023-06-26
WIP: Upgrade python to 3.11

maudetes:
Fix https://github.com/etalab/data.gouv.fr/issues/1079
Fix https://github.com/etalab/data.gouv.fr/issues/1062

We should target for support >=3.8 and make it work with 3.10 at least.

Add [flask-shell-ipython](https://pypi.org/project/flask-shell-ipython/) deps in develop.in in this PR since latest version requires python >= 3.8.

Taking care of https://github.com/opendatateam/udata/pull/2820 at the same time.

# TODO
- [ ] read changelogs for any unexpected breaking changes in deps upgrade
- [ ] test all udata plugins
    - [X] udata-front
    - [X] udata-ckan 
    - [X] udata-ods
    - [ ] ...
- [x] [update docker used in CI](https://github.com/opendatateam/udata-dockers/pull/6) to our target version
- [ ] check issue with  `pkg_resources.UnknownExtra` for `kombu[redis]` with python3.8 in https://github.com/opendatateam/udata/blob/819f8ebf8c997083e392c95fefeff4594bcd074f/udata/entrypoints.py#L55


.circleci/config.yml
```diff
@@ -24,11 +24,10 @@ jobs:
           - py3-cache-v9-{{ arch }}-{{ .Environment.BASE_BRANCH }}
       - run:
           name: Install python dependencies
-          # FIXME: For now, we fix setuptools due to https://github.com/etalab/data.gouv.fr/issues/1041
           command: |
-            virtualenv venv
+            python -m venv venv
             source venv/bin/activate
-            pip install --force-reinstall setuptools==66.1.1
+            pip install --upgrade wheel
```
maudetes, line 30: Not sure about the need to install wheel explicitly. See https://stackoverflow.com/questions/69369669/python-3-venv-and-the-wheel-package

ThibaudDauce:
I think we can close this one, see https://github.com/opendatateam/udata/pull/2992

PR #2862, opened 2023-07-03
Expose dataset's created_at_internal date

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/1103


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Improve DCAT harvest of mime type [#2857](https://github.com/opendatateam/udata/pull/2857)
 - Don't crash on files not found when purging resources [2858](https://github.com/opendatateam/udata/pull/2858)
+- Expose dataset's created_at_internal date in api endpoints [#2862](https://github.com/opendatateam/udata/pull/2862)
```
maudetes, line 7: I guess we want the the same modification for last_modified?
And I think this applies at the dataset and the resource model, right?

udata/core/dataset/api_fields.py
```diff
@@ -174,7 +174,9 @@
     'description': fields.Markdown(
         description='The dataset description in markdown', required=True),
     'created_at': fields.ISODateTime(
-        description='The dataset creation date', required=True),
+        description='This date is computed between harvested creation date if any and site\'s internal creation date' , required=True),
+    'created_at_internal': fields.ISODateTime(
```
maudetes, line 178: Could we have a minimal test? :pray:

udata/core/dataset/api_fields.py
```diff
@@ -228,6 +235,8 @@
     'quality': fields.Raw(description='The dataset quality', readonly=True),
     'last_update': fields.ISODateTime(
         description='The resources last modification date', required=True),
+    'internal': fields.Nested(
+        dataset_internal_fields, description='Site internal and specific dataset\'s data'),
```
maudetes, line 239: Should we add `readonly`, or isn't it needed?

maudetes (commented):
We should apply the same for resources in this PR as well probably

maudetes (commented):
I think `internals` are still missing at the dataset's level in apiv2, right?

maudetes (approved):
Nice ! :pray:

PR #2868, opened 2023-07-10
Fix dct:issued doc on resource

abulte:
Cf https://github.com/opendatateam/udata/blob/b6e4a159739efacfccce8be86344af7be6c3f4e3/udata/core/dataset/rdf.py#L136C21-L136C21


maudetes (approved):
I wonder if this doc should probably be merged with our user documentation instead of udata's: https://doc.data.gouv.fr/api/telecharger-un-catalogue-de-donnees/ (or [upcoming guides ](https://guides.data.gouv.fr/publier-des-donnees/guide-data.gouv.fr/jeux-de-donnees/telecharger-le-catalogue-de-donnees-de-data.gouv.fr)).

There are duplicates information I didn't have in mind.

PR #2873, opened 2023-07-20
Fix dataset with duplicate long title cannot be added

tboye:
This issue is related to slug generation in general but is easily seen with datasets.

In case a dataset with a title longer than max allowed slug length is added, an attempt of adding a dataset with an identical long title will result in an error: `mongoengine.errors.ValidationError: ValidationError (Dataset:None) (String value is too long: ['slug'])`

This is due to the fact we don't preserve some space for index postfix that will be added in case of duplicate slugs.

This PR adds an optional `max_identical_slugs` parameter to `SlugField` class that allows for limiting the number of identical slugs that will be handled without errors. Default value is set to 999 which should be more than sufficient for most use cases. This value can be exposed to settings in the future if needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Fix dataset with duplicate long title cannot be added
 
 ## 6.1.6 (2023-07-19)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Fix dataset with duplicate long title cannot be added
+- Fix dataset with duplicate long title cannot be added [#2873](https://github.com/opendatateam/udata/pull/2873)
 
 ## 6.1.6 (2023-07-19)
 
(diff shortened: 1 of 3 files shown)


maudetes:
Thank you a lot for this contribution! :pray: 

At first sight, this could do it but I'm wondering if instead of keeping an extra space, we could trim some chars when adding index values? We should be cautious to still ensure uniqueness though. This would prevent adding an arbitrary limit.

Let's see if I get time to give it a go next week, else we'll probably go for this solution. :blush:

tboye:
@maudetes Oh, I'm also not really fond of this arbitrary number so I'm very open to improve the approach.

Just thinking out loud, if we trim it when adding indexes, we might end up having "very different" slugs, i.e. we won't be able to assume that these items are somewhat relevant. Although I cannot think of any use case and probably we only want to ensure uniqueness here.

maudetes:
To me, duplicate slugs do not carry any meaning of relation between the two, it would not be a problem if they end up different, but maybe I'm missing some use cases?
I gave it a go [in this PR](https://github.com/opendatateam/udata/pull/2874) if you would like to take a look.

tboye:
@maudetes No use cases came to my mind either, closing this one.

PR #2874, opened 2023-07-25
Fix/fix slug overflow with index suffix

maudetes:
Alternative to https://github.com/opendatateam/udata/pull/2873.

On slug generation, we have an issue with slugs that are duplicates and that need a suffix but are already at `field.max_length`.

This PR adds a check to crop slug if needed before adding suffix. It ensures that slug size is always <= `field.max_length`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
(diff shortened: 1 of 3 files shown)


tboye (approved):
Looks all good to me 👍

PR #2876, opened 2023-07-31
Reuse topics as playlists

quaxsze:
related to https://github.com/etalab/data.gouv.fr/issues/1117

See relevant PR in search service: https://github.com/opendatateam/udata-search-service/pull/43


udata/core/topic/models.py
```diff
@@ -23,10 +26,30 @@ class Topic(db.Document):
...
+            reuses_list_dif = list(set(original_doc.reuses).symmetric_difference(set(document.reuses)))
+        except cls.DoesNotExist:
+            datasets_list_dif = document.datasets
+            reuses_list_dif = document.reuses
+        if datasets_list_dif:
+            for dataset in datasets_list_dif:
+                reindex.delay(*as_task_param(dataset))
+        if reuses_list_dif:
+            for reuse in reuses_list_dif:
+                reindex.delay(*as_task_param(reuse))
```
maudetes, line 48: We actually haven't discussed indexing reuses with topic info. It would makes sense since we do for datasets, but we also have a `topic` property in the Reuse object that might get in conflict.

udata/core/topic/api.py
```diff
@@ -41,6 +41,7 @@
     'page': fields.UrlFor(
         'topics.display', lambda o: {'topic': o},
         description='The topic page URL', readonly=True, fallback_endpoint='api.topic'),
+    'extras': fields.Raw(description='Extras attributes as key-value pairs'),
```
maudetes, line 44: You should add the field in forms as well?

maudetes (commented):
Nice! :clap:

udata/tests/test_topics.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class TopicModelTest:
+
+    def test_pre_save(self, job_reindex):
+        topic = TopicFactory()
+        dataset = DatasetFactory()
+        topic.datasets = [dataset]
+        topic.save()
+        job_reindex.assert_called()
```
maudetes, line 23: Could you test that job_reindex is not called if dataset list hasn't changed? But it called on dataset being removed from topics? :pray:

maudetes (approved):
:sparkle: I think we're good!

PR #2882, opened 2023-08-18
Filter out datapackage schema

maudetes:
Close https://github.com/etalab/data.gouv.fr/issues/1148

Add an IGNORED_SCHEMA_TYPES variable to ignore some schema types (datapackage). Felt more secure to have an exclude instead of an include logic.

:warning: The front filter doesn't use the /schema route. Instead it loads the catalog directly with `getCatalog`: https://github.com/etalab/udata-front/blob/38ed004ebb5df3dd2f93c749d25b0ff1e0440975/udata_front/theme/gouvfr/assets/js/components/search/schema-filter.vue#L33C11-L33C11.
- [ ] We need to use the /schema endpoint from the front

### DIFF
--- a/udata/core/dataset/models.py
+++ b/udata/core/dataset/models.py
@@ -113,6 +113,7 @@ CLOSED_FORMATS = ('pdf', 'doc', 'docx', 'word', 'xls', 'excel', 'xlsx')
 MAX_DISTANCE = 2
 
 SCHEMA_CACHE_DURATION = 60 * 5  # In seconds
+IGNORED_SCHEMA_TYPES = ('datapackage', )
 
 TITLE_SIZE_LIMIT = 350
 DESCRIPTION_SIZE_LIMIT = 100000
@@ -919,6 +920,7 @@ class ResourceSchema(object):
                     'label': s['title'],
                     'versions': [d['version_name'] for d in s['versions']],
                 } for s in schemas
+                if s.get('schema_type') not in IGNORED_SCHEMA_TYPES
             ]
             cache.set(cache_key, content)
         # no cached version or no content
--- a/udata/tests/dataset/test_dataset_model.py
+++ b/udata/tests/dataset/test_dataset_model.py
@@ -203,10 +203,12 @@ class DatasetModelTest:
         assert dataset.quality['score'] == Dataset.normalize_score(2)
 
     def test_quality_description_length(self):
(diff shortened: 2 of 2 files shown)


nicolaskempf57:
> ⚠️ The front filter doesn't use the /schema route. Instead it loads the catalog directly with `getCatalog`

We can call the `datasets/schemas` endpoint on the front side but for now we're using the `schema_type` to know if we should show a validation url and the `schema_url` from a schema version to make the validata URL. 

We can update this logic if there is a better way to get the validata URL, or add the required informations in the endpoint.

maudetes:
That's a good point! Should we add `schema_url` and `schema_type` to the endpoint? It would seem like a useful information to expose in any case

maudetes:
After taking a look at calls on `api/1/datasets/schemas/` endpoint, seems like no external users use this API.
We should indeed modify it to expose additional schema urls
```
{
    "id": "etalab/schema-irve-statique",
    "label": "IRVE statique",
    "schema_url": "https://schema.data.gouv.fr/schemas/etalab/schema-irve-statique/latest/schema-statique.json",
    "schema_type": "tableschema",
    "versions": {
        "version_name": "2.2.0",
        "schema_url": "https://schema.data.gouv.fr/schemas/etalab/schema-irve-statique/2.2.0/schema-statique.json"
    }
}
```

ThibaudDauce:
> After taking a look at calls on `api/1/datasets/schemas/` endpoint, seems like no external users use this API. We should indeed modify it to expose additional schema urls
> 
> ```
> {
>     "id": "etalab/schema-irve-statique",
>     "label": "IRVE statique",
>     "schema_url": "https://schema.data.gouv.fr/schemas/etalab/schema-irve-statique/latest/schema-statique.json",
>     "schema_type": "tableschema",
>     "versions": {
>         "version_name": "2.2.0",
>         "schema_url": "https://schema.data.gouv.fr/schemas/etalab/schema-irve-statique/2.2.0/schema-statique.json"
>     }
> }
> ```

Still no external user for this endpoint? Can we break it?

ThibaudDauce:
Feature added in this PR https://github.com/opendatateam/udata/pull/2989

maudetes:
> Still no external user for this endpoint? Can we break it?

I confirm this. It hasn't been used in the last days outside of our own admin calls.

ThibaudDauce:
Closing it for https://github.com/opendatateam/udata/pull/2989

PR #2886, opened 2023-08-29
Replace set() by dict comprehension in recipients for new comments

maudetes:
Member of an organization that have taken part in a discussion on one of its documents are notified twice by mail about a new comment.

Indeed, `set(recipients)` does not work as expected, probably because it uses `__hash__`, that [should be defined based on primary key](https://github.com/MongoEngine/mongoengine/blob/090e6e1efb32272f3ada0b9a8c70c16c6683a0b8/mongoengine/document.py#L190), but is [overridden in the context of UserMixin](https://github.com/maxcountryman/flask-login/blob/3ee566b5f324885491b688cd2e4a6dff49ebef2f/src/flask_login/mixins.py#L9).

We're replacing it by a dict comprehension on user.id to make sure we have only one occurrence of each user.

See https://mattermost.incubateur.net/betagouv/pl/q91rpukpatna9jfbryiafqq7rw for context.

### DIFF
--- a/udata/tests/test_discussions.py
+++ b/udata/tests/test_discussions.py
@@ -4,7 +4,7 @@ from flask import url_for
 
 from udata.models import Dataset, Member
 from udata.core.discussions.models import Message, Discussion
-from udata.core.discussions.metrics import update_discussions_metric
+from udata.core.discussions.metrics import update_discussions_metric  # noqa
 from udata.core.discussions.notifications import discussions_notifications
 from udata.core.discussions.signals import (
     on_new_discussion, on_new_discussion_comment,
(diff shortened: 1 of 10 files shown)


quaxsze (approved):

udata/tests/test_discussions.py
```diff
@@ -550,3 +551,76 @@ def test_notify_org_discussions(self):
             self.assertEqual(details['title'], discussion.title)
             self.assertEqual(details['subject']['id'], discussion.subject.id)
             self.assertEqual(details['subject']['type'], 'dataset')
+
+
+class DiscussionsMailsTest(APITestCase):
```
maudetes, line 556: Mostly copied from udata-front

PR #2889, opened 2023-08-31
Return 400 instead of 500 in case of not ObjectID arg

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/1125


maudetes (approved):
Thank you! :pray:

PR #2897, opened 2023-09-11
Migrate to Flask-Babel

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/1059
fix https://github.com/etalab/data.gouv.fr/issues/1131


maudetes (changes requested):
Thank you! However, I think we still need to make translations work with jinja templates

maudetes:
Could you make a review of my latest changes @quaxsze? :pray: Another fresh eye on this would be nice, to make sure everything works well!

maudetes (approved):

Issue #2877, opened 2023-08-02
Additional languages in themes


tboye:
## Detailed Description

**Disclaimer:** I might be overcomplicating/misunderstanding how it's supposed to work.

In my theme for `udata` I'd like to add a `.po` file for English language so that I could tweak text by changing translations rather than `msgid`s. But generally speaking, it's no different from a use case when someone wants to add a completely new language to their theme.

It seems to be not possible at the moment as only languages registered in `udata` will be taken into account and overridden by translations in the theme.

## Context
When I develop a theme, I want to add languages that aren't part of udata, so I can customize theme independently.

## Possible Implementation
- let `udata` pick up languages from the theme once it's injected
- let the theme add languages to `udata`
- extend list of languages in `udata`, including English (not desirable)

## Your Environment
N/A

maudetes:
Thank you for this issue (that wasn't on my radar, sorry).
I'm not sure I've understood it correctly though since I don't have all translation mechanic in mind.

If you want to use any language, you can override default `LANGUAGES` setting in your udata.cfg.
You should be able to use any of the languages that exist in `udata/translations/` and it will also load translations for any theme or plugin accordingly.
What would be the use case where you don't want to have existing translations in udata?

tboye:
Thanks for looking into that.

> If you want to use any language, you can override default LANGUAGES setting in your udata.cfg.
You should be able to use any of the languages that exist in udata/translations/ and it will also load translations for any theme or plugin accordingly.

Yes, that's clear and that works without issues!

> What would be the use case where you don't want to have existing translations in udata?

In general, my use case is to tweak translations in my theme to make them relevant to my project. Those are very small tweaks, yet still very important.

At the moment I can do it for languages that already exist in `udata`. If I ever come up with additional languages, they can and should be added to `udata` first -- it's not a problem either indeed, my ticket might sound a bit too dramatic 😅  The main problem I have at the moment is with English strings. Since English is the default language and message IDs are used directly, there's no way for me to change anything there at the moment.

Would there be any drawbacks if we register English in `udata` explicitly?

PR #2899, opened 2023-09-30
add German translation

jze:


maudetes (approved):
Thank you for this PR :)

Do you plan to add german translations on both udata and udata-front? We'll add German to Crowdin settings accordingly.

Additionally, I'm curious about your expected usage of udata, if you'd be willing to share :blush:

jze:
Yes, I will try to do both translations.

I really like the French open data system. In Germany we have mostly CKAN based open data portals and a view Piveau based ones. I think it is good have to have another open source choice for data portals. And I guess we both know the problem when a program is not in the local language. :wink:

tboye:
@maudetes This PR reminded me about [this issue](https://github.com/opendatateam/udata/issues/2877) which I still haven't found a solution for. Do you maybe have any feedback on that?

PR #2900, opened 2023-10-01
Fix site title and keywords never get updated

tboye:
Currently changing `SITE_TITLE` and/or `SITE_KEYWORDS` settings doesn't have any effect as initial values stored in Mongo are taken. I'm not sure if this was done intentionally, but in my opinion such update might be needed every now and then, at least for keywords.

**Note:** `upsert` option could've been used in `modify()` to do everything in a single line but unfortunately this method doesn't seem to return the object. Maybe there's still a more elegant way that I'm overlooking.

### DIFF
--- a/udata/core/site/models.py
+++ b/udata/core/site/models.py
@@ -1,3 +1,5 @@
+import logging
+
 from flask import g, current_app
 from werkzeug.local import LocalProxy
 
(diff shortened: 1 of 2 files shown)


udata/core/site/models.py
```diff
@@ -142,10 +142,15 @@ def count_max_org_datasets(self):
...
-            'title': current_app.config.get('SITE_TITLE'),
-            'keywords': current_app.config.get('SITE_KEYWORDS', []),
+        site_title = current_app.config.get('SITE_TITLE')
+        site_keywords = current_app.config.get('SITE_KEYWORDS', [])
+        g.site, created = Site.objects.get_or_create(id=site_id, defaults={
+            'title': site_title,
+            'keywords': site_keywords,
         })
+        if not created:
+            Site.objects(id=site_id).modify(set__title=site_title, set__keywords=site_keywords)
```
maudetes, line 152: This would imply a modification every time with access `current_site`, which may be overkill.
Instead of using a `not created` as condition which will be true most of the time, would it be more efficient to compare `if g.site.keyword != site_keyword` and same with title?

maudetes (commented):
Thank you for this PR. I don't think we've encountered the case recently, but we should indeed fix this behavior.

Maybe we could deal with creation at init time and only get the Site object in `get_current_site`?

udata/core/site/models.py
tboye: True, I somehow thought that it won't be called every time.

maudetes (approved):
Nice, thank you :) I've shortly looked around for a different place to init the Site object but did not find something obvious, let's go this way ! :+1:

PR #2904, opened 2023-10-10
Topics upgrades : add a created_at field and allow to filter on tags or name

quaxsze:
fix https://github.com/ecolabdata/ecospheres-front/issues/51

### DIFF
--- a/udata/core/topic/models.py
+++ b/udata/core/topic/models.py
@@ -1,5 +1,6 @@
+from datetime import datetime
 from flask import url_for
-
+from mongoengine.fields import DateTimeField
 from mongoengine.signals import pre_save
 from udata.models import db
 from udata.search import reindex
@@ -27,11 +28,15 @@ class Topic(db.Document, db.Owned):
     private = db.BooleanField()
     extras = db.ExtrasField()
 
+    created_at = DateTimeField(default=datetime.utcnow, required=True)
+
     meta = {
         'indexes': [
             '$name',
+            'created_at',
             'slug'
         ] + db.Owned.meta['indexes'],
+        'ordering': ['-created_at'],
         'auto_create_index_on_save': True
     }
 
--- a/udata/core/topic/models.py
+++ b/udata/core/topic/models.py
@@ -9,7 +9,7 @@ from udata.tasks import as_task_param
 __all__ = ('Topic', )
 
 
-class Topic(db.Document):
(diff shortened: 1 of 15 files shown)


udata/core/topic/models.py
```diff
@@ -27,6 +28,18 @@ class Topic(db.Document, db.Owned):
     private = db.BooleanField()
     extras = db.ExtrasField()
 
+    created_at = DateTimeField(default=datetime.utcnow, required=True)
```
maudetes, line 31: What about existing ones with `required=True`? Should we fill these somehow?

udata/forms/fields.py
```diff
@@ -540,8 +540,9 @@ def _value(self):
             return ''
 
     def process_formdata(self, valuelist):
-        if not valuelist:
```
maudetes, line 543: you may keep this condition. Else a case with valuelist = None won't match the new first condition but then `len(None)` will fail

maudetes (commented):
Nice, thank you! Could you add tests on new filters and remove all datasets bugs please? :pray:

udata/core/topic/models.py
quaxsze: `name` is filled by the user and `slug` is filled automaticaly as is `created_at`

maudetes: I meant existing Topics without a `created_at`

quaxsze: We can make a ligration but it won't be really relevant. Or remove the `required`

maudetes: I would say that it is nice to keep the `required`. I've tested your branch and it does fill the created_at value with default value, thus current utcnow.

udata/forms/fields.py
```diff
@@ -541,6 +541,8 @@ def _value(self):
 
     def process_formdata(self, valuelist):
         if not valuelist:
+            if valuelist == []:
```
maudetes, line 544: :warning:  valuelist is also `[]` if the field is simply missing! Thus, we're emptying `datasets` if the value is not set. I'm confused about how we're supposed to differentiate between an empty list and a missing value here?

We may de-prioritize this ticket if we go for https://github.com/ecolabdata/ecospheres-front/issues/52.
But I guess this issue is probably not encountered on Topics only?

maudetes (approved):
:pray:

PR #2905, opened 2023-10-10
Expose reuses extras

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/1180


maudetes (commented):
I think you'll need to update the form as well?

maudetes (approved):
:+1:

PR #2910, opened 2023-10-16
Add options to enhance cookies security

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/1167

### DIFF
--- a/udata/settings.py
+++ b/udata/settings.py
@@ -74,7 +74,7 @@ class Defaults(object):
 
     # Flask security settings
 
-    SESSION_COOKIE_SECURE = False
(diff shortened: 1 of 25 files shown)


udata/settings.py
```diff
@@ -74,6 +74,11 @@ class Defaults(object):
 
     # Flask security settings
 
+    SESSION_COOKIE_SECURE = False
```
maudetes, line 77: We may set it to True in defaults?

maudetes (commented):
Let's test it on our server with a `strict` config and secure cookie first :)

maudetes (approved):
Thank you!

PR #2913, opened 2023-10-18
feat(topics): apiv2

abulte:
Fix https://github.com/ecolabdata/ecospheres-front/issues/52

Adds the following endpoints:
- GET `/api/2/topics/` with a HATEOAS payload for datasets and reuses
- GET `/api/2/topics/{id}/` with same payload
- POST `/api/2/topics/{id}/datasets/` to add a list of datasets to a topic
- POST `/api/2/topics/{id}/reuses/` to add a list of reuses to a topic
- DELETE `/api/2/topics/{id}/datasets/{did}/` to delete a dataset from a topic
- DELETE `/api/2/topics/{id}/reuses/{rid}/` to delete a reuse from a topic

Endpoints are tested, and secured when needed.

Bonus: a conditional on search adapter index/unindex methods because I had some errors popping up from there when running the tests (which have `SEARCH_SERVICE_API_URL = None` setting by default).

### DIFF
--- a/udata/core/topic/apiv2.py
+++ b/udata/core/topic/apiv2.py
@@ -12,6 +12,7 @@ from udata.core.reuse.models import Reuse
 from udata.core.reuse.apiv2 import reuse_page_fields
 from udata.core.topic.models import Topic
 from udata.core.topic.parsers import TopicApiParser
+from udata.core.topic.permissions import TopicEditPermission
 from udata.core.user.api_fields import user_ref_fields
 
 DEFAULT_SORTING = '-created_at'
@@ -115,7 +116,7 @@ class TopicDatasetsAPI(API):
         return (Dataset.objects.filter(id__in=(d.id for d in topic.datasets))
                 .paginate(args['page'], args['page_size']))
 
-    # FIXME: secure
+    @apiv2.secure
     @apiv2.doc('topic_datasets_create')
     @apiv2.expect([dataset_add_fields])
     @apiv2.marshal_with(topic_fields)
@@ -124,9 +125,9 @@ class TopicDatasetsAPI(API):
     @apiv2.response(400, 'Expecting a list of dicts with id attribute')
     @apiv2.response(404, 'Topic not found')
     @apiv2.response(404, 'Dataset(s) not found')
+    @apiv2.response(403, 'Forbidden')
     def post(self, topic):
         '''Add datasets to a given topic from a list of dataset ids'''
-
         def add_dataset(topic, dataset):
             # TODO: maybe a mongo query on Dataset would be faster?
             if dataset.id not in (d.id for d in topic.datasets):
@@ -139,6 +140,9 @@ class TopicDatasetsAPI(API):
             except mongoengine.errors.ValidationError:
                 apiv2.abort(400, 'Malformed object id(s) in request')
             return dataset
+
+        if not TopicEditPermission(topic).can():
+            apiv2.abort(403, 'Forbidden')
 
         data = request.json
 
@@ -157,20 +161,24 @@ class TopicDatasetsAPI(API):
         return topic, 201
 
 
-# FIXME: secure
 @ns.route('/<topic:topic>/datasets/<dataset:dataset>/', endpoint='topic_dataset', doc={
     'params': {'topic': 'The topic ID', 'dataset': 'The dataset ID'}
 })
 class TopicDatasetAPI(API):
+    @apiv2.secure
     @apiv2.response(404, 'Topic not found')
     @apiv2.response(404, 'Dataset not found in topic')
     @apiv2.response(204, 'Success')
     def delete(self, topic, dataset):
         '''Delete a given dataset from the given topic'''
+        if not TopicEditPermission(topic).can():
+            apiv2.abort(403, 'Forbidden')
+
         if dataset.id not in (d.id for d in topic.datasets):
             apiv2.abort(404, 'Dataset not found in topic')
         topic.datasets = [d for d in topic.datasets if d.id != dataset.id]
         topic.save()
+
         return None, 204
 
 
--- a/udata/core/topic/apiv2.py
+++ b/udata/core/topic/apiv2.py
@@ -1,28 +1,33 @@
 import logging
 
-from flask import url_for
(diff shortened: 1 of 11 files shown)


quaxsze (approved):

udata/core/topic/apiv2.py
```diff
@@ -0,0 +1,258 @@
...
+    'id': fields.String(description='Id of the item to add', required=True),
+}, location='json')
+
+
+@ns.route('/<topic:topic>/datasets/', endpoint='topic_datasets', doc=common_doc)
+class TopicDatasetsAPI(API):
+    @apiv2.doc('topic_datasets')
+    @apiv2.expect(generic_parser)
+    @apiv2.marshal_with(dataset_page_fields)
+    def get(self, topic):
```
maudetes, line 113: Seems to work well!
But I'm scared that some users *will probably* want to add some filters or sort, redeveloping a more complex dataset list endpoint.

Could we somehow share some code with apiv1 `datasets` list endpoint by adding a topic filter? Or even imagine a redirect?

maudetes (commented):
Thank you a lot for this PR and sorry for the review delay.

Seems like your proposed POST/DELETE pattern is quite clear, and the endpoints are well documented, making for a readable swagger.

udata/core/topic/apiv2.py
```diff
@@ -0,0 +1,258 @@
...
+            except mongoengine.errors.ValidationError:
+                apiv2.abort(400, 'Malformed object id(s) in request')
+            return dataset
+
+        if not TopicEditPermission(topic).can():
+            apiv2.abort(403, 'Forbidden')
+
+        data = request.json
+
+        if not isinstance(data, list):
```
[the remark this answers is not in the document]
abulte: Don't want to if I don't have to 😅 

Look, [no Form](https://github.com/opendatateam/udata/blob/42cba4c39abcdab7bced628e99c7e37921e64470/udata/core/dataset/apiv2.py#L225C18-L225C18)!

udata/core/topic/apiv2.py
abulte: About adding filter: maybe simpler to keep it as is (2 lines of code, generic parser) and mutualize if and when more filters are needed?

About redirect: not sure how it could work 🤔 Won't be to v1 anyway.

udata/core/topic/apiv2.py
```diff
@@ -0,0 +1,258 @@
...
+            apiv2.abort(400, 'Expecting a list')
+        if not all(isinstance(d, dict) and d.get('id') for d in data):
+            apiv2.abort(400, 'Expecting a list of dicts with id attribute')
+
+        datasets = (get_dataset(d['id']) for d in data)
+        for dataset in datasets:
+            topic = add_dataset(topic, dataset)
+        topic.save()
+
+        # TODO: maybe we should return None, or the topics/datasets page
```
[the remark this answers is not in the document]
abulte: So... keep it like this?

udata/core/topic/apiv2.py
abulte: I'm weak, I used the `DatasetParser` from api v1 in the latest commit ;-)

We should use the topic arg when it's available from #2915.

Only drawback I see is mixing v1 and v2. I don't think this needs dedicated testing, the parser should already be tested elsewhere.

maudetes: Thank you! I've merged https://github.com/opendatateam/udata/pull/2915.
I think mixing both for now is not much of a problem, let's reuse what works :)

abulte: 🤝

udata/core/topic/apiv2.py
```diff
@@ -0,0 +1,261 @@
...
+@ns.route('/<topic:topic>/datasets/', endpoint='topic_datasets', doc=common_doc)
+class TopicDatasetsAPI(API):
+    @apiv2.doc('topic_datasets')
+    @apiv2.expect(dataset_parser)
+    @apiv2.marshal_with(dataset_page_fields)
+    def get(self, topic):
+        '''Get a given topic datasets, with filters'''
+        args = dataset_parser.parse()
+        args['topic'] = topic.id
+        datasets = dataset_parser.parse_filters(Dataset.objects.visible(), args)
```
abulte, line 119: @maudetes WDYT about `.visible()` here? Seems pretty legit but might lead to some surprises, such with `resources__0__exists=True`.

udata/core/topic/apiv2.py
```diff
@@ -0,0 +1,269 @@
...
+    @apiv2.expect(reuse_parser.parser)
+    @apiv2.marshal_with(reuse_page_fields)
+    def get(self, topic):
+        '''Get a given topic reuses, with filters'''
+        args = reuse_parser.parse()
+        reuses = Reuse.objects(deleted=None, private__ne=True).filter(
+            id__in=[d.id for d in topic.reuses]
+        )
+        # warning: topic in reuse_parser is different from Topic
+        args.pop('topic', None)
```
abulte, line 202: @maudetes :-/

abulte: Actually I don't need to pop it, it will filter properly on topicoThematique of the reuse. But how ugly this is 🙈

udata/core/topic/apiv2.py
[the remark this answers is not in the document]
maudetes: Let's keep it this way

udata/core/topic/apiv2.py
maudetes: I would use `Dataset.objects(archived=None, deleted=None, private=False)` instead indeed. Same as apiv1 dataset list endpoint

udata/core/topic/apiv2.py
maudetes: Oh my, quite ugly indeed! We'll need to fix this semantic issue shortly

abulte:
@maudetes I think we're good ✅  (I can't relaunch the build, but I think it should be alright, tests are passing).

maudetes (approved):
Thank you! :pray:

We can deploy the new routes shortly in our demo env for tests a bit more at scale.

PR #2914, opened 2023-10-19
Add contact points for Orgs and Users

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/1182


udata/core/organization/api.py
```diff
@@ -206,6 +208,21 @@ def delete(self, org, badge_kind):
         return badges_api.remove(org, badge_kind)
 
 
+@ns.route('/<org:org>/contact/', endpoint='org_contact_points')
+class OrgContactAPI(API):
+    @api.secure
```
abulte, line 213: Just passing by 👋 I have no context, I might be totally wrong.

From what I understand, this is not enough to secure a write endpoint, this only checks that the user is authenticated. You need to check specific permissions explicitly in the endpoint (CanEditOrganization I guess).

udata/core/dataset/api.py
```diff
@@ -234,6 +236,26 @@ def delete(self, dataset):
         return '', 204
 
 
+@ns.route('/<dataset:dataset>/contact/', endpoint='dataset_contact_points')
+class DatasetContactAPI(API):
+    @api.secure
+    @api.doc('create_dataset_contact_point')
+    @api.expect(contact_points_fields)
+    @api.marshal_list_with(contact_points_fields, code=201)
+    def post(self, dataset):
```
maudetes, line 245: We'll need a route to detach a contact point from a dataset?

maudetes (commented):
Thank you for this contribution!
I haven't gotten my head around this modeling. I did not find the routes intuitive to manipulate contact points. Maybe using a `db.Owned` pattern would make it easier (and more coherent with existing patterns in udata). It's an open question, would be curious about your opinion.

udata/core/dataset/api.py
```diff
@@ -245,6 +247,26 @@ def delete(self, dataset):
...
+        '''Assign a contact point to a dataset'''
+        data = request.json
+        contact_point_id = data.get('id')
+        if not contact_point_id:
+            api.abort(400, 'Wrong payload format, id expected')
+        DatasetEditPermission(dataset).test()
+        contact_point = ContactPoint.objects.get_or_404(id=id_or_404(contact_point_id))
```
maudetes, line 263: We need to make sure the contact point belongs to the owner of the dataset

udata/core/dataset/models.py
```diff
@@ -477,6 +477,8 @@ class Dataset(WithMetrics, BadgeMixin, db.Owned, db.Document):
 
     featured = db.BooleanField(required=True, default=False)
 
+    contact_points = db.ListField(db.ReferenceField(ContactPoint, reverse_delete_rule=db.PULL))
```
maudetes, line 480: We're modeling multiple contact points for one dataset here @agarrone.
It may be useful later if we want to have multiple roles (publisher, legal owner, etc.).

Is this an issue?

udata/core/dataset/rdf.py
```diff
@@ -312,6 +312,17 @@ def temporal_from_rdf(period_of_time):
...
 
 
+def contact_point_from_rdf(rdf):
+    contact_point = rdf.value(DCAT.contactPoint)
+    if contact_point:
+        name = contact_point.value(VCARD.fn)
+        email = (contact_point.value(VCARD.hasEmail)
+                 or contact_point.value(VCARD.email)
+                 or contact_point.value(DCAT.email))
+        contact = ContactPoint(name=name, email=email).save()
```
maudetes, line 322: We should either get or create the contact point from the owner/organization here.

udata/core/user/api.py
```diff
@@ -316,6 +316,25 @@ def delete(self, user):
...
 
+from udata.models import ContactPoint
+from udata.core.contact_point.api import ContactPointApiParser
+from udata.core.contact_point.api_fields import contact_point_page_fields
+
+
+contact_point_parser = ContactPointApiParser()
+
+
+@ns.route('/<user:user>/contact/', endpoint='user_contact_points')
```
maudetes, line 327: I'm a bit confused between `contact` vs `contacts`. Shouldn't it be plural for org and users?

maudetes (commented):
I have only minor comments.
As a more general note, I'm thinking we should use a new custom `fields.ContactField` in forms instead of adding new routes to attach or detach contact points to datasets. What do you think?

udata/core/dataset/models.py
```diff
@@ -477,6 +477,8 @@ class Dataset(WithMetrics, BadgeMixin, db.Owned, db.Document):
 
     featured = db.BooleanField(required=True, default=False)
 
+    contact_point = db.ReferenceField('ContactPoint', reverse_delete_rule=db.PULL)
```
maudetes, line 480: I had this issue occurring when deleting a contact point that had been attached to a dataset:
```
Update failed (Cannot apply $pull to a non-array value, full error: {'index': 0, 'code': 2, 'errmsg': 'Cannot apply $pull to a non-array value'})
```

udata/core/dataset/rdf.py
```diff
@@ -470,6 +489,9 @@ def dataset_from_rdf(graph, dataset=None, node=None):
     description = d.value(DCT.description) or d.value(DCT.abstract)
     dataset.description = sanitize_html(description)
     dataset.frequency = frequency_from_rdf(d.value(DCT.accrualPeriodicity))
+    bob = contact_point_from_rdf(d, dataset) or dataset.contact_point
+    print(bob)
```
maudetes, line 493: Hello bob :wave:

udata/core/contact_point/api.py
```diff
@@ -0,0 +1,68 @@
...
+        '''Updates a contact point given its identifier'''
+        form = api.validate(ContactPointForm, contact_point)
+        return form.save()
+
+    @api.secure
+    @api.doc('delete_contact_point')
+    @api.response(204, 'Contact point deleted')
+    def delete(self, contact_point):
+        '''Deletes a contact point given its identifier'''
+        contact_point.delete()
```
maudetes, line 67: You should remove reference to this contact point in impacted Datasets

udata/core/contact_point/api.py
```diff
@@ -0,0 +1,68 @@
...
+contact_point_parser = ContactPointApiParser()
+
+
+@ns.route('/', endpoint='contact_points')
+class ContactPointsListAPI(API):
+    '''Contact points collection endpoint'''
+    @api.doc('list_contact_points')
+    @api.marshal_with(contact_point_page_fields)
+    def get(self):
+        '''List all contact points'''
```
maudetes, line 27: I wonder if this list makes much sense? In particular since we don't have orga/owner in resulting fields

maudetes (approved):
I find the API nice now! :clap: Congrats!

PR #2915, opened 2023-10-20
fix(dataset): add topic filter on list

abulte:
Related to https://github.com/opendatateam/udata-search-service/pull/43

This adds the `topic` filter to the dataset list, thus mapping to the external search index capability linked above.

### DIFF
--- a/udata/core/dataset/api.py
+++ b/udata/core/dataset/api.py
@@ -54,6 +54,7 @@ from .api_fields import (
     schema_fields,
 )
 from udata.linkchecker.checker import check_resource
+from udata.core.topic.models import Topic
 from .models import (
     Dataset, Resource, Checksum, License, UPDATE_FREQUENCIES,
     CommunityResource, RESOURCE_TYPES, ResourceSchema, get_resource
@@ -95,6 +96,7 @@ class DatasetApiParser(ModelApiParser):
         self.parser.add_argument('format', type=str, location='args')
         self.parser.add_argument('schema', type=str, location='args')
         self.parser.add_argument('schema_version', type=str, location='args')
+        self.parser.add_argument('topic', type=str, location='args')
 
     @staticmethod
     def parse_filters(datasets, args):
@@ -132,6 +134,13 @@ class DatasetApiParser(ModelApiParser):
             datasets = datasets.filter(resources__schema__name=args['schema'])
         if args.get('schema_version'):
             datasets = datasets.filter(resources__schema__version=args['schema_version'])
+        if args.get('topic'):
+            try:
+                topic = Topic.objects.get(id=args['topic'])
+            except Topic.DoesNotExist:
+                pass
+            else:
+                datasets = datasets.filter(id__in=[d.id for d in topic.datasets])
         return datasets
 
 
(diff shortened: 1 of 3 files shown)


quaxsze (approved):

udata/core/dataset/api.py
```diff
@@ -132,6 +134,15 @@ def parse_filters(datasets, args):
...
             datasets = datasets.filter(resources__schema__version=args['schema_version'])
+        if args.get('topic'):
+            if not ObjectId.is_valid(args['topic']):
+                api.abort(400, 'Topic arg must be an identifier')
+            try:
+                topic = Topic.objects.get(id=args['topic'])
+            except Topic.DoesNotExist:
+                pass
+            else:
+                datasets = datasets.filter(id__in=[d.id for d in topic.datasets])
```
maudetes, line 145: There may be performance issue here. Not sure if adding datasets ids in Topics index could help?

maudetes (commented):
Nice! Could you add an entry in the changelog?

udata/core/dataset/api.py
abulte:
```diff
diff --git a/udata/core/topic/models.py b/udata/core/topic/models.py
index c43dec0e8..f076f1134 100644
--- a/udata/core/topic/models.py
+++ b/udata/core/topic/models.py
@@ -34,7 +34,9 @@ class Topic(db.Document, db.Owned):
         'indexes': [
             '$name',
             'created_at',
-            'slug'
+            'slug',
+            'datasets.id',
+            'reuses.id',
         ] + db.Owned.meta['indexes'],
         'ordering': ['-created_at'],
         'auto_create_index_on_save': True
```

Adding this to the model does not change anything on my local env for a Topic with 5000 Datasets attached. FYI the overhead of the topic filter is ~1 second on a total of 6 seconds (yeah local env is slow).

Not actually sure how/when indexes are supposed to be computed on this ~~mock of a~~ DB, tried this to be sure:

```
>>> Topic.compare_indexes()
{'missing': [], 'extra': [[('datasets._id', 1)], [('reuses._id', 1)]]}
>>> Topic.ensure_indexes()
>>> Topic.compare_indexes()
{'missing': [], 'extra': [[('datasets._id', 1)], [('reuses._id', 1)]]}
```

I want to believe that `id` based filtering is supposed to come pre-optimized but I actually have no idea.

maudetes: Why is your local env this slow?
Tested locally with a topic with 15K attached, the entire request takes up to 300ms. Let's skip the index idea!

maudetes (approved):

udata/core/dataset/api.py
abulte: I don't know 😅 Did not have time to dig.

What's your value of `FLASK_CACHE`?

I think I recall from 6 years ago that I had to dig through a confusing network configuration on my local env to speed things up, but I don't remember what it was. FYI I was a brand new env and machine since the last time I worked on udata.

maudetes: Not sure but I think @quaxsze had some performance issues at some point as well.
I have `CACHE_TYPE` set to `flask_caching.backends.null`.

abulte: For reference (not the best place for that I guess :p) I found the "network issue" I was talking about on local (I'm using `dev.local`): https://superuser.com/a/1596341

abulte: Never written so much doc for this project (should have done it 6 years ago!): https://github.com/opendatateam/udata/pull/2923

PR #2917, opened 2023-10-21
fix(topic): secure API v1

abulte:


maudetes (approved):
Oh thank you for the fix! :pray:

PR #2918, opened 2023-10-21
Consider acronym when suggesting organizations

tboye:
Some organizations are mostly known by their abbreviations, so it would be beneficial to allow for suggesting organizations based on both names and acronyms.

### DIFF
--- a/udata/tests/api/test_organizations_api.py
+++ b/udata/tests/api/test_organizations_api.py
@@ -643,19 +643,18 @@ class MembershipAPITest:
         for i in range(3):
             OrganizationFactory(
                 name=faker.word(),
-                acronym=f'TEST{i}' if i % 2 else faker.word(),
+                acronym=f'UDATA{i}' if i % 2 else faker.word(),
                 metrics={"followers": i})
         max_follower_organization = OrganizationFactory(
             name=faker.word(),
-            acronym='TEST4',
+            acronym='UDATA4',
             metrics={"followers": 10}
         )
         response = api.get(url_for('api.suggest_organizations'),
-                           qs={'q': 'tEsT', 'size': '5'})
+                           qs={'q': 'uDaTa', 'size': '5'})
         assert200(response)
 
-        assert len(response.json) <= 5
-        assert len(response.json) > 1
+        assert len(response.json) == 2
 
         for suggestion in response.json:
             assert 'id' in suggestion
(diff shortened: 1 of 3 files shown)


udata/tests/api/test_organizations_api.py
```diff
@@ -637,6 +637,35 @@ def test_suggest_organizations_homonyms(self, api):
         for suggestion in response.json:
             assert suggestion['name'] == 'homonym'
 
+    def test_suggest_organizations_acronym(self, api):
+        '''Should suggest organizations based on acronym'''
+
+        for i in range(3):
+            OrganizationFactory(
+                name=faker.word(),
+                acronym=f'TEST{i}' if i % 2 else faker.word(),
```
maudetes, line 646: `TEST` is a risky one since faker.word() sometimes output words that contains `test`.
Maybe replace by a rarer word: `udata`, `opendata`, etc.?

udata/tests/api/test_organizations_api.py
```diff
@@ -637,6 +637,35 @@ def test_suggest_organizations_homonyms(self, api):
...
+        max_follower_organization = OrganizationFactory(
+            name=faker.word(),
+            acronym='TEST4',
+            metrics={"followers": 10}
+        )
+        response = api.get(url_for('api.suggest_organizations'),
+                           qs={'q': 'tEsT', 'size': '5'})
+        assert200(response)
+
+        assert len(response.json) <= 5
```
maudetes, line 657: Shouldn't be `len(response.json)` precisely defined?

maudetes (commented):
Thank you for your contribution, quite useful indeed! :clap:
 I have minor comments on the test.

udata/tests/api/test_organizations_api.py
tboye: I had the same concern when I was first looking at the existing tests for inspiration. As far as I understand the idea, we assume that the faker may return something that will match, that's why this `assert` is not strict. In case we use 'UDATA' as acronym, we can probably make it strict here. Should I make this change?

maudetes: Mmm indeed, we should probably clean other tests to make the assert strict. And yes please you can change for 'UDATA' here, to prevent adding more precarious tests! 🙏

maudetes (approved):
:pray:

PR #2925, opened 2023-11-15
Access rights and index dataset without file

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/1186


udata/core/dataset/rdf.py
```diff
@@ -488,6 +489,7 @@ def dataset_from_rdf(graph, dataset=None, node=None):
     default_license = dataset.license or License.default()
     dataset_license = rdf_value(d, DCT.license)
     dataset.license = License.guess(dataset_license, *licenses, default=default_license)
+    dataset.access_rights = rdf_value(d, DCT.accessRights)
```
maudetes, line 492: We should map from and towards the European vocabulary: http://publications.europa.eu/resource/authority/access-right.

maudetes (commented):
Nice already!

udata/core/dataset/models.py
```diff
@@ -491,6 +492,8 @@ class Dataset(WithMetrics, BadgeMixin, db.Owned, db.Document):
     deleted = db.DateTimeField()
     archived = db.DateTimeField()
 
+    access_rights = db.StringField(choices=list(ACCESS_RIGHTS), default=DEFAULT_ACCESS_RIGHTS, required=True)
```
maudetes, line 495: access_rights is at the Dataset level, may not match with usage and DCAT?

udata/core/dataset/search.py
```diff
@@ -45,8 +45,11 @@ class DatasetSearch(ModelSearchAdapter):
 
     @classmethod
     def is_indexable(cls, dataset):
```
maudetes, line 47: We don't need to modify this anymore, see https://github.com/opendatateam/udata/pull/2997

PR #2928, opened 2023-11-21
feat: early work to handle addtionnal resources

Pierlou:
@maudetes

### DIFF
--- a/udata/core/dataset/rdf.py
+++ b/udata/core/dataset/rdf.py
@@ -389,17 +389,8 @@ def remote_url_from_rdf(rdf):
             except uris.ValidationError:
                 pass
 
-def theme_labels_from_rdf(rdf):
-    for theme in rdf.objects(DCAT.theme):
-        if isinstance(theme, RdfResource):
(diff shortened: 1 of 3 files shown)


udata/core/dataset/rdf.py
```diff
@@ -410,9 +411,12 @@ def resource_from_rdf(graph_or_distrib, dataset=None):
...
 
-    download_url = url_from_rdf(distrib, DCAT.downloadURL)
-    access_url = url_from_rdf(distrib, DCAT.accessURL)
-    url = safe_unicode(download_url or access_url)
+    if not is_additionnal:
+        download_url = url_from_rdf(distrib, DCAT.downloadURL)
+        access_url = url_from_rdf(distrib, DCAT.accessURL)
+        url = safe_unicode(download_url or access_url)
+    else:
+        url = distrib.identifier.toPython() if isinstance(distrib.identifier, URIRef) else None
```
Pierlou, line 419: So we could end up with additionnal resources without URL ?

Pierlou (commented):
Few changes were needed indeed, nice!

PR #2929, opened 2023-11-21
Allow for id attribute in a tags in markdown

maudetes:
This would allow for adding anchors in markdown contents.
Exemple :
```
# Cookie <a id="cookie"></a>

We want a section full of cookies (the edible ones).
```
We could then use the anchor as `/posts/my-post-about-food/#cookie`

### Notes
I haven't found any other simpler way to add anchors with our markdown parser.
I haven't checked yet for security warning on adding these.


maudetes:
Well, it is a complicated topic with security issues : https://github.com/markdown-it/markdown-it/issues/28.
Using `name` *would* be safer but not cleared of all risks?
The need for anchors may not be worth the trouble.

PR #2935, opened 2023-12-05
feat(oauth): allow wildcards in redirect_uris

abulte:
This add a setting and logic to allow wildcards in `redirect_uris` for an `OAuth2Client`.

On Ecosphères, it will be very useful to be able to use OAuth when testing PR previews that have URLs like https://deploy-preview-205--ecospheres.netlify.app. We could authorize `*ecospheres.netlify.app/login/callback` for our client.

Since this might be dangerous security-wise, it's behind a flag that is `False` by default. I suggest enabling it only on dev/demo instances and not in production.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
     - `udata spatial migrate` to migrate datasets geozones to new ones
     - Reindex datasets
   - Removed forgotten fields in search [#2934](https://github.com/opendatateam/udata/pull/2934)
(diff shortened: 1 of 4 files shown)


maudetes (approved):
Nice! Seems secure enough with default flag indeed.

Since it's a hassle to test it correctly locally, if you update the branch I can deploy it on our demo env before merging :)

abulte:
@maudetes updated! FYI I tested locally with a `*.local` pattern and `ecospheres-front`, but it never hurts to check again.

PR #2939, opened 2023-12-12
Don't use a pluggableDomain but init Babel properly instead

maudetes:
Completes https://github.com/opendatateam/udata/pull/2897

Flask-babel [supports multiple domains and translation directories](https://github.com/python-babel/flask-babel/pull/194) `;`-separated in config.

See CKAN migration for reference: https://github.com/ckan/ckan/pull/7436/.

Flask-security forms not working (password, etc.) was due to the config
```
SECURITY_I18N_DOMAIN = 'udata'
```


quaxsze (approved):

quaxsze:
🤩

PR #2941, opened 2023-12-15
Allow for being one day late on update fulfilled in time

maudetes:
We may have up to one day delay due to harvesting causing misleading warnings.

Indeed with a dataset harvested daily, we are not immediately aware of a resource update. It would be more cautious to wait for an additional day before lowering the quality score.

What do you think of this @agarrone ?


agarrone:
It sounds resonable :)

quaxsze (approved):

PR #2944, opened 2024-01-10
When topic is deleted, discussions are purged

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/1267


maudetes (approved):
:pray:

PR #2949, opened 2024-01-25
 Harvest dct:conformsTo for Dataset and Resource

ThibaudDauce:


udata/core/dataset/models.py
```diff
@@ -915,24 +989,45 @@ def objects():
...
+        if not schemas:
             log.error(f'No content found inc. from cache for schema catalog')
             raise SchemasCacheUnavailableException('No content in cache for schema catalog')
-        return content
-
 
+        return schemas
+    
+    @staticmethod
+    def get_mock_data():
```
maudetes, line 1004: Could you somehow move this mock logic in the test code?

udata/core/dataset/rdf.py
```diff
@@ -458,6 +458,7 @@ def resource_from_rdf(graph_or_distrib, dataset=None, is_additionnal=False):
     resource.filesize = rdf_value(distrib, DCAT.byteSize)
     resource.mime = mime_from_rdf(distrib)
     resource.format = format_from_rdf(distrib)
+    resource.schema = schema_from_rdf(distrib)
```
maudetes, line 461: Since schema is not properly exposed in most catalogs, it would be nice to assign it only if found, to prevent erasing what users might fill on datagouv direclty:
```suggestion
     schema = schema_from_rdf(distrib)
     if schema:
         resource.schema = schema
```

maudetes (commented):
Thank  you! I've made a first batch of review, but this seems promising!

udata/core/dataset/forms.py
```diff
@@ -32,46 +32,49 @@ def normalize_format(data):
...
+    model_class = Schema
+    url = fields.URLField()
+    name = fields.StringField(_('Name of the schema'))
+    version = fields.StringField(_('Version of the schema'))
 
-        if not bool('name' in schema) ^ bool('url' in schema):
-            raise validators.ValidationError(_('Schema must have at least a name or an url. Having both is not allowed.'))
+    def validate_url(form, field):
+        if not field.data: 
+            if form.name.data: return
```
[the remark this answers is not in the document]
ThibaudDauce: Yeah, it's unconventional :-) I find it more readable but I changed them if it's the current policy in the codebase. (maybe add the linter into the CI?)

udata/core/dataset/models.py
ThibaudDauce: Do you have a good place to put this? I thought of `udata/core/dataset/factories.py`?

udata/core/dataset/forms.py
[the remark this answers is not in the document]
maudetes: We plan to, but we have to fix existing linting issues first :p

udata/core/dataset/forms.py
```diff
@@ -32,46 +32,49 @@ def normalize_format(data):
...
-        allowed_schemas = [s['id'] for s in ResourceSchema.objects()]
+class SchemaForm(ModelForm):
+    model_class = Schema
+    url = fields.URLField()
+    name = fields.StringField(_('Name of the schema'))
+    version = fields.StringField(_('Version of the schema'))
 
-        if not bool('name' in schema) ^ bool('url' in schema):
-            raise validators.ValidationError(_('Schema must have at least a name or an url. Having both is not allowed.'))
+    def validate_url(form, field):
```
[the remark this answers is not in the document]
maudetes: :eyes: oooh! I find it elegant!
I think it hasn't been used in udata so far, so maybe adding it explicitly may be more in-line with the existing codebase, sorry!

udata/core/dataset/models.py
maudetes: Yes, I think moving it to factories could do! Thanks

udata/core/dataset/forms.py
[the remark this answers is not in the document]
ThibaudDauce: Can I just add a comment to explain?

udata/core/dataset/forms.py
```diff
@@ -32,46 +32,51 @@ def normalize_format(data):
...
+
+        if form.url.data: 
+            raise validators.ValidationError(_('Having both name and URL is not allowed.'))
+
+        name = field.data
+        version = form.version.data
+
+        # If there is no URL, the name must match a known schema from our catalog.
+        allowed_schemas_version_by_name = {schema['name']: schema['versions'] for schema in ResourceSchema.all()}
+        allowed_versions = allowed_schemas_version_by_name.get(name)
```
maudetes, line 64: I think allowed_versions is not a list of versions str at that time, but a list of objects.
Not sure why it doesn't show in tests when looking for `version not in allowed_versions`, but we may lack this case coverage

udata/core/dataset/models.py
```diff
@@ -150,6 +149,68 @@ class HarvestResourceMetadata(DynamicEmbeddedDocument):
...
+        '''
+        In the database, since the schemas were only simple dicts, there is
+        empty `{}` stored. To prevent problems with converting to bool 
+        (bool({}) and default bool(Schema) do not yield the same value), we transform
+        empty Schema() to False.
+        It's maybe not necessary but being paranoid here.
+        '''
+        return bool(self.name) or bool(self.url)
+
+    def get_url(self):
```
maudetes, line 167: I think we don't use `get_name` and `get_url` anymore since we've simplified the api logic (returning what's stored only)?

udata/rdf.py
```diff
@@ -218,6 +220,25 @@ def url_from_rdf(rdf, prop):
         return value.identifier.toPython()
 
 
+def schema_from_rdf(rdf):
```
maudetes, line 223: :+1:

udata/rdf.py
```diff
@@ -218,6 +220,25 @@ def url_from_rdf(rdf, prop):
...
+    Try to extract a schema from a conformsTo property.
+    Currently the "issued" property is not harvest.
+    '''
+    resource = rdf.value(DCT.conformsTo)
+    if not resource:
+        return None
+
+    schema = Schema()
+    if isinstance(resource, (URIRef, Literal)):
+        schema.url = resource.toPython()
```
maudetes, line 234: In the particular case of harvesting, could we imagine setting schema.name if the url namespace is from schema.data.gouv.fr? Or would that lead again to inconsistencies in the end?

udata/core/dataset/forms.py
```diff
@@ -32,46 +32,51 @@ def normalize_format(data):
...
-        if 'url' in schema:
-            try:
-                validate_url(schema.get('url'))
-            except ValidationError:
-                raise validators.ValidationError(_('Provided URL is not valid.'))
+        if form.name.data:
+            raise validators.ValidationError(_('Having both name and URL is not allowed.'))
 
-        if 'name' in schema and schema.get('name') not in allowed_schemas:
+    def validate_name(form, field):
```
maudetes, line 50: As discussed, we may want to move this validation logic at the model level, but call it explicitly from this form validation?

udata/core/dataset/api_fields.py
```diff
@@ -245,6 +253,8 @@
     'quality': fields.Raw(description='The dataset quality', readonly=True),
     'last_update': fields.ISODateTime(
         description='The resources last modification date', required=True),
+    'schema': fields.Nested(
```
maudetes, line 256: We may want to update `DEFAULT_MASK` on dataset fields to add schema? Not totally sure if we don't want to confuse unaware users

udata/core/dataset/forms.py
ThibaudDauce: Good catch! Done :-)

udata/rdf.py
ThibaudDauce: Yes, it's fine I think! Done.

udata/core/dataset/models.py
ThibaudDauce: Yes, we can add them back if needed (for reference the commit of deletion 75585fc64)

udata/core/dataset/forms.py
ThibaudDauce: Done

udata/core/dataset/models.py
```diff
@@ -289,8 +350,8 @@ class ResourceMixin(object):
     fs_filename = db.StringField()
     extras = db.ExtrasField()
     harvest = db.EmbeddedDocumentField(HarvestResourceMetadata)
-    schema = db.DictField()
-
+    schema = db.EmbeddedDocumentField(Schema)
```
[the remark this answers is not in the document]
maudetes: Minimal example to reproduce this issue:
```
POST http://dev.local:7000/api/1/datasets/

{
  "description": "bla",
  "title": "A newly created dataset without schema"
}
```

```
POST http://dev.local:7000/api/1/datasets/a-newly-created-dataset-without-schema/resources/

{
  "filetype": "remote",
  "title": "Link to data.gouv.fr",
  "url": "https://data.gouv.fr/"
}
```

Then go to newly created dataset: http://dev.local:7000/fr/datasets/a-newly-created-dataset-without-schema/

maudetes (approved):
Thank you for this PR, it was a bit of spaghetti :spaghetti: to get it right!
We may improve on schema/conformsTo logic with usage later on :)

PR #2950, opened 2024-01-29
Fix remove schema from admin

ThibaudDauce:
Fix https://github.com/etalab/data.gouv.fr/issues/876


maudetes (approved):
Thank you for taking a look at this dark magic, it seems to work nicely! :magic_wand:

PR #2951, opened 2024-01-29
Fix reuses and datasets with no private attribute

ThibaudDauce:
Fix https://github.com/etalab/data.gouv.fr/issues/1262


maudetes (commented):
Thank you! Don't forget the changelog ;)
I'm wondering if we can have the same issue with Datasets?

udata/migrations/2024-01-29-fix-reuse-and-dataset-with-private-None.py
```diff
@@ -0,0 +1,36 @@
...
+
+from udata.models import Reuse, Dataset
+
+log = logging.getLogger(__name__)
+
+
+def migrate(db):
+    log.info('Processing Reuse.')
+
+    reuses = Reuse.objects().no_cache().timeout(False)
```
maudetes, line 14: We can filter on private=None
```suggestion
    reuses = Reuse.objects(private=None).no_cache().timeout(False)
```

udata/migrations/2024-01-29-fix-reuse-and-dataset-with-private-None.py
```diff
@@ -0,0 +1,36 @@
...
+    count = 0
+    for reuse in reuses:
+        if reuse.private == None:
```
maudetes, line 17: Since it defaults to False, it doesn't show as `None` here.
We may need to manipulate collections directly as was done in https://github.com/opendatateam/udata/blob/master/udata/migrations/2023-02-08-rename-internal-dates.py

maudetes (commented):
Thank you for this fix!

I think the migration doesn't work as is due to mongoengine Document logic.

udata/migrations/2024-01-29-fix-reuse-and-dataset-with-private-None.py
ThibaudDauce: Thanks!

udata/migrations/2024-01-29-fix-reuse-and-dataset-with-private-None.py
```diff
@@ -0,0 +1,21 @@
...
+
+    count = Reuse.objects(private=None).update(private=False)
+    log.info(f'Fixed {count} Reuse objects from private None to private False.')
+
+    log.info('Processing Datasets…')
+    count = Dataset.objects(private=None).update(private=False)
+    log.info(f'Fixed {count} Dataset objects from private None to private False.')
+
+    log.info('Done')
+    raise 'oups'
```
maudetes, line 21: oups ;)
```suggestion
```

maudetes (approved):
Thank you for this fix!

PR #2953, opened 2024-01-31
Add downloads count to CSV

ThibaudDauce:
See https://github.com/etalab/data.gouv.fr/issues/1237


udata/core/dataset/models.py
```diff
@@ -718,6 +718,11 @@ def quality(self):
 
         result['score'] = self.compute_quality_score(result)
         return result
+    
+    @property
+    def downloads(self):
+        return sum(map(lambda o: o.metrics.get('views', 0), self.resources))
```
maudetes, line 724: We mostly use list comprehension in udata code
```suggestion
        return sum(resource.metrics.get('views', 0) for resource in self.resources)
```

udata/core/dataset/csv.py
```diff
@@ -33,6 +33,7 @@ class DatasetCsvAdapter(csv.Adapter):
         ('tags', lambda o: ','.join(o.tags)),
         ('archived', lambda o: o.archived or False),
         ('resources_count', lambda o: len(o.resources)),
+        'downloads',
```
maudetes, line 36: I wonder if naming it `resources_downloads` would make it more explicit, following `resources_count`?

maudetes (approved):
I have only cosmetic comments :)

udata/core/dataset/csv.py
ThibaudDauce: I use the term requested in the issue https://github.com/etalab/data.gouv.fr/issues/1237 but I can change it.

maudetes: Let's keep with `downloads` :+1:

maudetes (approved):
Don't forget the changelog ;)

PR #2955, opened 2024-02-08
Update private attribute

ThibaudDauce:


udata/forms/fields.py
```diff
@@ -180,6 +180,19 @@ def __init__(self, *args, **kwargs):
...
+            return 
+
+        if valuelist[0] is None:
+            self.data = self.default
+            return
+
+        if valuelist[0] in self.false_values:
+            self.data = False
+        else:
+            self.data = True
```
maudetes, line 194: Should we raise an error if valuelist[0] is not a bool value?

udata/forms/fields.py
```diff
@@ -180,6 +180,19 @@ def __init__(self, *args, **kwargs):
         self.stacked = kwargs.pop('stacked', False)
         super(BooleanField, self).__init__(*args, **kwargs)
 
+    def process_formdata(self, valuelist):
+        if not valuelist:
+            return 
+
+        if valuelist[0] is None:
```
maudetes, line 187: Do we need to deal with `None`? Seems like this code isn't used if value is None, right?

maudetes (commented):
It seems to work, I mostly have questions :)
It's hard though to grasp any unwanted side effects, however `BooleanField` isn't that used around anyway :)

Don't forget the changelog ;)

udata/tests/api/test_fields.py
```diff
@@ -3,11 +3,46 @@
...
+            'with_default': None,
+        }, obj=model_in_db, instance=model_in_db)
+        self.assertEqual(form.without_default.data, None)
+        self.assertEqual(form.with_default.data, False)
+
+        # Without any field in request it should be the values in the model
+        print('Without any field in request')
+        form = FormWithBoolean.from_json({}, obj=model_in_db, instance=model_in_db)
+        self.assertEqual(form.without_default.data, model_in_db['without_default'])
+        self.assertEqual(form.with_default.data, model_in_db['with_default'])
```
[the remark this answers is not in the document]
maudetes: Ah sorry, I wondered what was the expected behavior you told me I should look at, but did not see that the test failed :p
I think that when saving to a model with `form.save()`, the model default is actually used at this time.
I haven't checked the responsible code, but you can check with
```
(Pdb) form.private.data
None
(Pdb) form._obj.private
True
(Pdb) form.save()
<Dataset: Jeu de données pour audit d'accessibilité !>
(Pdb) form._obj.private
False
```
But as I've said I think None value doesn't seem to get process through `process_formdata`?

udata/forms/fields.py
ThibaudDauce: Fallback to the official implementation so it's not our problem :-)

udata/tests/api/test_fields.py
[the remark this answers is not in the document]
ThibaudDauce: I remove this test, it was too hard to make it work with all the different interactions but I add a few test cases on the API test to cover `None` and a correct value. The only test case not covered is the `default=True` one but we can test that when it comes (and based on your test during our visio this morning it's working)

quaxsze (approved):

PR #2957, opened 2024-02-13
[WIP] Test CSW ISO output with SEMIC XSLT to GeoDCAT-AP

maudetes:
Parse CSW graph querying ISO schema.
Use SEMIC GeoDCAT-AP XSLT to map it to a correct version.
See https://github.com/SEMICeu/iso-19139-to-dcat-ap for more information on the XSLT.


maudetes:
Replaced by https://github.com/opendatateam/udata/pull/2982.

PR #2960, opened 2024-02-15
Fix refuse second org access

ThibaudDauce:
The `v-if` destroy the `textarea` so when querying it with `this.$el.querySelectorAll('textarea')[index].value` there are not all present (so the index is wrong). Using `v-show` instead is a quick fix because it leave the `textarea` in the DOM, so we can query all of them and keep the correct index.

Fix https://github.com/etalab/data.gouv.fr/issues/664


maudetes (approved):
Thank you! Seems to work nicely :pray:

PR #2963, opened 2024-02-15
Fix link not containing the ID in spam report

ThibaudDauce:


maudetes (approved):
Thank you for taking a look at this and fixing this

PR #2965, opened 2024-02-19
Improve spam reports

ThibaudDauce:


udata/core/discussions/models.py
```diff
@@ -19,6 +19,26 @@ class Message(SpamMixin, db.EmbeddedDocument):
...
+            return message
+        
+        discussion = breadcrumb[0]
+        if not isinstance(discussion, Discussion):
+            log.warning(f"`spam_report_message` called on message with a breadcrumb not containing a Discussion at index 0.", extra={ 'breadcrumb': breadcrumb})
+            return message
+
+        message += f" sur la discussion « [{discussion.title}]({discussion.external_url}) »"
+        if discussion.user:
+            message += f" de [{discussion.user.fullname}]({discussion.user.external_url})"
```
maudetes, line 39: I don't think the user's discussion is needed?

maudetes (approved):
Don't forget the changelog :)

udata/core/spam/api.py
```diff
@@ -53,6 +53,5 @@ def get(self):
         discussions = Discussion.objects(Q(spam__status=POTENTIAL_SPAM) | Q(discussion__spam__status=POTENTIAL_SPAM))
 
         return [{
-            'title': discussion.spam_report_title(),
-            'link': discussion.spam_report_link(),
+            'message': discussion.spam_report_message([discussion]),
```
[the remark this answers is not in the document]
ThibaudDauce: Since we have only one model right now, we could have more information… I think we can wait to have multiple models to see if we can find a better response here.

udata/core/discussions/models.py
ThibaudDauce: Removed :-)

maudetes (approved):
:ok_hand:

PR #2976, opened 2024-02-23
Fix exception in harvest

ThibaudDauce:
Some schemas inside the database are unknown so we may disable the hard check and only warn the problem right now to have logs and allow to save these datasets.

The part in RDF parsing is to avoid failing to import a dataset if the schema is not correct. It doesn't do much right now (since I've disabled the raising of the exception inside the `clean()` but we want to keep this when we re-enable the exception)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
   - Organization can nom define a custom metadata of a choosen type
   - Dataset belonging to the organization can assign a value to the defined metadata
   - Metadata value must match the choosen type by the organization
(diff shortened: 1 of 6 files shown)


nicolaskempf57 (approved):

ThibaudDauce:
Error in sentry https://errors.data.gouv.fr/organizations/sentry/issues/134167/?alert_rule_id=3&alert_timestamp=1708645323772&alert_type=email&environment=production&project=3&referrer=alert_email

PR #2980, opened 2024-02-28
[WIP] Fix harvest big catalog

ThibaudDauce:


quaxsze:
Very interesting PR. I know that @maudetes has been trying to upgrade python version to `3.10` or `3.11` for some time maybe you could see with her to merge? Or this PR will maybe make the update easier from `3.9` to `3.11`.
WDYT @maudetes ?

ThibaudDauce:
> Very interesting PR. I know that @maudetes has been trying to upgrade python version to `3.10` or `3.11` for some time maybe you could see with her to merge? Or this PR will maybe make the update easier from `3.9` to `3.11`. WDYT @maudetes ?

This PR doesn't change anything about the bump to Python 3.10 but as soon as we are on Python 3.10 in CI we can mock S3 to have unit test on this feature.

quaxsze:
> > Very interesting PR. I know that @maudetes has been trying to upgrade python version to `3.10` or `3.11` for some time maybe you could see with her to merge? Or this PR will maybe make the update easier from `3.9` to `3.11`. WDYT @maudetes ?
> 
> This PR doesn't change anything about the bump to Python 3.10 but as soon as we are on Python 3.10 in CI we can mock S3 to have unit test on this feature.

Indeed you are right. A question in the PR you use boto3 to interface with S3. We are maintaining a lib called [Flask-storage](https://github.com/etalab/flask-storage) that is already used by `udata` to handle ressources storage in FS but can handle S3 aswell. Could we use this librairy to achieve what you did?

ThibaudDauce:
> > > Very interesting PR. I know that @maudetes has been trying to upgrade python version to `3.10` or `3.11` for some time maybe you could see with her to merge? Or this PR will maybe make the update easier from `3.9` to `3.11`. WDYT @maudetes ?
> > 
> > 
> > This PR doesn't change anything about the bump to Python 3.10 but as soon as we are on Python 3.10 in CI we can mock S3 to have unit test on this feature.
> 
> Indeed you are right. A question in the PR you use boto3 to interface with S3. We are maintaining a lib called [Flask-storage](https://github.com/etalab/flask-storage) that is already used by `udata` to handle ressources storage in FS but can handle S3 aswell. Could we use this librairy to achieve what you did?

Yes we may use `Flask-storage` but not sure if it's simpler? The documentation on the readme is broken http://flask-storage.readthedocs.io/en/latest/ and it states that "Flask-Storage requires 3.9+ and Flask 1.1.4." (we are on 3.7 right now). Is this lib maintained?

quaxsze (approved):

geoffreyaldebert:
`boto3` is very standard and largely used right ? To connect with S3 like system, for me I do not think it is a problem to use this library. I have more trust on it than `flask-storage`. But yes it is another library to add to our dependencies.

udata/harvest/backends/dcat.py
```diff
@@ -58,10 +63,28 @@ def initialize(self):
...
+            max_harvest_graph_size_in_mongo = 15 * 1000 * 1000
+
+        bucket = current_app.config.get('HARVEST_GRAPHS_S3_BUCKET')
+
+        if bucket is not None and sum([len(g.encode('utf-8')) for g in serialized_graphs]) >= max_harvest_graph_size_in_mongo:
+            # TODO: we could store each page in independant files to allow downloading only the require page in
+            # subsequent jobs. (less data to download in each job)
+            filename = f'harvest_{self.job.id}_{date.today()}.json'
+
+            store_as_json(bucket, filename, serialized_graphs)
```
geoffreyaldebert, line 83: can you modify this function to add a `HARVEST_GRAPHS_S3_PATH` in it. Indeed, we can use a bucket for multiple things other than harvester. We can imagine to store all of this in an `/harvest` path for instance.

udata/settings.py
```diff
@@ -257,6 +257,14 @@ class Defaults(object):
 
     HARVEST_VALIDATION_CONTACT_FORM = None
 
+    HARVEST_MAX_CATALOG_SIZE_IN_MONGO = None # Defaults to the size of a MongoDB document 
+    HARVEST_GRAPHS_S3_BUCKET = None # If the catalog is bigger than `HARVEST_MAX_CATALOG_SIZE_IN_MONGO` store the graph inside S3 instead of MongoDB
+
+    # S3 connection details
+    S3_URL = None
+    S3_ACCESS_KEY_ID = None
+    S3_SECRET_ACCESS_KEY = None
```
geoffreyaldebert, line 266: maybe rename this with something like `UDATA_ADMIN_S3_URL` (`UDATA_ADMIN_...`)

geoffreyaldebert (changes requested):
👏 thanks
udata user can use this feature or not, which is cool. If not, they will have a problem with huge catalogs but for me telling them that they need to use S3 for that is acceptable.

udata/settings.py
geoffreyaldebert: yes you're right, so `ADMIN_S3_URL` ? because maybe in the future, resources files will be stored directly in S3 also so we may need to differenciate them in the future.

udata/harvest/backends/dcat.py
ThibaudDauce: Done in af58134a9

udata/settings.py
geoffreyaldebert: ok let's do it that way

geoffreyaldebert (approved):

ThibaudDauce:
Linked to https://github.com/opendatateam/udata/pull/2859

PR #2983, opened 2024-03-04
Prevent from loading deleted geozones

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/1303


quaxsze:
For the data.gouv.fr 2024 zones files, we drop from **53751** zones to **39216**.
**14535** zones were marked as deleted.

udata/core/spatial/commands.py
```diff
@@ -44,7 +44,10 @@ def load_levels(col, json_levels):
 
 
 def load_zones(col, json_geozones):
+    loaded_geozones = 0
     for i, geozone in enumerate(json_geozones):
```
quaxsze, line 48: Here `i` is not a total, it is an index used as a total. This index is not relevant anymore if we do not load every zone returned by `enumerate`.

ThibaudDauce: Why do you keep the `i` if it is not used?

ThibaudDauce (approved):

PR #2984, opened 2024-03-05
Add 3 new badges to organizations

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/1306


ThibaudDauce (commented):
There is a lot of duplication inside notifications / mails, do you think it is possible to refactor this to only one mail / one notify_* function?

PR #2987, opened 2024-03-08
Topic: add last_modified field

abulte:
Add `last_modified` field to `Topic` and expose it in the APIs.

Auto-handled by `db.Datetimed` with `created_at`.

### DIFF
--- a/udata/core/topic/models.py
+++ b/udata/core/topic/models.py
@@ -1,6 +1,4 @@
-from datetime import datetime
 from flask import url_for
-from mongoengine.fields import DateTimeField
 from mongoengine.signals import pre_save
 from udata.models import db
 from udata.search import reindex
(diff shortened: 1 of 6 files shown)


udata/core/topic/models.py
```diff
@@ -10,7 +8,7 @@
 __all__ = ('Topic', )
 
 
-class Topic(db.Document, db.Owned):
+class Topic(db.Document, db.Owned, db.Datetimed):
```
maudetes, line 11: I think you need to explicitly manage the `last_modified` property on api calls right?

maudetes (commented):
Thank you for this PR! I think it is a nice idea to add and expose this field :)

udata/core/topic/models.py
abulte: Yup https://github.com/opendatateam/udata/pull/2987/files#diff-66c97efd0f9b3994aeeb90485bafc1d004099de07ddfec978388731f455d38fc

maudetes: I mean to update the `last_modified` value on a PUT

abulte: Oh, sorry. Isn't it handled by this signal? https://github.com/opendatateam/udata/blob/9142ed5ae2f4fa8258892da8d0b9ed70dba1997c/udata/models/datetime_fields.py#L66

maudetes: Indeed my bad! :ok_hand:
On topics that haven't been saved since this PR, last_modified will show as utcnow, but that's life!

PR #2988, opened 2024-03-11
Topic: add spatial field

abulte:
Add `spatial` field to `Topic` model and APIs, using datasets' data structures and form.


maudetes (approved):
Thank you for this PR :clap:

PR #2989, opened 2024-03-14
Migrate schema endpoint to the full version

ThibaudDauce:
Fix https://github.com/etalab/data.gouv.fr/issues/1148

Replace https://github.com/opendatateam/udata/pull/2882


udata/core/dataset/models.py
```diff
@@ -205,12 +207,17 @@ def clean(self, **kwargs):
...
         # or no schema in the catalog we do not check the validity
         # of the name and version
-        catalog_schemas = ResourceSchema.all()
+        catalog_schemas = ResourceSchema.assignable_schemas()
         if not catalog_schemas:
             return
 
         # We know this schema so we can do some checks
-        existing_schema = ResourceSchema.get_schema_by_name(self.name)
+        existing_schema = None
```
maudetes, line 215: You could go for something like:
```
existing_schema = next((schema for schema in catalog_schemas if schema['name'] == self.name), None)
```

udata/core/dataset/api_fields.py
```diff
@@ -282,9 +282,28 @@
...
-    'id': fields.String(description='The schema identifier'),
-    'label': fields.String(description='The schema display name'),
-    'versions': fields.List(fields.String, description='The available versions of the schema'),
+    'name': fields.String(),
+    'title': fields.String(),
+    'description': fields.String(),
+    'schema_url': fields.String(description="Often the link to the latest version"),
+    'schema_type': fields.String(enum=['tableschema', 'datapackage', 'jsonschema', 'other']),
+    'contact': fields.String(),
+    'examples': fields.List(fields.Nested(api.model('CatalogSchema.example', {
```
maudetes, line 293: I think we usually use PascalCase in models name, right? 
```suggestion
    'examples': fields.List(fields.Nested(api.model('CatalogSchemaExample', {
```

udata/tests/api/test_datasets_api.py
```diff
@@ -1820,7 +1826,7 @@ def test_dataset_schemas_api_list_error_w_cache(self, api, rmock, mocker):
         # Long term cache is used
         response = api.get(url_for('api.schemas'))
         assert200(response)
-        assert response.json == ResourceSchemaMockData.get_expected_v1_result_from_mock_data()
+        assert response.json == ResourceSchemaMockData.get_expected_assignable_schemas_from_mock_data()
```
maudetes, line 1829: I think we could add a test on filtering out `NON_ASSIGNABLE_SCHEMA_TYPES` in schemas list api?

maudetes (commented):
I think you'll need to update admin vue files to follow the new schema API :)

(dont forget the changelog ;))

udata/core/dataset/api_fields.py
ThibaudDauce: I wanted to be more explicit that it's inner models (and not used outside the parent). But I can change it if you want

udata/tests/api/test_datasets_api.py
ThibaudDauce: Hum, not sure about what you want. This test already check that the non assignable are not present in the response? And we have a more gobal test here https://github.com/opendatateam/udata/pull/2989/files#diff-36da3294e88ee041b97b82285b97d6d4b365e1835f506171ddab893ecd06b049R774-R778

maudetes (approved):
Looks good, thank you! :pray:

udata/core/dataset/api_fields.py
ThibaudDauce: Changed in ce72578f0 :-)

PR #2991, opened 2024-03-18
Parse all DCT spatial instead of the first one

ThibaudDauce:


maudetes (approved):
:clap:

*When* implementing named entities, we may raise the question of returning early or not. But let's wait for that iteration then.

ThibaudDauce:
> 👏
> 
> _When_ implementing named entities, we may raise the question of returning early or not. But let's wait for that iteration then.

Yes, we'll save all the spatial objects in an array and at the end do a priority list to pick the best one I think!

PR #2994, opened 2024-03-19
Fix complex DCT.conformsTo

ThibaudDauce:
Fix https://github.com/etalab/data.gouv.fr/issues/1322


maudetes (commented):
Thank you! Could you please add a test? :pray:

It might interest @streino

PR #2996, opened 2024-03-20
Fix duplicate log in console

ThibaudDauce:
Trying to remove the duplicate log in console. It comes from adding the `handler` to the default logger (that doesn't seems to be used) and to the `app.logger`. I don't know which one is best to remove…


maudetes (approved):
I am not sure either on which logger we should add the handler but since it seems to work correctly, let's go for this!
Don't forget the changelog ;)

PR #2998, opened 2024-03-21
Do not fail on spatial harvest exception

ThibaudDauce:
Fix https://github.com/etalab/data.gouv.fr/issues/1325

### DIFF
--- a/udata/core/dataset/rdf.py
+++ b/udata/core/dataset/rdf.py
@@ -370,7 +370,7 @@ def spatial_from_rdf(graph):
                         return spatial_coverage
                     except ValidationError:
                         continue
-        except ValidationError:
+        except:
             log.exception(f"Exception during `spatial_from_rdf` for term {term}", stack_info=True)
 
     return None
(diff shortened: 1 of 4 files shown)


udata/core/dataset/rdf.py
```diff
@@ -339,36 +339,68 @@ def contact_point_from_rdf(rdf, dataset):
...
-                    try:
-                        # .upper() si here because geomet doesn't support Polygon but only POLYGON
-                        geojson = wkt.loads(object.toPython().strip().upper())
-                    except ValueError as e:
-                        log.warning(f"Invalid JSON in spatial WKT {object.toPython()} {e}")
+        try:
+            # This may not be official in the norm but some ArcGis return 
+            # bbox as literal directly in DCT.spatial.
+            if isinstance(term, Literal):
+                bbox = term.toPython().strip().split(',')
```
maudetes, line 348: Maybe you could move this is in a dedicated function? :pray:

ThibaudDauce: Done :-)

PR #2999, opened 2024-03-21
Refactor `Activity.kwargs` into `Activity.extras` to facilitate its usage

tboye:
`Activity` model had a field called `kwargs` that was intended to hold additional info about events. By the looks of the code and [data.gouv.fr](https://www.data.gouv.fr/api/1/activity?page=1&page_size=20), it hasn't been really used while it could be very helpful.

When trying to incorporate this field, it appeared that its name clashes with `**kwargs` argument used in dependent libraries. To avoid issues, this field was renamed into `extras` along with changing its type to `ExtrasField` to make it aligned with other models.

Migration shouldn't be a problem due to the fact this field is not in use.

@maudetes Please let me know if you agree with this change and I can add a migration script and adjust tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current (in progress)
 
 - Fix, do not fail on spatial coverage harvesting exception and allow literal spatial BBOX from Arcgis [2998](https://github.com/opendatateam/udata/pull/2998)
+- `Activity`: rename `kwargs` into `extras` and make use of it [2999](https://github.com/opendatateam/udata/pull/2999)
 
 ## 7.0.5 (2024-03-20)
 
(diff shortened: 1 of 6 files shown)


maudetes (commented):
It sounds like a good idea! Thank you for this PR :pray:
Do you have an idea of the kind of extras you would add?

tboye:
@maudetes A few things I currently have in mind:
- attaching a diff to dataset updated activities
- attaching a reason to dataset deletion activities

maudetes:
Nice! I'll let you add the migration script and some tests and you can ask me again for review then :)

tboye:
Ready for review @maudetes

maudetes (approved):
Thank you again for this PR! Looks good to me

PR #3001, opened 2024-03-28
Move constants to files

ThibaudDauce:
Reduce the number of modules depending on `models.py`


udata/core/reuse/forms.py
```diff
@@ -1,8 +1,9 @@
 from udata.forms import ModelForm, fields, validators
 from udata.i18n import lazy_gettext as _
-from udata.models import Reuse, REUSE_TYPES, REUSE_TOPICS
+from udata.models import Reuse
+from udata.core.reuse.constants import REUSE_TYPES, REUSE_TOPICS
 
-from .models import IMAGE_SIZES, TITLE_SIZE_LIMIT, DESCRIPTION_SIZE_LIMIT
+from .constants import IMAGE_SIZES, TITLE_SIZE_LIMIT, DESCRIPTION_SIZE_LIMIT
```
maudetes, line 6: Shouldn't we import all from `.constants`?

maudetes (approved):
:ok_hand:

Do we import all constants in models files to prevent breaking imports in other existing plugins?

udata/core/reuse/forms.py
ThibaudDauce: We could, I'm not sure what's best here…

ThibaudDauce:
> 👌
> 
> Do we import all constants in models files to prevent breaking imports in other existing plugins?

I think I prefer to fix the plugins (it's a quick fix) because leaving constants in models will encourage to use them from the models and leave the problematic dependencies…

PR #3002, opened 2024-03-28
Move DB fields to separate module than models

ThibaudDauce:
Cyclic score (mean of number of imports times the number of imported): 42.57 -> 35.53

All the Mongo fields and `db.*` elements doesn't depend on anything from `udata` (except translations, utils, URIs…). On `master`, when trying to get a field type (frequent in the codebase) it's bad to depend on all models. So split models from mongo DB / fields to avoid some unused deep dependencies.


maudetes:
Replaced by https://github.com/opendatateam/udata/pull/3005

PR #3005, opened 2024-04-02
Move Mongo Engine outside models

ThibaudDauce:
Cyclic score (mean of number of imports times the number of imported): 42.57 -> 35.31

All the Mongo fields and `db.*` elements doesn't depend on anything from `udata` (except translations, utils, URIs…). On `master`, when trying to get a field type (frequent in the codebase) it's bad to depend on all models. So split models from mongo DB / fields to avoid some unused deep dependencies.


maudetes (approved):
Thank you! Quite efficient on reducing cyclic score! :clap:
It doesn't seem to break many things since importing db from `udata.models` still work.

maudetes:
Don't forget to add a changelog entry for the different cyclic reduction refactoring

quaxsze (approved):

PR #3006, opened 2024-04-02
Remove unused dependencies

ThibaudDauce:


maudetes (approved):
:pray: Thank you!

Could you please check that not any of these deps were used in udata-front but pinned in udata instead?
And don't forget the changelog entry ;)

PR #3007, opened 2024-04-03
Topic: add filters in API

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/164

Add the following filter to Topic list APIs (v1 and v2):

- `include_private`
- `geozone`
- `granularity`
- `organization`
- `owner`

About `include_private`:

- called `include_private` instead of `private` because it either returns all topics, including privates, or only non-private ones
- I did not find any `private` or similar existing filter in other API, so I guess coherence is not a problem
- ⚠️ by default, the list will now only contains non-private topics: this is a change from previous behavior (but probably more coherent w/ the rest of data.gouv.fr)

This also adds a `last_modifed` sort.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Topic: add filters in API [#3007](https://github.com/opendatateam/udata/pull/3007)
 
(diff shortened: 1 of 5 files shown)


ThibaudDauce (approved):
Thanks!

PR #3008, opened 2024-04-03
Move `Owned` to its own module

ThibaudDauce:
`Owned` is more a model than a field (it has references to `User` and `Organization` model). It shouldn't be inside DB to not create a dependency between modules requiring DB fields depending on some models (especially a big one like `User`)

It's the same for `Datetimed` (not fix in this PR) but less important because `Datetimed` doesn't have any dependencies on our application (via some models like `Owned`)


udata/tests/test_owned.py
```diff
@@ -6,15 +6,15 @@
 from udata.core.user.models import User
 
 from udata.mongo import db
-
+import udata.core.owned as owned
```
maudetes, line 9: Why not import `from udata.core.owned import Owned`?

maudetes (approved):
Apart from my latest question, I approve this PR :)

udata/tests/test_owned.py
ThibaudDauce: Because we have a class named `Owned` in this file :-)

PR #3012, opened 2024-04-08
Better `FieldValidationError`

ThibaudDauce:


maudetes (commented):
It seems clean! Do you have an example of before/after for me to test?

ThibaudDauce:
> It seems clean! Do you have an example of before/after for me to test?

I think this error is never returned from the API so I don't have a before/after (but I try to extract everything I can from the dataservices PR).

maudetes (approved):

PR #3013, opened 2024-04-08
Move `import User` to runtime

ThibaudDauce:


maudetes (commented):
Do you have some context on this change for future reference? :)

I wonder if it would lead to some performance downgrades for importing the User models at runtime for each request?

ThibaudDauce:
> Do you have some context on this change for future reference? :)
> 
> I wonder if it would lead to some performance downgrades for importing the User models at runtime for each request?

In the dataservice PR I have some problems while importing the API force the import of `User` (you need the `api` module to create `fields` with `api.model()` and having `User` imported put a unwanted constraint. `import` are cache by Python, the performance downgrades should be negligeable I think.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Move constants outside `models.py` files to `constants.py` [#3001](https://github.com/opendatateam/udata/pull/3001)
 - Move `db` and Mongo fields classes outside `udata.models` [#3005](https://github.com/opendatateam/udata/pull/3005)
 - Remove unused dependencies [#3006](https://github.com/opendatateam/udata/pull/3006)
+- Move some `import User` to runtime imports [#3013](https://github.com/opendatateam/udata/pull/3013)
```
MurdoqZodoq, line 11: Kuluwm/ All

MurdoqZodoq (approved):
(ANU)

maudetes (approved):
:+1:

PR #3018, opened 2024-04-16
Fix GeoLevel fetching perfs

ThibaudDauce:
Fix https://github.com/etalab/data.gouv.fr/issues/1315


udata/core/metrics/commands.py
```diff
@@ -114,4 +115,18 @@ def update(site=False, organizations=False, users=False, datasets=False,
                 except Exception as e:
                     log.info(f'Error during update: {e}')
                     continue
+
+    if do_all or geozones:
```
maudetes, line 119: I think this is a punctual command to compute all explicitly. However, it isn't scheduled on contrary to `compute_site_metrics` task.

All `count_objects` methods here are instead triggered at runtime on object creation/update/deletion with signals.
Maybe we could apply the same logic as `update_datasets_metrics` for example?

maudetes: My point was that it won't get udpated at all if we deploy it as is (or am I missing something?). `update` command is not a scheduled job.
I agree about the questionable need about having real time metrics for this, it was just a suggestion to copy other metrics behavior for coherence.

ThibaudDauce: Yes I understood :-) I was wondering if I needed to implement the update of the metrics with the signals or with a job updating them every night :-)

ThibaudDauce: Done in [6556e07](https://github.com/opendatateam/udata/pull/3018/commits/6556e07bdaa2566b6e50ce316b46a2bae0e10f99)

udata/core/spatial/models.py
```diff
@@ -101,6 +101,11 @@ def url(self):
     def external_url(self):
         return endpoint_for('territories.territory', territory=self, _external=True)
 
+    def count_datasets(self):
+        from udata.models import Dataset
+        self.metrics['datasets'] = Dataset.objects(spatial__zones__contains=self.id).visible().count()
```
maudetes, line 106: Why the contains? This job seems to take a lot of time (~1h).

udata/core/spatial/tasks.py
```diff
@@ -0,0 +1,7 @@
+from udata.core.spatial.models import GeoZone
+from udata.tasks import job
+
+@job('compute-geozones-metrics')
```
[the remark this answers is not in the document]
ThibaudDauce: Done in faadb595d

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Fix spatial coverage fetching perfs [#3018](https://github.com/opendatateam/udata/pull/3018)
```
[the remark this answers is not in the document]
ThibaudDauce: Done in 253fadf7b

udata/core/spatial/models.py
maudetes: `__contains` is a string operator I think. You can just do `spatial_zones=` to [query a list element](https://docs.mongoengine.org/guide/querying.html#querying-lists).

ThibaudDauce: Hum… Maybe it'll do a `contains` on all element of the list to check… It could explain the job long duration time…

maudetes (approved):
Thank you :pray:

PR #3019, opened 2024-04-17
Fix `datetime` serialization in extras

ThibaudDauce:
Attemps to fix https://errors.data.gouv.fr/organizations/sentry/issues/138221 but cannot reproduce…


maudetes (approved):
Let's test this in our demo env to make sure it doesn't have side effects

PR #3027, opened 2024-04-25
Add geoDCAT-AP schema option to CswDcatBackend

maudetes:
Allow to test geoDCAT-AP (http://data.europa.eu/930/) schema for the CSW DCAT backend.

This schema is currently being worked on by GeoNetwork in https://github.com/geonetwork/core-geonetwork/pull/7600.

It is not available yet in GeoNetwork

### DIFF
--- a/js/components/harvest/feature-field.vue
+++ b/js/components/harvest/feature-field.vue
@@ -21,7 +21,11 @@ export default {
     },
     computed: {
         checked() {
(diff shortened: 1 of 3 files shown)


maudetes:
Not relevant anymore since we're investing primalarly in CSW-ISO-19139 using the SEMIC EU XLST.

PR #3028, opened 2024-04-26
Allow for series in CSW ISO 19139 DCAT backend

maudetes:
Following https://github.com/opendatateam/udata/pull/2982, we've added some filters on records to fetch dataset only.

It was added due to [this discussion](https://github.com/opendatateam/udata/pull/2982#discussion_r1513353040), where we decided to filter out records that could cause issues but that wouldn't get used.

In this PR, we add `series` records that get converted to dcat:Dataset in the current SEMIeu XSLT. See the logic [here](https://github.com/SEMICeu/iso-19139-to-dcat-ap/blob/34b64006053087e7b2674604312419d09a7af826/iso-19139-to-dcat-ap.xsl#L1100). They are currently harvested as a standard dataset.

Since [dcat:DatasetSeries](https://w3c.github.io/dxwg/dcat/#Class:Dataset_Series) have been added in DCAT 3, this mapping will probably be updated in the upcoming revision.

### DIFF
--- a/udata/harvest/backends/dcat.py
+++ b/udata/harvest/backends/dcat.py
@@ -268,7 +268,6 @@ class CswDcatBackend(DcatBackend):
                           headers=headers).content)
 
         return graphs
-    
 
 
 class CswIso19139DcatBackend(DcatBackend):
@@ -295,6 +294,7 @@ class CswIso19139DcatBackend(DcatBackend):
         transform = ET.XSLT(xsl)
 
         # Start querying and parsing graph
+        # Filter on dataset or serie records
         body = '''<csw:GetRecords xmlns:csw="http://www.opengis.net/cat/csw/2.0.2"
                                   xmlns:gmd="http://www.isotc211.org/2005/gmd"
                                   service="CSW" version="2.0.2" resultType="results"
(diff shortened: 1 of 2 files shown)


udata/harvest/backends/dcat.py
```diff
@@ -304,10 +304,16 @@ def parse_graph(self, url: str, fmt: str) -> List[Graph]:
...
+                                <ogc:Or xmlns:ogc="http://www.opengis.net/ogc">
                                 <ogc:PropertyIsEqualTo>
                                     <ogc:PropertyName>dc:type</ogc:PropertyName>
                                     <ogc:Literal>dataset</ogc:Literal>
                                 </ogc:PropertyIsEqualTo>
+                                <ogc:PropertyIsEqualTo>
+                                    <ogc:PropertyName>dc:type</ogc:PropertyName>
+                                    <ogc:Literal>series</ogc:Literal>
+                                </ogc:PropertyIsEqualTo>
+                                </ogc:Or>
```
ThibaudDauce, line 316: Could you indent inside the `Or`?

ThibaudDauce (approved):
Looks good, do you have an URL to test?

maudetes:
> Looks good, do you have an URL to test?

You could test on https://geo.compiegnois.fr/geonetwork/srv/eng/csw with 507 datasets and 7 series.

streino (approved):
> In this PR, we add series records that get converted to dcat:Dataset in the current SEMIeu XSLT. See the logic [here](https://github.com/SEMICeu/iso-19139-to-dcat-ap/blob/34b64006053087e7b2674604312419d09a7af826/iso-19139-to-dcat-ap.xsl#L1100). They are currently harvested as a standard dataset.

Nice catch!

PR #3038, opened 2024-05-13
Add email and since to org members/requests

ThibaudDauce:
Fix https://github.com/datagouv/data.gouv.fr/issues/1356

### DIFF
--- a/udata/core/organization/api_fields.py
+++ b/udata/core/organization/api_fields.py
@@ -44,7 +44,7 @@ def check_can_access_email():
     
     return OrganizationPrivatePermission(org).can()
 
-user_in_org_with_email_fields = api.inherit('UserReference', user_ref_fields, {
+member_user_with_email_fields = api.inherit('MemberUserWithEmail', user_ref_fields, {
     'email': fields.Raw(
         attribute=lambda o: o.email if check_can_access_email() else None,
         description='The user email (only present on show organization endpoint if the current user has edit permission on the org)', readonly=True),
@@ -52,7 +52,7 @@ user_in_org_with_email_fields = api.inherit('UserReference', user_ref_fields, {
 
 request_fields = api.model('MembershipRequest', {
     'id': fields.String(readonly=True),
-    'user': fields.Nested(user_in_org_with_email_fields),
+    'user': fields.Nested(member_user_with_email_fields),
     'created': fields.ISODateTime(
         description='The request creation date', readonly=True),
     'status': fields.String(
@@ -63,7 +63,7 @@ request_fields = api.model('MembershipRequest', {
 })
 
 member_fields = api.model('Member', {
-    'user': fields.Nested(user_in_org_with_email_fields),
+    'user': fields.Nested(member_user_with_email_fields),
     'role': fields.String(
         description='The member role in the organization', required=True,
         enum=list(ORG_ROLES), default=DEFAULT_ROLE),
--- a/udata/core/organization/api_fields.py
+++ b/udata/core/organization/api_fields.py
@@ -27,9 +27,14 @@ org_ref_fields = api.inherit('OrganizationReference', base_reference, {
 
 from udata.core.user.api_fields import user_ref_fields  # noqa: required
 
+user_in_org_with_email_fields = api.inherit('UserWithEmail', user_ref_fields, {
+    'email': fields.String(
+        description='The user email', readonly=True),
+})
+
 request_fields = api.model('MembershipRequest', {
     'id': fields.String(readonly=True),
-    'user': fields.Nested(user_ref_fields),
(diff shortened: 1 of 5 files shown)


udata/core/organization/api_fields.py
```diff
@@ -43,7 +48,18 @@
...
+    'since': fields.ISODateTime(
+        description='The date the user joined the organization', readonly=True),
+})
+
+private_member_fields = api.model('PrivateMember', {
+    'user': fields.Nested(user_in_org_with_email_fields),
+    'role': fields.String(
+        description='The member role in the organization', required=True,
+        enum=list(ORG_ROLES), default=DEFAULT_ROLE),
+    'since': fields.ISODateTime(
```
maudetes, line 61: I'm thinking `since` could be a public info available to all.

maudetes (commented):
I'm not sure about the approach of having a dedicated endpoint. I think it adds some
confusion with members being returned as part of the organization.
Could we copy the `user_fields` approach on email, having it set to None if permissions are missing?

udata/core/organization/api_fields.py
ThibaudDauce: It is :-) It's present in the public field above

ThibaudDauce:
> I'm not sure about the approach of having a dedicated endpoint. I think it adds some confusion with members being returned as part of the organization. Could we copy the `user_fields` approach on email, having it set to None if permissions are missing?

Yes me too, I first try this approach but didn't find a way to get the correct permissions… Because the field lambda is present in the `User`, but the permissions are linked to the `Organization` two levels up (`Organization` has `Member`, and `Member` has `User`). I didn't find a way to get the parent `Organization` from the `email` lambda inside `User` fields…

ThibaudDauce:
> > the field lambda is present in the `User`, but the permissions are linked to the `Organization` two levels up (`Organization` has `Member`, and `Member` has `User`). I didn't find a way to get the parent `Organization` from the `email` lambda inside `User` fields…
> 
> Indeed! I don't know if we can send additional info to Nested fields using kwargs or something? We could also consider showing the email whenever the current_user is admin (or member?) of an organization with user as member. We would not know if we are in the context of this particular organization, but the user would be allowed to see this email nonetheless. What do you think?

I thought about it, but I think the query to fetch this information could be complex (and run each time we use `user_fields`?). I also thought about trying to get the organization from the Flask request but it seems a little bit too magic and complex to put in `user_fields`. So I decided to create a new endpoint, in my opinion it's simpler and clearer than having too much logic inside fields but If you want I can try to code one of the other solution to see the difference and the complexity?

maudetes:
Thank you for the explanation. I think exploring the two alternatives would be interesting.
 something like `Organization.objects(members__user__all=[current_user, target_user])` would be interesting. I agree though about

ThibaudDauce:
> Thank you for the explanation. I think exploring the two alternatives would be interesting. Maybe something like `Organization.objects(members__user__all=[current_user, target_user])` may work. I agree though about performance issues that may arise -> we may want to create an index on organization > members > users. Fetching the organization may be magic indeed but would be quite efficient if somehow not so complex.

Thanks for persisting, the current solution is much cleaner without the extra endpoint (and it's really easy to check for org permission in field)

udata/core/organization/api_fields.py
```diff
@@ -27,9 +30,31 @@
...
+
+# To use on public endpoint to show the email only to editor of the org (currently working only with `api.organization` endpoint, see above)
+user_in_org_with_email_fields_if_permissions = api.inherit('UserReference', user_ref_fields, {
+    'email': fields.Raw(
+        attribute=lambda o: o.email if check_can_access_email() else None,
+        description='The user email (only present on show organization endpoint if the current user has edit permission on the org)', readonly=True),
+})
+
+# To use on private endpoints where the current user is already checked to be admin/editor of the org
+user_in_org_with_always_email_fields = api.inherit('UserReference', user_ref_fields, {
```
maudetes, line 51: Maybe having only `if_permissions` differing between both is clear enough?
We could consider using only the public one, making an unnecessary check but making code simpler.
```suggestion
user_in_org_with_email_fields = api.inherit('UserReference', user_ref_fields, {
```

maudetes (commented):
Thank you for taking a look in this direction! Indeed, seems way more easier :clap:

I think the additional `inherit` (or something else?) may break the JS swagger client used in admin with a `InternalError: too much recursion`. Do you reproduce this behavior?

udata/core/organization/api_fields.py
ThibaudDauce: Done in 880ed6935. Much simpler!

maudetes (approved):
I think we're clear, thank you :pray:

PR #3039, opened 2024-05-13
Merge multiple polygons during spatial coverage harvesting 

ThibaudDauce:
Fix https://github.com/datagouv/data.gouv.fr/issues/1362

The goal of this PR is to support when there is multiple geometries. Instead of saving the first one, save a `MultiPolygon`  merging all geometries. Deduplicates the geometries because some backends provide the same geometry in multiple formats.


maudetes (approved):
This seems to work as expected, thank you! :pray:

We can iterate on this later if we have more real-world examples

PR #3043, opened 2024-05-21
Fix extras overrides from admin

ThibaudDauce:
Fix https://github.com/datagouv/data.gouv.fr/issues/1364


maudetes (approved):
:ok_hand: Thank you for the quick fix!

PR #3045, opened 2024-05-23
Remove path converter in spatial routes

ThibaudDauce:


udata/core/spatial/api.py
```diff
@@ -85,7 +85,7 @@ def get(self):
     location='args', default=25)
 
 
-@ns.route('/zones/<pathlist:ids>/', endpoint='zones')
```
maudetes, line 88: As discussed, we should make sure zones don't have `/`?

ThibaudDauce: On dev there is none:
```
>>> GeoZone.objects(id__contains='/')
[]
```

PR #3046, opened 2024-05-23
CORS: always returns 204 on OPTIONS request

ThibaudDauce:
Fix https://github.com/datagouv/data.gouv.fr/issues/1359


ThibaudDauce:
Will do it via Nginx instead

PR #3047, opened 2024-05-27
Add admin permissions for /users endpoint

ThibaudDauce:
Fix https://github.com/opendatateam/udata/issues/3044

### DIFF
--- a/udata/tests/api/test_user_api.py
+++ b/udata/tests/api/test_user_api.py
@@ -153,6 +153,8 @@ class UserAPITest(APITestCase):
 
     def test_user_api_full_text_search_first_name(self):
         '''It should find users based on first name'''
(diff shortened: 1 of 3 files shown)


udata/tests/api/test_user_api.py
```diff
@@ -153,6 +153,8 @@ def test_suggest_users_api_no_dedup(self):
 
     def test_user_api_full_text_search_first_name(self):
```
maudetes, line 154: Could you add a test that we cant get user list without the login as admin? :pray:

maudetes (approved):
:pray:

udata/tests/api/test_user_api.py
maudetes: Yes, sorry!

PR #3049, opened 2024-05-31
Add api and methods to allow followers and discussions

nicolaskempf57:
Required for https://github.com/datagouv/data.gouv.fr/issues/1369


ThibaudDauce:
I don't see the controller for `Discussion` (only for `Followers`), am I missing something? (but I can add latter it if you've only added the minimal requirements to not crash on the frontend I understand :-D )

nicolaskempf57:
> I don't see the controller for `Discussion` (only for `Followers`), am I missing something? (but I can add latter it if you've only added the minimal requirements to not crash on the frontend I understand :-D )

You're right, there is no controller for discussions, it's handled in `udata/core/discussions/api.py` only.

maudetes:
As a reference, when creating discussions on topics object, we had to make these two fixes:
* https://github.com/opendatateam/udata/pull/2922
* https://github.com/opendatateam/udata/pull/2944

ThibaudDauce (approved):

maudetes (approved):
Thank you, I think it works nicely!

PR #3050, opened 2024-06-04
Add DCAT-AP HVD properties in RDF output if the dataset is tagged hvd

maudetes:
* DCATAP.applicableLegislation (at the dataset and distribution level). See in particular this [DCAT-AP HVD specification](https://semiceu.github.io/DCAT-AP/releases/2.2.0-hvd/#denoting-a-hvd-dataset).
* DCATAP.hvdCategory (at the dataset level). See in particular this [DCAT-AP HVD specification](https://semiceu.github.io/DCAT-AP/releases/2.2.0-hvd/#c2).

These properties should be added on dataservices when exposed as well.

We've added a support for tag filtering on the catalog, in order to do http://dev.local:7000/api/1/site/catalog.xml?tag=hvd.
You can test using this output locally, and by having datasets tagged `hvd` and `mobilite` for example.

> [!NOTE]  
> We rely heavily on `EU_HVD_CATEGORIES` french vocabulary as pivot for keywods. Should we move it to `settings.py` to be overriden in other udata instances?

Iterate on https://github.com/datagouv/data.gouv.fr/issues/1371.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Add DCAT-AP HVD properties in RDF output if the dataset is tagged hvd [#3050](https://github.com/opendatateam/udata/pull/3050)
 
 ## 8.0.1 (2024-05-28)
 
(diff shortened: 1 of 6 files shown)


ThibaudDauce (approved):
Good! We could extract the `EU_HVD_CATEGORIES` inside settings in another PR.

PR #3051, opened 2024-06-04
Add resources to ISO backend using XSLT fork

ThibaudDauce:


ThibaudDauce:
We're working on fixing the issue at the source

PR #3052, opened 2024-06-06
Catalog refactor

Pierlou:
Close https://github.com/datagouv/data.gouv.fr/issues/979


ThibaudDauce (approved):
:+1:

Pierlou:
Will do 🫡 but are we OK with only these changes? Anything else to add/refactor?

ThibaudDauce:
> Will do 🫡 but are we OK with only these changes? Anything else to add/refactor?

Yes :-) Let's keep it small, if we need something else we could always do another PR.

PR #3053, opened 2024-06-06
Save and show harvest logs

ThibaudDauce:
There is some ANSI colors codes inside the log messages… Not sure about the best way to remove them…

I'm not a big fan of how it works… But I don't have a better idea of how to add this feature…

### DIFF
--- a/js/components/harvest/job.vue
+++ b/js/components/harvest/job.vue
@@ -85,6 +85,9 @@ export default {
                 sort: 'remote_id',
                 ellipsis: true
             }, {
+                label: this._('Logs'),
+                key: (item) => item.logs.length ? item.logs.length + ' ' + this._('logs') : '',
+            }, {
                 label: this._('Status'),
                 key: 'status',
                 sort: 'status',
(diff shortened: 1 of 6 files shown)


js/components/harvest/job.vue
```diff
@@ -84,6 +84,9 @@ export default {
                 key: 'remote_id',
                 sort: 'remote_id',
                 ellipsis: true
+            }, {
```
maudetes, line 87: Maybe we could move `Logs` after Status and add `width` : 100 as is done for status?

maudetes (approved):
Nice! I think this first iteration will do!

Don't forget the changelog!

js/components/harvest/job.vue
ThibaudDauce: I'm not sure about that… It'll move status for every line (even if the page doesn't have any logs), with this order, the page without logs will look exactly the same as now. And only item with logs will display the info. But as you wish :-)

maudetes: It was a suggestion because I found the absence of border between fields made it less readable. Could you try with a small width?

ThibaudDauce: Done in ec1800133

PR #3054, opened 2024-06-07
Fetch dct:license, dct:rights and dct:accessRights from proper locations

streino:
Fix https://github.com/ecolabdata/ecospheres/issues/260
Fix https://github.com/ecolabdata/ecospheres/issues/351

Licenses, rights and access rights can be properties of [`Resource`](https://www.w3.org/TR/vocab-dcat-3/#Property:resource_access_rights) (super-class of `Dataset`) or [`Distribution`](https://www.w3.org/TR/vocab-dcat-3/#Property:distribution_access_rights).

[Some conditions apply](https://www.w3.org/TR/vocab-dcat-3/#Property:distribution_rights) : 
```
Information about licenses and rights SHOULD be provided on the level of Distribution.
Information about licenses and rights MAY be provided for a Dataset in addition to but
not instead of the information provided for the Distributions of that Dataset. Providing
license or rights information for a Dataset that is different from information provided
for a Distribution of that Dataset SHOULD be avoided as this can create legal conflicts.
```

Also, in the case of DCAT resources coming from the SEMICeu transformation, if I'm not mistaken, the `license`/`accessRights` (no `rights` in XSLT 2.0, but will be introduced in 3.0) are taken at the dataset level from the ISO input and copied at the distribution level in the DCAT output. So we should be able to assume that what we find at the distribution level is in fact applying to the entire record.

_(Original PR proposal below, revisited in comments)_

Since udata doesn't display distribution-level licenses/rights/accessRights, we'll have to approximate at the dataset-level. It's probably better to be overly careful than miss some restrictions. So we could go with : 
- Collect all access rights found at `Resource` and `Distribution` levels.
- Uniqify the collection.
- Assign/display the set of rights at the dataset level.

The other approach would be to store all information in the extras as-is and retrieve what we're interested in displaying when needed.

Finally, the same sort of logic should apply to `license`, but that part needs to be articulated with existing license logic in udata. ~~So probably another PR.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Current (in progress)
+## 9.2.1 (2024-09-23)
 
 - Enable basic search on dataservices [#3148](https://github.com/opendatateam/udata/pull/3148)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Add resources formats and harvest remote_url on dataset catalog [#3159](https://github.com/opendatateam/udata/pull/3159)
 - Add contact form in contact point model [#3164](https://github.com/opendatateam/udata/pull/3164)
 - Make base_api_url optional in dataservice [https://github.com/opendatateam/udata/pull/3163](#3163)
+- DCAT harvest : fetch licenses and access rights from proper locations [#3054](https://github.com/opendatateam/udata/pull/3054)
 
 ## 9.2.1 (2024-09-23)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - Add migration to delete duplicate resources due to ODS harvesting [#3158](https://github.com/opendatateam/udata/pull/3158)
 - Add discussion.posted_on in discussion sort choices [3168](https://github.com/opendatateam/udata/pull/3168)
+- DCAT harvest : Refactor DCAT extras handling in rdf parsing [#3054](https://github.com/opendatateam/udata/pull/3054). `accesRights`, `rights` and `license` are now in `extras["dcat"]` for both datasets and resources. `accessRights` can be infered for a dataset from its resources.
 
 ## 9.2.2 (2024-10-08)
 
@@ -16,7 +17,6 @@
 - Add resources formats and harvest remote_url on dataset catalog [#3159](https://github.com/opendatateam/udata/pull/3159)
 - Add contact form in contact point model [#3164](https://github.com/opendatateam/udata/pull/3164)
 - Make base_api_url optional in dataservice [https://github.com/opendatateam/udata/pull/3163](#3163)
-- DCAT harvest : Refactor DCAT extras handling in rdf parsing [#3054](https://github.com/opendatateam/udata/pull/3054)
 
 ## 9.2.1 (2024-09-23)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Fix undelete reuse and dataservices [#3141](https://github.com/opendatateam/udata/pull/3141)
 
 ## 9.1.4 (2024-08-26)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Current (in progress)
+## 9.2.0 (2024-09-13)
 
 - Allow OAuth clients without secrets [#3138](https://github.com/opendatateam/udata/pull/3138)
 - Add a `archived` button for datasets and reuses on frontend admin [#3104](https://github.com/opendatateam/udata/pull/3104)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Add a minimal publiccode.yml [#3144](https://github.com/opendatateam/udata/pull/3144)
 - Fix the boolean filters in the API for the "new system" endpoints [#3139](https://github.com/opendatateam/udata/pull/3139)
 - Update authlib dependency from 0.14.3 to 1.3.1 [#3135](https://github.com/opendatateam/udata/pull/3135)
+- Use hydra's RESTful endpoint URLs [#3142](https://github.com/opendatateam/udata/pull/3142)
 
 ## 9.1.4 (2024-08-26)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Current (in progress)
 
+- Add a `archived` button for datasets and reuses on frontend admin [#3104](https://github.com/opendatateam/udata/pull/3104)
 - **breaking change** Return all the reuses available to a user on the /reuses endpoint, including the private and deleted ones they own [#3140](https://github.com/opendatateam/udata/pull/3140).
 - Fix undelete reuse and dataservices [#3141](https://github.com/opendatateam/udata/pull/3141)
 - Add a minimal publiccode.yml [#3144](https://github.com/opendatateam/udata/pull/3144)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Current (in progress)
+
+- Nothing yet
+
 ## 9.2.0 (2024-09-13)
 
 - Allow OAuth clients without secrets [#3138](https://github.com/opendatateam/udata/pull/3138)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Enable basic search on dataservices [#3148](https://github.com/opendatateam/udata/pull/3148)
 
 ## 9.2.0 (2024-09-13)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Add a minimal publiccode.yml [#3144](https://github.com/opendatateam/udata/pull/3144)
 - Fix the boolean filters in the API for the "new system" endpoints [#3139](https://github.com/opendatateam/udata/pull/3139)
 - Update authlib dependency from 0.14.3 to 1.3.1 [#3135](https://github.com/opendatateam/udata/pull/3135)
+- Add CORS on resource redirect [#3145](https://github.com/opendatateam/udata/pull/3145)
 - Use hydra's RESTful endpoint URLs [#3142](https://github.com/opendatateam/udata/pull/3142)
 
 ## 9.1.4 (2024-08-26)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Current (in progress)
 
+- Allow OAuth clients without secrets [#3138](https://github.com/opendatateam/udata/pull/3138)
 - Add a `archived` button for datasets and reuses on frontend admin [#3104](https://github.com/opendatateam/udata/pull/3104)
 - **breaking change** Return all the reuses available to a user on the /reuses endpoint, including the private and deleted ones they own [#3140](https://github.com/opendatateam/udata/pull/3140).
 - Fix undelete reuse and dataservices [#3141](https://github.com/opendatateam/udata/pull/3141)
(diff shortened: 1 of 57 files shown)


ThibaudDauce:
Why not put accesRight information inside `Resource` extras directly?

streino:
> Why not put accesRight information inside `Resource` extras directly?

Right. I used what was already there, but it makes sense to move it outside of the `harvest` section into `resource` and/or `dataset` extras directly (depending on the implementation we go with).

maudetes:
> The other approach would be to store all information in the extras as-is and retrieve what we're interested in displaying when needed.

I would rather store the information as-is for now instead of consolidating it at the dataset level.

streino:
> I would rather store the information as-is for now instead of consolidating it at the dataset level.

We (ping @alhyss) were leaning the other way for the following reasons:
- Udata doesn't have licence/access-rights at the distribution level, so having it in extra is inconsistent with the udata model and won't be usable in the front.
- That's possibly quite a bit of duplicate information stored in extras, with no clear use.
- Spec is complex so I'm worried the cases of specific distribution-level conditions we might encouter would be caused by poor implementation of the spec rather than actual distribution-level conditions.
- Avoid putting consolidation logic in the front-end to keep it as dumb as possible.

That said, if there are plans to use distribution-level licence/access-rights in udata : 
- I can work directly on that instead.
- We can go for dataset-level now and switch to distribution-level later by simply removing the consolidation logic.

alhyss:
> I would rather store the information as-is for now instead of consolidating it at the dataset level.

I agree that distinct distribution-level rights and licenses should be stored as they are. If these properties are not the same for all distributions, this is the only way to prevent losing an important information. Legal conditions are meaningless (and confusing) if users can't tell which distribution they're applicable to.

That being said: 
- udata already has a standardized dataset-level license property. Unless you decide not to use it for DCAT metadata records, some kind of consolidation work will be necessary there.
- As @streino explained, rights and licenses will always be the same for all distributions if the metadata record was transformed from ISO.
- Rights and licenses are generally the same for all distributions. Forcing udata users to enter the same information multiple times will bother them and (probably) affect metadata quality.
- Allowing dataset-level and distribution-level properties to coexist in a metadata record would be hazardous without consistency checks.

Using only distribution-level properties is arguably simpler, but wouldn't be very efficient.

Ideally, you could allow both dataset and distribution level properties, though not together:
- Distribution-level properties would be used in the (rare) case of harvested metadata with distribution-level rights or licenses that are not exactly the same for all distributions. If there are more than one license, I wouldn't use the standardized dataset-level license property either. 
   Example of metadata record that would require distribution-level access rights and license properties:
   ```turtle
   [] a dcat:Dataset ;
      dcat:distribution [ a dcat:Distribution ;
            dct:accessRights [ a dct:RightsStatement ;
                  rdfs:label "Accès restreint"@fr ] ] ,
         [ a dcat:Distribution ;
            dct:accessRights [ a dct:RightsStatement ;
                  rdfs:label "Pas de limitation d'accès"@fr ] ;
            dct:license <https://www.etalab.gouv.fr/licence-ouverte-open-licence> ] .
   ```
- Dataset-level properties would be used in all other cases.
   Example:
   ```turtle
   [] a dcat:Dataset ;
      dcat:distribution [ a dcat:Distribution ;
            dct:accessRights [ a dct:RightsStatement ;
                  rdfs:label "Pas de limitation d'accès"@fr ] ;
            dct:license <https://www.etalab.gouv.fr/licence-ouverte-open-licence> ] ,
         [ a dcat:Distribution ;
            dct:accessRights [ a dct:RightsStatement ;
                  rdfs:label "Pas de limitation d'accès"@fr ] ;
            dct:license <https://www.etalab.gouv.fr/licence-ouverte-open-licence> ] .
   ```

If all distributions have the same access rights property but not the same license, a dataset-level property would be used for access rights and distribution-level properties for the licenses.

Finding a way to display the distribution-level properties doesn't seem urgent at all, considering this case might never happen. But when it does, we'll be able to detect it and decide how to show it.

streino:
@maudetes, we rediscussed the topic with @alhyss so I'm go with her suggestion. Ok on your side?

maudetes:
Thank you for the detailed explanation of the different considerations. This seems like the best approach to cover the main usages as well as possible tricky ones. :+1:
I think we can also iterate quite easily on this if needed since these are extras properties.

abulte:
I gave a go at `accessRights` handling in https://github.com/opendatateam/udata/pull/3054/commits/0cf8fdddab07e89148cec2236e8be8f0cf124259.

It's not exactly what has been discussed above. This solution is easier to implement given the current code:
- if `dct:accessRights` is present on a `Distribution`, store it in `resource.extras["harvest"]["accessRights"]`
- if `dct:accessRights` is present on a `Dataset`, store it in `dataset.extras["harvest"]["accessRights"]`
- if the `Dataset` does not have a `dct:accessRights` directly set on it _and_ the resources have one and only one unique `dct:accessRights` associated, copy it to the dataset extras

I think it does the main job: get the access rights from the `Distribution` to the `Dataset`. It's also generic enough that it can be applied to the shared RDF parsing of a dataset. And it's easy enough to reason about and debug: no information is lost and the logic can be followed by looking at the various extras.

The logic is tested on the geo2france `CswIso19139DcatBackendTest`, with access rights on a `Distribution` that find their way into the `Dataset`.

The considerations above related to `Licence` do not apply here, we're only talking about access rights that are stored in `extras`.

Re `Licence`, I think we need a separate discussion. There's already a mechanism in place that looks a lot like what is described above: licenses infos are fetched from the `Distributions` and fed to the "guessing" algorithm as alternative propositions here https://github.com/opendatateam/udata/blob/0cf8fdddab07e89148cec2236e8be8f0cf124259/udata/core/dataset/rdf.py#L626 (`*licenses` are from the resources). If the dataset has a matchable license, it will take precedence.

streino:
Thanks for picking this up!

IIUC the differences with the previous proposition are:
1. We drop the requirement that ALL distributions have the same `accessRights` (only "those that have an `accessRights`").
2. We restrict the notion of `accessRights` to a single occurrence per distribution (not sure if this was a previous requirement or not).
3. We lose the link between a given `accessRights` and the Distribution(s) it came from.

Point 1 seems fine but may have consequences for non-ISO DCAT harvests, so leaving it up to @maudetes and @ThibaudDauce.

Point 2 seems to be in line with the [0..1 constraint for `dct:accessRights` for Distribution](https://semiceu.github.io/GeoDCAT-AP/releases/2.0.0/#optional-properties-for-distribution), but the [SEMIC xslt doesn't look like it guarantees it, with that `<xsl:for-each select="gmd:otherConstraints[../gmd:accessConstraints]">`](https://github.com/SEMICeu/iso-19139-to-dcat-ap/blob/master/iso-19139-to-dcat-ap.xsl#L2827). In ISO, we can definitely have multiple `gmd:otherConstraints` and even multiple `gmd:accessConstraints` within the same `gmd:resourceConstraints` (not a bug, there are examples in the INSPIRE and CNIG guidelines). So unless I'm mistaken we'll end up with more `accessRights` than we'd like.

Point 3 seems fine since at the moment we don't plan to display distribution-level `accessRights`. We can revist if/when needed.

abulte:
Re licenses, I gave it go here https://github.com/ecolabdata/udata/pull/1 (based on this PR, so can not do a PR on this repo yet)

Pretty much what was done before (there's already a mechanism in place that looks a lot like what is described above: licenses infos are fetched from the Distributions and fed to the "guessing" algorithm as alternative propositions), with modifications:
- handle nested `rdfs:label`, not only literals
- handle a list of values for each predicate, not only a single value
- add a test from a geo-ide metadata record

udata/core/dataset/rdf.py
```diff
@@ -468,6 +468,13 @@ def title_from_rdf(rdf, url):
             return i18n._("Nameless resource")
 
 
+def access_rights_from_rdf(resource: RdfResource):
+    top_el = resource.value(DCT.accessRights)
+    if top_el:
+        # try to find a structured label or fallback on the raw value
+        return rdf_value(top_el, RDFS.label) or rdf_value(resource, DCT.accessRights)
```
maudetes, line 475: I don't think it works on graph such as
```
	"dct:accessRights": {
		"@type": "dct:RightsStatement",
		"@label": {
			"@value": "public"
		}
	},
```
as in https://opendata.normandie.fr/api/feed/dcat-ap/2.0.1.json?

udata/core/dataset/rdf.py
```diff
@@ -468,6 +468,13 @@ def title_from_rdf(rdf, url):
             return i18n._("Nameless resource")
 
 
+def access_rights_from_rdf(resource: RdfResource):
+    top_el = resource.value(DCT.accessRights)
+    if top_el:
```
maudetes, line 473: I think `top_el` could be a `Literal`. Example: https://data.lillemetropole.fr/geonetwork/srv/fre/csw?SERVICE=CSW&VERSION=2.0.2&REQUEST=GetRecords&outputSchema=http://www.w3.org/ns/dcat%23&typenames=gmd:MD_Metadata&elementSetName=full&resultType=results

maudetes (commented):
I think the access rights storing approach works!

I have some comments on being more resilient on accessRights specifications

udata/core/dataset/rdf.py
abulte: Quand je convertis le flux en XML (j'ai plus l'habitude de regarder des XML ces temps-ci), ça me donne ça :-/ J'utilise https://www.easyrdf.org/converter depuis JSON-LD vers XML.

```
    <dc:accessRights>
      <dc:RightsStatement>
      </dc:RightsStatement>
    </dc:accessRights>
```

udata/core/dataset/rdf.py
abulte: I pretty much reused the license detection from the other PR, WDYT? https://github.com/opendatateam/udata/pull/3054/commits/c979244bef0be95d4f5668751d76de7f605f409a

udata/core/dataset/rdf.py
```diff
@@ -594,6 +612,24 @@ def dataset_from_rdf(graph: Graph, dataset=None, node=None):
             elif isinstance(value, RdfResource):
                 licenses.add(value.identifier.toPython())
 
+    # assign the distribution accessRights to the dataset if all distribs have the same accessRights
```
streino, line 615:
```suggestion
    # assign the distribution accessRights to the dataset if all distribs that have an accessRights have the same accessRights
```

streino:
> Licence handled in [ecolabdata#1](https://github.com/ecolabdata/udata/pull/1) and merged here.

Fixed title to reflect that.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - Add resources formats and harvest remote_url on dataset catalog [#3159](https://github.com/opendatateam/udata/pull/3159)
 - Add contact form in contact point model [#3164](https://github.com/opendatateam/udata/pull/3164)
 - Make base_api_url optional in dataservice [https://github.com/opendatateam/udata/pull/3163](#3163)
+- DCAT harvest : Refactor DCAT extras handling in rdf parsing [#3054](https://github.com/opendatateam/udata/pull/3054)
```
streino, line 19: Should the entry mention what extras are available?

udata/core/dataset/rdf.py
```diff
@@ -468,6 +468,62 @@ def title_from_rdf(rdf, url):
...
+
+
+def infer_dataset_access_rights(resources_access_rights: list[set]) -> set | None:
+    """
+    Infer the dataset access rights from a list of resources access rights.
+    If all resources have the same set of access rights return it.
+    """
+    if not resources_access_rights:
+        return
+    if resources_access_rights[0] == set.intersection(*resources_access_rights):
```
streino, line 511:
```suggestion
    if set.union(*resources_access_rights) == set.intersection(*resources_access_rights):
```

My bad. I mistakenly put `resources_access_rights[0] == set.intersection(*resources_access_rights)` in a short-lived comment but seems I was too slow to fix it :/

Counter-example: `[{1,2}, {1,2,3}]`

streino:
@abulte Just like I change the title/description for you, now you can approve the PR for me :p

abulte (approved):

udata/core/dataset/rdf.py
```diff
@@ -570,36 +638,41 @@ def dataset_from_rdf(graph: Graph, dataset=None, node=None):
...
+        # include both dct:license and dct:rights as licenses hints from resources
+        resources_licenses_hints |= licenses_from_rdf(distrib)
+        resources_licenses_hints |= rights_from_rdf(distrib)
 
     for additionnal in d.objects(DCT.hasPart):
         resource_from_rdf(additionnal, dataset, is_additionnal=True)
 
+    dataset_access_rights = access_rights_from_rdf(d)
+    if not dataset_access_rights and resources_access_rights:
+        dataset_access_rights = infer_dataset_access_rights(resources_access_rights)
```
maudetes, line 659: We could have
```
dataset_access_rights = infer_dataset_access_rights(dataset_access_rights, resources_access_rights)
```
with the fallback logic on resources access rights inside the function, as we have for `License.guess`?

maudetes (approved):
Thank you again for this nice PR :star_struck:

udata/core/dataset/rdf.py
abulte: https://github.com/opendatateam/udata/pull/3054/commits/61548df9c9aba23919e14dc87beb229ce8513cf0

PR #3060, opened 2024-06-12
Validate `owner` as an `ObjectId` in /sources endpoint

ThibaudDauce:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/141574


maudetes (approved):
Thank you! Could you maybe add a changelog entry?

PR #3061, opened 2024-06-12
Prevent users with no_org role to create organizations

ThibaudDauce:
Require to add the `no_org` role to the database


ThibaudDauce:
Closing for now

PR #3067, opened 2024-06-20
Update Markdown base settings

nicolaskempf57:
Allows div and class in markdown config to make https://github.com/datagouv/udata-front/pull/432 works.

This PR could be closed if you prefer to have the changes only in `udata.cfg`.


ThibaudDauce (approved):
Seems ok to have `div` in allowed HTML tags :-)

PR #3069, opened 2024-06-24
Add reports backend

ThibaudDauce:


udata/core/reports/api.py
```diff
@@ -0,0 +1,34 @@
...
+
+ns = api.namespace('reports', 'User reported objects related operations (beta)')
+
+@ns.route('/', endpoint='reports')
+class ReportsAPI(API):
+    @api.doc('list_reports')
+    @api.expect(Report.__index_parser__)
+    @api.marshal_with(Report.__page_fields__)
+    def get(self):
+        query = Report.objects.visible()
```
nicolaskempf57, line 17: I'm getting an error `AttributeError: 'UDataQuerySet' object has no attribute 'visible'` on this line.

But with only `.objects` it's working great !

nicolaskempf57 (commented):
As we discuss yesterday, it would be nice for the front-end to have an new endpoint to get the list of report reasons (and later of reportable models). Otherwise it's working great without the `visible()` call.

It seems much easier to use than the old `issue` model from 2021.

udata/core/reports/api.py
ThibaudDauce: Oups… Done in 7c2556828

nicolaskempf57 (approved):

PR #3072, opened 2024-07-01
feat(discussions): allow discussion notification customization

abulte:
Adds customizations for discussion notification emails:
- `external_url`: the URL the notification email points to
- `model_name`: the model name used in the notification email

Also allows notifications for `Topic`.

The expected payload is in `Discussion.extras`. It will be used for subsequent notifications of comments on discussion and discussion closing. A comment made on a discussion pointing to an external site will always point to this external site, even if made on another frontend.

```json
{
  "notification": {
    "model_name": "bouquet",
    "external_url": "https://ecologie.data.gouv.fr/bouquets/xxx"
  }
}
```

Both parameters are protected through an access list, to avoid ill-intentioned people injecting links or other arbitrary text in the notification emails:
- `DISCUSSION_ALLOWED_EXTERNAL_DOMAINS` (accepts wildcards)
- `DISCUSSION_ALTERNATE_MODEL_NAMES`

Both parameters can be omitted, the default behavior will apply. For `Topic`, since there's no detail page on data.gouv.fr (yet), no notification will be sent if `external_url` is not supplied.

NB1: there's a grammatical error in the English notification text `Your %(type)s have a new discussion`, not fixed here because it would involve updating the translations.
NB2: the translation of `model_name` is not handled. This is not a problem for ecologie.data.gouv.fr and might not be easy to do (how to add a translation for something variable? or include translations in payload?).

Cf https://github.com/ecolabdata/ecospheres/issues/263 for requirements and more details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Prevent tagging as spam owners' messages [#3071](https://github.com/opendatateam/udata/pull/3071)
 - Add api endpoint /me/org_topics/ [#3070](https://github.com/opendatateam/udata/pull/3070)
 - Expose dataservices in RDF catalog [#3058](https://github.com/opendatateam/udata/pull/3058)
+- Allow discussion notification customization [#3072](https://github.com/opendatateam/udata/pull/3072)
 
 ## 9.0.0 (2024-06-07)
 
--- a/udata/core/discussions/tasks.py
+++ b/udata/core/discussions/tasks.py
@@ -1,11 +1,18 @@
+import fnmatch
+
+from urllib.parse import urlparse
+
+from flask import current_app
+
 from udata import mail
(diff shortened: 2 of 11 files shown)


ThibaudDauce:
@abulte Is this still something we want to do? Can we close this PR? We may reopen it latter if prioritize? (ping @maudetes / @agarrone )

agarrone:
I think we can deal with this during the notifications sprint

ThibaudDauce:
> I think we can deal with this during the notifications sprint

Can you create an issue and link it here and with https://github.com/ecolabdata/ecospheres/issues/263 so we remember it?

abulte:
@Thesauruv do we still need that? cf https://github.com/ecolabdata/ecospheres/issues/263

PR #3074, opened 2024-07-01
New CORS system

ThibaudDauce:
See https://github.com/datagouv/data.gouv.fr/issues/1359

I copied the implementation of the biggest PHP framework ([the middleware](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Http/Middleware/HandleCors.php) and [the CORS logic](https://github.com/fruitcake/php-cors/blob/master/src/CorsService.php)) while removing all the generics conditions (since we are an API we have a lot of stuff allowed to the maximum)

Exemple de test CORS:
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Document</title>
</head>
<body>
    <div data-udata-dataset="666c037b54009239e0a974ad"></div><script  src="http://dev.local:7000/static/oembed.js" async defer></script>

    <div id="results"></div>

    <script>
        const results = document.getElementById('results')
        const get = async (url) => {
            const response = await fetch(url, {
                headers: {
                    'XXXX': 'xxx',
                },
            })
            const body = await response.text()
            results.innerHTML = `${results.innerHTML}<br><br>${body}`
        }

        (async () => {
            await get('http://dev.local:7000/api/1/datasets')
        })()
    </script>
</body>
</html>
```


udata/api/__init__.py
```diff
@@ -15,11 +15,10 @@
     request,
     url_for,
 )
-from flask_cors import CORS
```
maudetes, line 18: We can remove dependency?

ThibaudDauce: Done in ece3ad4fe

udata/cors.py
```diff
@@ -0,0 +1,79 @@
...
+
+def is_preflight_request() -> bool:
+    return request.method == 'OPTIONS' and request.headers.get('Access-Control-Request-Method', None) is not None
+
+def is_api():
+    return request.path.startswith('/api') or request.path.startswith('/oauth')
+
+def add_preflight_request_headers(headers: Headers) -> Headers:
+    origin = request.headers.get('Origin', None)
+    if origin:
```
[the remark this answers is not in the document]
ThibaudDauce: I look into this and I don't like the idea of having a function to hide the "has origin" part and then here do a "get origin" without any check that it's present, no?

maudetes (approved):
:ok_hand:

PR #3080, opened 2024-07-10
Another docs cleaning PR

bolinocroustibat:
Minor docs fix PR:

- Fix wrong GitHub links in docs
- Fix wrong Python version in docs

Can be left as draft for later cleaning tasks if needed.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -10,7 +10,7 @@ We’ll use the following repositories:
 # Check the system requirements
 
 !!! info
-    Be aware that udata now requires Python **>3.7,<3.10** to work. 
+    Be aware that udata now requires Python **>3.7,<=3.11** to work. 
 
 udata requires several libraries to be installed to work. You can see them on the udata documentation link below.
 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -10,7 +10,7 @@ We’ll use the following repositories:
 # Check the system requirements
 
 !!! info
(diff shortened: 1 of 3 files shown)


docs/getting-started.md
```diff
@@ -10,7 +10,7 @@ We’ll use the following repositories:
 # Check the system requirements
 
 !!! info
-    Be aware that udata now requires Python **>3.7,<3.10** to work. 
+    Be aware that udata now requires Python **>3.7,<=3.11** to work.
```
ThibaudDauce, line 13: I don't think we support 3.7 anymore :-) I think it's at least 3.9

udata/settings.py
```diff
@@ -132,7 +132,7 @@ class Defaults(object):
     SITE_KEYWORDS = ['opendata', 'udata']
     SITE_AUTHOR_URL = None
     SITE_AUTHOR = 'Udata'
-    SITE_GITHUB_URL = 'https://github.com/etalab/udata'
+    SITE_GITHUB_URL = "https://github.com/opendatateam/udata"
```
maudetes, line 135: Single quotes are preferred (for now :p)
```suggestion
    SITE_GITHUB_URL = 'https://github.com/opendatateam/udata'
```

docs/getting-started.md
bolinocroustibat: Great, updated.

udata/settings.py
bolinocroustibat: Fixed

PR #3086, opened 2024-07-11
feat: improve Sentry logging

bolinocroustibat:
Closes https://github.com/datagouv/data.gouv.fr/issues/1389:

- Update Sentry module and fix deps conflicts (see note)
- Send environment (using `SITE_ID`) and release info to Sentry
- Send profiling and traces to Sentry for performance analyzing
- Add an app-scoped config var `SENTRY_SAMPLE_RATE`

**Note:**
This PR adds `sentry-sdk[flask]` as an install dependency. I'm not sure why it wasn't there before - let me know if this is a mistake!
So since `sentry-sdk[flask]` depends on `urllib3` and `urllib3` depends on `requests` which were incompatible with newer versions of `sentry-sdk`, both pinned versions of `urllib3` and `requests` have been updated in this PR.


udata/sentry.py
```diff
@@ -56,10 +62,24 @@ def init_app(app):
...
+            release=f"udata@{package_version('udata')}",
+            environment=app.config.get('SITE_ID', None),
+            # Set traces_sample_rate to 1.0 to capture 100%
+            # of transactions for performance monitoring.
+            # Sentry recommends adjusting this value in production.
+            traces_sample_rate=sentry_sample_rate,
+            # Experimental profiling
+            _experiments={
+                'profiles_sample_rate': sentry_sample_rate,
+            },
```
maudetes, line 82: I would say that we should log 100% of errors by default for now (and make it configurable) but only a portion of profiling indeed. Ignore this comment if this is what it does ;)

maudetes (approved):
Thank you! We can probably test it on our dev envs :)

Don't forget to add a changelog entry ;)

udata/sentry.py
bolinocroustibat: Yes after some inquiry, this is only for perf so let's keep as it is.

bolinocroustibat: But: Sentry sample rate has been changed as an app config.

bolinocroustibat:
> Thank you! We can probably test it on our dev envs :)
> 
> Don't forget to add a changelog entry ;)

Indeed. Added a changelog entry.

requirements/install.pip
```diff
@@ -279,7 +281,7 @@ uritools==4.0.3
     # via urlextract
 urlextract==0.14.0
     # via -r requirements/install.in
-urllib3==1.25.11
+urllib3==1.26.19
```
maudetes, line 284: I think you should compute pip-compile on all impacted dep files.
It should have gotten triggered by the pip-tool pre-commit hook.
I've updated its version in https://github.com/opendatateam/udata/pull/3086/commits/5982e2a8f5ee0f24ac779887870f2415527bd9cc.

magopian (approved):
LGTM, nice work 👍

PR #3088, opened 2024-07-11
Add 'archived' field for reuse model and endpoints

bolinocroustibat:
Closes https://github.com/datagouv/data.gouv.fr/issues/1273

Also:
- Minor .gitignore update
- Minor translation typo fix


udata/core/reuse/api.py
```diff
@@ -101,7 +101,7 @@ class ReuseListAPI(API):
     @api.marshal_with(reuse_page_fields)
     def get(self):
         args = reuse_parser.parse()
-        reuses = Reuse.objects(deleted=None, private__ne=True)
+        reuses = Reuse.objects(archived=None, deleted=None, private__ne=True)
```
maudetes, line 104: For now, we can keep returning archived reuses on list and suggest endpoints

udata/core/reuse/models.py
```diff
@@ -113,7 +114,7 @@ def is_visible(self):
 
     @property
     def is_hidden(self):
-        return len(self.datasets) == 0 or self.private or self.deleted
+        return len(self.datasets) == 0 or self.private or self.deleted or self.archived
```
maudetes, line 117: Same here, I would not modify it before having discussed product strategy towards archived reuses visibility.

maudetes (commented):
Thank you for this PR! For I think we shouldn't filter out archived reuses yet :)

udata/core/reuse/api.py
bolinocroustibat: Thanks, that makes sense. Fixed

udata/core/reuse/models.py
bolinocroustibat: Indeed, fixed!

udata/core/reuse/search.py
```diff
@@ -43,7 +44,7 @@ def is_indexable(cls, reuse):
 
     @classmethod
     def mongo_search(cls, args):
-        reuses = Reuse.objects(deleted=None, private__ne=True)
+        reuses = Reuse.objects(archived=None, deleted=None, private__ne=True)
```
maudetes, line 47: Woops, here too
```suggestion
        reuses = Reuse.objects(deleted=None, private__ne=True)
```

udata/core/reuse/api.py
```diff
@@ -254,7 +254,7 @@ class ReusesSuggestAPI(API):
     def get(self):
         '''Reuses suggest endpoint using mongoDB contains'''
         args = suggest_parser.parse_args()
-        reuses = Reuse.objects(deleted=None, private__ne=True, title__icontains=args['q'])
+        reuses = Reuse.objects(archived=None, deleted=None, private__ne=True, title__icontains=args['q'])
```
maudetes, line 257:
```suggestion
        reuses = Reuse.objects(deleted=None, private__ne=True, title__icontains=args['q'])
```

maudetes (approved):
Looks good to me!

I think we can prioritize https://github.com/datagouv/data.gouv.fr/issues/1393 next to add this functionality in the admin.

js/views/reuse.vue
```diff
@@ -194,6 +205,24 @@ export default {
...
+      archived_reuse.archived = new Date().toISOString();
+      API.reuses.update_reuse({reuse: this.reuse.id, payload: archived_reuse},
+          (response) => {
+              this.reuse.on_fetched(response);
+          }
+      );
+    },
+    unarchive() {
+      const archived_reuse = this.reuse;
+      archived_reuse.archived = null;
```
ThibaudDauce, line 219: Why not modify the `this.reuse` directly?

js/views/reuse.vue
```diff
@@ -194,6 +205,24 @@ export default {
     edit() {
       this.$go({ name: 'reuse-edit', params: { oid: this.reuse.id } })
     },
+    archive() {
+      const archived_reuse = this.reuse;
+      archived_reuse.archived = new Date().toISOString();
```
ThibaudDauce, line 210: Why not modify the `this.reuse` directly?

ThibaudDauce (changes requested):
:+1: 

Could you add a test for getting archived reuses? I know I will modify this code in my PR https://github.com/opendatateam/udata/pull/3066 so it could be cool to have a test checking this new feature :-)

bolinocroustibat:
> 👍
> 
> Could you add a test for getting archived reuses? I know I will modify this code in my PR #3066 so it could be cool to have a test checking this new feature :-)

OK I'll try! Not sure I know how frontend tests work yet though :)

bolinocroustibat:
> > 👍
> > Could you add a test for getting archived reuses? I know I will modify this code in my PR #3066 so it could be cool to have a test checking this new feature :-)
> 
> OK I'll try! Not sure I know how frontend tests work yet though :)

Since we want the API to return even archived reuses, we cannot test on the API return, but added a test for archived model field behaviour. LMKWYT

ThibaudDauce (approved):

js/views/reuse.vue
```diff
@@ -194,6 +205,22 @@ export default {
     edit() {
       this.$go({ name: 'reuse-edit', params: { oid: this.reuse.id } })
     },
+    archive() {
+      this.reuse.archived = new Date().toISOString();
```
maudetes, line 209: Reacting late on this PR, but I think we would like to add a watch on `reuse.archived` to show a label when archived (similarly to datasets) for a feedback to the user on the action.

PR #3089, opened 2024-07-12
Update pinned dependencies

bolinocroustibat:
Update pinned dependencies according to project dependencies, without updating any project dependencies.

Compatibility between dependencies versions in this PR have been checked by a package manager, and updated to the latest ones when possible.

_**Note:**_
_It would be also useful to update `cryptography` package in project dependencies, since 1) old versions are often causes of system incompatibilities issues and 2) since it's a security dependency and 3) since the project version is very old (`2.8`, we're now at `42.0.8`). This can be done in another PR, or in this PR, or later. Let me know what you think!_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Update pinned dependencies according to project dependencies, without updating any project dependencies [#3089](https://github.com/opendatateam/udata/pull/3089)
 
 ## 9.1.0 (2024-07-11)
 
--- a/requirements/install.pip
+++ b/requirements/install.pip
@@ -50,7 +50,7 @@ celery==5.3.1
     #   celerybeat-mongo
 celerybeat-mongo==0.2.0
     # via -r requirements/install.in
-certifi==2024.2.2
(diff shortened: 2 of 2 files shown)


bolinocroustibat:
> This looks good to me! Don't forget to add a changelog entry before merging.
Indeed! Done

> Have you updated pinned dependencies for other .pip files also?
Nope, I would say to do it in another PR, let me know

> I would say updating cryptography is a bigger jump and require testing more thoroughly. We can update it in another PR I would say
Sure, it makes sense.

ThibaudDauce:
I think this PR breaks `udata-front` with dependencies conflicts (or maybe it's a problem with my cache venv…)

maudetes:
> I think this PR breaks `udata-front` with dependencies conflicts (or maybe it's a problem with my cache venv…)

@nicolaskempf57 has updated udata-front indeed following this PR : https://github.com/datagouv/udata-front/pull/448

bolinocroustibat:
@ThibaudDauce @maudetes Since https://github.com/datagouv/udata-front/pull/448 was merged (thanks @nicolaskempf57 ), are we ok to merge this one?

PR #3091, opened 2024-07-16
Add correct owner permissions on harvest sources

maudetes:


udata/harvest/tests/test_api.py
```diff
@@ -307,11 +307,31 @@ def test_update_source(self, api):
...
+        response = api.put(api_url, data)
         assert200(response)
 
-        source = response.json
-        assert source['url'] == new_url
+    def test_update_source_require_permission(self, api):
+        '''It should not update a source if not the owner'''
+        api.login()
+        source = HarvestSourceFactory()
+        new_url = faker.url()
```
bolinocroustibat, line 324: I would add a little type hint here so that reader knows what's the type without having to inspect `faker.url()` method, but might be because I'm a type hint nazi :)

udata/harvest/tests/test_api.py
```diff
@@ -307,11 +307,31 @@ def test_update_source(self, api):
...
+        data = {
+            'name': source.name,
+            'description': source.description,
+            'url': new_url,
+            'backend': 'factory',
+        }
+        api_url = url_for('api.harvest_source', ident=str(source.id))
```
bolinocroustibat, line 331: Same here

udata/harvest/api.py
```diff
@@ -241,6 +243,8 @@ def put(self, ident):
     @api.doc('delete_harvest_source')
     @api.marshal_with(source_fields)
     def delete(self, ident):
+        source = actions.get_source(ident)
```
bolinocroustibat, line 246: I would add a little type hint here so that reader knows what's the type without having to inspect `faker.url()` method, but might be because I'm a type hint nazi :)

bolinocroustibat (approved):
LGTM

PR #3100, opened 2024-07-24
Fix missing title on new dataset preview

ThibaudDauce:
When previewing a new dataset (not yet in database) the `full_title` is not set.


maudetes (approved):
:pray: so fast!

PR #3101, opened 2024-07-25
Tests not running, fix and refactor generate-fixtures

magopian:
Some tests were not running because they were not following the naming expected by pytest: `test_...`.

This PR changes this by:
- removing the leftover test `udata/tests/features/territories/commands.py` from when the commands were moved to `flask-cli`
- renaming the `udata/commands/tests/fixtures.py` file to `udata/commands/tests/test_fixtures.py`
- fixing the test which was failing
- fixing the fixtures command which was failing to import newly created fixtures

### ATTENTION

This PR also adds a new pytest mark `datagouvfr` for tests that are accessing the http://data.gouv.fr website (which is both slow, and specific to the french usage, so the test runner can exclude those if needed).

### Discussion

The current way we get fixtures from the production website and then load them locally is thus:
1/ pull data from the website and strip some information from it before storing it on disk
2/ load the file and create the objects in the database

We could instead choose to do it the other way round:
1/ pull data from the website and store it verbatim on disk
2/ load the file and create the objects in the database, previously stripping useless or problematic information from it if needed

Doing it this second way would allow us to load more information from the fixtures saved on disk in the future if needed, without changing the "import" phase.

This would be a minor change, probably best suited for a further PR if we believe it's worth it.


udata/commands/fixtures.py
```diff
@@ -30,11 +30,21 @@
...
+    obj["last_modified_internal"] = obj["internal"]["last_modified_internal"]
+    del obj["internal"]
+    del obj["created_at"]
 
 
 @cli.command()
 @click.argument("data-source")
-def generate_fixtures_file(data_source):
+@click.argument("results-file", default=DEFAULT_FIXTURES_RESULTS_FILE)
+def generate_fixtures_file(data_source, results_file):
```
bolinocroustibat, line 47: I would type hint the args and output of the function to comply with the rest of the code which is mostly type hinted  or aiming to be fully type hinted. In this function for example, it could really help understanding it for a newbie of this code in a blink of an eye. But maybe that's because I'm a type hint nazi :)

Like 

```python
def generate_fixtures_file(data_source: str, results_file: str) -> None:
    ...
```

udata/commands/fixtures.py
```diff
@@ -30,11 +30,21 @@
 DEFAULT_FIXTURE_FILE = (
     "https://raw.githubusercontent.com/opendatateam/udata-fixtures/main/results.json"  # noqa
 )
+DEFAULT_FIXTURES_RESULTS_FILE = "results.json"
+
+
+def fix_dates(obj):
```
bolinocroustibat, line 36: Same here - `obj` could be type hinted

udata/commands/fixtures.py
```diff
@@ -30,11 +30,21 @@
 DEFAULT_FIXTURE_FILE = (
     "https://raw.githubusercontent.com/opendatateam/udata-fixtures/main/results.json"  # noqa
 )
+DEFAULT_FIXTURES_RESULTS_FILE = "results.json"
```
bolinocroustibat, line 33: Could be nice/safer/more readable to use `pathlib.Path` instead of strings filenames for files?

bolinocroustibat (approved):
LGTM with a few remarks! 

LGTM

udata/commands/fixtures.py
bolinocroustibat: Also:

If we keep it as a string and not a Path, I would name it DEFAULT_FIXTURES_RESULTS_FILENAME instead of DEFAULT_FIXTURES_RESULTS_FILE, and I would name results_filename var instead of results_file. Let me know if that makes sense.

udata/commands/fixtures.py
```diff
@@ -112,13 +124,14 @@ def generate_fixtures_file(data_source):
...
-        json.dump(json_result, f)
+    with results_file.open("w") as f:
+        json.dump(json_result, f, indent=2)
+        print(f"Fixtures saved to file {results_filename}")
 
 
 @cli.command()
 @click.argument("source", default=DEFAULT_FIXTURE_FILE)
-def generate_fixtures(source):
+def generate_fixtures(source: str) -> None:
```
maudetes, line 134: I think we may want to rename `generate_fixtures` by something like `import_fixtures` at some point

maudetes (commented):
Thank you for this PR! Totally agreed for the stripping at the import time instead of when storing.

Well seen for the tests files. I think the PR name should mention fixtures refactoring though.

udata/commands/tests/test_fixtures.py
```diff
@@ -0,0 +1,26 @@
...
+from udata.tests import DBTestMixin, TestCase
+
+
+class FixturesTest(DBTestMixin):
+    @pytest.mark.datagouvfr
+    @pytest.mark.options(FIXTURE_DATASET_SLUGS=["nombre-de-personnes-rickrollees-sur-data-gouv-fr"])
+    def test_generate_fixtures_file_and_generate_fixtures(self, cli):
+        with NamedTemporaryFile(mode="w+", delete=True) as fixtures_fd:
+            # Get the fixtures from http://data.gouv.fr.
+            result = cli("generate-fixtures-file", "http://data.gouv.fr", fixtures_fd.name)
```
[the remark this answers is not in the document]
magopian: I'm not exactly clear what you would like to see to move this PR forward. I agree that testing against the prod env isn't ideal. Also note that the previous test wasn't testing the fixtures generation at all.

So maybe, i could:

1. remove the part of the test that generates the fixtures (back to the original code which wasn't testing it)
2. modify the part of the test that imports those fixtures to read from a local `results_test.json` file (generated from prod for now, and committed to the repository): the previous code was failing, the fixtures from the test files won't load in the db as is
3. create a future PR that generates this `results_test.json` file from the current env, and makes sure the newly created `results_test.json` file is no different than the local one

This way, once the new PR is landed, we'd have two decoupled tests, and the fixtures would be generated from a local env.

What do you think?

udata/commands/fixtures.py
magopian: Yes definitely! The current naming is confusing 😬 
I can change it right away if you think it still makes sense in the context of this PR

udata/commands/tests/test_fixtures.py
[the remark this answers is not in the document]
magopian: > For 2., I would use the DEFAULT_FIXTURE_FILE for now.
Yes indeed, that's simpler, and it would make sure the "official" fixtures file is up to date and still working as expected, good idea 👍 

>For 3., I'm not sure about comparing the generated file with a snapshot one, we just want to make sure that importing it succeed with the correct content being imported. It would prevent making the test fail on any model modification as long as fixtures get generated and loaded without issues.
I see, if I understand that correctly, we would have two tests, one that is making sure that the `DEFAULT_FIXTURE_FILE` is still compatible, and a second one that would first generate some data in the db, then `generate_fixtures_file` and then `import_fixtures` with it (to make sure the data we extract can be re-imported).

Is that what you had in mind?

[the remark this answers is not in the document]
maudetes: Yes exactly! :pray: Seems to cover any modifications that would break the fixtures generation or importation :)

udata/commands/fixtures.py
maudetes: Maybe in the refactoring PR instead? :)

udata/commands/tests/test_fixtures.py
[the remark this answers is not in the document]
magopian: Ok, so I created a simpler, more focused PR #3105 to fix the non running tests for 1.
And a PR (draft) #3106 to address the rest of this comment (2. and 3.)

magopian:
I'm closing this one now that it's being superseded by #3105 and #3106.

PR #3103, opened 2024-07-25
clean: remove unused JS imports

bolinocroustibat:
Remove unused JS imports.


magopian (approved):
LGTM

I'm curious, did you use eslint or a similar tool to detect and remove those unused imports? If so, it might make sense to have a CI action, or a pre-commit that would prevent pushing code with unused imports, what do you think?

bolinocroustibat:
> LGTM
> 
> I'm curious, did you use eslint or a similar tool to detect and remove those unused imports? If so, it might make sense to have a CI action, or a pre-commit that would prevent pushing code with unused imports, what do you think?

I used a VScode plugin... But yeah that's a good idea! I think on the frontend side we could use [Biome](https://biomejs.dev/) for that, which is a bit like what Ruff is to Python (Biome replaces ESlint and many other linter/formatter while being way faster and becoming some standard recently). Anyway since frontend admin is bein refactored maybe we can wait to do that on the new stack. We can do that in another PR and create a backlog ticket about that

maudetes:
> Anyway since frontend admin is bein refactored maybe we can wait to do that on the new stack. We can do that in another PR and create a backlog ticket about that

See also: https://github.com/datagouv/data.gouv.fr/issues/392

PR #3104, opened 2024-07-25
feat: add un/archive dataset and reuse button

bolinocroustibat:
Closes https://github.com/datagouv/data.gouv.fr/issues/1393

EDIT: also added a "archived" badge UI on reuse card.


magopian (approved):
LGTM, nicely done 👍 

Do we need/want tests for that? I couldn't run tests locally, they fail with various phantomjs errors, i'll have to check that later.

magopian:
Quick heads up, I believe there's a mistake in the PR description, #3104 is the PR number, not the original issue number?

js/views/dataset.vue
```diff
@@ -169,6 +182,22 @@ export default {
         edit() {
             this.$go({name: 'dataset-edit', params: {oid: this.dataset.id}});
         },
+        archive() {
+        this.dataset.archived = new Date().toISOString();
```
maudetes, line 186: Somehow the `archive` quickly shows but then disappears in the second (probably due to a watch effect)?

maudetes (commented):
Thank you! :pray:

Could update the changelog entry made for reuse with this PR as well?

js/views/dataset.vue
bolinocroustibat: Couldn't reproduce... I lack a bit experience on this project JS part, any idea on why and how to fix that?

bolinocroustibat:
> for that? I couldn't run tests locally, they fail with various phantomjs errors, i'll have to check that later.

It seems there is no tests for the JS part, aren't there?
Also I think the team is working on a complete frontend admin refactoring right now, so maybe it's not the right time to start to add tests to the front admin?

js/views/dataset.vue
bolinocroustibat: I've been debugging this issue but couldn't figure it out yet for now.
Notes:
- it seems the watcher is successfully triggered on any update on dataset.archived
- it seems the function `addOrRemoveBadge`works as expected as it add the "archived" badge to the badges list and removes it as expected
Could it be some display logic error elsewhere in a badges component?

bolinocroustibat: BTW logic to add a "archived" badge on reuse page as well as been added to the PR.

maudetes: Finally understood the issue! It came from the return value beeing not formatted exactly as ISO and thus triggering the watch effect again. Fixed it by checking if the value is None or not when removing the badge : https://github.com/opendatateam/udata/pull/3104/commits/6107ab52c9f13318597818db6c3d75563f923b1c.

bolinocroustibat: @maudetes Amazing, well done! Shall I merge from main and merge?

maudetes (approved):
I think we're good now! Thank you :pray:

PR #3105, opened 2024-07-26
tests not running (bis): remove leftover test, rename and fix fixtures test

magopian:
This is a rework of #3101 which will be broken down in two parts:

1. this PR, simply fixing the fixtures test, making sure it's running, and removing a leftover test
2. a new PR which will implement the [fixtures refactoring](https://github.com/opendatateam/udata/pull/3101#discussion_r1692798253)


udata/commands/tests/test_fixtures.py
```diff
@@ -0,0 +1,15 @@
...
+from tempfile import NamedTemporaryFile
+
+from udata import models
+from udata.commands.fixtures import generate_fixtures
+from udata.tests import DBTestMixin, TestCase
+
+
+class FixturesTest(DBTestMixin):
+    def test_generate_fixtures(self, cli):
+        cli("generate-fixtures")
```
maudetes, line 11: Is there a reason why you prefered to use cli instead of importing the function?

maudetes (approved):
Thank you for this new focused PR! Looks good to me :)

Don't forget to [add a changelog entry](https://github.com/opendatateam/udata/blob/master/CHANGELOG.md)

udata/commands/tests/test_fixtures.py
magopian: The `generate_fixtures` being a flask-cli, it has some prerequisites:
`Error: Could not locate a Flask application. You did not provide the "FLASK_APP" environment variable, and a "wsgi.py" or "app.py" module was not found in the current directory.`

I saw that the jobs and datasets commands were tested this way, would you rather we do it differently?

PR #3109, opened 2024-07-30
Add dataservice Elasticsearch

ThibaudDauce:


maudetes:
Let's close for now and re-open when we'll work on the search service.

PR #3112, opened 2024-07-30
Add dataservices to fixtures

magopian:
Fixes #1400

This PR will have one failing test as long as the [PR on udata-fixtures](https://github.com/opendatateam/udata-fixtures/pull/2) isn't merged.

PR #3106, opened 2024-07-26
Fixtures refactoring

magopian:
This is a rework of https://github.com/opendatateam/udata/pull/3101 which will be broken down in two parts:

1. #3105 about simply fixing the fixtures test, making sure it's running, and removing a leftover test
 2. this PR which implements the https://github.com/opendatateam/udata/pull/3101#discussion_r1692798253

PR #3113, opened 2024-07-30
Version fixtures

magopian:
Since #3106 the fixtures are generated differently (the whole json is stored, instead of filtering it before saving it to file), and since #3112 the `dataservices` are expected in the fixtures.

This causes issues in other services and repositories counting on the [udata-fixtures](https://github.com/opendatateam/udata-fixtures/blob/main/results.json) to stay compatible.

This PR does two things:

1. add a tag to the default `results.json` url, so udata-fixtures can be updated without impacting other repositories
2. treat the `dataservices` entry in the `results.json` optional, so newer versions of the `import-fixtures` command can still function properly with the "legacy" `results.json`.

In the future, when all repositories have been updated to use the version of udata that has the "fixed tag for fixtures", we'll be able to update the udata-fixtures.

### DIFF
--- a/udata/commands/fixtures.py
+++ b/udata/commands/fixtures.py
@@ -39,9 +39,8 @@ COMMUNITY_RES_URL = "/api/1/datasets/community_resources"
 DISCUSSION_URL = "/api/1/discussions"
 
 
-DEFAULT_FIXTURE_FILE: str = (
-    "https://raw.githubusercontent.com/opendatateam/udata-fixtures/main/results.json"  # noqa
-)
+DEFAULT_FIXTURE_FILE_TAG: str = "v1.0.0"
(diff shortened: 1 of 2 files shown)


udata/commands/fixtures.py
```diff
@@ -199,6 +198,9 @@ def import_fixtures(source):
                         MessageDiscussionFactory(**message, posted_by=user) for message in messages
                     ],
                 )
+            if "dataservices" not in fixture:
+                # For backwards compatibility.
+                continue
             for dataservice in fixture["dataservices"]:
```
maudetes, line 204: Maybe use dict.get ?
```suggestion
            # We may be missing dataservices in older fixtures
            for dataservice in fixture.get("dataservices", []):
```

maudetes (approved):
Thank you!

udata/commands/fixtures.py
magopian: yep, even better, thanks 🙏

PR #3115, opened 2024-07-30
Add report get endpoint

maudetes:
Additionally to the list endpoint, add a get endpoint to get all details about a specific report.

Add a `self_api_url ` to easily get access to this new get endpoint in list endpoint for example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - Adds latest `sentry-sdk[flask]` as an install dependency, and update Sentry logic to be able to send environment, app version and profiling/performance info [#3086](https://github.com/opendatateam/udata/pull/3086)
 - Add report get endpoint and make report get and list admin only [#3115](https://github.com/opendatateam/udata/pull/3115)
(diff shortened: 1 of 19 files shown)


udata/core/reports/api.py
```diff
@@ -37,6 +37,14 @@ def post(self):
         return report, 201
 
 
+@ns.route("/<report:report>/", endpoint="report")
+class DataserviceAPI(API):
+    @api.doc("get_report")
+    @api.marshal_with(Report.__read_fields__)
+    def get(self, report):
+        return report
```
ThibaudDauce, line 45: Admin only? (maybe I didn't do it for list endpoint :-) )

ThibaudDauce (approved):
:+1:

udata/core/reports/api.py
maudetes: Indeed, we were talking about it with @Pierlou yesterday! I've made these routes admin only in https://github.com/opendatateam/udata/pull/3115/commits/8196473956f37ee98664ba524ad005e4c0d66cd7

PR #3116, opened 2024-07-31
Use requests==2.32.3 everywhere for consistency

magopian:
This fixes the "urllib3 is the wrong version down the line" issue when installing the project from scratch.
This goes hand in hand with the commit on udata-front: https://github.com/datagouv/udata-front/pull/455

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Fix the version of udata-fixtures used by `udata import-fixtures` [#3114](https://github.com/opendatateam/udata/pull/3114)
 - Update to the version v2.0.0 of udata-fixtures (with the dataservices)
 - Add type hints [#3111](https://github.com/opendatateam/udata/pull/3111)
+- Make sure requests v2.32.3 is used everywhere consistently [#3116](https://github.com/opendatateam/udata/pull/3116)
 
 ## 9.1.2 (2024-07-29)
 
--- a/requirements/develop.pip
+++ b/requirements/develop.pip
@@ -68,6 +68,7 @@ blinker==1.4
     #   flask-principal
     #   flask-security-too
     #   flask-sitemap
+    #   sentry-sdk
 boto3==1.26.102
     # via
     #   -c requirements/install.pip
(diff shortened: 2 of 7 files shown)


bolinocroustibat (changes requested):
Thanks for the fix!

One remark, there shouldn't be any need to add sentry as dependency in dev dependencies, isn't it? I wonder why it has been added by pip-compile?

magopian:
> One remark, there shouldn't be any need to add sentry as dependency in dev dependencies, isn't it? I wonder why it has been added by pip-compile?

This is compiled automatically with `pip-compile`, which is run automatically with `pre-commit` if it's installed on each commit that modifies a `.in` file, or can be run manually.

So, to compile the dev dependencies, found in `develop.in`, it'll need
- [requirements from `test.in`](https://github.com/opendatateam/udata/blob/master/requirements/develop.in#L1)
- which in turn needs [requirements from `install.in`](https://github.com/opendatateam/udata/blob/master/requirements/test.in#L1)
- which in turn uses [sentry-sdk](https://github.com/opendatateam/udata/blob/master/requirements/install.in#L45)

The dependencies are then compiled, and the results are stored in `.pip` files, which are used to install requirements.

So the requirements are duplicated in the various `.pip` files, but this is by design (and it even says from [where it got the dependency](https://github.com/opendatateam/udata/pull/3116/files#diff-3f663ebdc2f319909ce37441dd0bb29889f427c6cb7d50c277865562ba31c7fbR594-R597)).

I believe this ensures that we don't encounter the issue fixed by this PR, where different versions of a lib are used throughout the project, and so the final version of the lib isn't always what you expect it to be.

bolinocroustibat (approved):
LGTM, thanks for the explanation!

I'm quite confused by all the different requirements files in this project... I wish we could use a single `pyproject.toml` requirements file with section for each group dependencies (dev, test, report, etc...). The team has been trying to switch to `pyproject.toml` several times (including me: https://github.com/opendatateam/udata/pull/3090/files) but for this project it seems quite complicated...

PR #3118, opened 2024-07-31
Fix the `extend-select` ruff.toml config

magopian:
My comprehension from https://docs.astral.sh/ruff/settings/#lint_select is that it defines the rules ruff should care about.

We do want all the default rules that ruff cares about, and add the https://docs.astral.sh/ruff/rules/#isort-i ones (missing required imports, and unsorted imports), and thus
https://docs.astral.sh/ruff/settings/#lint_extend-select should be used instead.

### DIFF
--- a/udata/api/fields.py
+++ b/udata/api/fields.py
@@ -4,7 +4,28 @@ import logging
 import pytz
 from dateutil.parser import parse
 from flask import request, url_for
-from flask_restx.fields import *  # noqa
+
+# Explicitly import all of flask_restx fields so they're available throughout the codebase as api.fields
+from flask_restx.fields import Arbitrary as Arbitrary
+from flask_restx.fields import Boolean as Boolean
+from flask_restx.fields import ClassName as ClassName
+from flask_restx.fields import Date as Date
+from flask_restx.fields import DateTime as DateTime
+from flask_restx.fields import Fixed as Fixed
+from flask_restx.fields import Float as Float
+from flask_restx.fields import FormattedString as FormattedString
+from flask_restx.fields import Integer as Integer
+from flask_restx.fields import List as List
+from flask_restx.fields import MarshallingError as MarshallingError
+from flask_restx.fields import MinMaxMixin as MinMaxMixin
+from flask_restx.fields import Nested as Nested
+from flask_restx.fields import NumberMixin as NumberMixin
+from flask_restx.fields import Polymorph as Polymorph
+from flask_restx.fields import Raw as Raw
+from flask_restx.fields import String as String
+from flask_restx.fields import StringMixin as StringMixin
+from flask_restx.fields import Url as Url
+from flask_restx.fields import Wildcard as Wildcard
 
 from udata.uris import endpoint_for
 from udata.utils import multi_to_dict
--- a/udata/app.py
+++ b/udata/app.py
@@ -206,7 +206,6 @@ def register_extensions(app):
         mail,
         models,
         mongo,
-        notifications,
         routing,
         search,
         sentry,
--- a/udata/app.py
+++ b/udata/app.py
@@ -206,6 +206,7 @@ def register_extensions(app):
         mail,
         models,
         mongo,
+        notifications,  # noqa
         routing,
         search,
         sentry,
--- a/udata/auth/__init__.py
+++ b/udata/auth/__init__.py
@@ -2,19 +2,26 @@ import logging
 
 from flask import current_app, render_template
 from flask_principal import Permission as BasePermission
-from flask_principal import (
-    PermissionDenied,  # noqa: facade pattern
-    RoleNeed,
-    UserNeed,  # noqa: facade pattern
-    identity_loaded,  # noqa: facade pattern
-)
-from flask_security import (  # noqa
-    Security,  # noqa
-    current_user,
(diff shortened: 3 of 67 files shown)


bolinocroustibat (approved):
LGTM, thank you for noticing it, but I wonder if we shouls also update this in the docs?
https://github.com/opendatateam/udata/blob/master/CONTRIBUTING.md?plain=1#L91

I guess Ruff will now take into account the config (but not 100% sure) so it could be simply `ruff check --fix .` from now on?
Even `ruff check --fix && fruff format`, by the way.

If so, feel free to update it and merge without further PR approval.

magopian:
@bolinocroustibat we have a problem houston: because the ruff config changed, we have 191 linting errors on the codebase.
Do we want to commit those fixes? In the same PR? Another one?

bolinocroustibat:
> @bolinocroustibat we have a problem houston: because the ruff config changed, we have 191 linting errors on the codebase.
> Do we want to commit those fixes? In the same PR? Another one?

Haha is this even better linting?
If yes ok to commit those in same PR.

magopian:
Ok, I've added fixes (most are automatic fixes, I've spread the automatic safe fixes from the automatic unsafe fixes from the manual fixes in separate commits), which makes this PR rather large sadly.

Good luck for the reviews 😬

udata/api/fields.py
```diff
@@ -4,7 +4,28 @@
 import pytz
 from dateutil.parser import parse
 from flask import request, url_for
-from flask_restx.fields import *  # noqa
+
+# Explicitly import all of flask_restx fields so they're available throughout the codebase as api.fields
+from flask_restx.fields import Arbitrary as Arbitrary
```
maudetes, line 9: Unsure about the `as` with same value?
```suggestion
from flask_restx.fields import Arbitrary as Arbitrary
```

udata/app.py
```diff
@@ -206,7 +206,6 @@ def register_extensions(app):
         mail,
         models,
         mongo,
-        notifications,
```
maudetes, line 209: I would pay attention when removing this kind of imports this because there are some signals registering in notifications when loading it.

udata/api/fields.py
```diff
@@ -4,7 +4,28 @@
 import pytz
 from dateutil.parser import parse
 from flask import request, url_for
-from flask_restx.fields import *  # noqa
```
maudetes, line 7: Is this due to ruff complaint? Shouldn't it be bypass with a #noqa mechanic?

udata/auth/__init__.py
```diff
@@ -2,20 +2,27 @@
...
+from flask_principal import identity_loaded as identity_loaded
+from flask_security import Security as Security
+from flask_security import current_user as current_user
+from flask_security import login_required as login_required
+from flask_security import login_user as login_user
 from werkzeug.utils import import_string
 
+__all__ = [
+    "PermissionDenied",
+    "UserNeed",
```
maudetes, line 17: Why not `RoleNeed`? Why the need for this `__all__`?

udata/core/organization/api.py
```diff
@@ -207,10 +210,6 @@ def delete(self, org, badge_kind):
         return badges_api.remove(org, badge_kind)
 
 
-from udata.core.contact_point.api import ContactPointApiParser
```
maudetes, line 210: I  had in mind that @ThibaudDauce put the import here for circular imports reason, but it may not be the case anymore

maudetes (approved):
Thank you for this cleaning!

I have minor questions, the main point of attention may be the import deletion that may have unexpected side effects.

bolinocroustibat (approved):
LGTM, wow looks even more neat now! Good for my code OCD 😅

udata/api/fields.py
magopian: This is a "trick": it allows the linter to understand that this import is unused locally, but still needed.
As explained by the comment, we had a `from flask_restx.fields import *  # noqa` so the `flask_restx` fields were available as `api.fields`, but that would trip the linter:
- not being able to check if the imports were used, AND
- not being able to understand that the `flask_restx` fields used in the file were coming from the `import *`

udata/app.py
magopian: Ok, i'll do that 👍

udata/api/fields.py
magopian: It's bypassed on this line, but as explained above, it'll then complain about all the `String` and other `flask_restx` fields in the file, not being able to make sure it's coming from `flask_restx.fields`: https://docs.astral.sh/ruff/rules/undefined-local-with-import-star-usage/

udata/auth/__init__.py
magopian: Oops, sorry, this is a leftover from an experiment, it shouldn't be there, I removed it 🙏

udata/core/organization/api.py
maudetes: Hmm I guess it's okay with what you've described

udata/api/fields.py
maudetes: Fine with me for explicit import instead of `*` but I would use a `# noqa` that seems more explicit than this trick.

PR #3119, opened 2024-08-01
Remove the failing legacy frontend jstests from bumpr

magopian:
The js tests for the legacy admin frontend have been failing for a very long time now, and probably won't ever be fixed, as the admin is being re-implemented in udata-front.

So let's remove the jstests from bumpr, our release helper.


maudetes (approved):
Thank you!

PR #3122, opened 2024-08-02
Expose the dataservices in their organization's catalog and their own catalog export

magopian:
Fixes https://github.com/datagouv/data.gouv.fr/issues/1392

- [x] in their organization's RDF catalog
- [x] in a direct export of the dataservice

This reproduces very closely what's done for datasets.


udata/core/organization/api.py
```diff
@@ -176,7 +177,8 @@ def get(self, org, format):
         page = int(params.get("page", 1))
         page_size = int(params.get("page_size", 100))
         datasets = Dataset.objects(organization=org).visible().paginate(page, page_size)
-        catalog = build_org_catalog(org, datasets, format=format)
+        dataservices = Dataservice.objects(organization=org).visible().paginate(page, page_size)
```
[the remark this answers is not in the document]
magopian: I refactored the Site's catalog implementation out into the dataservice's queryset, and reused it. What do you think @maudetes, does that look better?

udata/core/dataservices/models.py
```diff
@@ -29,6 +31,42 @@ def visible(self):
...
+        # if somebody is doing the same thing as us (keeping the list of all the datasets IDs for the entire catalog then
+        # listing all dataservices in a second pass)
+        # Another option is to do some tricky Mongo requests to order/group datasets by their presence in some dataservices but
+        # it could be really hard to do with a n..n relation.
+        # Let's keep this solution simple right now and iterate on it in the future.
+        dataservices_filter = Q(datasets__in=[d.id for d in datasets])
+
+        # On the first page, add all dataservices without datasets
+        if page == 1:
+            dataservices_filter = dataservices_filter | Q(datasets__size=0)
```
maudetes, line 66: Not introduced in this PR, but dataservice.dataset may not exist (and not be an empty list).
I think we need to add `| Q(datasets__exists=False)`.

udata/core/dataservices/models.py
```diff
@@ -29,6 +31,42 @@ def visible(self):
     def hidden(self):
         return self(db.Q(private=True) | db.Q(deleted_at__ne=None) | db.Q(archived_at__ne=None))
 
+    def filter_by_dataset_pagination(self, datasets, page):
```
maudetes, line 34: Thank you for moving this logic in a queryset! :pray:

udata/tests/organization/test_organization_rdf.py
```diff
@@ -103,10 +109,12 @@ def test_catalog_pagination(self):
             _external=True,
         )
         DatasetFactory.create_batch(total, organization=origin_org)
+        DataserviceFactory.create_batch(total, organization=origin_org)
 
         # First page
         datasets = Dataset.objects.paginate(1, page_size)
-        catalog = build_org_catalog(origin_org, datasets, format="json")
+        dataservices = Dataservice.objects.paginate(1, page_size)
```
maudetes, line 116: Do we need to update this logic in test?

udata/core/dataservices/models.py
magopian: The final result would be:
```python
        if page == 1:
            dataservices_filter = (
                dataservices_filter | Q(datasets__size=0) | Q(datasets__exists=False)
            )
```

Is that what you were thinking?

udata/tests/organization/test_organization_rdf.py
magopian: ah, yes, indeed, good catch, i'll update the tests 🙏

bolinocroustibat (approved):
Not sure I get everything as I don't know the codebase sufficiently yet, but LGTM!

magopian:
@maudetes thanks for the review, I believe I've updated the PR with your comments, let me know what you think 🙏

udata/core/dataservices/models.py
ThibaudDauce: I don't know why `dataservice.datasets` could be `None`. For me it's always a list… If it's not wanted we could add a migration to convert all `None` to `[]`. If it's wanted maybe add a test with a `dataservice.datasets = None` to be sure this new condition `Q(datasets__exists=False)` is working as intended (and that someone doesn't remove it in 6 months thinking it's weird to have both).

magopian: From a few tests I made locally, it does seem that even when a Dataservice is created with no datasets (or with its `datasets` set to `None`), the model still ends up with an empty list in `Dataservice.datasets`, probably because the default value of `MongoEngine.ListField.default` is `[]`.

Also, if we wanted to allow non existing datasets, then we'd fail on [this line in udata/core/dataservices/rdf.py](https://github.com/opendatateam/udata/blob/730426458014867b8a7c247dc088b87f876379c8/udata/core/dataservices/rdf.py#L126).

So i'll just revert the change adding ` | Q(datasets__exists=False)` as it seems useless.

@maudetes if you feel we should change this behavior in the future, I think we should address it in a separate issue?

maudetes: I haven't delved into, but we somehow have None value in prod instead of empty list on this dataservice https://www.data.gouv.fr/fr/dataservices/6661e09c7230a8fc3b1ec229/.

We've already faced this situation before, ex: https://github.com/datagouv/data.gouv.fr/issues/1088.
Agreed for tackling it in a separate issue: https://github.com/datagouv/data.gouv.fr/issues/1430

Issue #1398, opened 2018-01-25
Add a "functional" type to resources


abulte:
First part of the overhaul of the resources display / navigation.

# What

We should be able to add a "functional" type to resources: is it documentation, stock data, incremental update data...?

# Why

Clarify the display of resources (doc on top?). Pave the way for full resources overhaul (space and time projection...).

# How

1. Identify possible types https://github.com/opendatateam/udata/issues/1398#issuecomment-360478270
2. How to store them?
3. How to display them on front? https://github.com/opendatateam/udata/issues/1398#issuecomment-361247945
4. How do producers fill them on back?
5. How to migrate current data?
6. How to make sure producers do fill the attributes, instead of relying on the default value?

abulte:
# Possible types

- Documentation
- Stock data
- Incremental update data (w/ a frequency?)
- ...

noirbizarre:
## Proposal

Reuse the existing `type` field on resources to add these extra types

abulte:
This is what I have in mind ATM about the display:

<img width="834" alt="paper work 3" src="https://user-images.githubusercontent.com/119625/35513018-f6a4feac-0501-11e8-9df3-db8771ffbb1b.png">

Is this compatible w/ reusing the `filetype` field as @noirbizarre suggested? What do we do with `file`, `remote` and `api` in this case? Display them in their own block? And what about resources that are both `file` and `documentation` for example?

noirbizarre:
Maybe we can have a backward compatibility mapping like it's done for frequencies [here](https://github.com/opendatateam/udata/blob/master/udata/core/dataset/models.py#L56-L61)
To me, the ideal solution is 2 fields:
- a *where* field, ie. `location = {local,remote}`
- a *what* field, ie. `type = {documentation,full,partial,misc}`
The current type field is composite from these 2 concepts

abulte:
Do we really need to ask the producer about `{local, remote}`? It seems like something we can infer from the resource creation / edition alone.

Also, shouldn't API be a possible `what` type field?

In this case we could replace the current field with the new type, with a migration before hand.

noirbizarre:
Yes we can infer this. But this need some UI/UX work on the resource creation and edition.
Currently, the `type` field purpose is to switch the create/edit form fields depending on it.
If we always display the url field and the upload widget, this is not necessary anymore.
The API form is another case but in fact this is the couple `location=remote,type=api`.

[...]

PR #3124, opened 2024-08-06
Add sort and pagination to the organization's discussions route

magopian:
Fixes [#1398](https://github.com/datagouv/data.gouv.fr/issues/1398)

Note: this add more than just the `sort` and `page*` parameters, it adds all the parameters available on the `/discussions/` route.
It also adds an `org` parameter on the `/discusions/` route.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Convert reuse to new API system [#3066](https://github.com/opendatateam/udata/pull/3066)
 - Fix circular import error [#3128](https://github.com/opendatateam/udata/pull/3128)
 - Add an option to specify the port when using `inv serve` [#3123](https://github.com/opendatateam/udata/pull/3123)
+- Add sort and pagination parameters to the organization discussions API [#3124](https://github.com/opendatateam/udata/pull/3124)
 
 ## 9.1.3 (2024-08-01)
 
--- a/udata/api/__init__.py
+++ b/udata/api/__init__.py
@@ -15,7 +15,9 @@ from flask import (
     url_for,
 )
 from flask_restx import Api, Resource
+from flask_restx._http import HTTPStatus
 from flask_restx.reqparse import RequestParser
+from flask_restx.utils import merge
 from flask_storage import UnauthorizedFileType
 
 from udata import entrypoints, tracking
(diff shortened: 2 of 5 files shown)


nicolaskempf57:
It seems that the route is returning an array of discussions and note a `pager` with all related informations (data, page, page_size, total, next_page and previous_page)

magopian:
I thought we wanted to stay backwards compatible, and if I understand correctly, at the moment [we get a simple list](https://www.data.gouv.fr/api/1/organizations/base-adresse-nationale/discussions/), not the "pager" mode that you could get for example on [the discussions API itself](https://www.data.gouv.fr/api/1/discussions/?org=base-adresse-nationale).

Should we change that, and make a breaking changea release?

magopian:
Having discussed this with @nicolaskempf57 , I now understand that we want two different return types, given the `page_size` existence in the query:

- `page_size` present (eg `?page_size=20`): return a pager
- `page_size` absent: return a list of all the discussions

magopian:
So... the current implementation "works", as in it returns the proper results: a pager if the `page_size` is provided, or a simple list (non paginated) for the old admin.

What DOESN'T work, though, is the api docs: it's set using the `marshal_with` helper, which is now "wrapped" in another decorator (which will call it using the needed fields given the result from the `get` function).

As the api doc isn't set in the proper place, the swagger generated doesn't show it. 
What should it be set to anyway? When the function is decorated, at startup, and the swagger file is generated, we don't know if the function will be called with a `page_size` parameter, and thus we don't know what the format of the response will be.

I'm not sure what the next step should be. I guess we could:
1. find a way to force the apidoc to be the "pager" version, as the swagger doc should display the most up to date version, and the old admin doesn't use swagger anyway ?
2. add a new endpoint called `get_with_filters_and_sorts` or something

I believe option 2. is simpler (way less convoluted code), and makes sense to keep the API as backwards compatible as possible?

Not sure what @maudetes @ThibaudDauce @nicolaskempf57 think?

magopian:
With the latest commit, the swagger documentation is back, and set to the "admin version" (list, not pager). We can choose to document using the pager version instead, it's a simple change.

ThibaudDauce:
Same for the PR https://github.com/opendatateam/udata/pull/3126, it adds a lot of complexity for supporting the old admin… Why the new admin cannot use the `/discussions` endpoint with an `org=xxx` filter?

magopian:
Closing this in favor of #3147

PR #3125, opened 2024-08-06
Properly set a reference to a user in Discussion.closed_by when importing the fixtures

magopian:
When running `udata import-fixtures`, the fixtures were creating discussions which were closed with a `closed_by` filled with the dictionnary of the user closing the discussion, instead of a reference to the user, causing server errors when listing such discussions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Fix circular import error [#3128](https://github.com/opendatateam/udata/pull/3128)
 - Add an option to specify the port when using `inv serve` [#3123](https://github.com/opendatateam/udata/pull/3123)
 - Add a new `related_to` filter parameter to the activities API endpoint [#3127](https://github.com/opendatateam/udata/pull/3127)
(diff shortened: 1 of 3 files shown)


maudetes (approved):
Don't forget to add a changelog entry! Was this problem not occurring before the refacto?

magopian:
I believe it was occuring before the refacto, but it wasn't easily spotted, as not many discussions were closed: I can see that 11 out of the nearly 150 discussions are closed in the [legacy fixtures](https://github.com/opendatateam/udata-fixtures/blob/v1.0.0/results.json) (search for `"closed_by": {`).

I'm adding the CHANGELOG and rebasing the branch on master, and then will merge it, thanks for the review 🙏

magopian:
Oops, I see that I got lost in the refactoring to properly set users everywhere... and forgot to properly test (and set) the `closed_by` user... I rebased the PR, and added the test and code needed, would you mind reviewing it again @maudetes ? 😬

maudetes (approved):
I think we're clear? :D

PR #3126, opened 2024-08-09
New parameters for the organization reuses API

magopian:
Fixes [#1397](https://github.com/datagouv/data.gouv.fr/issues/1397)

Note: this add more than just the sort and page* parameters, it adds all the parameters available on the `/reuses/` route.

The implementation doesn't take into account the indications from #1397, but rather implements it by reusing Reuse's API parser which is already present.

It should give the same results. The same logic was applied in #3124.

~~IMPORTANT NOTE: this PR is based on the PR #3066 that needs to be merged first.~~(merged)


bolinocroustibat (approved):
LGTM

ThibaudDauce:
Hi,

Thanks for looking into this, it wasn't easy…

I'm a little bit concerned about the amount of complexity added… Maybe we can just update the `/reuses` endpoint to allow to return `private` reuse if the user has the correct rights. It avoids to update the `/org/{id}/reuses` endpoint and leave it be for the current old admin, no?

magopian:
We're closing this one after the discussion in https://github.com/datagouv/data.gouv.fr/issues/1397#issuecomment-2310429729, too much complexity for too little gain, we're going in another direction.

PR #3127, opened 2024-08-13
Add a new `related_to` filter parameter to the activities endpoint

magopian:
Fixes [#1414](https://github.com/datagouv/data.gouv.fr/issues/1414)

Warning: the [return type of the activity list API endpoint](https://github.com/opendatateam/udata/blob/master/udata/core/activity/api.py#L70) seems wrong: it returns a list of pagers, instead of just a list, or just a pager.
Is it related to the legacy admin not being able to handle pagination?


udata/tests/api/test_activities_api.py
```diff
@@ -0,0 +1,63 @@
...
+
+class FakeReuseActivity(Activity):
+    key = "fakeReuse"
+    related_to = db.ReferenceField(Reuse, required=True)
+
+
+class ActivityAPITest:
+    modules = []
+
+    def test_activity_api_list(self, api):
```
bolinocroustibat, line 31: Just for my personal culture, you decided not to type hint the tests functions? I'm fine with either option, just so I can comply to some team rule :)

bolinocroustibat (approved):
LGTM

udata/tests/api/test_activities_api.py
magopian: I just forgot, thanks a lot for the reminder! 🙏

udata/core/activity/api.py
```diff
@@ -61,6 +61,12 @@
     help="Filter activities for that particular organization",
     location="args",
 )
+activity_parser.add_argument(
+    "related_to",
+    type=str,
+    help="Filter activities for that particular reuse or dataset id",
```
maudetes, line 67: It can currently also be organization. But in any case, it is generic and could be applied to dataservices & co in the future.
```suggestion
    help="Filter activities for that particular object id (ex : reuse, dataset, etc.)",
```

udata/core/activity/api.py
```diff
@@ -79,6 +85,9 @@ def get(self):
         if args["user"]:
             qs = qs(actor=args["user"])
 
+        if args["related_to"]:
+            qs = qs(related_to=args["related_to"])
```
maudetes, line 89: We may want to test for a valid ObjectId (ex using `ObjectId.is_valid`) to prevent an internal server error.

maudetes (approved):
Thank you for this param and the tests! :clap:

maudetes:
> Warning: the [documented return type of the activity list API endpoint](https://github.com/opendatateam/udata/blob/master/udata/core/activity/api.py#L70) seems wrong: it says that it returns a list of pagers, but it returns a pager (not a list). Is it related to the legacy admin not being able to handle pagination? I'm not sure if it's safe to fix?

Well spotted! I think the documentation is wrong, and we can probably fix it without side effects since it does not modify the payload itself.

magopian:
Fixed the PR according to your review (thanks!), what do you think @maudetes ?

maudetes (approved):
I think we're clear! Thank you for the updates :sparkles:

PR #3130, opened 2024-08-21
Use api token for hydra auth

magopian:
Fixes [#1424](https://github.com/datagouv/data.gouv.fr/issues/1424)

### DIFF
--- a/docs/adapting-settings.md
+++ b/docs/adapting-settings.md
@@ -644,7 +644,7 @@ FS_ROOT = '/srv/http/www.data.dev/fs'
 
 ## Resources modifications publishing
 
-Udata may publish modifications to resources, for example to an [hydra](https://github.com/datagouv/hydra)
+udata may notify external services (ex: [hydra](https://github.com/datagouv/hydra)) about resources modification on the platform.
 install, or any other service.
 
 
--- a/docs/adapting-settings.md
+++ b/docs/adapting-settings.md
@@ -641,3 +641,27 @@ FS_ROOT = '/srv/http/www.data.dev/fs'
 [flask-mongoengine-doc]: https://flask-mongoengine.readthedocs.org/
 [authlib-doc]: https://docs.authlib.org/en/latest/flask/2/authorization-server.html#server
 [udata-search-service]: https://github.com/opendatateam/udata-search-service
+
+## Resources modifications publishing
+
+Udata may publish modifications to resources, for example to an [hydra](https://github.com/datagouv/hydra)
+install, or any other service.
+
+
+### PUBLISH_ON_RESOURCE_EVENTS
+
+**default**: `False`
+
+Publish resource events to an external service.
+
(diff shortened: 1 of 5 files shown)


bolinocroustibat (approved):
LGTM, great!

udata/core/dataset/events.py
```diff
@@ -49,7 +49,8 @@ def publish(url, document, resource_id, action):
         "dataset_id": str(document.id),
         "document": resource,
     }
-    r = requests.post(url, json=payload)
+    headers = {"Authorization": f"Bearer {current_app.config['RESOURCES_ANALYSER_API_KEY']}"}
+    r = requests.post(url, json=payload, headers=headers)
```
bolinocroustibat, line 53: For my information, this method is used for all actions (creating, updating and deleting) about a resource on hydra, right?

magopian: yep, it's called by
- `publish_added_resource_message`
- `publish_updated_resource_message`
- `publish_removed_resource_message`
in the same file.

udata/core/dataset/events.py
```diff
@@ -49,7 +49,8 @@ def publish(url, document, resource_id, action):
         "dataset_id": str(document.id),
         "document": resource,
     }
-    r = requests.post(url, json=payload)
+    headers = {"Authorization": f"Bearer {current_app.config['RESOURCES_ANALYSER_API_KEY']}"}
```
maudetes, line 52: I would add a `if current_app.config['RESOURCES_ANALYSER_API_KEY']`

docs/adapting-settings.md
```diff
@@ -641,3 +641,27 @@ FS_ROOT = '/srv/http/www.data.dev/fs'
 [flask-mongoengine-doc]: https://flask-mongoengine.readthedocs.org/
 [authlib-doc]: https://docs.authlib.org/en/latest/flask/2/authorization-server.html#server
 [udata-search-service]: https://github.com/opendatateam/udata-search-service
+
+## Resources modifications publishing
+
+Udata may publish modifications to resources, for example to an [hydra](https://github.com/datagouv/hydra)
```
maudetes, line 647:
```suggestion
udata may notify external services (ex: [hydra](https://github.com/datagouv/hydra)) about resources modification on the platform.
```

maudetes (approved):
Looks good! I've added minor comments :)

udata/core/dataset/events.py
magopian: Sure!
I'm now wondering if we should have the `RESOURCES_ANALYSER_API_KEY` be `""` in the settings.py file... because right now it's set to `api_key_to_change`.

So unless someone sets it on purpose to `""`, they'll always send the auth header.

What do you think @maudetes?

maudetes: Yes indeed, agreed for changing it to an empty value!

PR #3131, opened 2024-08-22
Limit the number of user suggestions

magopian:
Fixes [#1425](https://github.com/datagouv/data.gouv.fr/issues/1425)

This adds a validation on the `size` parameter (must be between 1 and 100), and on the `q` parameter (must be at least two characters long).

### DIFF
--- a/udata/core/user/api.py
+++ b/udata/core/user/api.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from flask_security import current_user, logout_user
 from slugify import slugify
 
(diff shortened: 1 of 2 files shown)


ThibaudDauce (approved):
Have you try with the frontend to see if no queries are done for 1 char suggest? Because if the frontend send 1 char requests it will spam 400 errors?

magopian:
> Have you try with the frontend to see if no queries are done for 1 char suggest? Because if the frontend send 1 char requests it will spam 400 errors?

It does return a 400 error that I can see in the network console of my browser, but no apparent error or failure in the admin itself, not sure if it's an issue?

If we want to fix that, I guess there's two obvious solutions:
1/ revert the limit on the query length
2/ return an empty list if the query is too short (less than 2 characters)

maudetes:
Hmm to prevent breaking current behavior, I would go with  1/ revert the limit on the query length, but limit the `size` to 20 maximum to compensate ? I don't think there would be any valid usage for greater size on a suggest endpoint.

magopian:
Deal! I've reverted the commit on adding the restriction on the query size, and reduced the suggest size to 20 max instead of 100 👍

maudetes (approved):
Neat :ok_hand:

PR #3132, opened 2024-08-22
Always add Vary even for non CORS requests

ThibaudDauce:
If the first request doesn't have an `Origin` we don't set the `Vary: Origin` so Nginx put the response in cache. Then if a cross origin request the same URL, Nginx returns the same response without the `Access-Control-Allow-Origin`.

The other solution could be to always return `Access-Control-Allow-Origin: *` for all API route instead of dynamic responding with the `Access-Control-Allow-Origin: doc.data.gouv.fr` for each origin.


maudetes (approved):
Thank you!

Maybe a naive question, but if we set `Vary: Origin` every time, we won't have any Nginx cache anymore on api route right? Would that lead to performance issues? And was it the case previously with flask cors?

ThibaudDauce:
No it just indicate to Nginx to cache by the origin header. So if two request with the same `Origin` header are sent, Nginx is allowed to reuse the cached response (as I understand it)

maudetes:
Perfect, then! Thank you

PR #3133, opened 2024-08-22
Add last login at to org members API

ThibaudDauce:
See https://github.com/datagouv/data.gouv.fr/issues/1411

### DIFF
--- a/udata/core/organization/api_fields.py
+++ b/udata/core/organization/api_fields.py
@@ -66,12 +66,12 @@ member_user_with_email_fields = api.inherit(
     {
         "email": fields.Raw(
             attribute=lambda o: o.email if check_can_access_user_private_info() else None,
-            description="The user email (only present on show organization endpoint if the current user has edit permission on the org)",
+            description="The user email (only present on show organization endpoint if the current user is member of the organization: admin or editor)",
             readonly=True,
         ),
         "last_login_at": fields.Raw(
             attribute=lambda o: o.last_login_at if check_can_access_user_private_info() else None,
-            description="The user last connection date (only present on show organization endpoint if the current user is member of the organization)",
+            description="The user last connection date (only present on show organization endpoint if the current user is member of the organization: admin or editor)",
             readonly=True,
         ),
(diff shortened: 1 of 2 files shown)


udata/core/organization/api_fields.py
```diff
@@ -65,10 +65,15 @@ def check_can_access_email():
...
     {
         "email": fields.Raw(
-            attribute=lambda o: o.email if check_can_access_email() else None,
+            attribute=lambda o: o.email if check_can_access_user_private_info() else None,
             description="The user email (only present on show organization endpoint if the current user has edit permission on the org)",
             readonly=True,
         ),
+        "last_login_at": fields.Raw(
+            attribute=lambda o: o.last_login_at if check_can_access_user_private_info() else None,
+            description="The user last connection date (only present on show organization endpoint if the current user has edit permission on the org)",
```
maudetes, line 74: I think all members have access (same for email)
```suggestion
            description="The user last connection date (only present on show organization endpoint if the current user is member of the organization)",
```

maudetes (approved):
Nice, thank you! :+1:

udata/core/organization/api_fields.py
ThibaudDauce: The check is on `OrganizationAdminNeed(org.id), OrganizationEditorNeed(org.id)`, is it the edit permission? Or any member is `Editor`?

maudetes: Any member is `Editor` indeed (poorly named indeed)!
There is a dedicated `EditOrganizationPermission` with `OrganizationAdminNeed` only need.

ThibaudDauce: Do we want to change it to allow admin only access?

maudetes: I think going with `Editor` is a good idea for now!

maudetes: I think we need to reword the description if we want to go with all members (i.e. `Editor`) :)

PR #3135, opened 2024-08-26
Update authlib from 0.14.3 to 1.3.1

magopian:
This is needed by [udata#482](https://github.com/datagouv/udata-front/pull/482) to change the SSO login button from MonComptePro to AgentConnect

### DIFF
--- a/udata/api/oauth2.py
+++ b/udata/api/oauth2.py
@@ -15,6 +15,7 @@ As well as a sample application:
 """
 
 import fnmatch
+import time
 from datetime import datetime, timedelta
 
 from authlib.integrations.flask_oauth2 import AuthorizationServer, ResourceProtector
@@ -111,7 +112,7 @@ class OAuth2Client(ClientMixin, db.Datetimed, db.Document):
     def check_client_secret(self, client_secret):
         return self.secret == client_secret
 
-    def check_token_endpoint_auth_method(self, method):
+    def check_endpoint_auth_method(self, method, _endpoint):
         if not self.has_client_secret():
             return method == "none"
         return method in ("client_secret_post", "client_secret_basic")
@@ -149,6 +150,9 @@ class OAuth2Token(db.Document):
     def __str__(self):
         return "<OAuth2Token({0.client.name})>".format(self)
 
+    def check_client(self, client):
+        return self.client == client
+
     def get_scope(self):
         return self.scope
 
@@ -160,6 +164,13 @@ class OAuth2Token(db.Document):
 
     def get_client_id(self):
         return str(self.client.id)
+
+    def is_expired(self):
+        now = time.time()
+        return self.get_expires_at() < now
+
+    def is_revoked(self):
+        return self.revoked
 
     def is_refresh_token_valid(self):
         if self.revoked:
@@ -252,17 +263,17 @@ class RefreshTokenGrant(grants.RefreshTokenGrant):
 
 
 class RevokeToken(RevocationEndpoint):
-    def query_token(self, token, token_type_hint, client):
-        qs = OAuth2Token.objects(client=client)
+    def query_token(self, token_string, token_type_hint):
+        qs = OAuth2Token.objects()
         if token_type_hint == "access_token":
-            return qs.filter(access_token=token).first()
+            return qs.filter(access_token=token_string).first()
         elif token_type_hint == "refresh_token":
-            return qs.filter(refresh_token=token).first()
+            return qs.filter(refresh_token=token_string).first()
         else:
-            qs = qs(db.Q(access_token=token) | db.Q(refresh_token=token))
+            qs = qs(db.Q(access_token=token_string) | db.Q(refresh_token=token_string))
             return qs.first()
 
-    def revoke_token(self, token):
+    def revoke_token(self, token, _request):
         token.revoked = True
         token.save()
 
@@ -295,7 +306,7 @@ def revoke_token():
 def authorize(*args, **kwargs):
     if request.method == "GET":
         try:
-            grant = oauth.validate_consent_request(end_user=current_user)
+            grant = oauth.get_consent_grant(end_user=current_user)
         except OAuth2Error as error:
             return error.error
         # Bypass authorization screen for internal clients
@@ -324,13 +335,9 @@ def query_client(client_id):
 
 def save_token(token, request):
     scope = token.pop("scope", "")
-    if request.grant_type == "refresh_token":
-        credential = request.credential
-        credential.update(scope=scope, **token)
-    else:
-        client = request.client
-        user = request.user or client.owner
-        OAuth2Token.objects.create(client=client, user=user.id, scope=scope, **token)
+    client = request.client
+    user = request.user or client.owner
+    OAuth2Token.objects(client=client, user=user, scope=scope, upsert=True, **token)
 
 
 def check_credentials():
--- a/udata/api/oauth2.py
+++ b/udata/api/oauth2.py
@@ -337,7 +337,11 @@ def save_token(token, request):
     scope = token.pop("scope", "")
     client = request.client
     user = request.user or client.owner
-    OAuth2Token.objects(client=client, user=user, scope=scope, upsert=True, **token)
+    if request.grant_type == "refresh_token":
+        old_token = OAuth2Token.objects(client=client, user=user, scope=scope).first()
+        old_token.update(**token)
+    else:
+        OAuth2Token.objects.create(client=client, user=user, scope=scope, **token)
 
 
 def check_credentials():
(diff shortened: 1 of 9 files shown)


magopian:
Oops, I need to update the udata code (and the tests) which are now failing, let me address that before reviewing please

bolinocroustibat (approved):
Great, thanks, LGTM!

maudetes:
I tried a login process locally with an Oauth2 client (publier.etalab.studio) but it doesn't seem to work anymore (return a 401 on a /api/1/me/ call with an `Authorization` header).

Steps to reproduce:
* git clone https://github.com/etalab/publier.etalab.studio/
* `cp .env.sample .env`
* update .env with
```
VUE_APP_OAUTH_BASE_URL=http://dev.local:7000/fr
VUE_APP_OAUTH_CALLBACK=http://localhost:8080/
VUE_APP_DATAGOUV_API_URL=http://dev.local:7000
```
* npm install & npm run serve
* go to http://localhost:8080/fr/
* start a udata env with `AUTHLIB_INSECURE_TRANSPORT=1 inv serve` (on master VS this branch)
* click on login -> gets redirected to dev.local -> accept -> gets redirected to http://localhost:8080/fr/
* look at the `/api/1/me/` being made

magopian:
Thanks a lot @maudetes for the reproduction steps.
It's my bad, the `save_token` code was completely bogus, and didn't actually save the token.
This should now be fixed, and I updated the tests so we don't run into this issue again in the future ;) (regression tests ftw).

I tested locally and all seems to be working as expected. Could you please do the same on your end and make sure everything is fixed?

udata/api/oauth2.py
```diff
@@ -111,7 +112,7 @@ def check_redirect_uri(self, redirect_uri):
     def check_client_secret(self, client_secret):
         return self.secret == client_secret
 
-    def check_token_endpoint_auth_method(self, method):
+    def check_endpoint_auth_method(self, method, _endpoint):
```
maudetes, line 115: Those additional endpoints and renaming were needed for the new version?

udata/api/oauth2.py
```diff
@@ -161,6 +165,13 @@ def get_expires_at(self):
     def get_client_id(self):
         return str(self.client.id)
 
+    def is_expired(self):
```
maudetes, line 168: I'm slightly confused between `is_expired` VS `is_refresh_token_valid` logic

udata/api/oauth2.py
```diff
@@ -295,7 +306,7 @@ def revoke_token():
 def authorize(*args, **kwargs):
     if request.method == "GET":
         try:
-            grant = oauth.validate_consent_request(end_user=current_user)
+            grant = oauth.get_consent_grant(end_user=current_user)
```
maudetes, line 309: does `get_consent_grant` also `validate`?

udata/tests/api/test_auth_api.py
```diff
@@ -582,7 +584,7 @@ def test_invalid_implicit_grant_token(self, client, oauth):
         )
 
         assert_status(response, 400)
-        assert response.json["error"] == "invalid_grant"
+        assert response.json["error"] == "unsupported_response_type"
```
maudetes, line 587: Was this error modification expected?

udata/tests/api/test_auth_api.py
```diff
@@ -606,6 +608,11 @@ def test_refresh_token(self, client, oauth):
         assert200(response)
         assert response.content_type == "application/json"
         assert "access_token" in response.json
+        new_access_token = response.json["access_token"]
+        assert new_access_token != "access-token"  # The access token has been refreshed.
+        tokens = OAuth2Token.objects(client=oauth, user=user)
+        assert len(tokens) == 1  # No new token has been created.
+        assert tokens.first().access_token == new_access_token
```
maudetes, line 615: Thank you for this additional test! :pray:

udata/api/oauth2.py
```diff
@@ -324,13 +335,13 @@ def query_client(client_id):
 
 def save_token(token, request):
     scope = token.pop("scope", "")
+    client = request.client
+    user = request.user or client.owner
     if request.grant_type == "refresh_token":
-        credential = request.credential
-        credential.update(scope=scope, **token)
+        old_token = OAuth2Token.objects(client=client, user=user, scope=scope).first()
```
maudetes, line 341: Shouldn't `refresh_token` param be the identifier? Ex: `OAuth2Token.objects(refresh_token=request.refresh_token).first()` or similar?

maudetes (commented):
It seems to work as expected! Thank you for the update!

Since I am not familiar with authlib, I have some questions about some modifications I am not sure to understand. You don't need to answer in details if not relevant!

udata/api/oauth2.py
magopian: So what happened is when running the tests after the authlib upgrade, I had a bunch of errors. For each of those errors, I tracked down the initial cause, and search the old and new docs, and try to find how the lib was modified.

This is one of those, but an easy one, as it was right there [in the new docs](https://docs.authlib.org/en/v1.3.1/django/2/authorization-server.html#client)

udata/api/oauth2.py
magopian: `is_expired` seems to be a new method that needs to be implemented in 1.3.1, and is related to the token: is the (access) token still good, or is it expired?

The **refresh** token itself needs to be checked to see if it's still valid. It's valid as long as it's not expired (it's longer lived than the access token), AND has not been revoked.

This allows the revoking mecanism to work properly. Without a short lived access token, even if it's revoked on the auth server (eg AgentConnect's server), the client would still have access to the service (eg datagouv) as long as the access token is alive.

Making the access token short lived forces the client to query for a new one every so often, or to refresh it, at which point the revocation is checked.

I'm not sure if I'm being clear here, but I would be glad to explain this in more details if needed.

udata/api/oauth2.py
magopian: I believe this is a simple function renaming between [0.14.3](https://docs.authlib.org/en/v0.14.3/flask/2/api.html?highlight=validate_consent_request#authlib.integrations.flask_oauth2.AuthorizationServer.validate_consent_request) and [1.3.1](https://docs.authlib.org/en/v1.3.1/specs/rfc6749.html#authlib.oauth2.rfc6749.AuthorizationServer.get_consent_grant), as both their documentation are exactly the same.

I double checked, and indeed in the `get_consent_grant` function, there is a call to `validate_content_request`: https://github.com/lepture/authlib/blob/v1.3.1/authlib/oauth2/rfc6749/authorization_server.py#L217

udata/api/oauth2.py
magopian: Very good point indeed! I modified the code, and the tests to make sure we only update this token, and not any other token.

I've even tested that if a user has two tokens (which shouldn't happen?), only the correct one is refreshed.

udata/tests/api/test_auth_api.py
magopian: From what i can see from the source code, the error returned has been changed between [0.14.3](https://github.com/lepture/authlib/blob/v0.14.3/authlib/oauth2/rfc6749/authorization_server.py#L142-L151)  and [1.3.1](https://github.com/lepture/authlib/blob/v1.3.1/authlib/oauth2/rfc6749/authorization_server.py#L198-L207) in the same `get_authorization_grant` function.

udata/api/oauth2.py
```diff
@@ -324,13 +335,15 @@ def query_client(client_id):
...
     scope = token.pop("scope", "")
+    client = request.client
+    user = request.user or client.owner
     if request.grant_type == "refresh_token":
-        credential = request.credential
-        credential.update(scope=scope, **token)
+        old_token = OAuth2Token.objects(
+            refresh_token=request.refresh_token.refresh_token, client=client, user=user, scope=scope
+        ).first()
+        old_token.update(**token)
```
maudetes, line 344: Reading it again, should we return an HTTP error check if old_token is None?

udata/api/oauth2.py
maudetes: Yes I think I get the idea now, thank you! :pray:

maudetes (approved):
Thank you for all the detailed and sourced explanations!

I've added a suggestion, but else I think we're good to go!

udata/api/oauth2.py
magopian: Slightly unrelated: I went back through the documentation, and saw [this](https://docs.authlib.org/en/latest/flask/2/grants.html#refresh-token-grant):
> By default, RefreshTokenGrant will not issue a refresh_token in the token response. Developers can change this behavior with:
>
> class RefreshTokenGrant(grants.RefreshTokenGrant):
>     INCLUDE_NEW_REFRESH_TOKEN = True

And I believe this is a good idea, so the refresh token is refreshed each and every time the token is refreshed.

More related to your point: I haven't seen anywhere in the code of the lib that would use the return of this function, and I feel reluctant to possibly change the behaviour.

What eases my mind is that we should never reach a state where we're trying to save a token that has been previously validated and accepted through `RefreshTokenGrant.authenticate_refresh_token`.

If my explanation has convinced you, let's land this, and if not, I'd be happy to dig further in this issue ;)

PR #3136, opened 2024-08-29
Fix fake tags min length (again)

magopian:
This follows https://github.com/opendatateam/udata/pull/3129 which wasn't fixing all the instances of `faker.word(s)` being used instead of `faker.tag(s)`.

### DIFF
--- a/udata/tests/frontend/test_csv.py
+++ b/udata/tests/frontend/test_csv.py
@@ -59,7 +59,7 @@ class FakeFactory(MongoEngineFactory):
 
     title = factory.LazyAttribute(lambda o: faker.sentence())
     description = factory.LazyAttribute(lambda o: faker.paragraph())
-    tags = factory.LazyAttribute(lambda o: [faker.word() for _ in range(1, randint(1, 4))])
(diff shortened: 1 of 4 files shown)


udata/tests/frontend/test_csv.py
```diff
@@ -59,7 +59,7 @@ class Meta:
 
     title = factory.LazyAttribute(lambda o: faker.sentence())
     description = factory.LazyAttribute(lambda o: faker.paragraph())
-    tags = factory.LazyAttribute(lambda o: [faker.word() for _ in range(1, randint(1, 4))])
+    tags = factory.LazyAttribute(lambda o: [faker.tag() for _ in range(1, randint(1, 4))])
```
ThibaudDauce, line 62: Why not use `faker.tags()` here?

ThibaudDauce (approved):
:+1:

udata/tests/frontend/test_csv.py
magopian: I didn't want to change the code, but I sure could, let me do that quickly ;)

PR #3140, opened 2024-09-03
Add a "private" filter on the reuses endpoint

magopian:
This is a new take on [#1397](https://github.com/datagouv/data.gouv.fr/issues/1397)

In this PR, we'll have the following behavior:
1. always return each and every reuse available to the logged in user (if any), including private, deleted and archived ones
2. if the filter is false, eg `private=False`, only return the reuses that don't have this filter (no private reuses)
3. if the filter is true, eg `private=True`, only return the reuses that have this filter (only private reuses)

### DIFF
--- a/udata/api_fields.py
+++ b/udata/api_fields.py
@@ -305,8 +305,6 @@ def generate_fields(**kwargs):
             parser.add_argument(
                 filterable["key"],
                 type=filterable["type"],
-                default=filterable.get("default"),
-                choices=filterable.get("choices"),
                 location="args",
             )
 
--- a/udata/api_fields.py
+++ b/udata/api_fields.py
@@ -302,7 +302,13 @@ def generate_fields(**kwargs):
             parser.add_argument("q", type=str, location="args")
 
         for filterable in filterables:
-            parser.add_argument(filterable["key"], type=filterable["type"], location="args")
+            parser.add_argument(
+                filterable["key"],
+                type=filterable["type"],
+                default=filterable.get("default"),
+                choices=filterable.get("choices"),
+                location="args",
+            )
 
         cls.__index_parser__ = parser
 
--- a/udata/api_fields.py
+++ b/udata/api_fields.py
@@ -209,6 +209,8 @@ def generate_fields(**kwargs):
                     filterable["type"] = str
                     if isinstance(field, mongo_fields.BooleanField):
                         filterable["type"] = boolean
+                if filterable["type"] == boolean and isinstance(field, mongo_fields.DateTimeField):
+                    filterable["datetime_as_boolean"] = True
 
                 # We may add more information later here:
                 # - type of mongo query to execute (right now only simple =)
@@ -342,11 +344,19 @@ def generate_fields(**kwargs):
                         ):
                             api.abort(400, f'`{filterable["key"]}` must be an identifier')
 
-                    base_query = base_query.filter(
-                        **{
-                            filterable["column"]: args[filterable["key"]],
-                        }
-                    )
+                    filter_value = args[filterable["key"]]
+                    filter_column = filterable["column"]
+
+                    if filterable.get("datetime_as_boolean"):
+                        # It's a datetime field, but we filter it as a bool, so we convert
+                        # - True -> field__ne=None (objects which have a `deleted` datetime set)
+                        # - False -> field=None (objects without a `deleted` datetime set)
+                        if filter_value:
+                            filter_column += "__ne"
+                        filter_value = None
+
+                    filter = {filter_column: filter_value}
+                    base_query = base_query.filter(**filter)
 
             if paginable:
                 base_query = base_query.paginate(args["page"], args["page_size"])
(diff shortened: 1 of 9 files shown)


ThibaudDauce (approved):
Nice!

udata/core/reuse/models.py
```diff
@@ -23,6 +25,20 @@ class ReuseQuerySet(OwnedQuerySet):
     def visible(self):
         return self(private__ne=True, datasets__0__exists=True, deleted=None)
 
+    def visible_by_user(self, user: User) -> OwnedQuerySet:
+        """Return EVERYTHING visible to the user."""
+        public_qs: OwnedQuerySet = db.Q(private__ne=True, deleted=None, archived=None)
```
maudetes, line 30: Could we use `visible`?
```suggestion
        public_qs: OwnedQuerySet = db.Q(private__ne=True, deleted=None, archived=None)
```

udata/api_fields.py
```diff
@@ -302,7 +302,13 @@ def make_lambda(method):
             parser.add_argument("q", type=str, location="args")
 
         for filterable in filterables:
-            parser.add_argument(filterable["key"], type=filterable["type"], location="args")
+            parser.add_argument(
+                filterable["key"],
+                type=filterable["type"],
+                default=filterable.get("default"),
+                choices=filterable.get("choices"),
```
maudetes, line 309: Do we need these additions for this PR?

udata/core/reuse/models.py
```diff
@@ -23,6 +25,20 @@ class ReuseQuerySet(OwnedQuerySet):
...
+        if user.is_anonymous:
+            return self(public_qs)
+
+        owners: list[User | Organization] = []
+        owned_qs: OwnedQuerySet = db.Q()
+        owners = list(user.organizations) + [user.id]
+        for owner in owners:
+            owned_qs |= db.Q(owner=owner) | db.Q(organization=owner)
```
maudetes, line 38: Couldn't we use [owned_by](https://github.com/opendatateam/udata/blob/c42e85883d5f796fe01ebb5acad8de979a4d040f/udata/core/owned.py#L21)?

maudetes (approved):
Thank you! I have minor suggestions/questions but looks good overall

udata/core/reuse/models.py
bolinocroustibat, line 38: Can the DB use `user` instead of `owner`, so that the logic could be like `owned_qs |= db.Q(user=owner) | db.Q(organization=owner)`?

bolinocroustibat (approved):
LGTM, with a question!

udata/core/reuse/models.py
magopian: We could! It's a bit more work, but I believe it's less duplication this way, and I even moved the `visible_by_user` method to the `OwnedQuerySet` which is used by many models, so it'll be easier to reuse (pun not intended) in the future.

udata/core/reuse/models.py
magopian: Same answer as above, we reuse it in `OwnedQuerySet.visible_by_user`

magopian: I'm not sure I understand your point? `owner` is a field on the `Reuse` model, but `user` is not, what are you suggesting?

udata/core/reuse/api.py
```diff
@@ -99,8 +99,7 @@ class ReuseListAPI(API):
     @api.expect(Reuse.__index_parser__)
     @api.marshal_with(Reuse.__page_fields__)
     def get(self):
-        query = Reuse.objects(deleted=None, private__ne=True)
-
+        query = Reuse.objects.visible_by_user(current_user)
```
maudetes, line 102: Woops, we're changing the behavior since using `visible` that now excludes reuses without datasets.
Is that the reason why visible wasn't used there before?

In any case, I think we want to keep including reuses without datasets here

magopian: Indeed, and after some investigation, we don't want to return reuses with empty datasets in the `visible()` queryset, as it would break some other places in the code.
I've pushed some partial reverts and rework.

udata/core/owned.py
```diff
@@ -24,6 +24,19 @@ def owned_by(self, *owners):
             qs |= Q(owner=owner) | Q(organization=owner)
         return self(qs)
 
+    def visible_by_user(self, user: User, visible_query: Q):
+        """Return EVERYTHING visible to the user."""
```
maudetes, line 28: I think we may want to add an admin case.
```
if user.sysadmin:
    return self()
```

maudetes (approved):
Thank you! A late suggestion for sysadmin

udata/core/owned.py
magopian: Nice addition, indeed. I added it and the test that goes with it 👍

maudetes: Thank you for the fix when user isn't authenticated! :p

maudetes (approved):

PR #3141, opened 2024-09-03
Fix undelete reuses and dataservices

ThibaudDauce:


udata/tests/api/test_dataservices_api.py
```diff
@@ -101,8 +101,29 @@ def test_dataservice_api_create(self):
         self.assertEqual(dataservice.base_api_url, "https://example.org")
         self.assertIsNotNone(dataservice.deleted_at)
 
-        # response = self.get(url_for('api.dataservice', dataservice=dataservice))
-        # self.assert410(response)
```
maudetes, line 105: I don't see where you've modified this in this PR?

maudetes (approved):
Thank you for the fix! :pray:

udata/tests/api/test_dataservices_api.py
ThibaudDauce: It was already working (only with a new user logged in, so I added it back)

maudetes: Oh my bad, I hadn't see it was commented! Thank you :pray:

Issue #1426, opened 2018-02-07
The ressource markdown description


damonzon33:
Hi
When creating a dataset (remote file), the ressource markdown description is not displayed correctly. The description is displayed correctly only during the preview. See screenshot.

![image](https://user-images.githubusercontent.com/30826928/35920560-78c652ec-0c18-11e8-8b12-bb2b758da219.png)

![image](https://user-images.githubusercontent.com/30826928/35920652-b3624aaa-0c18-11e8-9c19-e6d6b0432b4e.png)

![image](https://user-images.githubusercontent.com/30826928/35920864-649bc968-0c19-11e8-9607-c52d4c1ae04f.png)

Udata 1.2.10

Test on data.gouv.fr (1.2.11) same issue

Thanks for your support

noirbizarre:
Yes, this is a known issue and this will be fixed soon.
We are currently refactoring the ressources handling and display, this issue will be fixed in the process.

damonzon33:
thanks

noirbizarre:
Fixed by #1547

PR #3142, opened 2024-09-04
Use restful hydra enpoint URLs

magopian:
Fixes [#1426](https://github.com/datagouv/data.gouv.fr/issues/1426)

### DIFF
--- a/udata/tests/dataset/test_dataset_events.py
+++ b/udata/tests/dataset/test_dataset_events.py
@@ -31,7 +31,7 @@ class DatasetEventsTest:
         }
 
         mock_req.assert_called_with(
-            f"{current_app.config['RESOURCES_ANALYSER_URI']}/api/resource/created/",
+            f"{current_app.config['RESOURCES_ANALYSER_URI']}/api/resources/",
             json=expected_value,
             headers={},  # No RESOURCES_ANALYSER_API_KEY, no headers.
         )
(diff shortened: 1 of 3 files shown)


udata/tests/dataset/test_dataset_events.py
```diff
@@ -77,7 +77,7 @@ def test_publish_message_resource_modified(self, mock_req):
             dataset.update_resource(resource)
 
         mock_req.assert_called_with(
-            f"{current_app.config['RESOURCES_ANALYSER_URI']}/api/resource/updated/",
+            f"{current_app.config['RESOURCES_ANALYSER_URI']}/api/resources/",
```
bolinocroustibat, line 80: For information, it's not part of this issue but at some point I'd like to make the _routes_ really REST, meaning on this call it would be a `PUT` on `/api/resources/{resource_id}`, with still the same payload.

Does that make sense to you?

Either we prepare it in this PR, either in another one, as you wish.

udata/tests/dataset/test_dataset_events.py
```diff
@@ -104,7 +104,7 @@ def test_publish_message_resource_removed(self, mock_req):
             dataset.remove_resource(resource)
 
         mock_req.assert_called_with(
-            f"{current_app.config['RESOURCES_ANALYSER_URI']}/api/resource/deleted/",
+            f"{current_app.config['RESOURCES_ANALYSER_URI']}/api/resources/",
```
bolinocroustibat, line 107: For information, it's not part of this issue but at some point I'd like to make the _routes_ really REST, meaning on this call it would be a `DELETE` on `/api/resources/{resource_id}`, with no payload.

Does that make sense to you?

Either we prepare it in this PR, either in another one, as you wish.

bolinocroustibat (approved):
LGTM, with a few remarks about resources analyser's side that would have impact on udata.

Another remark: trailing slash on resource analyser URLs will soon be non mandatory. I'm not sure which is is the more RESTful though, trailing slash or not, but you can adapt whichever you prefer in this PR.

udata/tests/dataset/test_dataset_events.py
magopian: I think it would make sense to do this in two steps to ease the "synchronization" of the PRs in udata and hydra. So i think i'd rather merge this PR as is, if it looks good to you, and then create two new linked issues in udata and hydra, that'll be prioritized and fixed later.

bolinocroustibat: Yes I agree!

bolinocroustibat:
@magopian Is this on production? If yes, maybe we can remove the [legacy endpoints](https://github.com/datagouv/hydra/blob/ecbe288c115284026288ca60d6ff607815867548/udata_hydra/routes/__init__.py#L89) on Hydra, what do you think?

PR #3147, opened 2024-09-12
Add the "org" and "user" filter on the `/discussions` endpoint

magopian:
Fixes [#1398](https://github.com/datagouv/data.gouv.fr/issues/1398)

### DIFF
--- a/udata/core/discussions/api.py
+++ b/udata/core/discussions/api.py
@@ -6,7 +6,10 @@ from flask_security import current_user
 
 from udata.api import API, api, fields
 from udata.auth import admin_permission
+from udata.core.dataservices.models import Dataservice
+from udata.core.dataset.models import Dataset
 from udata.core.organization.models import Organization
+from udata.core.reuse.models import Reuse
 from udata.core.spam.api import SpamAPIMixin
 from udata.core.spam.fields import spam_fields
 from udata.core.user.api_fields import user_ref_fields
@@ -211,7 +214,11 @@ class DiscussionsAPI(API):
             discussions = discussions.generic_in(subject=args["for"])
         if args["org"]:
             org = Organization.objects.get(id=ObjectId(args["org"]))
-            discussions = discussions(subject=org)
+            reuses = Reuse.objects(organization=org).only("id")
+            datasets = Dataset.objects(organization=org).only("id")
+            dataservices = Dataservice.objects(organization=org).only("id")
+            subjects = list(reuses) + list(datasets) + list(dataservices)
+            discussions = discussions(subject__in=subjects).order_by("-created")
         if args["user"]:
             discussions = discussions(discussion__posted_by=ObjectId(args["user"]))
         if args["closed"] is False:
--- a/udata/core/discussions/api.py
+++ b/udata/core/discussions/api.py
@@ -6,6 +6,7 @@ from flask_security import current_user
 
 from udata.api import API, api, fields
 from udata.auth import admin_permission
+from udata.core.organization.models import Organization
 from udata.core.spam.api import SpamAPIMixin
 from udata.core.spam.fields import spam_fields
 from udata.core.user.api_fields import user_ref_fields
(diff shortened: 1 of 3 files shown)


udata/tests/test_discussions.py
```diff
@@ -358,6 +362,44 @@ def test_list_discussions_for(self):
 
         self.assertEqual(len(response.json["data"]), len(discussions))
 
+    def test_list_discussions_org(self) -> None:
+        organization: Organization = OrganizationFactory()
+        user: User = UserFactory()
+        _discussion: Discussion = DiscussionFactory(user=user)
+        discussion_for_org: Discussion = DiscussionFactory(subject=organization, user=user)
```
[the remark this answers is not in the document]
magopian: oh, right, I didn't realize that! I'll update the code and tests, thanks 🙏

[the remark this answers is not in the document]
magopian: It's now listing discussions related to the org's datasets, dataservices and reuses. What do you think?

udata/core/discussions/api.py
```diff
@@ -198,6 +212,13 @@ def get(self):
         discussions = Discussion.objects
         if args["for"]:
             discussions = discussions.generic_in(subject=args["for"])
+        if args["org"]:
+            org = Organization.objects.get(id=ObjectId(args["org"]))
+            reuses = Reuse.objects(organization=org).only("id")
+            datasets = Dataset.objects(organization=org).only("id")
+            dataservices = Dataservice.objects(organization=org).only("id")
+            subjects = list(reuses) + list(datasets) + list(dataservices)
+            discussions = discussions(subject__in=subjects).order_by("-created")
```
maudetes, line 221: I don't think we want to apply order here?
```suggestion
            discussions = discussions(subject__in=subjects)
```

udata/tests/test_discussions.py
```diff
@@ -358,6 +364,57 @@ def test_list_discussions_for(self):
...
+        response: TestResponse = self.get(url_for("api.discussions", org=organization.id))
+        self.assert200(response)
+        self.assertEqual(len(response.json["data"]), 3)
+        self.assertIdIn(response.json["data"], str(discussion_for_dataset.id))
+        self.assertIdIn(response.json["data"], str(discussion_for_dataservice.id))
+        self.assertIdIn(response.json["data"], str(discussion_for_reuse.id))
+
+    def test_list_discussions_sort(self) -> None:
+        user: User = UserFactory()
+        sorting_keys_dict: dict = {
```
maudetes, line 393: Was it too cumbersome to parametrize the tests?
It would additionally prevent cleaning the discussions manually

udata/core/discussions/api.py
```diff
@@ -198,6 +212,13 @@ def get(self):
         discussions = Discussion.objects
         if args["for"]:
             discussions = discussions.generic_in(subject=args["for"])
+        if args["org"]:
+            org = Organization.objects.get(id=ObjectId(args["org"]))
```
maudetes, line 216: We may want to fail properly if `org` is not an ObjectId

maudetes (approved):
Thank you! I have minor comments, but it seems to work as expected

udata/tests/test_discussions.py
magopian: It's not a matter of being cumbersome, but a matter of being in a file and class that doesn't use pytest, but rather unittest, and pytest's parametrization [isn't compatible](https://docs.pytest.org/en/latest/how-to/unittest.html#pytest-features-in-unittest-testcase-subclasses) :/

I didn't want to create another class that would test the same stuff, but that wouldn't inherit from unittest, as it seemed a bit clunky to me... What do you think?

Maybe I could rewrite the `DiscussionsTest` class to not use unittest, maybe in another PR?

maudetes: Ahah I think it is the kind of things I meant behind cumbersome.

Indeed, it would be the best to make these tests not use unittest in an upcoming PR :)
Maybe done class by class or by updating `TestCase` directly.

PR #3152, opened 2024-09-24
Move discussion catalog creation and add fields

Pierlou:
Linked to https://github.com/datagouv/udata-front/pull/543


udata/core/dataset/csv.py
```diff
@@ -1,4 +1,3 @@
-from udata.core.discussions.models import Discussion
 from udata.frontend import csv
```
maudetes, line 1: You could actually import `DiscussionCsvAdapter` here for backward compatibility

udata/core/discussions/tasks.py
```diff
@@ -64,3 +73,112 @@ def notify_discussion_closed(discussion_id, message=None):
...
+        if created:
+            dataset.add_resource(resource)
+        dataset.last_modified_internal = datetime.utcnow()
+        dataset.save()
+    finally:
+        csvfile.close()
+        os.unlink(csvfile.name)
+
+
+@job("export-csv")
```
maudetes, line 161: I don't think you should copy any of `export-csv` logic and helpers.
It is the job to export csv for all models, discussion included.

It may not be located at the most logic place right now, but I am unsure where we could move it

udata/core/discussions/csv.py
```diff
@@ -0,0 +1,26 @@
...
+        "user",
+        "subject",
+        ("subject_class", "subject.class"),
+        ("subject_id", "subject.id"),
+        "title",
+        ("size", lambda o: len(o.discussion)),
+        (
+            "participants",
+            lambda o: ",".join(
+                getattr(msg.posted_by, "class") + "_" + msg.posted_by.id for msg in o.discussion
```
maudetes, line 19: Why using a `getattr`? For now posted_by is only a user, I'm unsure about the way it would evolve if it can be an Organization, but I think you should not anticipate this here yet.

Maybe add a comment on https://github.com/datagouv/data.gouv.fr/issues/1410 to mention to update the adapter at that time?

maudetes (commented):
Thank you for updating the discussion adapter!

udata/core/discussions/csv.py
Pierlou: When using the usual form `msg.posted_by.class` my editor displayed an error because `class` was recognized as a reserved keyword (but I am not sure it would have failed actually). I'll only get the ids, as indeed we only have users for now

udata/core/dataset/csv.py
Pierlou: ruff doesn't like that it is imported but unused 🙃

udata/core/dataset/csv.py
```diff
@@ -1,4 +1,4 @@
-from udata.core.discussions.models import Discussion
+from udata.core.discussions.csv import DiscussionCsvAdapter  # noqa: F401
```
maudetes, line 1: Could you add a comment that it's for backward compatibility (maybe with a link to this PR?)

udata/core/discussions/csv.py
```diff
@@ -0,0 +1,21 @@
...
+class DiscussionCsvAdapter(csv.Adapter):
+    fields = (
+        "id",
+        "user",
+        "subject",
+        ("subject_class", "subject.class"),
+        ("subject_id", "subject.id"),
+        "title",
+        ("size", lambda o: len(o.discussion)),
+        ("participants", lambda o: ",".join(set(msg.posted_by.id for msg in o.discussion))),
```
maudetes, line 16: I think ObjectId needs to be casted to str
```suggestion
        ("participants", lambda o: ",".join(set(str(msg.posted_by.id) for msg in o.discussion))),

```

udata/core/discussions/csv.py
```diff
@@ -0,0 +1,21 @@
...
+from .models import Discussion
+
+
+@csv.adapter(Discussion)
+class DiscussionCsvAdapter(csv.Adapter):
+    fields = (
+        "id",
+        "user",
+        "subject",
+        ("subject_class", "subject.class"),
```
maudetes, line 12:
```suggestion
        ("subject_class", "subject._class_name"),

```

udata/core/discussions/csv.py
```diff
@@ -0,0 +1,21 @@
...
+        ("subject_id", "subject.id"),
+        "title",
+        ("size", lambda o: len(o.discussion)),
+        ("participants", lambda o: ",".join(set(msg.posted_by.id for msg in o.discussion))),
+        ("messages", lambda o: "\n".join(msg.content.replace("\n", " ") for msg in o.discussion)),
+        "created",
+        "closed",
+        "closed_by",
```
maudetes, line 20: It may be worth noting that `closed_by` currently shows the user name (and not the id as in participants). It seems more user-friendly but less automatisation friendly.

maudetes (approved):
I've added some upgrades on the new fields

udata/core/discussions/csv.py
Pierlou: I had not noticed (and I was not expecting that just mentioning `closed_by` would pick the user name, as it is an object). We could change it for `("closed_by", "closed_by.id"),`, what do you think?

maudetes: I think it comes from https://github.com/opendatateam/udata/blob/5f54568240ed430726bb68e0a307415e9870aa8e/udata/core/user/models.py#L105

maudetes: And yes, let's split the closed by name and id

udata/core/discussions/csv.py
```diff
@@ -0,0 +1,22 @@
...
+        "subject",
+        ("subject_class", "subject._class_name"),
+        ("subject_id", "subject.id"),
+        "title",
+        ("size", lambda o: len(o.discussion)),
+        ("participants", lambda o: ",".join(set(str(msg.posted_by.id) for msg in o.discussion))),
+        ("messages", lambda o: "\n".join(msg.content.replace("\n", " ") for msg in o.discussion)),
+        "created",
+        "closed",
+        ("closed_by_fullname", "closed_by"),
```
maudetes, line 20: Woops, coming after the battle but `closed_by_fullname` breaks compatibility. I think we can keep `closed_by` and `closed_by_id`, similarly to what is used in other csv adapters, ex: `organization` (name) and `organization_id`.

Pierlou: https://github.com/opendatateam/udata/pull/3153

PR #3156, opened 2024-09-30
Expose dataservices' datasets by link instead of list

ThibaudDauce:
Related to https://github.com/datagouv/data.gouv.fr/issues/1467


udata/core/dataset/api.py
```diff
@@ -148,6 +150,15 @@ def parse_filters(datasets, args):
                 pass
             else:
                 datasets = datasets.filter(id__in=[d.id for d in topic.datasets])
+        if args.get("dataservice"):
+            if not ObjectId.is_valid(args["dataservice"]):
+                api.abort(400, "Dataservice arg must be an identifier")
+            try:
+                dataservice = Dataservice.objects.get(id=args["dataservice"])
+            except Dataservice.DoesNotExist:
+                pass
```
magopian, line 159: Should we abort 404 instead of passing?

udata/tests/api/test_dataservices_api.py
```diff
@@ -61,18 +61,19 @@ def test_dataservice_api_create(self):
...
-            },
-        )
+        self.assertEqual(response.json["extras"], {"foo": "bar"})
         self.assertEqual(response.json["license"], license.id)
         self.assertEqual(
             response.json["self_api_url"], "http://local.test/api/1/dataservices/updated-title/"
         )
+
+        self.assertEqual(response.json["datasets"]["total"], 2)
+        response_datasets = self.get(response.json["datasets"]["href"])
```
magopian, line 71: Is it on purpose that you test the "by link" feature on the test PATCH response instead of a POST? Anything special about the PATCH?
Would it make sense to also test it on the POST response?

udata/tests/api/test_dataservices_api.py
```diff
@@ -91,6 +92,26 @@ def test_dataservice_api_create(self):
...
+
+        response = self.delete(
+            url_for("api.dataservice_dataset", dataservice=dataservice, dataset=datasets[0])
+        )
+        self.assert204(response)
+        dataservice.reload()
+        self.assertEqual(len(dataservice.datasets), 2)
+        self.assertEqual(dataservice.datasets[0].title, datasets[2].title)
+        self.assertEqual(dataservice.datasets[1].title, datasets[1].title)
+
```
magopian, line 114: Is this related to the PR, adding the `by_link` feature?

udata/core/dataservices/api.py
```diff
@@ -95,6 +97,82 @@ def delete(self, dataservice):
...
+        OwnablePermission(dataservice).test()
+
+        if dataset.id not in (d.id for d in dataservice.datasets):
+            api.abort(404, "Dataset not found in dataservice")
+        dataservice.datasets = [d for d in dataservice.datasets if d.id != dataset.id]
+        dataservice.save()
+
+        return None, 204
+
+
```
magopian, line 175: I'm a bit confused, is this related to the PR, adding the `by_link` feature?

ThibaudDauce: Yes, we add two new endpoint to manipulate the list without sending it (the `by_link` feature allows to not return during reads the list of datasets, but to modify the list we need to have an endpoint to add or remove an item of the list without sending the full list of thousands of datasets)

udata/core/dataset/api.py
ThibaudDauce: It's copy/paste from the topic logic above. We could change this in another PR if it's not wanted…

udata/tests/api/test_dataservices_api.py
ThibaudDauce: It could be tested on either one, since both requests returns the same object structure. I tested it here because it was here that I tested that the datasets are correctly modified.

udata/tests/api/test_dataservices_api.py
ThibaudDauce: It's a similar fix, the `by_link` feature "requires" to allow to add/remove item alone without sending the list

udata/tests/api/test_dataservices_api.py
magopian: Understood thanks, it makes sense, thanks for the explanation 🙏

udata/tests/api/test_dataservices_api.py
magopian: oooooooh, right, i understand the link now, makes sense, thanks! 🙏

udata/core/dataservices/models.py
```diff
@@ -181,6 +182,9 @@ class Dataservice(WithMetrics, Owned, db.Document):
         filterable={
             "key": "dataset",
         },
+        by_link=lambda dataservice: url_for(
```
maudetes, line 185: Should we name it `href` instead of `by_link`? Since it appears as a `href` subsection?

udata/api_fields.py
```diff
@@ -87,9 +105,13 @@ def convert_db_to_field(key, field, info):
             f"{key}.inner", field.field, {**info, **inner_info, **info.get("inner_field_info", {})}
         )
 
-        def constructor_read(**kwargs):
-            return restx_fields.List(field_read, **kwargs)
+        if constructor_read is None:
```
maudetes, line 108: can't we replace it by `else`?

maudetes (approved):
Thank you! Seems to work nicely :)

I have very cosmetic suggestions

maudetes:
Testing this PR on dev env with a dataservice linked to 20K datasets does not scale at the moment.
See the [sentry error](https://errors.data.gouv.fr/organizations/sentry/issues/147024/?environment=dev.data.gouv.fr&query=is%3Aunresolved&referrer=issue-stream) when accessing https://dev.data.gouv.fr/api/1/dataservices/api-tabulaire-data-gouv-fr-beta/ (:warning: only 1000 associated datasets have been kept to dataservice now).
I think we could try using `db.LazyReferenceField` as in the [topic model](https://github.com/opendatateam/udata/pull/2924)?

udata/core/dataservices/api.py
```diff
@@ -95,6 +97,82 @@ def delete(self, dataservice):
...
+    doc=common_doc,
+)
+class DataserviceDatasetAPI(API):
+    @api.secure
+    @api.response(404, "Dataservice not found")
+    @api.response(404, "Dataset not found in dataservice")
+    def delete(self, dataservice, dataset):
+        if dataservice.deleted_at and not (
+            # Allow requests containing "deleted_at: None" to undelete.
+            "deleted_at" in request.json and request.json.get("deleted_at") is None
```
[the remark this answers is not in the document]
ThibaudDauce: Fixed in 03757ee38

udata/core/dataservices/models.py
ThibaudDauce: Changed in 704d75bb5

udata/core/dataservices/api.py
```diff
@@ -95,6 +97,82 @@ def delete(self, dataservice):
...
+        ):
+            api.abort(410, "Dataservice has been deleted")
+
+        OwnablePermission(dataservice).test()
+
+        if dataset.id not in (d.id for d in dataservice.datasets):
```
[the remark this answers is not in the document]
ThibaudDauce: Simplified in 89cec3d28

maudetes (approved):
I think we're clear! Thank you a lot :pray:

PR #3159, opened 2024-10-02
Add resources formats and harvest remote_url on dataset catalog

maudetes:
* format as a comma-separated set of resources formats
* harvest remote_url if exists

This format information is useful at the dataset level to prevent having to join with the resources catalog.


Pierlou (approved):
nice additions 🚀

PR #3160, opened 2024-10-03
Filter by org badge complex

ThibaudDauce:
Probably not worth it :-) But we tried! :-D

Modifications:
- [x] Allow to add a `default_filterable_field` on classes. For exemple if we activate the new system for the `Organization`, we could filter by `badge__kind` automatically by adding a `filterable` on the `badges` field.
- [x] Allow to use custom query for filters (instead of simple equals)
- [ ] `Badge` are now separate classes for each model (`Organization`, `Reuse`…). It enables to use the `choices` of the `StringField` instead of custom validation check

### DIFF
--- a/udata/api_fields.py
+++ b/udata/api_fields.py
@@ -1,3 +1,6 @@
+import functools
+from typing import Dict, List
+
 import flask_restx.fields as restx_fields
 import mongoengine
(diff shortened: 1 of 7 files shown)


maudetes:
This was another go at https://github.com/opendatateam/udata/pull/3155, partly integrated.

PR #3162, opened 2024-10-07
Add a job to bind Tabular API to its datasets

maudetes:
Related to https://github.com/datagouv/data.gouv.fr/issues/1467.

The purpose is to add all the datasets served by tabular API to the corresponding dataservice object.
It is based on resources extras to know if a dataset is served by tabular API.

The job could be scheduled to run daily.

We're adding the job in udata directly instead of the [tabular preview plugin](https://github.com/opendatateam/udata-tabular-preview/).
Both use the same logic : https://github.com/opendatateam/udata-tabular-preview/blob/master/udata_tabular_preview/preview.py#L18-L21


udata/core/dataset/tasks.py
```diff
@@ -231,3 +233,38 @@ def export_csv(self, model=None):
...
+    datasets = Dataset.objects(
+        **{
+            "resources__extras__analysis:parsing:finished_at__exists": True,
+            "resources__extras__analysis:parsing:error": None,
+        }
+    ).visible()
+
+    dataservice.datasets = []
+    for dat in datasets:
+        dataservice.datasets.append(dat)
```
maudetes, line 263: I guess we could do `dataservice.datasets = datasets` directly

PR #3166, opened 2024-10-08
Fix/csv export resources formats

maudetes:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/146772/?environment=dev.data.gouv.fr&project=12&query=is%3Aunresolved&referrer=issue-stream


Pierlou (approved):
Oops I forgot that this metadata is not mandatory, thanks for the fix!

PR #3170, opened 2024-10-16
Add default class sort

maudetes:
Set dataservice default sort as `-created_at`


ThibaudDauce:
Is it still relevant?

PR #3172, opened 2024-10-18
fix(harvest): use dct.created for creation date and store dct.issued

abulte:
Related https://github.com/ecolabdata/ecospheres/issues/422

Adds `issued_at` from `dct.issued` to `harvest` informations for:
- dataset
- dataservice
- resource

Also switch `harvest.created_at` to `dct.created` instead of `dct.issued` when there's one (if not, fallback to `issued`).

Only a proposal for now, some more tests will need to be written for example.

### DIFF
--- a/udata/core/dataset/models.py
+++ b/udata/core/dataset/models.py
@@ -71,6 +71,7 @@ def get_json_ld_extra(key, value):
 class HarvestDatasetMetadata(DynamicEmbeddedDocument):
     backend = db.StringField()
     created_at = db.DateTimeField()
+    issued_at = db.DateTimeField()
     modified_at = db.DateTimeField()
     source_id = db.StringField()
     remote_id = db.StringField()
@@ -86,6 +87,7 @@ class HarvestDatasetMetadata(DynamicEmbeddedDocument):
 class HarvestResourceMetadata(DynamicEmbeddedDocument):
     created_at = db.DateTimeField()
     modified_at = db.DateTimeField()
+    issued_at = db.DateTimeField()
     uri = db.StringField()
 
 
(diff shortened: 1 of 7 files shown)


abulte:
@maudetes WDYT?

udata/core/dataset/rdf.py
maudetes, line 1: We should also expose these to `dataset_to_rdf`?

udata/core/dataset/models.py
```diff
@@ -86,6 +87,7 @@ class HarvestDatasetMetadata(DynamicEmbeddedDocument):
 class HarvestResourceMetadata(DynamicEmbeddedDocument):
     created_at = db.DateTimeField()
     modified_at = db.DateTimeField()
+    issued_at = db.DateTimeField()
```
maudetes, line 90: do we need to have both created_at and issued_at at the resource level?

maudetes (commented):
Yes, agreed with @streino with the direction

udata/core/dataset/rdf.py
```diff
@@ -678,7 +683,11 @@ def dataset_from_rdf(graph: Graph, dataset=None, node=None):
     identifier = rdf_value(d, DCT.identifier)
     uri = d.identifier.toPython() if isinstance(d.identifier, URIRef) else None
     remote_url = remote_url_from_rdf(d)
-    created_at = rdf_value(d, DCT.issued)
+    created_at = rdf_value(d, DCT.created)
```
maudetes, line 686: `DCT.created` [comes from GeoDCAT-AP](https://semiceu.github.io/GeoDCAT-AP/releases/2.0.0/) only, if you wanna add a comment

udata/core/dataset/models.py
abulte: `issued` and `modified` are probably enough, no `created` in GeoDCAT-AP on Distributions https://semiceu.github.io/GeoDCAT-AP/releases/2.0.0/#optional-properties-for-distribution

abulte: Same in GeoDCAT-AP 3.0 https://semiceu.github.io/GeoDCAT-AP/releases/3.0.0/#Distribution

abulte:
Replace by https://github.com/opendatateam/udata/pull/3352

PR #3176, opened 2024-10-28
Fix flaky "duplicated email" tests

magopian:
Fixes [#1466](https://github.com/datagouv/data.gouv.fr/issues/1466)

@nicolaskempf57 found the root cause [here](https://github.com/datagouv/data.gouv.fr/issues/1466#issuecomment-2338006617)


maudetes (approved):
:pray: thank you for the fix!

PR #3177, opened 2024-10-28
Use trailing slashes for the upload files URLs

magopian:
Fixes [#1423](https://github.com/datagouv/data.gouv.fr/issues/1423)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current (in progress)
 
 - Add contact_form in ContactPoint api fields [#3175](https://github.com/opendatateam/udata/pull/3175)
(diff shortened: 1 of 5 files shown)


maudetes (approved):
Thank you! :pray:

I think we should always add trailing slash to allow for a redirect.
See [flask documentation](https://flask-docs.readthedocs.io/en/latest/api/#url-route-registrations) on this:
> An important detail to keep in mind is how Flask deals with trailing slashes. The idea is to keep each URL unique so the following rules apply:
If a rule ends with a slash and is requested without a slash by the user, the user is automatically redirected to the same page with a trailing slash attached.
If a rule does not end with a trailing slash and the user requests the page with a trailing slash, a 404 not found is raised.
This is consistent with how web servers deal with static files. This also makes it possible to use relative link targets safely.

PR #3180, opened 2024-10-29
Return dataservices visible to the user

maudetes:
Close https://github.com/datagouv/data.gouv.fr/issues/1549

Apply a similar logic as [for reuses](https://github.com/opendatateam/udata/pull/3140) without the private param.

### DIFF
--- a/udata/tests/api/test_dataservices_api.py
+++ b/udata/tests/api/test_dataservices_api.py
@@ -2,6 +2,7 @@ from xml.etree.ElementTree import XML
 
 import pytest
 from flask import url_for
+from werkzeug.test import TestResponse
 
 from udata.core.dataservices.factories import DataserviceFactory
 from udata.core.dataservices.models import Dataservice
@@ -13,6 +14,11 @@ from udata.i18n import gettext as _
 from udata.tests.helpers import assert200, assert_redirects
 
 from . import APITestCase
+
+
+def dataservice_in_response(response: TestResponse, dataservice: Dataservice) -> bool:
+    only_dataservice = [r for r in response.json["data"] if r["id"] == str(dataservice.id)]
+    return len(only_dataservice) > 0
 
 
 class DataserviceAPITest(APITestCase):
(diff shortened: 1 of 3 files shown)


udata/tests/api/test_dataservices_api.py
```diff
@@ -192,6 +244,9 @@ def test_dataservice_api_index(self):
 
         self.assertEqual(Dataservice.objects.count(), 4)
 
+        # Login with a distinct user
+        self.login(UserFactory())
+
```
magopian, line 249: What is that for? I'm also not sure what the assertion below checks, other than "this url exists"?

udata/tests/api/test_dataservices_api.py
```diff
@@ -15,6 +16,11 @@
 from . import APITestCase
 
 
+def dataservice_in_response(response: TestResponse, dataservice: Dataservice) -> bool:
+    only_dataservice = [r for r in response.json["data"] if r["id"] == str(dataservice.id)]
```
magopian, line 20: I was thinking "meh, this name feels a bit off, i'm not sure what's the logic behind it", and then I saw it was my code ... guess I couldn't come up with a better one, and still can't 🙄

udata/tests/api/test_dataservices_api.py
```diff
@@ -151,6 +157,52 @@ def test_dataservice_api_create(self):
         self.assertEqual(dataservice.title, "Undeleted title")
         self.assertIsNone(dataservice.deleted_at)
 
+    def test_dataservice_api_list_owned(self) -> None:
+        """Should filters out private dataservices if not owner"""
```
magopian, line 161: Nit: typo `Should filter out...`

udata/tests/api/test_dataservices_api.py
```diff
@@ -151,6 +157,52 @@ def test_dataservice_api_create(self):
...
+        assert200(response)
+        assert len(response.json["data"]) == 2  # Return everything
+        assert dataservice_in_response(response, public_dataservice)
+        assert dataservice_in_response(response, private_user_dataservice)
+
+        # Authenticated user is now also member of the organization
+        member = Member(user=owner, role="editor")
+        org.members = [member]
+        org.save()
+        del owner.organizations  # clear user.organizations cached property
```
magopian, line 195: Wow didn't know we had to do that 🤯

magopian (approved):
Excellent!
Looks perfect, with a couple of minor nits that you may ignore.

udata/tests/api/test_dataservices_api.py
maudetes: Well, I had to debug to understand what was happening :scream:

udata/tests/api/test_dataservices_api.py
maudetes: If we don't change the User we're logged in with, we would have dataservice being returned since the user would now get the private dataservice in the response.
I've updated the comment

Issue #1534, opened 2018-03-26
Listing paginator allows to advance to empty results set 


petzlux:
## Detailed Description
Listing pages (Datasets,Reuses,Organisations) Paginator allows at last page of results to click on last page and next page arrows, to advance to empty results page beyond max page. 

![listing paginator](https://user-images.githubusercontent.com/2540428/37913945-62997de6-3116-11e8-8b47-56817a0c131d.gif)

PR #3179, opened 2024-10-29
Expose contact point in DCAT RDF

maudetes:
Close https://github.com/datagouv/data.gouv.fr/issues/1520
Close https://github.com/datagouv/data.gouv.fr/issues/1519

Example:
```
  <rdf:Description rdf:about="http://dev.local:7000/dataservices/6661e3b3ac0706063080aa40/">
    <rdf:type rdf:resource="http://www.w3.org/ns/dcat#DataService"/>
    <dcat:contactPoint rdf:resource="http://dev.local:7000/api/1/contacts/6661de3080af2e6f478a53f7/"/>
  </rdf:Description>
  <rdf:Description rdf:about="http://dev.local:7000/api/1/contacts/6661de3080af2e6f478a53f7/">
    <rdf:type rdf:resource="http://www.w3.org/2006/vcard/ns#Kind"/>
    <vcard:fn>Support API Météo-France</vcard:fn>
    <vcard:hasEmail rdf:resource="mailto:contact.api@meteo.fr"/>
    <vcard:hasUrl rdf:resource="https://ceci.est.un.test.com/"/>
  </rdf:Description>
```

PR #3186, opened 2024-11-05
Expose the alternate identifiers in RDF

magopian:
Fixes [#1534](https://github.com/datagouv/data.gouv.fr/issues/1534)

### DIFF
--- a/udata/core/dataset/rdf.py
+++ b/udata/core/dataset/rdf.py
@@ -192,24 +192,19 @@ def dataset_to_graph_id(dataset: Dataset) -> URIRef | BNode:
         return BNode()
 
 
-def set_identifier(dataset: Dataset, graph: Graph, dataset_rdf: Resource) -> None:
-    # Expose upstream identifier if present
-    if dataset.harvest and dataset.harvest.dct_identifier:
-        dataset_rdf.set(DCT.identifier, Literal(dataset.harvest.dct_identifier))
-        alternative_id = URIRef(
-            endpoint_for(
-                "datasets.show_redirect",
-                "api.dataset",
-                dataset=dataset.id,
-                _external=True,
-            )
+def alternate_identifier(dataset: Dataset, graph: Graph) -> None:
+    alternative_id = URIRef(
+        endpoint_for(
+            "datasets.show_redirect",
+            "api.dataset",
+            dataset=dataset.id,
+            _external=True,
         )
-        adms_identifier = graph.resource(BNode())
-        adms_identifier.set(DCT.creator, URIRef("https://data.gouv.fr"))
-        adms_identifier.set(SKOS.notation, alternative_id)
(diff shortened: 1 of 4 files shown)


udata/core/dataset/rdf.py
```diff
@@ -200,16 +201,25 @@ def dataset_to_rdf(dataset, graph=None):
...
     id = dataset_to_graph_id(dataset)
 
+    graph = graph or Graph(namespace_manager=namespace_manager)
+    d = graph.resource(id)
+
     # Expose upstream identifier if present
     if dataset.harvest and dataset.harvest.dct_identifier:
-        identifier = dataset.harvest.dct_identifier
+        d.set(DCT.identifier, Literal(dataset.harvest.dct_identifier))
+        alternate_identifier = URIRef(
```
maudetes, line 210: Shouldn't the identifier be a literal? because else it may be a resource described in the graph.

magopian: I'm confused, didn't we decide to use [URIRef for the `endpointURL`](https://github.com/opendatateam/udata/pull/3179/files#r1821117993)? Is this any different?

I'm happy to change it, I would just love to understand the logics of when to use a Literal (simple string, right?) or a URIRef (I thought `ref` in `URIRef` meant "reference to another resource described in the graph", but then we used it for the `endpointURL` which is just a URL, not another resource described in the graph).

maudetes: I'm not 100% sure about the logic either.
I think that endpointURL is an identified resource on the web whereas an identifier is an identifying string?
However, we can base our decision on:
* the spec, which shows [endpoint URL](https://www.w3.org/TR/vocab-dcat-3/#Property:data_service_endpoint_url) in range Resource and [identifier having a notation](https://semiceu.github.io/DCAT-AP/releases/3.0.0/#Identifier) in range Literal
* usage in real life examples, ex: [this dataset on data.europa.eu](https://data.europa.eu/data/datasets/34fc887b-be29-47bf-a2b3-67428d47d434?locale=en) (with both adms:identifier and dcat:endpointUrl)

udata/core/dataset/rdf.py
```diff
@@ -207,15 +207,20 @@ def dataset_to_rdf(dataset, graph=None):
...
             endpoint_for(
                 "datasets.show_redirect",
                 "api.dataset",
                 dataset=dataset.id,
                 _external=True,
             )
         )
-        d.add(ADMS.identifier, alternate_identifier)
+        alt.set(RDF.type, ADMS.Identifier)
+        alt.set(DCT.creator, Literal("https://data.gouv.fr"))
```
maudetes, line 221: I think we should use `SITE_TITLE` config

maudetes (approved):
I think it works nicely, thank you!

Issue #1533, opened 2018-03-26
Add Dataset/Reuse Dialog not filtering Organisations


petzlux:
## Detailed Description
On trying to add a new reuse or dataset , the first page under which org to publish, allows user to search for and select orgs to which the publisher has no rights.

This is with a non admin user. Only on the add dataset entry page on submit is there a 400 error created, without notification to the user.

See also https://github.com/opendatalu/gouvlu/issues/77

## Your Environment
udata 1.3.3

PR #3187, opened 2024-11-06
Add DCAT-AP HVD properties in RDF output if the dataservice is tagged hvd

magopian:
Fixes [#1553](https://github.com/datagouv/data.gouv.fr/issues/1553)

### DIFF
--- a/udata/core/dataservices/rdf.py
+++ b/udata/core/dataservices/rdf.py
@@ -157,6 +157,8 @@ def dataservice_to_rdf(dataservice: Dataservice, graph=None):
     if is_hvd:
         # We also want to automatically add any HVD category tags of the dataservice's datasets.
         for dataset in dataservice.datasets:
+            if "hvd" not in dataset.tags:  # Only check HVD datasets for their categories.
+                continue
             for tag in dataset.tags:
                 if tag in TAG_TO_EU_HVD_CATEGORIES:
                     hvd_category_tags.add(tag)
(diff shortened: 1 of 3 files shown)


maudetes (commented):
I think I have some questions on applying HVD applicableLegislation based on serving an hvd dataset, since the dataservice may not be sufficient yet.
It may be more cautious to tag dataservices explicitly for now?

Same for HVD category, should they depend HVD datasets being served or be explicit on the dataservice?

I would be interested to discuss this live!

magopian:
I don't have an opinion on the matter, as I don't know enough! I implemented it like that because that's what I understood from [#1533](https://github.com/datagouv/data.gouv.fr/issues/1553)

> La catégorie peut peut-être être déduite des datasets HVD servis par ce dataservice.

I'll change it to only apply it when there's a direct badge on the dataservice, it'll also simplify the code ;)

bolinocroustibat (approved):
LGTM!

udata/core/dataservices/rdf.py
```diff
@@ -137,8 +141,17 @@ def dataservice_to_rdf(dataservice: Dataservice, graph=None):
     if dataservice.endpoint_description_url:
         d.set(DCAT.endpointDescription, URIRef(dataservice.endpoint_description_url))
 
+    # Add DCAT-AP HVD properties if the dataservice is tagged hvd.
+    # See https://semiceu.github.io/DCAT-AP/releases/2.2.0-hvd/
+    is_hvd = current_app.config["HVD_SUPPORT"] and "hvd" in dataservice.tags
+    if is_hvd:
+        d.add(DCATAP.applicableLegislation, URIRef(HVD_LEGISLATION))
+
     for tag in dataservice.tags:
```
maudetes, line 150: I think we should look at tags from HVD datasets served by this dataservice instead of its own tags.

PR #3188, opened 2024-11-07
Minor refactor of resource catalog

Pierlou:
Add extras column and fix preview url


Pierlou:
So we're fine with the `extras` column?

PR #3189, opened 2024-11-08
Allow override of dataset badges

maudetes:
Fix https://github.com/datagouv/data.gouv.fr/issues/1564

This is a quickfix for the dataset case, we may want to find a cleaner solution.

### DIFF
--- a/udata/core/badges/tests/test_model.py
+++ b/udata/core/badges/tests/test_model.py
@@ -1,5 +1,6 @@
 from udata.api_fields import field
 from udata.auth import login_user
+from udata.core.dataset.models import Dataset, DatasetBadge
 from udata.core.user.factories import UserFactory
 from udata.mongo import db
 from udata.tests import DBTestMixin, TestCase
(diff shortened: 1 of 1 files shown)


geoffreyaldebert (approved):

magopian:
Sure thing @maudetes!

udata/core/dataset/models.py
```diff
@@ -521,8 +521,13 @@ def save(self, *args, **kwargs):
...
 
 
+def validate_badge(value):
+    if value not in Dataset.__badges__.keys():
+        raise db.ValidationError("Unknown badge type")
+
+
 class DatasetBadge(Badge):
-    kind = db.StringField(required=True, choices=list(BADGES.keys()))
+    kind = db.StringField(required=True, validation=validate_badge)
```
magopian, line 530: using `choices` makes it use a dropdown in swagger, which is convenient, but I don't know any existing way to have a dynamic list of choices... so I guess we have to let go of that convenience ;)

magopian (approved):
Thank you very much for this, indeed, it's a bit frightening to see we can break udata-front with what felt like basic changes on the backend 😱 

Nice code, nice test, excellent work 💯

udata/core/dataset/models.py
maudetes: I've actually tried keeping `choices` with a lambda to compute the value dynamically but I could not make it work sadely

magopian: Yep, I tried also in the past, and couldn't make it work. There doesn't seem to be a way to do it, other than hacking/monkeypatching the lib itself, which seemed a little bit too much at the time.

I'll have a deeper look in the future if it's prioritized 👍

maudetes:
If we keep this direction, we may want to apply the same logic to other badges type

PR #3190, opened 2024-11-12
Add matomo campaign parameters on links in emails

magopian:
Fixes [#1445](https://github.com/datagouv/data.gouv.fr/issues/1445)

To be used in production, this needs a new setting to be set:

`MAIL_CAMPAIGN = "campaign-name-for-matomo"`

### DIFF
--- a/udata/templates/mail/badge_added_association.html
+++ b/udata/templates/mail/badge_added_association.html
@@ -6,15 +6,14 @@
     _('%(user)s has identified your organization "%(name)s" as an association',
     user=(
         '<a href="'|safe
-        + url_for('api.user', user=badge.created_by, _external=True, **extras)
+        + badge.created_by.external_url_with_campaign
         + '">'|safe
         + badge.created_by.fullname
         + '</a>'|safe
     ),
     name=(
         '<a href="'|safe
-        + organization.external_url
-        + ('?mtm_campaign=' + extras.get('mtm_campaign')) if extras.get('mtm_campaign') else ''
+        + organization.external_url_with_campaign
         + '">'|safe
         + organization.name|string
         + '</a>'|safe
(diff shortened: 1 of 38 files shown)


udata/templates/mail/badge_added_association.html
```diff
@@ -6,14 +6,15 @@
...
-        + url_for('api.user', user=badge.created_by, _external=True)
+        + url_for('api.user', user=badge.created_by, _external=True, **extras)
         + '">'|safe
         + badge.created_by.fullname
         + '</a>'|safe
     ),
     name=(
         '<a href="'|safe
         + organization.external_url
+        + ('?mtm_campaign=' + extras.get('mtm_campaign')) if extras.get('mtm_campaign') else ''
```
maudetes, line 17: I think some html links get broken on some mails

maudetes (commented):
Thank you!

It would be nice if we could test it with matomo before pushing it into production.
I can give you the matomo config to use locally?

udata/templates/mail/badge_added_association.html
magopian: What do you mean? Do you have an example I could try?
I'll spend some time soon to try setting up the actual sending of emails locally, to make sure the emails end up looking correct in my proper inbox.

magopian:
> I can give you the matomo config to use locally?

Please do! Do you mean a matomo config to push pings to a dev instance of matomo?

magopian:
Ok, I pushed a big change: most urls in the email templates are now using a new `external_url_with_campaign` helper, and for the few where it's not possible, we still use the `extras` provided in the template context.

We shouldn't have broken links anymore as we're not building them manually. Not sure how to test them all though, some of them I'm not even sure how to have them sent, but I'll spend some time making sure most of them are working correctly.

maudetes (approved):
Thank you! :clap: good work!

Issue #1564, opened 2018-04-06
Add a preview URL for a resource


abulte:
This a feature idea to be discussed.

## First iteration

- add a field "preview URL" field on the resource model, display it on resource form
- if the field is present, a new button "preview" appears on the resource card — on click it opens the preview URL in a new tab, or an iframe somewhere on the site

This field could be used by resource owners to link to whatever tool/page they wish.

## Next iterations

- some harvesters can fill this field automatically: both CKAN and ODS have a preview page
- `geo.data.gouv.fr` can also fill this field
- some plugins can declare they can take care of the preview of a particular format (eg: `udata-preview-csv`) and fill the field automatically
- some additional fields could be required to configure the preview plugin properly (eg: separator, encoding, column names...)

PR #3191, opened 2024-11-12
Allow override of badges

magopian:
Follows up on #3189, fixes [#1564](https://github.com/datagouv/data.gouv.fr/issues/1564)

### DIFF
--- a/udata/core/badges/tests/test_model.py
+++ b/udata/core/badges/tests/test_model.py
@@ -1,5 +1,6 @@
 from udata.api_fields import field
 from udata.auth import login_user
+from udata.core.dataset.models import Dataset, DatasetBadge
 from udata.core.user.factories import UserFactory
 from udata.mongo import db
 from udata.tests import DBTestMixin, TestCase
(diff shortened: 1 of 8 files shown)


udata/core/badges/tests/test_model.py
```diff
@@ -155,3 +156,21 @@ def test_create_disallow_unknown_badges(self):
         with self.assertRaises(db.ValidationError):
             fake = Fake.objects.create()
             fake.add_badge("unknown")
+
+
+class DatasetBadgeTest(DBTestMixin, TestCase):
```
maudetes, line 161: Do you want to move this in test_dataset_model?

maudetes (approved):
Thank you! I think we're good :clap:

Don't forget the changelog ;)

udata/core/badges/tests/test_model.py
magopian: Fixed the badge tests ;) What do you think?

maudetes: :ok_hand: looks good, thank you!

PR #3192, opened 2024-11-13
Temporal coverage with only start date

magopian:
Fixes [#1268](https://github.com/datagouv/data.gouv.fr/issues/1268)

Code from @thibaudDauce:
https://github.com/opendatateam/udata/compare/temporal_coverage_with_only_start_date?expand=1

Was copy/pasted from his branch, because cherry-picking led to humongous conflicts for some reason...


maudetes (approved):
Thank you for this PR! I think it works nicely but we should probably check for side effects a bit :detective: 

On udata-front, we should probably re-open https://github.com/datagouv/udata-front/pull/411/ and I think the udata (old) admin daterange may crash as well?

Don't forget the changelog ;)

magopian:
> udata (old) admin daterange may crash as well?

Regarding the (old) admin daterange, I've just tested and it seems to work as usual, no crashes if you provide both start and end date (this behavior wasn't changed by this PR as far as I can tell).

Not providing an end date shows a frontend validation error asking for both the start and end date.

One thing that fails in the old admin: editing a dataset that has no end date. In this case, the dataset can't be updated unless an end date is provided.

I'll work on allowing that in this PR.

magopian:
Updated the PR with a change to the old admin so it also allows setting/editing a temporal coverage with only a start date.

bolinocroustibat (approved):
Nice, LGTM!

js/components/form/daterange-picker.vue
```diff
@@ -131,11 +130,11 @@ export default {
         $(this.$els.endHidden).rules('add', {
             dateGreaterThan: this.$els.startHidden.id,
             required: (el) => {
-                return (this.startValue && !this.endValue) || (this.endValue && !this.startValue);
+                return (this.endValue && !this.startValue);
             },
```
maudetes, line 134: I think it somehow does not work as expected? Maybe it should be cast to Boolean?

maudetes (approved):
Minor comment on the admin, but else I think we're fine!

js/components/form/daterange-picker.vue
maudetes: If I have an existing dataset with a complete temporal coverage and I delete the end date, I can't submit:
image

magopian: It took me a while to understand what I was checking wrong, but I finally managed to reproduce and fix it ;)

Thanks, it was indeed needing a proper bool value (`null` didn't cut it).

maudetes (approved):
We'll need to be ready to merge https://github.com/datagouv/udata-front/pull/411 before merging this PR.

I think we will probably need to update the front-end as well. cc @nicolaskempf57

PR #3198, opened 2024-11-18
Fix the Badge.badge_label method

### LINKED ISSUE #3197: AttributeError: 'Organization' object has no attribute 'badge'
After [this change](https://github.com/opendatateam/udata/pull/3155/files#diff-da2e07ff2c5e7a01cc03139038bdd40e90840c8c36d1db5fd515e823199ec10fR86) I'm getting `AttributeError: 'Organization' object has no attribute 'badge'` error when calling `badge_label()`.

## Detailed Description
I'm not sure about the motivation for this change and where `self.badge` is supposed to be set, but it seems that it's not set anywhere at the moment, thus the error.

## Possible Implementation
`self.badge` could be replaced with `self._fields["badges"].field.document_type` like in `add_badge()`.

## Your Environment
udata `v10.0.1`

magopian:
Fixes #3197

### DIFF
--- a/udata/core/badges/models.py
+++ b/udata/core/badges/models.py
@@ -83,5 +83,6 @@ class BadgeMixin:
 
     def badge_label(self, badge):
         """Display the badge label for a given kind"""
-        kind = badge.kind if isinstance(badge, self.badge) else badge
+        badge_model = self._fields["badges"].field.document_type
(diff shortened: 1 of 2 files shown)


maudetes (approved):
Don't forget the changelog ;)

PR #3199, opened 2024-11-18
Use correct DCAT.service predicate

maudetes:
Our DataServices were not harvested by data.europa.eu (cf [0 results](https://data.europa.eu/data/datasets?locale=en&minScoring=0&catalog=plateforme-ouverte-des-donnees-publiques-francaises&page=1&dataServices=true)).

The predicate to use on the catalog to link to a dataservice is `DCAT.service` (as specified in https://w3c.github.io/dxwg/dcat/#Property:catalog_service).


magopian (approved):
Very good, thanks for the fix! This is ha(RD)F.

PR #3200, opened 2024-11-19
Only keep the "local authority" org badge if it's also a "public service"

magopian:
Fixes [11433](https://github.com/datagouv/data.gouv.fr/issues/1433)

Also needs the [updated fixtures](https://github.com/opendatateam/udata-fixtures/pull/6)

### DIFF
--- a/udata/commands/fixtures.py
+++ b/udata/commands/fixtures.py
@@ -180,7 +180,11 @@ def get_or_create_user(data):
 def import_fixtures(source):
     """Build sample fixture data (users, datasets, reuses, dataservices) from local or remote file."""
     if source.startswith("http"):
-        json_fixtures = requests.get(source).json()
+        response = requests.get(source)
+        if response.ok:
+            json_fixtures = response.json()
+        else:
+            response.raise_for_status()
     else:
         with open(source) as f:
             json_fixtures = json.load(f)
--- a/udata/commands/fixtures.py
+++ b/udata/commands/fixtures.py
@@ -181,10 +181,8 @@ def import_fixtures(source):
     """Build sample fixture data (users, datasets, reuses, dataservices) from local or remote file."""
     if source.startswith("http"):
         response = requests.get(source)
-        if response.ok:
-            json_fixtures = response.json()
-        else:
-            response.raise_for_status()
+        response.raise_for_status()
+        json_fixtures = response.json()
     else:
         with open(source) as f:
             json_fixtures = json.load(f)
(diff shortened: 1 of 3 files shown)


magopian:
The failing test is because https://github.com/opendatateam/udata-fixtures/pull/6 is not merged yet, wich means that the tag `v4.0.0` isn't available yet, and returns a 404.

udata/migrations/2024-11-19-keep-only-local_authority-if-also-public_service-organization-badges.py
```diff
@@ -0,0 +1,22 @@
...
+from udata.models import Organization
+
+log = logging.getLogger(__name__)
+
+
+def migrate(db):
+    log.info("Processing Organizations...")
+    query = Q(badges__kind=LOCAL_AUTHORITY) & Q(badges__kind=PUBLIC_SERVICE)
+
+    count = Organization.objects(query).update(set__badges__S__kind=LOCAL_AUTHORITY)
```
maudetes, line 19: I'm not sure the `__S__` will work as expected since we have two queries badges__kind?
In any case, it will replace the `PUBLIc_SERVICE` badge by another `LOCAL_AUTHORITY` leaving the orga with two `LOCAL_AUTHORITY` badges right? They don't seem to get de-duplicated

udata/commands/fixtures.py
```diff
@@ -180,7 +180,11 @@ def get_or_create_user(data):
 def import_fixtures(source):
     """Build sample fixture data (users, datasets, reuses, dataservices) from local or remote file."""
     if source.startswith("http"):
-        json_fixtures = requests.get(source).json()
+        response = requests.get(source)
+        if response.ok:
+            json_fixtures = response.json()
+        else:
+            response.raise_for_status()
```
maudetes, line 187: I think you can do
```suggestion
        response = requests.get(source)
        response.raise_for_status()
        json_fixtures = response.json()
```

maudetes (commented):
Thank you for this migration!

udata/commands/fixtures.py
magopian: Ah yes, way better, thanks!

udata/migrations/2024-11-19-keep-only-local_authority-if-also-public_service-organization-badges.py
magopian: Yes indeed, I don't know what i've done to test this out and found it working, ... I guess my query was wrong, I indeed wasn't having any org with both the badges, but they did have their badges overwritten with `local-authority`, which made them have it twice... dumb :/

maudetes (approved):
Looks all clear! :pray: :classical_building:

PR #3201, opened 2024-11-19
feat: create custom Exception with contextual Sentry tags

bolinocroustibat:
This a a PR following a [discussion on improving information within Sentry errors on Hydra and suggesting creating a similar feature on udata](https://github.com/datagouv/hydra/pull/212).

This is a proposal and totally open for discussion, as I am not fully aware of this udata codebase!

Main changes:

- Remove `SENTRY_TAGS` setting in settings, which was unused (this has been checked on infra config as well), and since we assume Sentry tags are most of a contextual use than of a global application use.
- Add a `ExceptionWithSentryDetails` custom exception  in `udata/errors.py`, to which one can pass specific contextual tags for a better context when debugging using Sentry.

This PR also embarks the following cleaning proposal:

- Remove the differentiation between `SENTRY_PUBLIC_DSN` and `SENTRY_DSN` as well as the `public_dsn()` method to convert a secret DSN to a public one, since secret Sentry DSN are deprecated and since we only use public `SENTRY_DSN` in our configs nowadays. This has also been checked on infra's side, where `SENTRY_PUBLIC_DSN` is nowhere being used anymore.
- Minor cleaning: use dedicated and shared `package_version()` method instead of `pkg_resources.get_distribution().version` method for udata app tag in Sentry inititalization.

Note: should this PR should be updated with proposals of where to use `ExceptionWithSentryDetails` custom exceptions?


udata/sentry.py
```diff
@@ -79,13 +50,9 @@ def init_app(app: UDataApp):
...
-        tags = app.config["SENTRY_TAGS"]
-        for tag_key in tags:
-            sentry_sdk.set_tag(tag_key, tags[tag_key])
         # Versions Management: uData and plugins versions as tags.
         for dist in entrypoints.get_plugins_dists(app):
             if dist.version:
                 sentry_sdk.set_tag(dist.project_name, dist.version)
-        # Do not forget udata itself
-        sentry_sdk.set_tag("udata", pkg_resources.get_distribution("udata").version)
+        # Do not forget udata itself (is that necessary since we already have the release?)
```
magopian, line 57: This might indeed not be needed

magopian (approved):
Such cleanup, nice!
Thank you very much for porting your changes from hydra to udata, not sure yet where we'll use this, but it's going to be useful for sure, maybe when we go through the actual sentry exceptions?
🚢

udata/sentry.py
bolinocroustibat: Indeed I also think we can remove it. Done

bolinocroustibat:
> Such cleanup, nice! Thank you very much for porting your changes from hydra to udata, not sure yet where we'll use this, but it's going to be useful for sure, maybe when we go through the actual sentry exceptions? 🚢

Only when we need to send more info to identify the error, like `resource_id`, `url` etc, we could use that custom exception and send them as tags. Otherwise, I think it's not necessary.
I leave it to you to use it wherever you think it might be necessary to send tags!

udata/errors.py
```diff
@@ -1,4 +1,27 @@
+from flask import current_app
+
+
 class ConfigError(ValueError):
     """Exception raised when an error is detected into the configuration"""
 
     pass
+
+
+class ExceptionWithSentryDetails(Exception):
```
maudetes, line 10: I'm not sure of the usage of this Exception, since we mostly don't know when they'll occur.
Shouldn't we use [set_context](https://docs.sentry.io/platforms/python/enriching-events/context/) at the right point in time (fetching the user, etc.)

maudetes (approved):
Thank you for cleaning the sentry configuration!

udata/errors.py
bolinocroustibat: Tags and context seem quite similar indeed, but tags might a bit better as they are searchable in the Sentry UI, it would seem: https://docs.sentry.io/platforms/javascript/enriching-events/context/

Not sure yet when this exception could be used indeed, it is mostly to align with hydra features, isn't there some events where we would want to send the associated `resource_id` when an exception occurs?

bolinocroustibat:
Should we merge that one @magopian @maudetes ?

That was intended to reproduce in udata a useful error class which was used in hydra, but if we don't have the use for it, no problem, we can close this PR and not merge it!

maudetes:
I think we can close for now. We can re-open when the need arises.

PR #3202, opened 2024-11-19
Convert bool filters to 0 or 1 when querying the search service

magopian:
Fixes [#1456](https://github.com/datagouv/data.gouv.fr/issues/1456)

### DIFF
--- a/udata/search/query.py
+++ b/udata/search/query.py
@@ -41,7 +41,12 @@ class SearchQuery:
             if self.sort:
                 url = url + f"&sort={self.sort}"
             for name, value in self._filters.items():
-                url = url + f"&{name}={value}"
+                param_value = value
+                # HACK: use the `.__class__.__name__` to avoid having to import `BoolFilter` here, as importing it at the top would make a import loop.
+                if self.adapter.filters[name].__class__.__name__ == "BoolFilter":
+                    # The search service uses 1 and 0 for booleans.
+                    param_value = 1 if value == "true" else 0
+                url = url + f"&{name}={param_value}"
             r = requests.get(url, timeout=current_app.config["SEARCH_SERVICE_REQUEST_TIMEOUT"])
             r.raise_for_status()
             result = r.json()
--- a/udata/tests/search/test_query.py
+++ b/udata/tests/search/test_query.py
@@ -1,3 +1,8 @@
+from unittest.mock import patch
+
+from flask import current_app
+
+from udata.core.dataset.search import DatasetSearch
 from udata.search.query import DEFAULT_PAGE_SIZE, SearchQuery
 from udata.tests.api import APITestCase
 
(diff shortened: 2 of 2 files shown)


udata/tests/apiv2/test_datasets.py
```diff
@@ -53,6 +53,8 @@ def test_search_dataset(self):
         _dataset_org = DatasetFactory(organization=org)
         dataset_org_public_service = DatasetFactory(organization=org_public_service)
 
+        # This is needed or the test will fail if you have a search service configured in your udata.cfg file.
```
magopian, line 56: This is unrelated to this PR, but this test will fail if you have a search service configured, which I needed to fix this PR ;)

udata/tests/dataset/test_dataset_tasks.py
```diff
@@ -22,6 +23,9 @@ def test_purge_datasets():
         Dataset.objects.create(title="keep me"),
     ]
 
+    # This is needed or the test will fail if you have a search service configured in your udata.cfg file.
+    current_app.config["SEARCH_SERVICE_API_URL"] = ""
+
```
magopian, line 28: This is unrelated to this PR, but this test will fail if you have a search service configured, which I needed to fix this PR ;)

udata/tests/apiv2/test_datasets.py
```diff
@@ -53,6 +53,8 @@ def test_search_dataset(self):
...
+        current_app.config["SEARCH_SERVICE_API_URL"] = ""
```
magopian, line 57: This is unrelated to this PR, but this test will fail if you have a search service configured, which I needed to fix this PR ;)

udata/search/query.py
```diff
@@ -41,7 +41,12 @@ def execute_search(self):
             if self.sort:
                 url = url + f"&sort={self.sort}"
             for name, value in self._filters.items():
-                url = url + f"&{name}={value}"
+                param_value = value
+                # HACK: use the `.__class__.__name__` to avoid having to import `BoolFilter` here, as importing it at the top would make a import loop.
+                if self.adapter.filters[name].__class__.__name__ == "BoolFilter":
+                    # The search service uses 1 and 0 for booleans.
+                    param_value = 1 if value == "true" else 0
```
maudetes, line 48: Somehow on udata-search-service side, featured is not set to 0 or 1 [but 1 or 4](https://github.com/opendatateam/udata-search-service/blob/5bac31a90710c3281ffcf48198ed844aa3583041/udata_search_service/infrastructure/consumers.py#L36) (probably for poorly decided scoring reasons) :smiling_face_with_tear:

maudetes, line 48: I think if value is `True` and not `true` when it comes from the udata-front search view: `/fr/datasets/`

maudetes (commented):
Thank you for this PR!
I think the multiplication of endpoints and API shows all the small consistencies issues sadly

udata/search/query.py
ThibaudDauce: What do we need to merge this? (If we want to merge it…) Just  change the 1 to 4 here for the filter to work?

maudetes: I'd rather not add this hack on udata side as this isn't used at the moment anyway. I think we should fix the boolean score approach in the search service when we'll work on it.
Ticket created: https://github.com/datagouv/data.gouv.fr/issues/1909

PR #3203, opened 2024-11-20
Expose OGC services distributions as dataservice

maudetes:
Close https://github.com/datagouv/data.gouv.fr/issues/1538

Inspired from the conversion in the SEMIC XSLT: https://github.com/SEMICeu/iso-19139-to-dcat-ap/blob/f61b2921dd398b90b2dd2db14085e75687f7616b/iso-19139-to-dcat-ap.xsl#L1419

If the resource is an OGC Service, we add a `DCAT.accessService` property on the distribution and describe the DataService accordingly.

We support resources that have a format of `(ogc:)wms` or `(ogc:)wfs`.
We don't base our decision on `request=GetCapabilities` as in the XSLT for now.

We may update the logic later, either by detecting the format or basing our decision on the `request=GetCapabilities`.
It must be aligned with [the frontend](https://github.com/datagouv/udata-front/pull/609).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Expose OGC services distributions as dataservice [#3203](https://github.com/opendatateam/udata/pull/3203)
 
 ## 10.0.2 (2024-11-19)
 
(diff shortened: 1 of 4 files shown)


magopian (approved):
This is excellent work, well done!

PR #3204, opened 2024-11-21
feat: handle multiple tags on datasets and topics apis

abulte:
Multiple tags filter is now supported and datasets and topics list APIs.

The filter is a `AND`, i.e. `?tag=tag1&tag=tag2` will return objects that have both the `tag1` and `tag2` tags.

- [x] datasets api v1
- [x] datasets api v2 (`datasets/search` through MongoEngine)
- [x] topics api v1
- [x] topics api v2 
- [x] datasets api v2 w/ search service https://github.com/opendatateam/udata-search-service/pull/50

### DIFF
--- a/udata/core/dataset/apiv2.py
+++ b/udata/core/dataset/apiv2.py
@@ -267,7 +267,7 @@ class DatasetSearchAPI(API):
         """List or search all datasets"""
         args = search_parser.parse_args()
         try:
-            return search.query(Dataset, **args)
+            return search.query(Dataset, **{k: v for k, v in args.items() if v is not None})
         except NotImplementedError:
             abort(501, "Search endpoint not enabled")
         except RuntimeError:
--- a/udata/core/dataset/apiv2.py
+++ b/udata/core/dataset/apiv2.py
@@ -9,7 +9,7 @@ from udata.api import API, apiv2, fields
 from udata.core.contact_point.api_fields import contact_point_fields
 from udata.core.organization.api_fields import member_user_with_email_fields
 from udata.core.spatial.api_fields import geojson
-from udata.utils import get_by, multi_to_dict
+from udata.utils import get_by
 
 from .api import ResourceMixin
 from .api_fields import (
@@ -265,9 +265,9 @@ class DatasetSearchAPI(API):
     @apiv2.marshal_with(dataset_page_fields)
     def get(self):
         """List or search all datasets"""
-        search_parser.parse_args()
+        args = search_parser.parse_args()
         try:
-            return search.query(Dataset, **multi_to_dict(request.args))
+            return search.query(Dataset, **args)
         except NotImplementedError:
             abort(501, "Search endpoint not enabled")
         except RuntimeError:
--- a/udata/core/dataset/apiv2.py
+++ b/udata/core/dataset/apiv2.py
@@ -73,7 +73,7 @@ DEFAULT_MASK_APIV2 = ",".join(
 log = logging.getLogger(__name__)
 
 ns = apiv2.namespace("datasets", "Dataset related operations")
-search_parser = DatasetSearch.as_request_parser()
+search_parser = DatasetSearch.as_request_parser(store_missing=False)
 resources_parser = apiv2.parser()
 resources_parser.add_argument(
     "page", type=int, default=1, location="args", help="The page to fetch"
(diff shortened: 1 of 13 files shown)


udata/core/dataset/apiv2.py
```diff
@@ -73,7 +73,7 @@
 log = logging.getLogger(__name__)
 
 ns = apiv2.namespace("datasets", "Dataset related operations")
-search_parser = DatasetSearch.as_request_parser()
+search_parser = DatasetSearch.as_request_parser(store_missing=False)
```
abulte, line 76: We use `store_missing=False` because we don't want to send filters set to `None` to the search service (Mongo does not care).

udata/core/dataset/apiv2.py
```diff
@@ -265,9 +265,9 @@ class DatasetSearchAPI(API):
     @apiv2.marshal_with(dataset_page_fields)
     def get(self):
         """List or search all datasets"""
-        search_parser.parse_args()
+        args = search_parser.parse_args()
         try:
-            return search.query(Dataset, **multi_to_dict(request.args))
+            return search.query(Dataset, **args)
```
magopian, line 270: Makes more sense indeed.

Note for self: we have the same pattern elsewhere 
- https://github.com/opendatateam/udata/blob/bfeae1b6a1d0d33c29485a2feae012db69eec38f/udata/core/reuse/apiv2.py#L29-L30
- https://github.com/opendatateam/udata/blob/bfeae1b6a1d0d33c29485a2feae012db69eec38f/udata/core/organization/apiv2.py#L34-L35
- https://github.com/opendatateam/udata/blob/bfeae1b6a1d0d33c29485a2feae012db69eec38f/udata/core/dataservices/apiv2.py#L28-L29

udata/search/query.py
```diff
@@ -37,11 +37,7 @@ def execute_search(self):
         # If SEARCH_SERVICE_API_URL is set, the remote search service will be queried.
         # Otherwise mongo search will be used instead.
         if current_app.config["SEARCH_SERVICE_API_URL"]:
-            url = f"{current_app.config['SEARCH_SERVICE_API_URL']}{self.adapter.search_url}?q={self._query}&page={self.page}&page_size={self.page_size}"
-            if self.sort:
-                url = url + f"&sort={self.sort}"
-            for name, value in self._filters.items():
-                url = url + f"&{name}={value}"
+            url = self.to_search_service_url()
```
magopian, line 40: 🙏

udata/search/query.py
```diff
@@ -59,6 +55,19 @@ def execute_search(self):
...
+            url = url + f"&sort={self.sort}"
+        for name, value in self._filters.items():
+            if isinstance(value, (list, tuple)):
+                for v in value:
+                    url = url + f"&{name}={v}"
+            else:
+                url = url + f"&{name}={value}"
+        return url
+
+    # FIXME: unused?
```
magopian, line 70: Note to self: check, and remove in another cleanup PR if indeed unused.

udata/tests/apiv2/test_topics.py
```diff
@@ -19,7 +19,8 @@ def test_topic_api_list(self):
         org = OrganizationFactory()
         paca, _, _ = create_geozones_fixtures()
 
-        tag_topic = TopicFactory(tags=["energy"])
+        tag_topic_1 = TopicFactory(tags=["my-tag-shared", "my-tag-1"])
+        tag_topic_2 = TopicFactory(tags=["my-tag-shared", "my-tag-2"])
```
magopian, line 23: I'm confused, is this implemented only for the mongo search? I didn't see it in [the udata-search-service PR](https://github.com/opendatateam/udata-search-service/pull/50), am I missing something?

magopian (approved):
Excellent work! Very well executed 🙏 
I'm slightly confused as it seems we don't have the multi-tags implemented for the topics on the udata-search-service side, am I missing something?

udata/tests/apiv2/test_topics.py
abulte: Yup, mongo only. Topics are never handled by the search-service (unless I missed something huge :p) https://github.com/opendatateam/udata-search-service/blob/main/udata_search_service/domain/entities.py.

magopian: I see, makes sense! Thanks 🙏

maudetes (approved):
Thank you a lot for this PR :rocket: 
I think we can add a changelog entry and we're all clear!

PR #3207, opened 2024-11-26
Fix dataservice metadata_modified_at update in API

maudetes:
Use the correct wording when updating the `metadata_modified_at` field in API PATCH and DELETE calls.

This field is also used as a key in [the cache value in udata-front](https://github.com/datagouv/udata-front/blob/f22bfaf72fdb2b7b24a15a6992eaf1d15c95e585/udata_front/theme/gouvfr/templates/dataservice/display.html#L63).

Fixes https://github.com/opendatateam/udata/pull/3207

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add a matomo "campaign" parameter on links in emails if `MAIL_CAMPAIGN` is configured [#3190](https://github.com/opendatateam/udata/pull/3190)
 - Add DCAT-AP HVD properties in RDF output if the dataservice or its datasets are tagged hvd [#3187](https://github.com/opendatateam/udata/pull/3187)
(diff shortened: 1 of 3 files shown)


magopian (approved):
Nice! So many different `modified` and `created` variations throughout the code 😱 
One day, for sure, we'll fix all of them, right? :P

🚢

maudetes:
> Nice! So many different `modified` and `created` variations throughout the code 😱 One day, for sure, we'll fix all of them, right? :P

I actually hesitated between fixing this by using the actual field or by renaming the fields to the one used in other main collections (using db.Datetimed). I did not go down this path for this quickfix, but I think we should seriously take a look as long as dataservices are marked as beta.
I think @Pierlou would be glad too.

magopian:
Yeah, this quickfix makes total sense, and yes, fixing the incoherences will probably take some tinkering and some time ... in the future ;)

PR #3208, opened 2024-11-28
Add dataservice csv adapter

maudetes:
Close https://github.com/datagouv/data.gouv.fr/issues/1556

We split apply sort & filters and apply pagination in https://github.com/opendatateam/udata/pull/3208/commits/bcde4a3bd58c2785e62dbceda215d98f550e4302.
We want to apply sort and filters but not pagination in the context of the csv adapter (see [view in udata-front](https://github.com/datagouv/udata-front/pull/612))
This may be done more elegantly perhaps? cc @ThibaudDauce

### DIFF
--- a/udata/api_fields.py
+++ b/udata/api_fields.py
@@ -251,7 +251,8 @@ def generate_fields(**kwargs) -> Callable:
 
     It will also
     - generate an API parameter parser
-    - sort and filter a list of documents with the provided params using the `apply_sort_filters_and_pagination` helper
+    - sort and filter a list of documents with the provided params using the `apply_sort_filters` helper
+    - apply a pagination with page and page size args with the `apply_pagination` helper
 
     """
 
@@ -414,7 +415,7 @@ def generate_fields(**kwargs) -> Callable:
 
         cls.__index_parser__ = parser
 
-        def apply_sort_filters_and_pagination(base_query) -> DBPaginator:
+        def apply_sort_filters(base_query) -> UDataQuerySet:
             args = cls.__index_parser__.parse_args()
 
             if sortables and args["sort"]:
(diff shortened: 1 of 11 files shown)


ThibaudDauce (approved):
Seems ok!

Maybe we could generate the CSV class from the model with some magic but not sure if it's better…

udata/api_fields.py
```diff
@@ -457,12 +458,17 @@ def apply_sort_filters_and_pagination(base_query) -> DBPaginator:
                             }
                         )
 
+            return base_query
+
+        def apply_pagination(base_query) -> DBPaginator:
+            args = cls.__index_parser__.parse_args()
+
```
magopian, line 465: Oooooh I like that, splitting those into their own functions, nice! 💯

maudetes:
> Maybe we could generate the CSV class from the model with some magic but not sure if it's better…

Yep, I thought that too to reduce redundancy, but I think the 30 lines weren't worth the trouble for now :p

udata/tests/test_api_fields.py
```diff
@@ -267,19 +267,44 @@ def test_sortable_field(self, app) -> None:
...
+            assert results.page == page_arg.default
+
+    def test_paginate(self, app) -> None:
+        """Results should be returned paginated."""
+        [FakeFactory() for _ in range(100)]
+
+        with app.test_request_context("/foobar", query_string={"page": 3, "page_size": 5}):
+            results: DBPaginator = Fake.apply_pagination(Fake.apply_sort_filters(Fake.objects))
+            assert results.page_size == 5
+            assert results.page == 3
```
magopian, line 310: Way easier and cleaner to test both helper separatly, i'm loving this 👍

magopian (commented):
A bit late to the party, very nice work, thank you!
🚢

PR #3211, opened 2024-12-02
Add list transfers endpoint

ThibaudDauce:


udata/features/transfer/api.py
```diff
@@ -81,14 +85,34 @@
...
 @ns.route("/", endpoint="transfers")
 class TransferRequestsAPI(API):
     @api.doc("list_transfers")
     @api.marshal_list_with(transfer_fields)
     def get(self):
-        """List all transfer requests"""
-        pass
+        args = requests_parser.parse_args()
+
+        transfers = Transfer.objects
```
maudetes, line 104: Don't we want to filter on `Transfer.objects(status="pending")` by default for example?
Else we have more than 60K existing transfer objects

maudetes (approved):
Thank you for this PR!

I have a comment regarding scaling with many transfers

udata/features/transfer/api.py
ThibaudDauce: For now, I added a check for at least subject or recipient to prevent big responses d02b8a8633f1b6d249ebe876b7d2294942379444

PR #3212, opened 2024-12-02
Update default method blocklist for READ ONLY mode

maudetes:
Update default `METHOD_BLOCKLIST` with new post endpoints.
It is used for READ ONLY mode implemented in https://github.com/opendatateam/udata/pull/2565.


magopian (approved):
Nice! 🚢

PR #3219, opened 2024-12-05
Don’t duplicate datasets with a `harvest.remote_id` URI when harvesting

magopian:
Fixes [#1587](https://github.com/datagouv/data.gouv.fr/issues/1587)

### DIFF
--- a/docs/harvesting.md
+++ b/docs/harvesting.md
@@ -88,7 +88,7 @@ optional arguments:
 
 ## Backends
 
-`udata` comes with 3 harvest backends but you can implement your own backend.
+`udata` comes with 3 harvest backends (listed below) but you can implement your own backend.
 In order for `udata` to be able to use any of those backends, they first need to be enabled
 in the `udata.cfg` `PLUGINS` section, like so:
 
@@ -262,15 +262,14 @@ You may take a look at the [existing backends][backends-repository] to see exiti
 Debugging the harvesting code may be difficult as it's run in Celery, asynchronously, and using a `breakpoint` (to drop into a pdb)
 is [kind of a pain](https://docs.celeryq.dev/en/stable/userguide/debugging.html).
 
-Another trick could be calling the `udata.harvest.task.harvest` function straight from a `udata shell`.
-To do that:
+Another trick could be calling the synchronous version of the harvesting command:
 
-```python
+```shell
 # Drop a `breakpoint()` somewhere in your harvesting code, then:
-$ udata shell
->>> from udata.harvest.tasks import harvest
->>> harvest("<id of your job here, taken from the (old) admin at http://dev.local:7000/fr/admin/system/>")
+$ udata harvest run <source_id>
 ```
+
+`source_id` is the id of your job, taken from the (old) admin at http://dev.local:7000/fr/admin/system/.
 
 
 [DCAT]: https://www.w3.org/TR/vocab-dcat/
--- a/docs/harvesting.md
+++ b/docs/harvesting.md
@@ -89,6 +89,12 @@ optional arguments:
 ## Backends
 
 `udata` comes with 3 harvest backends but you can implement your own backend.
+In order for `udata` to be able to use any of those backends, they first need to be enabled
+in the `udata.cfg` `PLUGINS` section, like so:
+
+```cfg
+PLUGINS = ['dcat']
+```
 
 ### DCAT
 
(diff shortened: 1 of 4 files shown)


docs/harvesting.md
```diff
@@ -89,6 +89,12 @@ optional arguments:
 ## Backends
 
 `udata` comes with 3 harvest backends but you can implement your own backend.
```
bolinocroustibat, line 91: Would be nice to list the existing backends in the README here, no?

bolinocroustibat (approved):
LGTM, looks like it makes the code more robust. Always painful to implement a test, except maybe when it is to test a very new feature :)

docs/harvesting.md
```diff
@@ -251,6 +257,22 @@ This will create a new package with a harvester skeleton on which you can start
 You may take a look at the [existing backends][backends-repository] to see exiting implementations.
 
 
+## Debugging
```
maudetes, line 260: I think you may want to document this dedicated command instead?
```
udata harvest run <source_id>
```

udata/harvest/tests/test_base_backend.py
```diff
@@ -281,6 +287,78 @@ def test_harvest_datasets_get_deleted(self):
         for item in job.items:
             assert item.dataset is None
 
+    def test_no_datasets_duplication(self, app):
+        limit = app.config["HARVEST_AUTOARCHIVE_GRACE_DAYS"]
+        last_update = datetime.utcnow() - timedelta(days=limit - 1)
+        reuse_remote_id_uri = "http://example.com/reuse-this-uri"
```
maudetes, line 293: `reuse` means something else in the udata context, it may be misinterpreted :stuck_out_tongue: 
`duplicated_remote_id_uri` maybe?

udata/harvest/tests/test_base_backend.py
```diff
@@ -281,6 +287,78 @@ def test_harvest_datasets_get_deleted(self):
...
+        nb_datasets = 3
+        source = HarvestSourceFactory(
+            config={
+                "nb_datasets": nb_datasets,
+                "dataset_remote_ids": [reuse_remote_id_uri],
```
maudetes, line 298: Don't we want to have either `nb_datasets` OR `dataset_remote_ids`?
Else I find it hard to follow what happens when we have both :)

udata/harvest/tests/test_base_backend.py
```diff
@@ -281,6 +287,78 @@ def test_harvest_datasets_get_deleted(self):
...
+            harvest={
+                "domain": source.domain,
+                "remote_id": "fake-0",  # the FakeBackend harvest should reuse this dataset
+                "source_id": str(source.id),
+                "last_update": last_update,
+            },
+        )
+
+        # Create a dataset that should be reused even though it's a different `remote_id` and `domain,
+        # because it's an URI.
```
maudetes, line 316: Did you mean?
```suggestion
        # Create a dataset that should be reused even though it's a different `source_id` and `domain,
        # because the remote_id is the same and an URI.
```

udata/harvest/tests/test_base_backend.py
```diff
@@ -281,6 +287,78 @@ def test_harvest_datasets_get_deleted(self):
         for item in job.items:
             assert item.dataset is None
 
+    def test_no_datasets_duplication(self, app):
+        limit = app.config["HARVEST_AUTOARCHIVE_GRACE_DAYS"]
+        last_update = datetime.utcnow() - timedelta(days=limit - 1)
```
maudetes, line 292: I think bringing `HARVEST_AUTOARCHIVE_GRACE_DAYS` here makes it seem like it has some impact, but it should not right?

maudetes (approved):
Thank you!
And thank you for adding tests, it's not always easy with harvesting.
I have minor comments only

Don't forget the changelog ;)

docs/harvesting.md
magopian: yes indeed, i'll add the ones i'm aware of ;)

docs/harvesting.md
magopian: Incredible, this was there all along, I didn't realize 😬

udata/harvest/tests/test_base_backend.py
magopian: yes indeed, confusing, i'll change it, thanks!

udata/harvest/tests/test_base_backend.py
magopian: Yeah, it's wayyyyy clearer that way, thanks 😅

udata/harvest/tests/test_base_backend.py
magopian: Totally right! It was some leftover from a previous copy-paste, well spotted, thanks, i'll remove it straight away 🙏

udata/harvest/tests/test_base_backend.py
magopian: I updated the tests: https://github.com/opendatateam/udata/pull/3219/commits/aec0a6a5bf36e5a866bb7b4b6c73f50db47786aa
What do you think?

maudetes: Yes, I find it better, thank you!

PR #3224, opened 2024-12-09
doc: add Windows reference to add etc/hosts rules

rv2931:
For dev stack and overall Windows dev stack.
Adding the equivalent method for adding local DNS solution on Windows. As Linux you need administrator permissions but it is working perfectly and is a not very well known solution of Windows users/developpers that simplify developments with docker


maudetes (approved):
Thank you for this documentation!

PR #3227, opened 2024-12-09
Organization url is now a URLField and should be a valid URL

maudetes:
Same as https://github.com/opendatateam/udata/pull/3222.

We should also apply a script to list existing invalid URL.


magopian (approved):
Nice! I wonder how many bogus organization URLs we'll have to fix in production this time around 😬 
🚢

maudetes:
On production env, `https://` has been added on all organizations with a failing URL validation :heavy_check_mark:

PR #3228, opened 2024-12-13
feat(topic): add datasets clear API

abulte:
This adds a `POST /api/2/topics/{id}/datasets/clear/` route which empties the datasets list of the given topic.

This is useful to sync a huge topic against a list of datasets, instead of relying on [ugly hacks](https://github.com/ecolabdata/ecospheres-universe/blob/31af696b47f71fa2e154ab6c38b55cc4a4f64233/feed-universe.py#L185) or [a very slow process](https://github.com/ecolabdata/ecospheres-universe/blob/31af696b47f71fa2e154ab6c38b55cc4a4f64233/feed-universe.py#L213). 

This also optimizes the datasets reindexation of a topic, by not calling `Dataset.fetch()` when triggering the reindex (I don't know why I did not do this before 🤔). This a performance gain on all topics write operations involving datasets.

Tested locally on a 66k datasets topic (no search service).

With optim:
```
❯ time curl -X POST -H "x-api-key: xxx"  http://dev.local:7000/api/2/topics/change-title/datasets/clear/

________________________________________________________
Executed in    9.14 secs      fish           external
   usr time    5.98 millis    0.11 millis    5.87 millis
   sys time   12.40 millis    1.07 millis   11.32 millis
```

Without optim:
```
❯ time curl -X POST -H "x-api-key: xxx"  http://dev.local:7000/api/2/topics/change-title/datasets/clear/

________________________________________________________
Executed in  133.37 secs      fish           external
   usr time    7.22 millis    0.10 millis    7.12 millis
   sys time   14.90 millis    1.16 millis   13.74 millis
```

Fix https://github.com/ecolabdata/ecospheres/issues/510

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current (in progress)
 
 - Fix the `parse-url` command [#3225](https://github.com/opendatateam/udata/pull/3225)
+- feat(topic): add datasets clear API [#3228](https://github.com/opendatateam/udata/pull/3228)
 
 ## 10.0.5 (2024-12-09)
 
--- a/udata/core/topic/apiv2.py
+++ b/udata/core/topic/apiv2.py
@@ -188,6 +188,22 @@ class TopicDatasetsAPI(API):
         return topic, 201
 
 
+@ns.route("/<topic:topic>/datasets/clear/", endpoint="topic_datasets_clear", doc=common_doc)
+class TopicDatasetsClearAPI(API):
+    @apiv2.secure
+    @apiv2.doc("topic_datasets_clear")
+    @apiv2.response(403, "Forbidden")
+    @apiv2.response(404, "Topic not found")
+    def post(self, topic):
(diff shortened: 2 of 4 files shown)


abulte:
`/<topic:topic>/datasets/clear/` overlaps `/<topic:topic>/datasets/<dataset:dataset>/`, even though the latter is only a DELETE and `clear` can not be an object id. Alternatively, we could use a POST parameter on `/<topic:topic>/datasets/` (`operation = delete` or smtg), or use use DELETE on `/<topic:topic>/datasets/` (not very rest-y).

maudetes:
I may miss something here, but can't we clear datasets already with API v1, as now tested in https://github.com/opendatateam/udata/pull/3233? :)

abulte:
@maudetes it works 🤯 And the fact that the API v1 payload is bloated is not a problem, since the topic will be empty 👌 

I think we still need the optim, it's quite slow (95 seconds on 66k datasets). I'll make this PR only about the optim.

PR #3229, opened 2024-12-17
Add with_drafts arg to posts API

ThibaudDauce:
Useful for https://github.com/datagouv/front-end/pull/217

### DIFF
--- a/udata/core/post/api.py
+++ b/udata/core/post/api.py
@@ -61,7 +61,11 @@ parser.add_argument(
     "sort", type=str, default="-created_at", location="args", help="The sorting attribute"
 )
 parser.add_argument(
-    "with_drafts", type=bool, default=False, location="args", help="`True` also returns the unpublished posts (only for super-admins)"
+    "with_drafts",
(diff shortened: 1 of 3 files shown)


magopian (approved):
Nice!
Would it make sense to have a test to avoid a regression in the future?

udata/core/post/api.py
```diff
@@ -70,11 +78,13 @@ class PostsAPI(API):
...
-        return (
-            Post.objects.published()
-            .order_by(args["sort"])
-            .paginate(args["page"], args["page_size"])
-        )
+
+        posts = Post.objects()
+
+        if not (AdminPermission().can() and args["with_drafts"]):
+            posts = posts.published()
```
maudetes, line 85: I think we could have this behavior by default, not needing a `with_drafts` parameter?
Since we're mostly going for a logic of getting objects to which you have access by default

maudetes (approved):
Thank you for this PR! :rocket: 
Minor comment that you may ignore

PR #3230, opened 2024-12-18
Fix failing dataset save in update reuses metrics

ThibaudDauce:


maudetes (approved):
Could you maybe add a bit of context on the PR description? :)

PR #3231, opened 2024-12-19
Fetch LazyReference manually in dataservice RDF

maudetes:
Related to https://github.com/datagouv/data.gouv.fr/issues/1562#issuecomment-2553183350

We want to prevent RAM increase that seems to occur when using implicit `LazyReference.fetch`.

We can deploy this fix on dev.data.gouv.fr to see if it fixes the [500 also occuring there](https://dev.data.gouv.fr/api/1/site/catalog.xml).

### DIFF
--- a/udata/core/dataservices/rdf.py
+++ b/udata/core/dataservices/rdf.py
@@ -156,8 +156,11 @@ def dataservice_to_rdf(dataservice: Dataservice, graph=None):
 
     if is_hvd:
         # We also want to automatically add any HVD category tags of the dataservice's datasets.
-        for dataset in dataservice.datasets:
-            if "hvd" not in dataset.tags:  # Only check HVD datasets for their categories.
+        for dataset_lazyref in dataservice.datasets:
+            # We fetch dataset manually to prevent RAM increase with LazyReference.fetch()
+            dataset = Dataset.objects.get(id=dataset_lazyref.id)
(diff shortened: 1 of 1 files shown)


udata/core/dataservices/rdf.py
```diff
@@ -156,12 +156,10 @@ def dataservice_to_rdf(dataservice: Dataservice, graph=None):
...
         # We also want to automatically add any HVD category tags of the dataservice's datasets.
-        for dataset in dataservice.datasets:
-            if "hvd" not in dataset.tags:  # Only check HVD datasets for their categories.
-                continue
-            for tag in dataset.tags:
-                if tag in TAG_TO_EU_HVD_CATEGORIES:
-                    hvd_category_tags.add(tag)
+        dataset_ids = [dat.id for dat in dataservice.datasets]
+        for tag in TAG_TO_EU_HVD_CATEGORIES:
+            if Dataset.objects(id__in=dataset_ids, tags="hvd").filter(tags=tag).first():
```
magopian, line 161: Very clever to use `.first()` to avoid returning everything when we just want to check if there's at least one 💯

magopian (approved):
It makes sense to indeed only get the `dataservice.datasets`' IDs, as it's a lazy reference field, which won't dereference the datasets, clever.

I'll have to remember to use this trick in the future 👍 

🚢

PR #3235, opened 2024-12-20
ci: faster ci

bolinocroustibat:
- Parallelise the execution of `assets` job among 3 CircleCI executors, similar to what have been done in https://github.com/datagouv/hydra/pull/238
- Add fallback cache

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,6 +9,7 @@ jobs:
       - image: redis:alpine
     environment:
       BASH_ENV: /root/.bashrc
+    parallelism: 4
     steps:
       - checkout
       - run:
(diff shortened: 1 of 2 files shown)


maudetes (commented):
I'm unclear about the way it works, but don't we now have 1,4K tests instead of 2K?

.circleci/config.yml
```diff
@@ -69,7 +68,62 @@ jobs:
...
+              "tests/forms/test_reference_field.py"
+              "tests/frontend/test_markdown.py"
+              "tests/test_model.py"
+              "tests/test_rdf.py"
+              "tests/test_routing.py"
+              "tests/test_uris.py"
+              "tests/test_utils.py"
+              # This test modifies the global app configuration
+              "tests/test_i18n.py"
+            )
```
ThibaudDauce, line 103: We need to maintain this list? Is it require?

ThibaudDauce (commented):
Same comment as @maudetes, I see 

```
Results (149.12s (0:02:29)):
     713 passed
       1 skipped
```

But on master we have:
```
=========== 2176 passed, 4 skipped, 33 warnings in 298.89s (0:04:58) ===========
```

Is it normal?

bolinocroustibat:
I suggest we abandon and close this PR — the feature would be difficult to maintain, and indeed, I’m not sure to what extent we can reliably ensure that all tests pass.

bolinocroustibat:
Appreciate you looking at it!

PR #3236, opened 2024-12-30
Fix function_field wrong name

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1615


ThibaudDauce (approved):
Good catch!

PR #3239, opened 2025-01-02
Fix updating ID for resource

ThibaudDauce:


bolinocroustibat (approved):
LGTM, that was quickly done!
Would that make sense, in the future, to create the ID on udata server side?

ThibaudDauce:
> LGTM, that was quickly done! Would that make sense, in the future, to create the ID on udata server side?

The front-end is using the client side ID generation to send in parallel all the chunks so not sure it's possible without changing the API (for exemple having a client side UUID then a real UUID saved in the database at the end of the upload process)… https://github.com/datagouv/front-end/blob/700fa2e57f9ae347c7b59e8eb1e5dbbff659a475/utils/datasets.ts#L175-L214

magopian (approved):
Nice! Thanks for the investigation and resolution, good job!

udata/core/dataset/api.py
```diff
@@ -379,6 +379,10 @@ def post(self, dataset):
         ResourceEditPermission(dataset).test()
         form = api.validate(ResourceForm)
         resource = Resource()
+
+        if form._fields.get("id").data in [r.id for r in dataset.resources]:
+            abort(400, "A resource with the same ID already exists.")
```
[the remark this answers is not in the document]
ThibaudDauce: Yes, we cannot check all ressource in the database is too costly since it's in multiple documents

udata/tests/api/test_datasets_api.py
```diff
@@ -558,6 +558,18 @@ def test_dataset_api_update(self):
...
+    def test_cannot_modify_dataset_id(self):
+        user = self.login()
+        dataset = DatasetFactory(owner=user)
+
+        data = dataset.to_dict()
+        data["id"] = "7776aa373aa050e302b5714d"
+
+        response = self.put(url_for("api.dataset", dataset=dataset.id), data)
+
+        self.assert200(response)
```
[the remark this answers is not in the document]
ThibaudDauce: Yes and no… I wanted to be sure we cannot update the ID in a "normal" document (not an embed)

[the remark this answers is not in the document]
ThibaudDauce: I could have try dataservice too since it used a different system

maudetes (commented):
Thank you for the investigation and proposed solution!

I don't think the resource id field is currently set by the frontend.
The `uuid` header in the multipart isn't currently used to create the Resource object, it is only used to identify the chunks the file belong to.
Comparing the uuid with the resource id shows they differ [in this chunk upload test case](https://github.com/opendatateam/udata/blob/f9370b3535476c499f87e83b8f13a101ecb42556/udata/tests/api/test_datasets_api.py#L1125).

I think we should make sure the resource id field is entirely readonly instead.

udata/core/dataset/api.py
```diff
@@ -377,8 +383,9 @@ class ResourcesAPI(API):
     def post(self, dataset):
         """Create a new resource for a given dataset"""
         ResourceEditPermission(dataset).test()
-        form = api.validate(ResourceForm)
-        resource = Resource()
+        form = api.validate(ResourceFormWithoutId)
```
maudetes, line 386: :clap:

udata/core/dataset/forms.py
```diff
@@ -104,6 +104,10 @@ class ResourceForm(BaseResourceForm):
     id = fields.UUIDField()
 
 
+class ResourceFormWithoutId(BaseResourceForm):
```
maudetes, line 107: I'm confused we use ResourceFormWithoutId in DatasetForm?

udata/core/dataset/models.py
```diff
@@ -638,6 +638,14 @@ def clean(self):
         if self.frequency in LEGACY_FREQUENCIES:
             self.frequency = LEGACY_FREQUENCIES[self.frequency]
 
+        resources_ids = set()
+        for resource in self.resources:
+            if resource.id in resources_ids:
+                raise MongoEngineValidationError(
+                    f"Duplicate resource ID {resource.id} in dataset #{self.id}."
+                )
+            resources_ids.add(resource.id)
```
maudetes, line 647: Could we do something like?
```
if len(set(self.resources)) != len(res.id for res in self.resources):
    raise...
```

udata/core/dataset/models.py
```diff
@@ -335,7 +335,7 @@ def to_mongo(self, *args, **kwargs):
 
 
 class ResourceMixin(object):
-    id = db.AutoUUIDField(primary_key=True)
+    id = db.AutoUUIDField(primary_key=True, unique=True)
```
maudetes, line 338: Do we need the manual check if we already have `unique=True`?

udata/core/dataset/models.py
```diff
@@ -638,6 +638,14 @@ def clean(self):
         if self.frequency in LEGACY_FREQUENCIES:
             self.frequency = LEGACY_FREQUENCIES[self.frequency]
 
+        resources_ids = set()
```
maudetes, line 641: These manual checks only care about the current dataset though?

maudetes (commented):
I think we'll need to deal with existing duplicated resources ID before merging & deploying?

udata/core/dataset/forms.py
ThibaudDauce: Because otherwise we don't have the ID for the reorder for exemple…

udata/core/dataset/models.py
ThibaudDauce: Yes I can remove this, I didn't find a way to make it work

udata/core/dataset/models.py
ThibaudDauce: Yes we cannot check for all resources in different dataset (but I think resources are always scoped by the dataset ID so it shouldn't be a problem, no?)

udata/core/dataset/models.py
ThibaudDauce: I removed it. Adding it fails a lot of tests I don't know why… https://app.circleci.com/pipelines/github/opendatateam/udata/5863/workflows/ee92a1f9-786c-42dc-a1e9-dc5b5a3be992/jobs/33495?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary

udata/core/dataset/models.py
maudetes: Sorry, I think I meant
```
if len(set(res.id for res in self.resources)) != len(self.resources):
    raise...
```

udata/core/dataset/models.py
maudetes: As discussed, I think there are some rare routes where it is the case, but let's deal with these later on.

udata/core/dataset/models.py
ThibaudDauce: Done in c69624788 (we just loose the `resource.id` in the exception) but it's simpler

udata/tests/dataset/test_dataset_model.py
```diff
@@ -59,6 +60,21 @@ def test_add_resource_without_checksum(self):
...
+        resource_a = ResourceFactory(id=uuid)
+        resource_b = ResourceFactory(id=uuid)
+
+        dataset.add_resource(resource_a)
+        dataset.add_resource(ResourceFactory())
+        try:
+            dataset.add_resource(resource_b)
+            assert False
+        except RuntimeError:
+            print("Ok!")
```
maudetes, line 76:
```suggestion
        with pytest.raises(RuntimeError):
            dataset.add_resource(resource_b)
```

udata/core/dataset/models.py
```diff
@@ -897,6 +900,9 @@ def get(cls, id_or_slug):
     def add_resource(self, resource):
         """Perform an atomic prepend for a new resource"""
         resource.validate()
+        if resource.id in [r.id for r in self.resources]:
+            raise RuntimeError("Cannot add resource with already existing ID")
```
maudetes, line 904: Should it be a RuntimeError by the way?

maudetes (approved):
I have minor comments but then we can merge I think!

udata/core/dataset/models.py
ThibaudDauce: I'm not sure about the policy on errors. Do we create a new error type? Do we use something generic?

maudetes: I would say we could use a `MongoEngineValidationError` as you did above?

udata/core/dataset/api.py
```diff
@@ -377,8 +383,9 @@ class ResourcesAPI(API):
...
+        resource = Resource(id=str(uuid4()))
```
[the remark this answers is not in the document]
ThibaudDauce: Removed in 35e775e1186ade0b607a3610fa251d262a8f487e

udata/tests/dataset/test_dataset_model.py
ThibaudDauce: Changed in 35e775e1186ade0b607a3610fa251d262a8f487e

udata/core/dataset/models.py
ThibaudDauce: Changed in 35e775e1186ade0b607a3610fa251d262a8f487e

PR #3241, opened 2025-01-06
Improve OGC service detection to expose in RDF

maudetes:
Follows on https://github.com/opendatateam/udata/pull/3203

Detect OGC Service on either of the following hints:
* the resource format is in the known list of OGC Service formats (already implemented in https://github.com/opendatateam/udata/pull/3203)
* the resource URL contains a REQUEST=GetCapabilities param and a corresponding SERVICE=format (new)

It used to expose an accessService on OGC distribution on [this distribution](https://www.data.gouv.fr/fr/datasets/66601a411fd7f3661df23bad/#/resources/39e6c520-48c4-43db-aac1-b8f07e44081a) (see [rdf](https://www.data.gouv.fr/api/1/datasets/66601a411fd7f3661df23bad/rdf.xml)).
We will also now exposes an accessService on [this distribution](https://www.data.gouv.fr/fr/datasets/installations-classees-pour-la-protection-de-lenvironnement-icpe-france-metropolitaine-et-drom-3/#/resources/cb550e4d-3c91-4209-b52c-b04c5cb43861) (see [rdf](https://www.data.gouv.fr/api/1/datasets/66631f6d5ff38efa62d385ac/rdf.xml)).

In an ideal world, we would like a more explicit information but the `request=getCapabilities` identification is actually the method used in the [SEMIC XLST conversion](https://github.com/SEMICeu/iso-19139-to-dcat-ap/blob/f61b2921dd398b90b2dd2db14085e75687f7616b/iso-19139-to-dcat-ap.xsl#L1419).

### DIFF
--- a/udata/core/dataset/rdf.py
+++ b/udata/core/dataset/rdf.py
@@ -131,7 +131,7 @@ def detect_ogc_service(resource):
     Detect if the resource points towards an OGC Service based on either
     * a known OGC Service format
     * a REQUEST=GetCapabilities param in url
-    It returns the OGC service type
+    It returns the OGC service type or None
     """
     if resource.format and resource.format.strip("ogc:") in OGC_SERVICE_FORMATS:
         return resource.format.strip("ogc:")
--- a/udata/core/dataset/rdf.py
+++ b/udata/core/dataset/rdf.py
@@ -6,6 +6,7 @@ import calendar
 import json
 import logging
 from datetime import date
+from typing import Optional
 
 from dateutil.parser import parse as parse_dt
 from flask import current_app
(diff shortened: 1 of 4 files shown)


udata/core/dataset/rdf.py
```diff
@@ -126,7 +126,23 @@ def owner_to_rdf(dataset, graph=None):
     return
 
 
-def ogc_service_to_rdf(dataset, resource, graph=None, is_hvd=False):
+def detect_ogc_service(resource):
```
magopian, line 129: Is it worth adding types?

magopian (approved):
Excellent work, well executed, thank!
🚢

udata/core/dataset/rdf.py
maudetes: Done in https://github.com/opendatateam/udata/pull/3241/commits/b7de4d998e93edecf402e5da35fabb3bf9acd1af

Also added on other `x_to_rdf` functions in this file since I was at it

PR #3242, opened 2025-01-08
Add integrity checks on non existent list field instead of empty

magopian:
Fixes [#1430](https://github.com/datagouv/data.gouv.fr/issues/1430)

## First approach (reverted)

The main issue seems to be because of a [bug in mongoengine](https://github.com/MongoEngine/mongoengine/issues/267).

This first draft is a hackish fix: when a `ListField` is updated, if it's changed to be the empty list (or if the result is an empty list, if you dropped the last element for example), instead of being set to `[]`, it's in fact unset (the field is removed from the document).

The added test showcases this.

This happens in [BaseDocument._delta](https://github.com/MongoEngine/mongoengine/blob/master/mongoengine/base/document.py#L766-L767): if the value isn't different from the default value (which is `[]` by default on a `ListField`), then the field is unset (instead of set to be empty).

The hackish fix updates the `datasets` field to be `[]` in the `post_save` signal. This means that we might want to do the same thing for `Dataset.resources`, and maybe for other `ListField`s just in case.

We might want to follow another route, maybe subclassing `MongoEngine.BaseDocument` or `MongoEngine.Document` to overwrite the `_delta` method, or the [`MongoEngine.Document._get_update_doc`](https://github.com/MongoEngine/mongoengine/blob/master/mongoengine/document.py#L528-L541) method to not unset a field if it's a `ListField`.

For the record, there's an existing [opened PR](https://github.com/MongoEngine/mongoengine/pull/2517/files) on mongoengine (that I didn't review, so no clue if it fixes it properly), but it's been open for 3 years already, so not sure if or when it'll be merged.

## Second approach (implemented here)

For now, instead of overriding mongoengine, we add a check on `check-integrity` command to spot issues with missing ListFields.
Since `check-integrity` runs exhaustively on models and their fields, it allows to monitor issues on any ListField.
We may want to update our strategy if there is some update at some point on the PR mentioned above.


udata/tests/dataservice/test_dataservice_model.py
```diff
@@ -0,0 +1,30 @@
...
+        # Setting it explicitely to a non-empty list should work as expected
+        dataservice.datasets = [dataset]
+        dataservice.save()
+        assert "datasets" in Dataservice._get_collection().find_one({"_id": dataservice.pk})
+        assert dataservice.datasets == [dataset]
+
+        # See https://github.com/MongoEngine/mongoengine/issues/267#issuecomment-283065318
+        # Setting it explicitely to an empty list should NOT remove the field.
+        dataservice.datasets = []
+        dataservice.save()
```
maudetes, line 27:
```suggestion
        dataservice.save()
        dataservice.reload()
```

udata/tests/dataservice/test_dataservice_model.py
```diff
@@ -0,0 +1,30 @@
...
+    def test_empty_datasets(self):
+        dataservice = DataserviceFactory()
+        dataset = DatasetFactory()
+        # If we're not setting the datasets ListField, it's set to the empty list.
+        assert "datasets" in Dataservice._get_collection().find_one({"_id": dataservice.pk})
+        assert dataservice.datasets == []
+
+        # Setting it explicitely to a non-empty list should work as expected
+        dataservice.datasets = [dataset]
+        dataservice.save()
```
maudetes, line 20:
```suggestion
        dataservice.save()
        dataservice.reload()
```

ThibaudDauce (approved):

PR #3243, opened 2025-01-09
More safeguards in the resource reorder endpoint

Pierlou:
Fix https://github.com/datagouv/data.gouv.fr/issues/1630


maudetes:
Checking for resources duplicates should already be done in https://github.com/opendatateam/udata/pull/3239.
However, we still need to make sure no resource is missing in the reorder body

maudetes:
> Checking for resources duplicates should already be done in #3239

Actually, it's a ValidationError that isn't properly returned to the user. Kept the explicit check in https://github.com/opendatateam/udata/pull/3243/commits/a559fd8bc2e4a0aa68a31e146678ef9636f909dd for better error response.

udata/tests/api/test_datasets_api.py
```diff
@@ -1366,17 +1414,10 @@ def test_bulk_update(self):
...
-                "title": faker.sentence(),
-                "description": faker.text(),
-                "url": faker.url(),
-            }
-        )
         response = self.put(url_for("api.resources", dataset=self.dataset), data)
         self.assert200(response)
         self.dataset.reload()
-        self.assertEqual(len(self.dataset.resources), 3)
+        self.assertEqual(len(self.dataset.resources), 2)
```
maudetes, line 1420: We used to be able to add resources on bulk update (PUT `/api/1/resources/`) somehow? We can't anymore.

Pierlou (commented):
Nice we can use it again without second thoughts! Do we want a test for the case of duplicated resources too?
(I can't approve because I am the author 🙃)

maudetes:
> Do we want a test for the case of duplicated resources too?

You mean [this test](https://github.com/opendatateam/udata/pull/3243/files#diff-36da3294e88ee041b97b82285b97d6d4b365e1835f506171ddab893ecd06b049R1316)?

maudetes (approved):

PR #3244, opened 2025-01-09
ci: fix version number building

bolinocroustibat:


.circleci/config.yml
```diff
@@ -142,31 +142,42 @@ jobs:
...
+                BASE_VERSION=$(python -c "from udata import __version__; print(__version__)")
+                if [[ "$CIRCLE_BRANCH" == feature/* ]]; then
+                    # This is a feature branch
+                    BUILDNO=$CIRCLE_BUILD_NUM+${CIRCLE_BRANCH#*/}
+                else
+                    # This is a simple development build
+                    BUILDNO=$CIRCLE_BUILD_NUM
+                fi
+                # Build the release version string with the build number, like "1.2.1.dev+1234"
+                RELEASE_VERSION="${BASE_VERSION}+${BUILDNO}"
```
maudetes, line 158: I think we don't have the plus sign
```suggestion
                # Build the release version string with the build number, like "1.2.1.dev1234"
                RELEASE_VERSION="${BASE_VERSION}${BUILDNO}"
```

maudetes (approved):
Thank you for this fix, if the CI shows value as expected, let's go with this

PR #3248, opened 2025-01-14
Add search to posts

ThibaudDauce:


maudetes (approved):
Thank you! Could you add a changelog and a minimal test maybe?

udata/core/post/models.py
```diff
@@ -42,6 +42,9 @@ class Post(db.Datetimed, db.Document):
         "indexes": [
             "-created_at",
             "-published",
+            "$name",
+            "$headline",
+            "$content",
```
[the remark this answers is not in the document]
maudetes: On udata restart, it occurs again.
Maybe it is related to multiple text index definition that should instead use this [`fields` structure example here](https://docs.mongoengine.org/guide/text-indexes.html#defining-a-document-with-text-index)?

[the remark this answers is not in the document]
maudetes: On udata restart, it occurs again.
Maybe it is related to multiple text index definition that should instead use this [`fields` structure example here](https://docs.mongoengine.org/guide/text-indexes.html#defining-a-document-with-text-index)?

maudetes (changes requested):

udata/core/post/models.py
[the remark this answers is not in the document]
ThibaudDauce: Done in d86a41755

udata/tests/plugin.py
```diff
@@ -130,6 +131,17 @@ def drop_db(app):
...
+        # When dropping the database, MongoEngine will keep the collection cached inside
+        # `_collection` (in memory). This cache is used to call `ensure_indexes` only on the
+        # first call to `_get_collection()`, on subsequent calls the value inside `_collection`
+        # is returned without calling `ensure_indexes`.
+        # In tests, the first test will have a clean memory state, so MongoEngine will initialise
+        # the collection and create the indexes, then the following test, with a clean database (no indexes)
+        # will have the collection cached, so MongoEngine will never create the indexes (except if `auto_create_index_on_save`
+        # is set on the model, which may be the reason it is present on most of the big models, we may remove it?)
+        model._collection = None
+
```
magopian, line 144: This must have been a nightmare to investigate, congratz for finding the issue!

udata/core/post/tests/test_api.py
```diff
@@ -1,55 +1,68 @@
...
+        """It should fetch a post list from the API"""
+        name_match = PostFactory(name="Foobar")
+        content_match = PostFactory(content="Foobar")
+        PostFactory(content="Something else")
+
+        response = self.get(url_for("api.posts", q="Foobar"))
+        assert200(response)
+        assert len(response.json["data"]) == 2
+
+        # Yes it's the wrong order but the search order (weights set in model) in not preserved
```
magopian, line 47: typo: `iS not preserved`

magopian (approved):
Nice! Thanks for this, wondering if we were expecting changes in the `udata/tests/features/territories/test_territories_api.py` test file which seems unrelated to this PR?

udata/core/post/api.py
```diff
@@ -84,6 +87,10 @@ def get(self):
         if not (AdminPermission().can() and args["with_drafts"]):
             posts = posts.published()
 
+        if args["q"]:
+            phrase_query = " ".join([f'"{elem}"' for elem in args["q"].split(" ")])
+            posts = posts.search_text(phrase_query)
+
         return posts.order_by(args["sort"]).paginate(args["page"], args["page_size"])
```
maudetes, line 94: I think we can use $text_score in case of a search text as in [dataset api](https://github.com/opendatateam/udata/blob/4b8fbb046bc84aa75f983e5f9adb504b4bcb7adf/udata/core/dataset/api.py#L221).

udata/tests/plugin.py
maudetes: +1

udata/tests/plugin.py
```diff
@@ -130,6 +131,17 @@ def drop_db(app):
...
     drop_db(app)
+    for model in get_all_models():
+        # When dropping the database, MongoEngine will keep the collection cached inside
+        # `_collection` (in memory). This cache is used to call `ensure_indexes` only on the
+        # first call to `_get_collection()`, on subsequent calls the value inside `_collection`
+        # is returned without calling `ensure_indexes`.
+        # In tests, the first test will have a clean memory state, so MongoEngine will initialise
+        # the collection and create the indexes, then the following test, with a clean database (no indexes)
+        # will have the collection cached, so MongoEngine will never create the indexes (except if `auto_create_index_on_save`
+        # is set on the model, which may be the reason it is present on most of the big models, we may remove it?)
```
maudetes, line 142: I think we can remove it! Maybe in a follow-up PR?

maudetes (approved):
Thank you for this PR that was somehow more complicated than expected!

udata/core/post/api.py
ThibaudDauce: Hum… Not sure to do that with the default value of the param… Because if someone is sorting by a column, we may want to keep his sort, no?

udata/tests/plugin.py
ThibaudDauce: Yes I'll do that in a following PR

udata/core/post/tests/test_api.py
ThibaudDauce: Done in a85684460

udata/core/post/api.py
ThibaudDauce: Done in 14f02a069 but not a big fan of how I implement it… But I don't see another way of doing this…

ThibaudDauce: I refactor a little bit in c840d701c

maudetes (approved):
Thank you! I have a minor comment

PR #3250, opened 2025-01-23
WIP : Use latest ISO to DCAT XSLT with saxonche

maudetes:
Related to https://github.com/datagouv/data.gouv.fr/issues/1512

Trying out if it works directly on our current tests.
We should however test for any unexpected diff in the harvesting with the new version.


maudetes:
We could split in two PRs (one for saxonche and one for testing the latest ISO to DCAT)

maudetes:
Will be done in https://github.com/opendatateam/udata/pull/3380 instead

PR #3254, opened 2025-01-30
Cannot update resource filetype after creation

ThibaudDauce:


udata/core/dataset/api.py
```diff
@@ -546,9 +546,17 @@ def put(self, dataset, rid):
         ResourceEditPermission(dataset).test()
         resource = self.get_resource_or_404(dataset, rid)
         form = api.validate(ResourceForm, resource)
+
+        if (
```
maudetes, line 550:
```suggestion
        # ensure filetype is not modified after creation
        if (
```

maudetes (approved):
:pray: thank you

PR #3258, opened 2025-01-31
Expose dataset_id for communautary resources

Pierlou:
Fixes https://github.com/datagouv/data.gouv.fr/issues/1645

Another possibility is to allow `Dataset.objects(resources__id=rid).first()` to work for communautary resources


maudetes:
Since you've worked on community resources @ThibaudDauce, how do you differentiate between community and embedded resource in API?

Should we expose this info explicitly in the API?

ThibaudDauce:
> Since you've worked on community resources @ThibaudDauce, how do you differentiate between community and embedded resource in API?
> 
> Should we expose this info explicitly in the API?

I think I checked for `owner` or `organization`

PR #3261, opened 2025-02-04
Fix wrong marshal for community resource uploads

ThibaudDauce:
When uploading a new community resource, we return only the fields of the resource (and not the owner for exemple, which is specific to community resources).

I need to have the owner to know if the response is from a community resource or a resource in the frontend. https://github.com/datagouv/front-end/pull/289


maudetes (approved):
:pray: thank you!

PR #3268, opened 2025-02-17
Add nested labels to dataset response

ThibaudDauce:
Don't know why it wasn't done this way at first. Is there a performance problem?


maudetes:
As discussed, I'm not convinced for now it's worth making the payload more complicated?

ThibaudDauce:
We can try to make it work with extra requests but I'm not sure it's a good solution either…

PR #3269, opened 2025-02-18
Allow a user to check its own membership requests

nicolaskempf57:
Required for https://github.com/datagouv/data.gouv.fr/issues/1646

### DIFF
--- a/udata/core/organization/api.py
+++ b/udata/core/organization/api.py
@@ -268,7 +268,9 @@ class MembershipRequestAPI(API):
                     403,
                     "You can only access your own membership requests or the one of your organizations.",
                 )
-            return [r for r in org.requests if r.user.id == args["user"]]
+            if args["status"]:
+                return [r for r in org.requests if (r.status == args["status"] and str(r.user.id) == args["user"])]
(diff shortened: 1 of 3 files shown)


udata/core/organization/api.py
```diff
@@ -255,8 +258,18 @@ class MembershipRequestAPI(API):
...
+        if args["user"]:
+            if not current_user.is_authenticated or (
+                str(current_user.id) != args["user"]
+                and not OrganizationPrivatePermission(org).can()
+            ):
+                api.abort(
+                    403,
+                    "You can only access your own membership requests or the one of your organizations.",
+                )
+            return [r for r in org.requests if r.user.id == args["user"]]
```
ThibaudDauce, line 271: You do not check the `status` param?

ThibaudDauce (commented):
Could you add a test?

udata/core/organization/api.py
nicolaskempf57: Indeed, after looking at the membership code, multiple requests are allowed for the same user and organization

ThibaudDauce (approved):

PR #3273, opened 2025-02-20
fix json dumps for non serializable stuff

Pierlou:


maudetes (approved):
I think you can add a link to this PR in the existing changelog entry :)

PR #3274, opened 2025-02-21
Add inactive users notification and deletion jobs

maudetes:
Closes https://github.com/datagouv/data.gouv.fr/issues/1529

## Process
We have two jobs, a notification and deletion job.

`notify-inactive-users` notifies all users by mail that will have their account deleted in 1 month (`DAYS_BEFORE_ACCOUNT_INACTIVITY_NOTIFY_DELAY`).
In this case, we set the notification date (`inactive_deletion_notified_at`) in User model.

`delete-inactive-users` deletes all users that are inactive and haven't logged in since notification date.
It also unsets the `inactive_deletion_notified_at` if the user has logged in since.

## Notes
1. :heavy_check_mark: We went with storing an `inactive_deletion_notified_at` to make sure we don't delete
   accounts that haven't been notified long enough in advance.
2. :heavy_check_mark: What about user contents (Dataset, Org, etc.)? Our confidentiality policy states : "Données relatives au Contributeur qui s’inscrit"
   We can keep user contents (discussion included).
   No need to be any more explicit than this.
3. :heavy_check_mark: For harmonization, kept discussion on user (self-)deletion. A new optionnal args has been added. See [udata-front related PR](https://github.com/datagouv/udata-front/pull/666).
4. :heavy_check_mark: Should we add a link to contact us if they can't connect? -> Nope
5. :heavy_check_mark:  Can we deal with ooooold inactive users with this system? Or should we do a specific Brevo campaign? -> Yes from a juridic point of view. Still need to see how to send these at a reasonable pace.
Added a `MAX_NUMBER_OF_USER_INACTIVITY_NOTIFICATIONS` config that should be set to a low value to get started and slowly increased. We currently have 70K+ user inactive since 3 years or more on the platform.


nicolaskempf57 (approved):
Working great except https://github.com/datagouv/data.gouv.fr/issues/1663

PR #3276, opened 2025-02-27
Add flask security 2FA

maudetes:
Part of https://github.com/datagouv/data.gouv.fr/issues/1674.

This is a first go at using Two-Factor Authentication with Flask Security.
See [the feature description](https://flask-security-too.readthedocs.io/en/stable/features.html#two-factor-authentication), [all existing configuration](https://flask-security-too.readthedocs.io/en/stable/configuration.html#two-factor) and a [code sample](https://flask-security-too.readthedocs.io/en/stable/two_factor_configurations.html) from their documentation.

## Notes
* upgraded some dependency -> need to look at the changelogs (should probably be done in a dedicated PR)
* templates are poorly copied and trimmed from [Flask Security templates folder](https://github.com/pallets-eco/flask-security/tree/main/flask_security/templates/security)
* it isn't clear yet how to enforce 2FA for some users only
* I haven't taken a look at recovery codes features yet
* we should make sure that it isn't useless due to ProConnect alternative login
* I played only with authenticator app for now


maudetes:
We can close for now and re-open when prioritize.

PR #3279, opened 2025-03-05
Investigate datasets' perfs with lots of resources

ThibaudDauce:
- [x] Limit the number of resources in JSON-LD in HTML (from 7s to 6,4s on http://dev.local:7000/fr/datasets/643671587b87c77bffaae082/)
- [x] Do not fetch the resources array from Mongo (from 1,4s to 0,16s on http://dev.local:7000/api/2/datasets/643671587b87c77bffaae082/)
- [x] Create /api/2/datasets (from 1min3s on v1 to 1,6s on v2 without showing resources, to 0,16s excluding resources from Mongo)
- [ ] Explored some alternative to re-order resources
    - ~~adding a `order` value in the resource but require to do migrations or having a default ordering? It's not easy to order the array during marshalling too… the reorder imply to update all nested resource (or having floating order value to insert value between to other?)~~
    - ~~using `$pull` `$push` in Mongo to reorder the array but it's not atomic so dangerous~~
    - [estelle] TODO in a dedicated PR
- ~~[ ] Fix HTML view by using the new `dataset_without_resources` converter~~ too much work because we need to convert fetch all the existing resources' types for this dataset to put the corrects `DatasetResources` components into the page, and the `noscript` tag is also using raw `resources` (I don't think the `max_resources_to_show` is working great for big arrays…)


ThibaudDauce:
Fix https://github.com/datagouv/data.gouv.fr/issues/1669

maudetes (approved):
Do you maybe wanna add some tests on new endpoint and resources length computation?

PR #3284, opened 2025-03-10
Add last_login_at for super admins

ThibaudDauce:
Fix https://github.com/datagouv/data.gouv.fr/issues/1660


maudetes:
Working on https://github.com/opendatateam/udata/pull/3274, I now think we should use `current_login_at` instead of `last_login_at` ([code](https://github.com/pallets-eco/flask-security/blob/4473f92db44f7c69575a16bd153cdb41872fe956/flask_security/utils.py#L197-L198)).
Do you understand the same? Could you modify the exposition in members as well if it's the case?

udata/core/user/api_fields.py
```diff
@@ -61,6 +63,13 @@
         "since": fields.ISODateTime(
             attribute="created_at", description="The registeration date", required=True
         ),
+        "last_login_at": fields.Raw(
+            attribute=lambda o: o.current_login_at
+            if current_user.is_authenticated and current_user.sysadmin
```
maudetes, line 68: Do you want to use `current_user_is_admin_or_self`? As was done for email

maudetes (approved):
Thank you! :pray:

PR #3285, opened 2025-03-13
Hard skip empty datasets nodes in dataservices

ThibaudDauce:


udata/core/dataset/rdf.py
```diff
@@ -751,6 +752,14 @@ def dataset_from_rdf(graph: Graph, dataset=None, node=None, remote_url_prefix: s
 
     dataset.title = rdf_value(d, DCT.title)
     if not dataset.title:
+        external_rdf = rdf_value(d, FOAF.isPrimaryTopicOf)
```
maudetes, line 755: I am unsure this is a valid way of identifying datasets that are just referenced.
I think updating the `missing title on dataset` with these considerations of why title is missing may be better?

udata/harvest/backends/dcat.py
```diff
@@ -154,11 +157,35 @@ def walk_graph(self, url: str, fmt: str) -> Generator[tuple[int, Graph], None, N
...
+        # These datasets are either present in the catalog in past or future pages or datasets
+        # external from the catalog we are currently harvesting (so we don't want to harvest them).
+        # First we thought of skipping them inside `dataset_from_rdf` (see :ExcludeExternalyDefinedDataset)
+        # but it creates a lot of "fake" items in the job and raising problems (reaching the max harvest item for
+        # example and not getting to the "real" datasets/dataservices in subsequent pages)
+        # We could check if the node is really empty before discarding it?
+        for node in page.subjects(RDF.type, DCAT.DataService):
+            dataservice = page.resource(node)
+            for dataset in dataservice.objects(DCAT.servesDataset):
+                if dataset_node == dataset.identifier:
```
maudetes, line 181: Are we sure it works if a dataset is properly described in the catalog but is also served by a dataservice?

maudetes (commented):
Thank you for taking a look at this! :pray:

udata/core/dataset/rdf.py
ThibaudDauce: It may not be the only way but it is a way. The only problem is if there is a `isPrimaryTopicOf` but it's not why there is no title? Otherwise we may show "missing title" for referenced dataset in some case but it's not a big deal? (we can add more conditions later if we encounter them?)

udata/harvest/backends/dcat.py
ThibaudDauce: No it doesn't work in some case (I was waiting for the tests and they fail…), need to find a better way to exclude them…

udata/harvest/backends/dcat.py
```diff
@@ -146,25 +157,51 @@ def walk_graph(self, url: str, fmt: str) -> Generator[tuple[int, Graph], None, N
...
+        # if `isPrimaryTopicOf` is present.
+        # `isPrimaryTopicOf` is the tag present in the first harvester raising the problem, it may exists other
+        # values of the same sort we need to check here.
+        # This is not dangerous because we check for missing title in `dataset_from_rdf` later so we would have skipped
+        # this dataset anyway.
+
+        resource = page.resource(node)
+
+        title = rdf_value(resource, DCT.title)
+        primary_topic = rdf_value(resource, FOAF.isPrimaryTopicOf)
```
maudetes, line 195: I don't trust the `isPrimaryTopicOf` heuristic approach. I think all datasets described by GeoNetwork and transformed with the XSLT end up with this property. It's not only the case in the `servesDataset` context.
See for example this [random dataset DCAT output](https://geodcat-ap.semic.eu/api/?outputSchema=core&src=https%3A%2F%2Fcatalogue-formation.sigloire.fr%2Fgeonetwork%2Fsrv%2Fapi%2Frecords%2Ffr-120066022-jdd-73fbab7e-e7cf-4b12-b88b-5b91efc7b6de%2Fformatters%2Fxml&outputFormat=text%2Fhtml).

Ideally, we would like to exclude datasets that are *only* predicate of `servesDataset`  and `hasPart` explicitly?
Indeed, they are not directly bound to the harvested catalogue at this moment, we wouldn't want to harvest them if they only appear as reference only?

ThibaudDauce: I didn't find another way to filter these nodes. But I don't exclude all the nodes with `isPrimaryTopicOf`, only nodes without title (so they will be skipped anyway), I don't think there is a lot of datasets without titles and `isPrimaryTopicOf` that are real datasets?

udata/harvest/backends/dcat.py
```diff
@@ -146,25 +156,51 @@ def walk_graph(self, url: str, fmt: str) -> Generator[tuple[int, Graph], None, N
...
+            if self.is_dataset_external_to_this_page(page, node):
+                continue
+
             self.process_dataset(remote_id, page_number=page_number, page=page, node=node)
 
-            if self.is_done():
+            if self.has_reached_max_items():
                 return
 
+    def is_dataset_external_to_this_page(self, page: Graph, node) -> bool:
```
maudetes, line 175: Let's check if the title is missing as well to make sure that we don't lose datasets that were currently harvested

udata/harvest/backends/dcat.py
```diff
@@ -77,9 +77,19 @@ def inner_harvest(self):
             self.process_one_datasets_page(page_number, page)
             serialized_graphs.append(page.serialize(format=fmt, indent=None))
 
+        # We do a second pass to have all datasets in memory and attach datasets
+        # to dataservices. It could be better to be one pass of graph walking and
+        # then one pass of attaching datasets to dataservices.
         for page_number, page in self.walk_graph(self.source.url, fmt):
             self.process_one_dataservices_page(page_number, page)
 
+        if not self.dryrun and self.has_reached_max_items():
```
maudetes, line 86: :+1:

udata/harvest/backends/base.py
```diff
@@ -256,7 +256,7 @@ def process_dataset(self, remote_id: str, **kwargs):
             ]
             self.save_job()
 
-    def is_done(self) -> bool:
```
maudetes, line 259: We'll need to update [ckan backend](https://github.com/opendatateam/udata-ckan/blob/b7cf3c680ca643eff8bb9d5ecf58dad25a05d721/udata_ckan/harvesters.py#L127)

udata/harvest/backends/dcat.py
maudetes: I prefer the alternative done in https://github.com/opendatateam/udata/pull/3285/commits/ca1bc595dcafbd9e818405336fe29ccd86752a05, in conjunction with a check on a missing title

udata/harvest/backends/dcat.py
ThibaudDauce: Done in 604e3c2f0

maudetes (approved):
Thank you a lot! :clap:

PR #3290, opened 2025-03-20
Compute quality score to show correct resources score in api v2

ThibaudDauce:


udata/core/dataset/models.py
```diff
@@ -824,8 +828,7 @@ def next_update(self):
         else:
             return self.last_update + delta
 
-    @cached_property
-    def quality(self):
+    def compute_quality(self):
```
maudetes, line 831: I think computing `update_fulfilled_in_time` at save time doesn't make sense though

udata/migrations/2025-03-20-save-quality-for-datasets.py
maudetes, line 1: We'll need to make the migration painless since it takes more than one hour to run

udata/core/dataset/models.py
ThibaudDauce: Yes… Don't know what to do about this…

udata/core/dataset/models.py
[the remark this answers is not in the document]
ThibaudDauce: Done in a60fe863c

udata/core/dataset/models.py
ThibaudDauce: Fixed in 06fa31458

udata/migrations/2025-03-20-save-quality-for-datasets.py
ThibaudDauce: Done in https://github.com/opendatateam/udata/pull/3290/commits/06fa314580cad89dcef4571bb7cd3830441d25ef

udata/core/dataset/models.py
```diff
@@ -824,8 +828,34 @@ def next_update(self):
         else:
             return self.last_update + delta
 
-    @cached_property
+    @property
     def quality(self):
```
maudetes, line 832: I wonder if keeping the existing `quality` cached_property but instead storing a `resource_quality` DictField on save wouldn't it make more readable. We wouldn't need a migration?

maudetes (commented):
I actually think we also have the issue for `last_update`, that is based on resources `last_modified`

ThibaudDauce:
> I actually think we also have the issue for `last_update`, that is based on resources `last_modified`

Hum… You're right there is a `datetime.utcnow()` in `last_modified` we need to not save in the cache…

udata/core/dataset/models.py
ThibaudDauce: Not sure how it solves our problem. In my new system, we don't need the migration (because we fallback to the computed version if there is no cache).

And not sure how `resource_quality` change something really major about the new system. Yes we could separate and have cache only for resources but it doesn't reduce the complexity? We still need a cache value, a computed value?

maudetes (approved):
Thank you!
I think as a good check of any score side effect, we could compare scores on [resource catalog](https://dev.data.gouv.fr/fr/datasets/catalogue-des-donnees-de-data-gouv-fr/) before and after this is deployed in dev :)

maudetes:
Migration applied successfully! See [errors that occurred due to invalid datasets](https://errors.data.gouv.fr/organizations/sentry/issues/?project=12&query=logger%3Audata.migrations.2025-03-20-save-quality-for-datasets&referrer=event-tags-table&statsPeriod=90d&utc=true)

PR #3292, opened 2025-04-01
Refactor error management

ThibaudDauce:
- [x] Add an `error` key to the response to know which fields have errors
- [x] Auto abort with 400 (it wasn't always the case, so we may have failed with 500 instead of 400 on `ValidationError`

Before:

```
{
    'message': "ValidationError (Organization:67ebc7aed08a1f46a0537d37) (custom.The dictionary does not contain the mandatory keys: 'title', 'description', 'type'. type.Type 'None' of 'color' should be one of: ['str', 'int', 'float', 'bool', 'datetime', 'date', 'choice']: ['extras'])"
}

```

After:

```
{
    'errors': {
        'extras': {
            'custom': "The dictionary does not contain the mandatory keys: 'title', 'description', 'type'.",
            'type': "Type 'None' of 'color' should be one of: ['str', 'int', 'float', 'bool', 'datetime', 'date', 'choice']"
        }
    },
    'message': "ValidationError (Organization:67ebc7aed08a1f46a0537d37) (custom.The dictionary does not contain the mandatory keys: 'title', 'description', 'type'. type.Type 'None' of 'color' should be one of: ['str', 'int', 'float', 'bool', 'datetime', 'date', 'choice']: ['extras'])"
}
```


udata/api/__init__.py
```diff
@@ -258,17 +258,57 @@ def handle_unauthorized_file_type(error):
...
+
+
+@api.errorhandler(mongoengine.errors.ValidationError)
+@api.marshal_with(validation_error_fields, code=400)
+def handle_validation_error(error: mongoengine.errors.ValidationError):
+    """Error returned when validation failed."""
+    return (
+        {
+            "errors": convert_object_of_exceptions_to_object_of_strings(error.errors),
+            "message": str(error),
```
maudetes, line 301: Do we have only duplicated info between `errors` and `message`?

udata/tests/api/test_dataservices_api.py
```diff
@@ -431,9 +431,7 @@ def test_dataservice_api_create_with_unkwown_license(self):
             },
         )
         self.assert400(response)
-        self.assertEqual(
-            response.json["errors"]["license"], ["Unknown reference 'unwkown-license'"]
-        )
+        self.assertEqual(response.json["errors"]["license"], "Unknown reference 'unwkown-license'")
```
maudetes, line 434: We used to have both `list[str]` and `str` in errors fields?
Do we now have `str` only everwhere?

maudetes (approved):
Thank you a lot for this error handler! :clap: 

My comments are mostly questions :blush:

udata/core/dataset/forms.py
```diff
@@ -54,7 +54,7 @@ def validate(self, extra_validators=None):
             )
         except FieldValidationError as err:
             field = getattr(self, err.field)
-            field.errors.append(err.message)
+            field.errors.append(str(err))
```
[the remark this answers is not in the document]
ThibaudDauce: Because str(err) use err.raw_message instead of err.message which is modified by ValidationError (to add the errors in fields)

udata/api/__init__.py
ThibaudDauce: Mostly yes. Message is formatted different for `FieldValidationError` and `ValidationError`

udata/api/__init__.py
```diff
@@ -258,17 +258,57 @@ def handle_unauthorized_file_type(error):
...
+)
+validation_error_fields = api.model(
+    "ValidationError",
+    {"errors": fields.Raw, "message": fields.String},
+)
+
+validation_error_fields_v2 = apiv2.model(
+    "ValidationError",
+    {"errors": fields.Raw, "message": fields.String},
+)
```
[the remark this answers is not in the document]
ThibaudDauce: Good catch, didn't know we can do this :-) Done in fa5983ff2

PR #3297, opened 2025-04-15
Exclude deleted dataservices in csv queryset

maudetes:
Dataservice has [a `deleted_at` property](https://github.com/opendatateam/udata/blob/9d2d993537061b6e57a142d4a5fb70e985183d5f/udata/core/dataservices/models.py#L205).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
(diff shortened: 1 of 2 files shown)


Pierlou (approved):
Thanks 🙏 but why do we have two attributes that state deletion, datasets for instance only have one right?

maudetes:
> Thanks 🙏 but why do we have two attributes that state deletion, datasets for instance only have one right?

The `get_queryset` is generic and most objects have a `deleted` property except dataservice that has a `deleted_at`.

PR #3214, opened 2024-12-03
Return all the datasets a user has access to

magopian:
Fixes [#1575](https://github.com/datagouv/data.gouv.fr/issues/1575)

PR #3298, opened 2025-04-16
Add archived, deleted and private filter in dataset list api

maudetes:


udata/core/dataset/api.py
```diff
@@ -116,6 +117,24 @@ def __init__(self):
...
+            "archived",
+            type=boolean,
+            location="args",
+            help="If set to true, it will filter on archived datasets only. If set to false, it will exclude archived datasets. User must be authenticated and results are limited to user visibility",
+        )
+        self.parser.add_argument(
+            "deleted",
+            type=boolean,
+            location="args",
+            help="If set to true, it will filter on deleted datasets only. If set to false, it will exclude deleted datasets. User must be authenticated and results are limited to user visibility",
```
ThibaudDauce, line 130: Deleted datasets are not filtered by default?

udata/core/dataset/api.py
```diff
@@ -116,6 +117,24 @@ def __init__(self):
         self.parser.add_argument("topic", type=str, location="args")
         self.parser.add_argument("credit", type=str, location="args")
         self.parser.add_argument("dataservice", type=str, location="args")
+        self.parser.add_argument(
+            "archived",
+            type=boolean,
+            location="args",
+            help="If set to true, it will filter on archived datasets only. If set to false, it will exclude archived datasets. User must be authenticated and results are limited to user visibility",
```
ThibaudDauce, line 124: Same here, archived dataset are not filtered by default?

udata/core/dataset/api.py
```diff
@@ -98,7 +99,7 @@ def __init__(self):
         super().__init__()
         self.parser.add_argument("tag", type=str, location="args", action="append")
         self.parser.add_argument("license", type=str, location="args")
-        self.parser.add_argument("featured", type=bool, location="args")
+        self.parser.add_argument("featured", type=boolean, location="args")
```
ThibaudDauce, line 102: Could you add the same `help` text as below?

udata/core/dataset/api.py
maudetes: not anymore since [this PR](https://github.com/opendatateam/udata/pull/3214) :)

udata/core/dataset/api.py
maudetes: same, not since [this PR](https://github.com/opendatateam/udata/pull/3214) :)

PR #3300, opened 2025-04-23
Migrate MAAF backend to udata

ThibaudDauce:
See https://github.com/datagouv/udata-front/pull/688

### DIFF
--- a/udata/harvest/backends/__init__.py
+++ b/udata/harvest/backends/__init__.py
@@ -15,3 +15,4 @@ def get_all(app):
 
 
 from .base import BaseBackend, HarvestFeature, HarvestFilter, HarvestExtraConfig  # noqa
+from .maaf import MaafBackend  # noqa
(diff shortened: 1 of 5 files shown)


udata/harvest/backends/__init__.py
```diff
@@ -15,3 +15,4 @@ def get_all(app):
 
 
 from .base import BaseBackend, HarvestFeature, HarvestFilter, HarvestExtraConfig  # noqa
+from .maaf import MaafBackend  # noqa
```
maudetes, line 18: Do we need this import here?

maudetes (approved):
:ok_hand: thank you!

udata/harvest/backends/__init__.py
ThibaudDauce: Doesn't seem… Removed in add6ca2ff

PR #3262, opened 2025-02-05
Add new dataservices fields

ThibaudDauce:
See https://github.com/datagouv/udata-front/pull/655
See https://github.com/datagouv/front-end/pull/338

PR #3301, opened 2025-04-24
feat(topics): add featured filter in API

abulte:
Adds a `featured` filter through query arg in the topics list API.

### DIFF
--- a/udata/core/topic/parsers.py
+++ b/udata/core/topic/parsers.py
@@ -1,4 +1,5 @@
 from bson.objectid import ObjectId
+from flask_restx.inputs import boolean
 
 from udata.api import api
 from udata.api.parsers import ModelApiParser
@@ -20,7 +21,7 @@ class TopicApiParser(ModelApiParser):
         self.parser.add_argument("granularity", type=str, location="args")
         self.parser.add_argument("organization", type=str, location="args")
         self.parser.add_argument("owner", type=str, location="args")
-        self.parser.add_argument("featured", type=bool, location="args")
+        self.parser.add_argument("featured", type=boolean, location="args")
 
     @staticmethod
     def parse_filters(topics, args):
@@ -39,7 +40,7 @@ class TopicApiParser(ModelApiParser):
             topics = topics.filter(spatial__zones=args["geozone"])
         if args.get("granularity"):
             topics = topics.filter(spatial__granularity=args["granularity"])
-        if args.get("featured"):
+        if args.get("featured") is not None:
             topics = topics.filter(featured=args["featured"])
         if args.get("organization"):
             if not ObjectId.is_valid(args["organization"]):
--- a/udata/core/topic/parsers.py
+++ b/udata/core/topic/parsers.py
@@ -20,6 +20,7 @@ class TopicApiParser(ModelApiParser):
         self.parser.add_argument("granularity", type=str, location="args")
         self.parser.add_argument("organization", type=str, location="args")
         self.parser.add_argument("owner", type=str, location="args")
+        self.parser.add_argument("featured", type=bool, location="args")
 
     @staticmethod
     def parse_filters(topics, args):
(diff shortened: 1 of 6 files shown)


udata/core/topic/parsers.py
```diff
@@ -20,6 +20,7 @@ def __init__(self, with_include_private=True):
         self.parser.add_argument("granularity", type=str, location="args")
         self.parser.add_argument("organization", type=str, location="args")
         self.parser.add_argument("owner", type=str, location="args")
+        self.parser.add_argument("featured", type=bool, location="args")
```
maudetes, line 23: We should use [flask_restx boolean](https://flask-restx.readthedocs.io/en/latest/api.html#flask_restx.inputs.boolean) I think.
I made the change for `/api/1/datasets` in https://github.com/opendatateam/udata/pull/3298

```suggestion
        self.parser.add_argument("featured", type=boolean, location="args")
```

ThibaudDauce:
Can you update the changelog?

udata/tests/api/test_topics_api.py
```diff
@@ -83,6 +84,11 @@ def test_topic_api_list(self):
         self.assertEqual(len(response.json["data"]), 1)
         self.assertIn(str(granularity_topic.id), [t["id"] for t in response.json["data"]])
 
+        response = self.get(url_for("api.topics", featured="true"))
+        self.assert200(response)
+        self.assertEqual(len(response.json["data"]), 1)
+        self.assertIn(str(featured_topic.id), [t["id"] for t in response.json["data"]])
+
```
maudetes, line 91: The way I see it, this should work but doesn't with `type=bool`
```suggestion

        response = self.get(url_for("api.topics", featured="false"))
        self.assert200(response)
        self.assertEqual(len(response.json["data"]), 7)
        self.assertIn(str(featured_topic.id), [t["id"] for t in response.json["data"]])
```

maudetes (commented):
Don't forget the changelog :)

udata/core/topic/parsers.py
abulte: Nice! Looks like much better parsing. Should I migrate `include_private` in the same PR so that we have the same behavior on both booleans?

udata/tests/api/test_topics_api.py
abulte: I had to change the model for this to work :-/ https://github.com/opendatateam/udata/pull/3301/files#diff-f3da585bd02f3f0b4060a6980b1a0ff7467e10d7bb1a19aaf8a044273547b27bR24

Should I write a migration? On my machine, all Topics created from API seem to have a boolean value, but shell created ones have `None`.

udata/core/topic/parsers.py
abulte: Yeah, let's keep the private mayhem for later :p

udata/tests/api/test_topics_api.py
abulte: I'll write one.

abulte: I guess I could force `featured=True`, save the objects, and then set `featured=False`? Since we don't care about featured Topics so far, it shouldn't be a problem. But it is ugly :-/

maudetes: Am I missing something?
On this branch, having a previously created topic without `featured` field, I can set `featured=False` successfully with `.update`.
```
In [2]: Topic.objects(featured__exists=False).count()
Out[2]: 1

In [3]: Topic.objects(featured__exists=False).update(featured=False)
Out[3]: 1

In [4]: Topic.objects(featured__exists=False).count()
Out[4]: 0
```

maudetes: And we've used a similar logic [in dataservices recently](https://github.com/opendatateam/udata/pull/3262/files#diff-8349b3199575575b59667edd4d444d44441ef9b2df958558b97b5c81a232f5de)

maudetes (approved):
Thank you! :rocket:

abulte:
@maudetes pas de merge pas de release 😢 Next time? 🥺

PR #3311, opened 2025-05-13
Private objects should return 404 by api

maudetes:
It would prevent showing private objects on cdata for non-authorized users.

It was already the case for RDF get endpoints (on datasets and dataservices).

Do we want to apply this logic to topics? I'm unsure about the usage of `private` by ecosphere (cc @streino)


nicolaskempf57 (approved):
I started to work on it [here](https://github.com/opendatateam/udata/pull/3312), I'll close mine :)

PR #3312, opened 2025-05-13
Send 404 for private non editable elements

nicolaskempf57:


nicolaskempf57:
Done in https://github.com/opendatateam/udata/pull/3311

PR #3318, opened 2025-05-16
feat: topic elements

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/594

Technical design doc: https://docs.numerique.gouv.fr/docs/ef75df93-54b8-4f26-a9df-fa16467b1427/

### DIFF
--- a/udata/core/topic/apiv2.py
+++ b/udata/core/topic/apiv2.py
@@ -7,11 +7,9 @@ from flask_security import current_user
 
 from udata.api import API, apiv2, fields
 from udata.core.dataset.api import DatasetApiParser
-from udata.core.dataset.apiv2 import dataset_page_fields
 from udata.core.dataset.models import Dataset
 from udata.core.organization.api_fields import org_ref_fields
 from udata.core.reuse.api import ReuseApiParser
-from udata.core.reuse.models import Reuse
 from udata.core.spatial.api_fields import spatial_coverage_fields
 from udata.core.topic.models import Topic
 from udata.core.topic.parsers import TopicApiParser
@@ -33,6 +31,7 @@ reuse_parser = ReuseApiParser()
 common_doc = {"params": {"topic": "The topic ID"}}
 
 
+# FIXME: move to fields submodule
 topic_fields = apiv2.model(
     "Topic",
     {
@@ -45,35 +44,20 @@ topic_fields = apiv2.model(
         "tags": fields.List(
             fields.String, description="Some keywords to help in search", required=True
         ),
-        "datasets": fields.Raw(
+        "elements": fields.Raw(
             attribute=lambda o: {
                 "rel": "subsection",
                 "href": url_for(
-                    "apiv2.topic_datasets",
+                    "apiv2.topic_elements",
                     topic=o.id,
                     page=1,
                     page_size=DEFAULT_PAGE_SIZE,
                     _external=True,
                 ),
                 "type": "GET",
-                "total": len(o.datasets),
+                "total": len(o.elements),
             },
-            description="Link to the topic datasets",
-        ),
-        "reuses": fields.Raw(
-            attribute=lambda o: {
-                "rel": "subsection",
-                "href": url_for(
-                    "apiv2.topic_reuses",
-                    topic=o.id,
-                    page=1,
-                    page_size=DEFAULT_PAGE_SIZE,
-                    _external=True,
-                ),
-                "type": "GET",
-                "total": len(o.reuses),
-            },
-            description="Link to the topic reuses",
+            description="Link to the topic elements",
         ),
         "featured": fields.Boolean(description="Is the topic featured"),
         "private": fields.Boolean(description="Is the topic private"),
@@ -109,6 +93,49 @@ topic_fields = apiv2.model(
 
 topic_page_fields = apiv2.model("TopicPage", fields.pager(topic_fields))
 
+nested_element_fields = apiv2.model(
+    "NestedTopicElement",
+    {
+        "class": fields.ClassName(description="The model class", required=True),
+        "id": fields.String(description="The object identifier", required=True),
+    },
+)
+
+element_fields = apiv2.model(
+    "TopicElement",
+    {
+        "id": fields.String(description="The element id"),
+        "title": fields.String(description="The element title"),
+        "description": fields.String(description="The element description"),
+        "element": fields.Nested(nested_element_fields, description="The element target object"),
+    },
+)
+
+element_page_fields = apiv2.model(
+    "TopicElementPage",
+    {
+        "data": fields.List(fields.Nested(element_fields, description="The topic elements")),
+        "next_page": fields.String(),
+        "previous_page": fields.String(),
+        "page": fields.Integer(),
+        "page_size": fields.Integer(),
+        "total": fields.Integer(),
+    },
+)
+
+# FIXME: move to parsers submodule
+elements_parser = apiv2.parser()
+elements_parser.add_argument("page", type=int, default=1, location="args", help="The page to fetch")
+elements_parser.add_argument(
+    "page_size", type=int, default=DEFAULT_PAGE_SIZE, location="args", help="The page size to fetch"
+)
+elements_parser.add_argument(
+    "type", type=str, location="args", help="The type of resources to fetch"
+)
+elements_parser.add_argument(
+    "q", type=str, location="args", help="query string to search through elements"
+)
+
 
 @ns.route("/", endpoint="topics_list")
 class TopicsAPI(API):
@@ -142,22 +169,56 @@ topic_add_items_fields = apiv2.model(
 )
 
 
-@ns.route("/<topic:topic>/datasets/", endpoint="topic_datasets", doc=common_doc)
-class TopicDatasetsAPI(API):
-    @apiv2.doc("topic_datasets")
-    @apiv2.expect(dataset_parser.parser)
-    @apiv2.marshal_with(dataset_page_fields)
+@ns.route("/<topic:topic>/elements/", endpoint="topic_elements", doc=common_doc)
+class TopicElementsAPI(API):
+    @apiv2.doc("topic_elements")
+    @apiv2.expect(elements_parser)
+    @apiv2.marshal_with(element_page_fields)
     def get(self, topic):
-        """Get a given topic datasets, with filters"""
-        args = dataset_parser.parse()
-        args["topic"] = topic.id
-        datasets = Dataset.objects(archived=None, deleted=None, private=False)
-        datasets = dataset_parser.parse_filters(datasets, args)
-        sort = args["sort"] or ("$text_score" if args["q"] else None) or "-created_at_internal"
-        return datasets.order_by(sort).paginate(args["page"], args["page_size"])
+        """Get a given topic elements, with filters"""
+        args = elements_parser.parse_args()
+        page = args["page"]
+        page_size = args["page_size"]
+        list_elements_url = url_for("apiv2.topic_elements", topic=topic.id, _external=True)
+        next_page = f"{list_elements_url}?page={page + 1}&page_size={page_size}"
+        previous_page = f"{list_elements_url}?page={page - 1}&page_size={page_size}"
+
+        # FIXME: this will probably fail with a lot of elements
+        # is there an efficient way to handle embedded documents pagination? or shall we use an Element document?
+        # -> aggregation pipeline is probably the way to go
+        res = topic.elements
+
+        # FIXME: implement class/type filter
+        if args["type"]:
+            res = [elem for elem in res if elem["type"] == args["type"]]
+            next_page += f"&type={args['type']}"
+            previous_page += f"&type={args['type']}"
+
+        # FIXME: implement on description too, and be smarter
+        if args["q"]:
+            res = [elem for elem in res if args["q"].lower() in elem["title"].lower()]
+            next_page += f"&q={args['q']}"
+            previous_page += f"&q={args['q']}"
+
+        if page > 1:
+            offset = page_size * (page - 1)
+        else:
+            offset = 0
+        paginated_result = res[offset : (page_size + offset if page_size is not None else None)]
+
+        print("elements", [e.id for e in topic.elements])
+
+        return {
+            "data": paginated_result,
+            "next_page": next_page if page_size + offset < len(res) else None,
+            "page": page,
+            "page_size": page_size,
+            "previous_page": previous_page if page > 1 else None,
+            "total": len(res),
+        }
 
     @apiv2.secure
-    @apiv2.doc("topic_datasets_create")
+    @apiv2.doc("topic_elements_create")
     @apiv2.expect([topic_add_items_fields])
     @apiv2.marshal_with(topic_fields)
     @apiv2.response(400, "Malformed object id(s) in request")
@@ -190,96 +251,23 @@ class TopicDatasetsAPI(API):
 
 
 @ns.route(
-    "/<topic:topic>/datasets/<dataset:dataset>/",
-    endpoint="topic_dataset",
-    doc={"params": {"topic": "The topic ID", "dataset": "The dataset ID"}},
+    "/<topic:topic>/elements/<uuid:element_id>/",
+    endpoint="topic_element",
+    doc={"params": {"topic": "The topic ID", "element": "The element ID"}},
 )
-class TopicDatasetAPI(API):
+class TopicElementAPI(API):
     @apiv2.secure
     @apiv2.response(404, "Topic not found")
-    @apiv2.response(404, "Dataset not found in topic")
+    @apiv2.response(404, "Element not found in topic")
     @apiv2.response(204, "Success")
-    def delete(self, topic, dataset):
+    def delete(self, topic, element_id):
         """Delete a given dataset from the given topic"""
         if not TopicEditPermission(topic).can():
             apiv2.abort(403, "Forbidden")
 
-        if dataset.id not in (d.id for d in topic.datasets):
-            apiv2.abort(404, "Dataset not found in topic")
-        topic.datasets = [d for d in topic.datasets if d.id != dataset.id]
-        topic.save()
-
-        return None, 204
-
-
-@ns.route("/<topic:topic>/reuses/", endpoint="topic_reuses", doc=common_doc)
-class TopicReusesAPI(API):
-    @apiv2.doc("topic_reuses")
-    @apiv2.expect(reuse_parser.parser)
-    @apiv2.marshal_with(Reuse.__page_fields__)
-    def get(self, topic):
-        """Get a given topic reuses, with filters"""
-        args = reuse_parser.parse()
-        reuses = Reuse.objects(deleted=None, private__ne=True).filter(
-            id__in=[d.id for d in topic.reuses]
-        )
-        # warning: topic in reuse_parser is different from Topic
-        reuses = reuse_parser.parse_filters(reuses, args)
-        sort = args["sort"] or ("$text_score" if args["q"] else None) or DEFAULT_SORTING
-        return reuses.order_by(sort).paginate(args["page"], args["page_size"])
-
-    @apiv2.secure
-    @apiv2.doc("topic_reuses_create")
-    @apiv2.expect([topic_add_items_fields])
-    @apiv2.marshal_with(topic_fields)
-    @apiv2.response(400, "Malformed object id(s) in request")
-    @apiv2.response(400, "Expecting a list")
-    @apiv2.response(400, "Expecting a list of dicts with id attribute")
-    @apiv2.response(404, "Topic not found")
-    @apiv2.response(403, "Forbidden")
-    def post(self, topic):
-        """Add reuses to a given topic from a list of reuses ids"""
-        if not TopicEditPermission(topic).can():
-            apiv2.abort(403, "Forbidden")
-
-        data = request.json
-
-        if not isinstance(data, list):
-            apiv2.abort(400, "Expecting a list")
-        if not all(isinstance(d, dict) and d.get("id") for d in data):
-            apiv2.abort(400, "Expecting a list of dicts with id attribute")
-
-        try:
-            reuses = Reuse.objects.filter(id__in=[r["id"] for r in data]).only("id")
-            diff = set(d.id for d in reuses) - set(d.id for d in topic.reuses)
-        except mongoengine.errors.ValidationError:
-            apiv2.abort(400, "Malformed object id(s) in request")
-
-        if diff:
-            topic.reuses += [ObjectId(rid) for rid in diff]
-            topic.save()
-
-        return topic, 201
-
-
-@ns.route(
-    "/<topic:topic>/reuses/<reuse:reuse>/",
-    endpoint="topic_reuse",
-    doc={"params": {"topic": "The topic ID", "reuse": "The reuse ID"}},
-)
-class TopicReuseAPI(API):
-    @apiv2.secure
-    @apiv2.response(404, "Topic not found")
-    @apiv2.response(404, "Reuse not found in topic")
-    @apiv2.response(204, "Success")
-    def delete(self, topic, reuse):
-        """Delete a given reuse from the given topic"""
-        if not TopicEditPermission(topic).can():
-            apiv2.abort(403, "Forbidden")
-
-        if reuse.id not in (d.id for d in topic.reuses):
-            apiv2.abort(404, "Reuse not found in topic")
-        topic.reuses = [d for d in topic.reuses if d.id != reuse.id]
+        if element_id not in (elt.id for elt in topic.elements):
+            apiv2.abort(404, "Element not found in topic")
+        topic.elements = [elt for elt in topic.elements if elt.id != element_id]
         topic.save()
 
         return None, 204
--- a/udata/core/topic/factories.py
+++ b/udata/core/topic/factories.py
@@ -2,10 +2,21 @@ import factory
 
 from udata import utils
 from udata.core.dataset.factories import DatasetFactory
-from udata.core.reuse.factories import VisibleReuseFactory
 from udata.factories import ModelFactory
 
-from .models import Topic
+from .models import Topic, TopicElement
+
+
+class TopicElementFactory(ModelFactory):
+    class Meta:
+        model = TopicElement
+
+    title = factory.Faker("sentence")
+    description = factory.Faker("text")
+
+
+class TopicElementDatasetFactory(TopicElementFactory):
+    element = factory.SubFactory(DatasetFactory)
 
 
 class TopicFactory(ModelFactory):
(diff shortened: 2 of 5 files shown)


udata/core/dataset/api.py
```diff
@@ -192,7 +192,13 @@ def parse_filters(datasets, args):
             except Topic.DoesNotExist:
                 pass
             else:
-                datasets = datasets.filter(id__in=[d.id for d in topic.datasets])
+                datasets = datasets.filter(
```
geoffreyaldebert, line 195: maybe add this to all elements supported by topics, not only datasets(or maybe in another PR but we can track an issue)

udata/core/dataset/search.py
```diff
@@ -68,7 +68,7 @@ def serialize(cls, dataset):
         organization = None
         owner = None
 
-        topics = Topic.objects(datasets=dataset).only("id")
+        topics = Topic.objects(elements__element=dataset).only("id")
```
geoffreyaldebert, line 71: same

udata/core/topic/factories.py
```diff
@@ -2,10 +2,39 @@
...
+            "description": elt["description"],
+        }
+
+    @classmethod
+    def as_payload(cls) -> dict:
+        elt = cls.as_dict()
+        return cls.element_as_payload(elt)
+
+
+class TopicElementDatasetFactory(TopicElementFactory):
```
geoffreyaldebert, line 32: maybe not in this PR but we will need to add dataservice elements (specifically for simplifions.data.gouv.fr) to topics, and maybe organizations ?

geoffreyaldebert (commented):
You make my dream come true. But I'll let our supreme tech lead @maudetes approve.

udata/core/dataset/api.py
abulte: My goal here was to stay isofunctionnal, i.e handling Datasets and Reuses in a Topic. The Reuse API does have a `topic=` filter, but it does not refer to Topics, it's a different concept 😬 Cf `REUSE_TOPICS`.

udata/core/topic/factories.py
abulte: Like I said in https://github.com/opendatateam/udata/pull/3318/files#r2137116502, I think a good goal for this PR is to implement what was previously possible, and we can add other objects later.

udata/core/dataset/api.py
abulte: @maudetes maybe we should rename this field? Something like `Reuse.theme` and then use `Reuse.topic` like `Dataset.topic`.

udata/core/topic/models.py
```diff
@@ -8,7 +8,16 @@
...
+__all__ = ("Topic", "TopicElement")
+
+
+class TopicElement(db.EmbeddedDocument):
+    id = field(db.AutoUUIDField(primary_key=True))
+    title = field(db.StringField(required=False))
+    description = field(db.StringField(required=False))
+    tags = field(db.ListField(db.StringField()))
+    extras = field(db.ExtrasField())
+    element = field(db.GenericReferenceField())
```
abulte:
```suggestion
    element = field(db.GenericReferenceField(choices=[Dataset, Reuse]))
```

Maybe we should be explicit about which objects are fully supported here. WDYT @maudetes?

udata/core/topic/factories.py
abulte: Dataservice support in https://github.com/ecolabdata/udata/pull/3

udata/core/topic/api_fields.py
```diff
@@ -0,0 +1,137 @@
...
+        "uri": fields.UrlFor(
+            "apiv2.topic", lambda o: {"topic": o}, description="The topic API URI", readonly=True
+        ),
+        "page": fields.UrlFor(
+            "topics.display",
+            lambda o: {"topic": o},
+            description="The topic page URL",
+            readonly=True,
+            fallback_endpoint="apiv2.topic",
+        ),
```
ThibaudDauce: `fields.UrlFor()` is removed in https://github.com/opendatateam/udata/pull/3345

udata/core/topic/models.py
```diff
@@ -21,8 +32,7 @@ class Topic(db.Datetimed, Auditable, db.Document, Owned):
     tags = field(db.ListField(db.StringField()))
     color = field(db.IntField())
 
-    datasets = field(db.ListField(db.LazyReferenceField("Dataset", reverse_delete_rule=db.PULL)))
-    reuses = field(db.ListField(db.LazyReferenceField("Reuse", reverse_delete_rule=db.PULL)))
+    elements = field(db.EmbeddedDocumentListField(TopicElement))
```
ThibaudDauce: Not a big fan of big list of embed documents in Mongo. We have a lot of problems with datasets with big list of resources. Have you try with a separate "table"?

udata/core/topic/apiv2.py
```diff
@@ -132,37 +64,156 @@ def get(self, topic):
...
-        args = dataset_parser.parse()
-        args["topic"] = topic.id
-        datasets = Dataset.objects(archived=None, deleted=None, private=False)
-        datasets = dataset_parser.parse_filters(datasets, args)
-        sort = args["sort"] or ("$text_score" if args["q"] else None) or "-created_at_internal"
-        return datasets.order_by(sort).paginate(args["page"], args["page_size"])
+        """
+        Get a given topic elements with pagination.
+        We're using an aggregation pipeline instead of high level MongoEngine API for performance reasons.
+        """
```
ThibaudDauce: I think changing embeds to real documents will solve this?

udata/core/topic/forms.py
```diff
@@ -2,9 +2,26 @@
...
+    title = fields.StringField(_("Title"))
+    description = fields.StringField(_("Description"))
+    tags = fields.TagField(_("Tags"))
+    extras = fields.ExtrasField()
+    element = fields.ModelField(_("Element"))
+
+    def validate(self, extra_validators=None):
+        validation = super().validate(extra_validators)
+        if not self.element.data and not self.title.data:
+            self.element.errors.append(_("A topic element must have a title or an element."))
```
ThibaudDauce, line 26: Could you explain in a comment the use case for a `TopicElement` with just a title and no element?

udata/core/topic/forms.py
```diff
@@ -2,9 +2,26 @@
 from udata.forms import ModelForm, fields, validators
```
ThibaudDauce, line 2: Have you try using our new api_fields system to avoid having to create forms?

udata/core/topic/api_fields.py
```diff
@@ -0,0 +1,137 @@
+from flask import url_for
```
ThibaudDauce, line 1: Have you try our new api_fields system to avoid having to create most of these fields manually?

udata/core/topic/apiv2.py
```diff
@@ -132,37 +64,156 @@ def get(self, topic):
...
-    @apiv2.expect([topic_add_items_fields])
+    @apiv2.doc("topic_elements_create")
+    @apiv2.expect([topic_add_elements_fields])
     @apiv2.marshal_with(topic_fields)
-    @apiv2.response(400, "Malformed object id(s) in request")
     @apiv2.response(400, "Expecting a list")
-    @apiv2.response(400, "Expecting a list of dicts with id attribute")
     @apiv2.response(404, "Topic not found")
     @apiv2.response(403, "Forbidden")
     def post(self, topic):
```
ThibaudDauce, line 115: Do you really need a batch insert endpoint? A simple insert one is not enough?

abulte: It shaves off a lot of time for operations on huge Topic, e.g. in https://github.com/ecolabdata/ecospheres-universe.

udata/core/topic/forms.py
abulte: I did not try, but I'd rather do that in another PR. It felt easier to migrate with the same pattern on source and destination.

udata/core/topic/api_fields.py
abulte: https://github.com/opendatateam/udata/pull/3318#discussion_r2209735151

udata/core/topic/models.py
abulte: Done ✅

udata/core/dataset/tasks.py
```diff
@@ -44,7 +44,7 @@ def purge_datasets(self):
         Activity.objects(related_to=dataset).delete()
         # Remove topics' related dataset
         # FIXME: is is acceptable to have a broken reference inbetween? seems similar to other objects.
-        Topic.objects(elements__element=dataset).update(set__elements__S__element=None)
+        TopicElement.objects(element=dataset).update(element=None)
```
ThibaudDauce, line 46: Why not delete here?

udata/core/reuse/tasks.py
```diff
@@ -23,7 +23,7 @@ def purge_reuses(self) -> None:
         # Remove transfers
         Transfer.objects(subject=reuse).delete()
         # Remove reuses references in Topics
-        Topic.objects(elements__element=reuse).update(set__elements__S__element=None)
+        TopicElement.objects(element=reuse).update(element=None)
```
ThibaudDauce, line 26: Same, why not delete here?

udata/core/topic/apiv2.py
```diff
@@ -127,37 +65,52 @@ def get(self, topic):
...
+    @apiv2.doc("delete_topic")
+    @apiv2.response(204, "Object deleted")
+    @apiv2.response(403, "Forbidden")
+    def delete(self, topic):
+        """Delete a given topic"""
+        if not TopicEditPermission(topic).can():
+            apiv2.abort(403, "Forbidden")
+        # Remove discussions linked to the topic
+        Discussion.objects(subject=topic).delete()
+        topic.delete()
```
ThibaudDauce, line 90: No soft delete? With a `deleted_at` and a `purge`? Not sure if it's worth it.

udata/core/topic/models.py
```diff
@@ -20,8 +39,11 @@ class Topic(db.Datetimed, Auditable, db.Document, Owned):
     tags = field(db.ListField(db.StringField()))
     color = field(db.IntField())
 
-    datasets = field(db.ListField(db.LazyReferenceField("Dataset", reverse_delete_rule=db.PULL)))
-    reuses = field(db.ListField(db.LazyReferenceField("Reuse", reverse_delete_rule=db.PULL)))
+    elements = field(
+        db.ListField(
+            db.LazyReferenceField("TopicElement", reverse_delete_rule=db.PULL, passthrough=True)
+        )
+    )
```
ThibaudDauce: I would go one step further and have a `topic` reference in `TopicElement` and nothing (no list) in `Topic`.

udata/core/topic/apiv2.py
```diff
@@ -127,37 +65,52 @@ def get(self, topic):
...
-        args = dataset_parser.parse()
-        args["topic"] = topic.id
-        datasets = Dataset.objects(archived=None, deleted=None, private=False)
-        datasets = dataset_parser.parse_filters(datasets, args)
-        sort = args["sort"] or ("$text_score" if args["q"] else None) or "-created_at_internal"
-        return datasets.order_by(sort).paginate(args["page"], args["page_size"])
+        """Get a given topic's elements with pagination."""
+        args = elements_parser.parse()
+        elements = elements_parser.parse_filters(
+            TopicElement.objects(id__in=[e.pk for e in topic.elements]),
```
ThibaudDauce: With a `Topic` reference inside `TopicElement` it would simply be `TopicElement.objects(topic=topic)`, no? See my comment on the model.

udata/core/topic/apiv2.py
```diff
@@ -168,113 +121,91 @@ def post(self, topic):
...
+        elements = []
+        for element_data in data:
+            form = TopicElementForm.from_json(element_data)
+            if not form.validate():
+                errors.append(form.errors)
+            else:
+                element = TopicElement()
+                form.populate_obj(element)
+                element.save()
+                elements.append(element)
```
ThibaudDauce, line 134: Not needed if you use a foreign reference in `TopicElement` instead of a list of reference in `Topic`

udata/core/topic/apiv2.py
```diff
@@ -168,113 +121,91 @@ def post(self, topic):
...
-            topic.datasets += [ObjectId(did) for did in diff]
-            topic.save()
+        if errors:
+            apiv2.abort(400, errors=errors)
 
-        return topic, 201
+        for element in elements:
+            topic.elements.insert(0, element)
 
+        topic.save()
```
ThibaudDauce: Not needed if you use a foreign reference in TopicElement instead of a list of reference in Topic

abulte:
@ThibaudDauce I refactored the TopicElement to Topic link here: https://github.com/ecolabdata/udata/pull/5, against this branch. I'm not totally convinced it's better. 

Cons:
- It feels less idiomatic re Mongo
- The form processing is even more complicated (this might not be a problem mid to long term if we switch to API fields)
- We loose the future ability to link a TopicElement to multiple Topics (not sure if we'd need that @streino @geoffreyaldebert)

Pros:
- Elements queries related to a topic are faster, this https://github.com/opendatateam/udata/pull/3318/files#diff-479230526cf856f3f5ac586257924b5a455a19fcaea8cdfdae47fa1d7a007d14R104 is heavy on a huge Topic
- Elements queries are slightly simpler (but not that much after all) since we don't manage an array of refs anymore

udata/core/dataset/tasks.py
abulte: A TopicElement without element can be valid. For example, on ecologie.data.gouv.fr, a Bouquet can link to a dataset that is not on data.gouv.fr, and thus have no element in database. If a dataset is removed on data.gouv.fr, we don't want to erase the title, description... and intention the user filled in.

udata/core/reuse/tasks.py
abulte: Same :-)

udata/core/topic/apiv2.py
abulte: I don't think we need that ATM. Could be useful in the future.

udata/core/topic/models.py
abulte: https://github.com/opendatateam/udata/pull/3318#issuecomment-3095748013

udata/core/topic/apiv2.py
abulte: Yep, this one is definitely better with the alternative model.

abulte:
> I'm not sure to understand why it should be more complicated. We replace list manipulation by simple post/delete on a new object. We could not accept the `elements = fields.NestedModelList(TopicElementForm)` in the `Topic` forms to simplify a little bit this. (and require a second call to create the `TopicElement` with the batch insert endpoint. But not sure it's worth it.

It's more complicated because we now have a chicken-and-egg problem: TopicElement needs a Topic to be fully valid, but form processing typically allows "dry run" where Topic is not saved before TopicElement is created.

The two-steps API could be a solution, but API usage would be more difficult. Creating and updating a Topic with its elements in one call is very nice. This makes state management easier on the frontend.

The current form implementation seems to work, although it's not great. We'll probably have a few bugs in the future, but I can handle them.

ThibaudDauce:
> > I'm not sure to understand why it should be more complicated. We replace list manipulation by simple post/delete on a new object. We could not accept the `elements = fields.NestedModelList(TopicElementForm)` in the `Topic` forms to simplify a little bit this. (and require a second call to create the `TopicElement` with the batch insert endpoint. But not sure it's worth it.
> 
> It's more complicated because we now have a chicken-and-egg problem: TopicElement needs a Topic to be fully valid, but form processing typically allows "dry run" where Topic is not saved before TopicElement is created.
> 
> The two-steps API could be a solution, but API usage would be more difficult. Creating and updating a Topic with its elements in one call is very nice. This makes state management easier on the frontend.
> 
> The current form implementation seems to work, although it's not great. We'll probably have a few bugs in the future, but I can handle them.

I don't undertand why we need this feature "but form processing typically allows "dry run" where Topic is not saved before TopicElement is created."

We can save a topic without `TopicElement` is perfectly valid? If we don't want to see the topic during the creation of a lot of `TopicElement` I more in favor of a `draft` solution like `Dataset`, `Dataservice` and `Reuse`

streino:
> * We loose the future ability to link a TopicElement to multiple Topics (not sure if we'd need that @streino @geoffreyaldebert)

I can't think of a case where we'd want this behavior. To me a TopicElement is comparable to a message in a discussion or a resource in a dataset, so something that has a single "owner" object.

We *might* want to link a single TopicElement to several Topics at the level of "bouquet templates", but it's not clear if we're ever going to implement templates, and if using instances of TopicElement would be the better way to do it.

udata/core/topic/api_fields.py
```diff
@@ -0,0 +1,103 @@
...
+            description="The topic API URI",
+            readonly=True,
+        ),
+        "extras": fields.Raw(description="Extras attributes as key-value pairs"),
+    },
+)
+
+topic_page_fields = apiv2.model("TopicPage", fields.pager(topic_fields))
+
+nested_element_fields = apiv2.model(
```
maudetes: It means we now only get class & id when calling the hateoas, in particular if the element itself doesn't have a description or title.
It is not as complete as [what we had before](https://dev.data.gouv.fr/api/2/topics/65b76f15d7874915c8e41298/datasets/?page=1&page_size=20), but it may not be an issue if front clients use the datasets or reuse api with a topic filter if they are in the context of one or the other

udata/core/topic/apiv2.py
```diff
@@ -127,37 +65,52 @@ def get(self, topic):
         """Get a given topic"""
         return topic
 
+    @apiv2.secure
+    @apiv2.doc("update_topic")
+    @apiv2.expect(topic_input_fields)
```
maudetes, line 69: Slightly confused about the fact that we expect `topic_input_fields` on PUT but `topic_fields` on POST?

udata/core/topic/apiv2.py
```diff
@@ -168,113 +121,91 @@ def post(self, topic):
...
-    @apiv2.response(404, "Dataset not found in topic")
-    @apiv2.response(204, "Success")
-    def delete(self, topic, dataset):
-        """Delete a given dataset from the given topic"""
+    @apiv2.response(403, "Forbidden")
+    def delete(self, topic):
+        """Delete all elements from a Topic
+
+        This a workaround for https://github.com/kvesteri/wtforms-json/issues/43
+        -> we can't use PUT /api/2/topics/{topic}/ with an empty list of elements
```
maudetes, line 155: Shouldn't we make this work instead of adding this api call?

udata/core/topic/api_fields.py
```diff
@@ -0,0 +1,103 @@
...
+    "NestedTopicElement",
```
maudetes: Could we use the existing `api.model_reference`? :)

udata/core/topic/apiv2.py
```diff
@@ -127,37 +65,52 @@ def get(self, topic):
...
+            TopicElement.objects(id__in=[e.pk for e in topic.elements]),
+            args,
+        )
+        return elements.paginate(args["page"], args["page_size"])
 
     @apiv2.secure
-    @apiv2.doc("topic_datasets_create")
-    @apiv2.expect([topic_add_items_fields])
+    @apiv2.doc("topic_elements_create")
+    @apiv2.expect([topic_add_elements_fields])
```
maudetes: I'm confused about this expect. `topic_add_elements_fields` is only composed of an `id` but we actually need a nested `element` with `class` + `id` infos?
Don't we want `[element_fields]`?

udata/core/topic/forms.py
```diff
@@ -2,9 +2,30 @@
...
+    tags = fields.TagField(_("Tags"))
+    extras = fields.ExtrasField()
+    element = fields.ModelField(_("Element"))
+
+    def validate(self, extra_validators=None):
+        """
+        Make sure that either title or element is set.
+        (Empty nested element is a valid use case for "placeholder" elements)
+        """
+        validation = super().validate(extra_validators)
```
maudetes, line 24: I have an issue with `csrf_token`missing, that I think is coming from this validation (either with curl or swaggerUI).

udata/migrations/2025-05-26-migrate-topics-to-elements.py
```diff
@@ -0,0 +1,64 @@
...
+            element_id = ObjectId()
+            element_doc = {
+                "_id": element_id,
+                "tags": [],
+                "extras": {},
+                "element": {"_cls": "Dataset", "_ref": DBRef("dataset", dataset_id)},
+            }
+
+            # Insert TopicElement document
+            get_db().topic_element.insert_one(element_doc)
```
maudetes, line 31: Could we create TopicElement using the mongoengine document instead?

maudetes (commented):
Thank you for this nice PR! I think this model and the alternative are quite straightforward.

I would say the API fields and forms are the most challenging, but it doesn't seem to differentiate greatly between the two.

Regarding the two approaches, I think the alternative solution preserves us with future performance issues? And the 1-1 relation seems a reasonable limit for now.

udata/migrations/2025-05-26-migrate-topics-to-elements.py
abulte: Is it a good idea to mix the paradigms in the same migration?

udata/core/topic/forms.py
abulte: What kind of request are you making? This works fine for me:

```
curl --request POST \
  --url http://dev.local:7000/api/2/topics/ \
  --header 'Content-Type: application/json' \
  --header 'x-api-key: xxx' \
  --data '{
    "name": "Bouquet de test API 1",
    "description": "Bouquet de test migration\n\n- Th\u00e9matique Mieux consommer\n- 4 facteurs, 1 pour chaque availabilty\n- 2 groupes, 1 pour \"missing\" et l'\''autre \"available\"\n- Couverture territoriale Ardennes\n- Organisation DDT Maine et Loire",
    "tags": [
        "ecospheres",
        "ecospheres-theme-mieux-consommer"
    ],
    "elements": [
      {
        "title": "xxx",
        "element": {
          "class": "Dataset",
          "id": "686b9feb3b039b0a3a146f6e"
        }
      }
    ]
}
'
```

udata/core/topic/apiv2.py
abulte: True! Fixed.

udata/core/topic/api_fields.py
abulte: Eagle eye 🦅 Fixed! I kept an alias for readability.

udata/core/topic/apiv2.py
abulte: Maybe, but the issue has been open for 9 years... This would probably mean forking wtforms-json, I don't feel like stepping into this rabbit hole right now 😬

udata/core/topic/apiv2.py
abulte: Right, should be `topic_input_fields` everywhere, fixed.

udata/core/topic/api_fields.py
abulte: True. I did the migration for frontends and it seems good enough. Having a payload that would work for all models and be useful would be challenging anyway.

udata/core/topic/api_fields.py
abulte: Actually, no :-) `model_reference` is not available on apiv2. I reverted with a better name. Not sure it's worth it declaring it in apiv2?

udata/core/topic/forms.py
maudetes: It's on `api/2/topics/<topic_id>/elements/`
```
curl -X 'POST' \
  'http://dev.local:7000/api/2/topics/682d942102634f7cc798472e/elements/' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -H "X-Api-Key: $API_KEY" \
  -d '[
  {
    "title": "mon élément"
  }
]'
```

udata/migrations/2025-05-26-migrate-topics-to-elements.py
maudetes: :shrug: Maybe not, I was thinking this because resulting underlying object might slightly differ (having an empty `tags = []` vs tags missing, etc.). Not sure it matters much

udata/core/topic/api_fields.py
maudetes: I think you can inherit from apiv1 ([example](https://github.com/opendatateam/udata/blob/91915ec268665f1650b5a792c00fc0b0b9d03644/udata/core/dataset/apiv2.py#L255))

udata/core/topic/forms.py
maudetes: Could you reproduce the issue?

udata/core/topic/apiv2.py
maudetes: Yes agreed! I was maybe thinking of a manual hack on our side, not relying on wtforms.
Example for [schema and checksum for datasets](https://github.com/opendatateam/udata/blob/4e748b5c8b19bf6457337b9cb77756560abc5d0a/udata/core/dataset/api.py#L695-L701).
If you think it's better to have a dedicated endpoint, we can keep it :)

udata/core/topic/forms.py
abulte: Nope, works fine for me 🤔

udata/core/topic/api_fields.py
abulte: A bit convoluted but I think it works now.

udata/core/topic/apiv2.py
abulte: We're already using the dedicated endpoint, so I think it's alright to keep it.

udata/core/topic/apiv2.py
```diff
@@ -168,113 +120,83 @@ def post(self, topic):
...
 
-        try:
-            datasets = Dataset.objects.filter(id__in=[d["id"] for d in data]).only("id")
-            diff = set(d.id for d in datasets) - set(d.id for d in topic.datasets)
-        except mongoengine.errors.ValidationError:
-            apiv2.abort(400, "Malformed object id(s) in request")
+        errors = []
+        elements = []
+        for element_data in data:
+            form = TopicElementForm.from_json(element_data)
```
maudetes: I think we should set meta csrf as False when using `ModelForm.from_json` directly (as in done in [api.validate](https://github.com/opendatateam/udata/blob/95dd559fb984d5a9d705f035fb89b811fafaa146/udata/api/__init__.py#L153-L154))
```suggestion
            form = TopicElementForm.from_json(element_data, meta={"csrf": False})
```
Not sure why you don't reproduce the issue though

abulte:
> I think we're landing on this PR! Thank you again for this hard work :)

🎉 

> We'll need an update on the [different front migrations](https://github.com/ecolabdata/ecospheres/issues/604) and to synchronize our deployment!

Yeah, I'll need at least to update them against main/master 🙈. Then we can have a quick meeting to organize the migration?

PR #3319, opened 2025-05-16
Remove `downloads` in dataset csv adapter

maudetes:
It was a duplicate of `metric.resources_downloads`, added in https://github.com/opendatateam/udata-metrics/pull/15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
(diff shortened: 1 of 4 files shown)


maudetes:
> Since we want to remove plugins, do we remove it here or from udata-metrics?

Since the other trafic metrics are computed [there](https://github.com/opendatateam/udata-metrics/blob/493af6ff2b627a41fa2fde9f3b953e92cbe71979/udata_metrics/tasks.py#L120), we should move them all at once I would say

ThibaudDauce (approved):

PR #3325, opened 2025-05-22
Add new global site metrics

ThibaudDauce:


udata/core/metrics/commands.py
```diff
@@ -60,6 +60,8 @@ def update(
             site.count_max_org_followers()
             site.count_max_org_reuses()
             site.count_max_org_datasets()
+            site.count_stock_metrics()
+            site.save()
```
maudetes: We don't need the `site.save()`?

udata/core/metrics/helpers.py
```diff
@@ -0,0 +1,102 @@
...
+def metrics_by_label(monthly_metrics: Dict, metrics_labels: List[str]) -> List[OrderedDict]:
+    metrics_by_label = []
+    for label in metrics_labels:
+        metrics_by_label.append(
+            OrderedDict((month, monthly_metrics[month][label]) for month in monthly_metrics)
+        )
+    return metrics_by_label
+
+
+def get_metrics_for_model(
```
maudetes, line 46: Maybe we should call it more explicitly `get_visit_metrics_for_model`?

udata/core/site/models.py
```diff
@@ -172,6 +187,24 @@ def count_max_org_datasets(self):
         self.metrics["max_org_datasets"] = org.metrics["datasets"] if org else 0
         self.save()
 
+    def count_stock_metrics(self):
```
maudetes, line 190: I think instead of counting the entire stock metrics for each object daily, we could only compute the value for the current month for example.
Since we're stocking it in the object, there is no need to re-compute it each time anymore.

udata/core/metrics/helpers.py
ThibaudDauce: It's not specific to visits, no? We call it with `download_resource` for example

udata/core/site/models.py
ThibaudDauce: Do you think it's costly? Can we optimise later if it becomes a problem? Because I'm afraid this way of doing is more code and I prefer to limit the complexity and the eventual inconsistency problems (it's more stable to get a fresh computation every night). If it's really slow we could do this in a second PR?

udata/core/metrics/commands.py
ThibaudDauce: Done in 438ab819e

udata/core/site/models.py
maudetes: I haven't tried to run it, but I think @nicolaskempf57 did?

udata/core/metrics/helpers.py
maudetes: Indeed! It was to differentiate it with "stock" metrics, but I'm not very inspired by the wording so this may do

udata/core/site/models.py
nicolaskempf57: udata is not really fast on mac but it was indeed really long like more than 1h

ThibaudDauce: On my computer it's instant

ThibaudDauce: Really weird… What command do you run? I run `udata metrics update -s`

nicolaskempf57: You're right, the site metrics are almost instant ! I even missed the first `Update site metrics` with `udata metrics update`

udata/core/metrics/helpers.py
ThibaudDauce: If I find a better name I will create a new PR it's not a big change :-)

udata/core/site/models.py
maudetes: :+1: I think it will be objects metrics that will cause trouble

PR #3333, opened 2025-06-04
Add topic activities

maudetes:
Closes https://github.com/datagouv/data.gouv.fr/issues/1735

Similarly to what has been done to dataservices in https://github.com/opendatateam/udata/pull/3308.

There isn't any `deleted` activity for topics as they are directly deleted and not set for purge later on.


ThibaudDauce (approved):
There is no test for activities?

maudetes:
> There is no test for activities?

There were only generic tests on activities, but not for each objects. I've added some for topics (since I'm at it) and datasets (since they are slightly more specific with resources management)

PR #3334, opened 2025-06-06
Add access audience to dataservices

nicolaskempf57:
Part of https://github.com/datagouv/data.gouv.fr/issues/1738


udata/core/dataservices/models.py
```diff
@@ -98,6 +102,17 @@ class HarvestMetadata(db.EmbeddedDocument):
     archived_reason = field(db.StringField())
 
 
+@generate_fields()
+class AccessAudience(db.EmbeddedDocument):
```
maudetes, line 108: I wonder if AccessAudience could be a list of ["role", "condition"], role being among three restricted values for now (compagny, local_auth, private)?
I think, it would allow to migrate role more easily than modifying the model itself?
What do you think?

udata/tests/test_api_fields.py
maudetes, line 1: I think we could add a test regarding audience in dataservice dedicated tests?

udata/tests/api/test_dataservices_api.py
```diff
@@ -127,6 +131,14 @@ def test_dataservice_api_create(self):
...
                     "foo": "bar",
                 },
+                "access_type": DATASERVICE_ACCESS_TYPE_RESTRICTED,
+                "access_audiences": [ {
+                    "role": DATASERVICE_ACCESS_AUDIENCE_ADMINISTRATION,
+                    "condition": DATASERVICE_ACCESS_AUDIENCE_YES,
+                }, {
+                    "role": DATASERVICE_ACCESS_AUDIENCE_COMPANY,
+                    "condition": DATASERVICE_ACCESS_AUDIENCE_UNDER_CONDITIONS,
+                } ]
```
ThibaudDauce: You never check this patch with asserts?

ThibaudDauce (approved):

PR #3335, opened 2025-06-13
Count dataservices for datasets

maudetes:
This adds a `dataservices` metrics in dataset objects, displayed in cdata in the reuses and API tabs.

### DIFF
--- a/udata/core/dataset/models.py
+++ b/udata/core/dataset/models.py
@@ -599,6 +599,7 @@ class Dataset(Auditable, WithMetrics, DatasetBadgeMixin, Owned, db.Document):
     __metrics_keys__ = [
         "discussions",
         "reuses",
+        "dataservices",
         "followers",
         "views",
         "resources_downloads",
@@ -610,6 +611,7 @@ class Dataset(Auditable, WithMetrics, DatasetBadgeMixin, Owned, db.Document):
             "created_at_internal",
             "last_modified_internal",
             "metrics.reuses",
+            "metrics.dataservices",
             "metrics.followers",
             "metrics.views",
             "slug",
@@ -1053,6 +1055,12 @@ class Dataset(Auditable, WithMetrics, DatasetBadgeMixin, Owned, db.Document):
         from udata.models import Reuse
 
         self.metrics["reuses"] = Reuse.objects(datasets=self).visible().count()
+        self.save(signal_kwargs={"ignores": ["post_save"]})
+
+    def count_dataservices(self):
+        from udata.core.dataservices.models import Dataservice
+
+        self.metrics["dataservices"] = Dataservice.objects(datasets=self).visible().count()
         self.save(signal_kwargs={"ignores": ["post_save"]})
 
     def count_followers(self):
--- a/udata/core/dataset/metrics.py
+++ b/udata/core/dataset/metrics.py
@@ -1,0 +1,12 @@
+from udata.core.dataservices.models import Dataservice
+
+from .models import Dataset
+
+
+@Dataservice.on_create.connect
+@Dataservice.on_update.connect
+@Dataservice.on_delete.connect
+def update_dataset_dataservices_metric(dataservice, **kwargs):
+    for dataset in dataservice.datasets:
+        # dataset is a LazyReferenceField
+        Dataset.get(dataset.pk).count_dataservices()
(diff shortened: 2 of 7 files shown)


nicolaskempf57:
> @ThibaudDauce & @nicolaskempf57 what do you think of this approach with the `Auditable.clean` and `post_save` method to keep previous changes? I can apply the same logic to reuses count if it fits

I didn't test the PR yet but it looks like a nice way to get previous changes ! This logic could be used in the future if we want to display the previous values before changes.

udata/core/activity/models.py
```diff
@@ -97,6 +118,7 @@ def post_save(cls, sender, document, **kwargs):
         if kwargs.get("created"):
             cls.on_create.send(document)
         elif len(changed_fields):
-            cls.on_update.send(document, changed_fields=changed_fields)
+            previous = getattr(document, "_previous_changed_fields", None)
+            cls.on_update.send(document, changed_fields=changed_fields, previous=previous)
```
ThibaudDauce, line 122: Why getting the `previous` variable here instead of directly in the listener?

udata/core/dataset/models.py
```diff
@@ -78,7 +77,7 @@ def get_json_ld_extra(key, value):
     }
 
 
-class HarvestDatasetMetadata(DynamicEmbeddedDocument):
```
ThibaudDauce: Why this change in this PR?

udata/core/dataset/models.py
```diff
@@ -91,12 +90,15 @@ class HarvestDatasetMetadata(DynamicEmbeddedDocument):
...
+    ckan_name = db.StringField()
+    ckan_source = db.StringField()
 
 
-class HarvestResourceMetadata(DynamicEmbeddedDocument):
+class HarvestResourceMetadata(db.EmbeddedDocument):
     created_at = db.DateTimeField()
     modified_at = db.DateTimeField()
     uri = db.StringField()
+    dct_identifier = db.StringField()
```
ThibaudDauce: Same?

ThibaudDauce (approved):

udata/core/dataset/models.py
maudetes: Ah sorry, I hadn't published my comment:
I don't think we want to bother with the DynamicEmbeddedDocument pattern since we don't have any other harvest-related plugin. Removed it in https://github.com/opendatateam/udata/pull/3335/commits/18d8101995d53ac7738b732474fd913d50b452e1.

ThibaudDauce: Yes but not really related to the changes in this PR? :-(

maudetes: But I can create another PR with these modifications indeed, it will be needed for this PR though since dynamic documents would [make tests fail](https://app.circleci.com/pipelines/github/opendatateam/udata/6982/workflows/b5d72cae-7de9-4a94-8478-918cd39e9bbf/jobs/36896?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary).

ThibaudDauce: Oh, so there is a link with this PR, so I'm fine with it being here :-)

maudetes: Woops, moved it here: https://github.com/opendatateam/udata/pull/3396 :p

udata/core/activity/models.py
maudetes: Hmm it made sense to me that it would be an arg on the `on_update` connecter? We also have this pattern [for owner change](https://github.com/opendatateam/udata/blob/f2b3d41a915f6a9144fbf9d0d1fa9c93a7b52155/udata/core/owned.py#L154).
I can change it if you think it would be better

ThibaudDauce: I think it's simpler but as you wish :-) You don't have a new argument to the signal and in the listener it's the same line except `.pop("previous")` is replaced by `getattr(document, "_previous_changed_fields", None))`

maudetes: I've actually tried to make the modification and ended up with non deterministic test fail!
It was tricky to understand why, and I think it comes from the fact that `Reuse.on_update` is connected multiple times. Some of which (`update_reuses_dataset_metric` in particular) trigger a `Auditable.clean()` modifying `self._previous_changed_fields` before the other connecter accessed the previous version.
In short, I chose to keep having the signal `on_update` being sent with the previous value at that time.

PR #3336, opened 2025-06-16
Expose permissions to cdata

ThibaudDauce:
- [x] datasets and resources
- [x] dataservices
- [x] reuses

### DIFF
--- a/udata/core/dataset/apiv2.py
+++ b/udata/core/dataset/apiv2.py
@@ -390,7 +390,7 @@ class ResourcesAPI(API):
     @apiv2.marshal_with(resource_page_fields)
     def get(self, dataset):
         """Get the given dataset resources, paginated."""
-        if dataset.permissions["edit"].can():
+        if not dataset.permissions["edit"].can():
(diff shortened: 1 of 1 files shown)


udata/api_fields.py
```diff
@@ -341,6 +341,9 @@ def query(filterable, query, value) -> UDataQuerySet:
                 continue  # Do not override if the attribute is also callable like for Extras
 
             method = getattr(cls, method_name)
+            if isinstance(method, property):
+                method = method.fget
```
maudetes, line 345: We didn't support `@property` fields before?

CHANGELOG.md
maudetes, line 1: I think we could add Organization as well

udata/api_fields.py
ThibaudDauce: No, only regular functions

CHANGELOG.md
ThibaudDauce: It's a little bit harder, will do it in a second PR.

PR #3337, opened 2025-06-16
Add endpoint to get datasets followed by user

ThibaudDauce:


udata/core/dataset/api.py
```diff
@@ -122,6 +123,7 @@ def __init__(self):
             location="args",
         )
         self.parser.add_argument("owner", type=str, location="args")
+        self.parser.add_argument("followed_by", type=str, location="args")
```
maudetes: Maybe we could specify that this is a beta arg? It may be done differently in the future

ThibaudDauce: Done in 56cd2068c

PR #3338, opened 2025-06-16
Try to lock the updates of the resources list

ThibaudDauce:
Test it by sending multiple files at once in the `/datasets/new`. It triggers on my machine with for exemple 17 small pictures. (duplicate ID in master, retry fail with this PR, all good with this PR and one from cdata https://github.com/datagouv/cdata/pull/502)


udata/core/dataset/models.py
```diff
@@ -943,36 +959,111 @@ def get(cls, id_or_slug):
...
+        # We loop below but if there is a duplicate, it should be alone otherwise it's a past problem…
+        if len(existing_resources) > 1:
+            log.warning(
+                f"{len(existing_resources)} resources already exists with ID '{resource.id}' in dataset '{self.id}'"
+            )
+
+        same_already_exists = False
+        for existing_resource in existing_resources:
+            # We don't want to crash if the user send twice the same resource in parallel.
+            if resource.is_equal_to(existing_resource):
```
ThibaudDauce: Do we need to check the creation date to see if it's recent or not?

udata/core/dataset/models.py
```diff
@@ -953,36 +969,92 @@ def get(cls, id_or_slug):
...
         """Perform an atomic prepend for a new resource"""
+        if retry == 0:
+            raise MongoEngineValidationError("Tried 5 time to add_resource without success")
+
+        self.reload()
         resource.validate()
-        if resource.id in [r.id for r in self.resources]:
-            raise MongoEngineValidationError("Cannot add resource with already existing ID")
 
+        existing_resource = next(r for r in self.resources if r.id == resource.id)
```
maudetes: Don't we get a `StopIteration` error if no result found?
```suggestion
        existing_resource = next((r for r in self.resources if r.id == resource.id), False)
```

udata/core/dataset/models.py
```diff
@@ -526,6 +526,22 @@ def save(self, *args, **kwargs):
             raise RuntimeError("Impossible to save an orphan resource")
         self.dataset.save(*args, **kwargs)
 
+    def is_equal_to(self, other: Self) -> bool:
```
maudetes: It's not needed anymore, is it?

udata/core/dataset/models.py
```diff
@@ -953,36 +969,92 @@ def get(cls, id_or_slug):
...
-                },
-            }
+
+        update_result = self.__class__.objects(
+            id=self.id, last_modified_internal=last_known_modified
+        ).update_one(
+            set__quality_cached=self.compute_quality(),
+            push__resources={"$each": [resource.to_mongo()], "$position": 0},
+            set__last_modified_internal=datetime.utcnow(),
         )
```
maudetes: It seems to me that we could only use `update_one` that would be the only transaction here.
We wouldn't have to compare for last_modified and retry.

The only (minor) issue could be `compute_quality` that may use a past version, but it seems a bit overkill though?

udata/core/dataset/models.py
```diff
@@ -953,36 +969,92 @@ def get(cls, id_or_slug):
...
         """Perform an atomic update for an existing resource"""
-        index = self.resources.index(resource)
-        data = {"resources__{index}".format(index=index): resource}
-        self.update(**data)
+        if retry == 0:
+            raise MongoEngineValidationError("Tried 5 time to update_resource without success")
+
+        self.reload()
+        last_known_modified = self.last_modified_internal
+        index = next(i for i, r in enumerate(self.resources) if r.id == resource.id)
```
maudetes: Instead of using the index strategy, you could use the positional operator ($ or S), example:
```
Dataset.objects(id=self.id, resources__id=resource.id).update_one(set__resources__S=resource)
```

maudetes: Instead of using the index strategy, you could use the positional operator ($ or S), example:
```
Dataset.objects(id=self.id, resources__id=resource.id).update_one(set__resources__S=resource)
```

udata/core/dataset/models.py
ThibaudDauce: Done in c2c03962d

udata/core/dataset/models.py
ThibaudDauce: Done in c2c03962d

udata/core/dataset/models.py
ThibaudDauce: If we add two resources with the same IDs it could break, no?

udata/core/dataset/models.py
ThibaudDauce: Done in a83947b3d

ThibaudDauce: Done in a83947b3d

udata/core/dataset/models.py
ThibaudDauce: Done in 24568934fdd605b3d6e61d0e99e53ebb9ac0bb8e

PR #3339, opened 2025-06-16
Fix old references errors in HarvestItem

ThibaudDauce:


ThibaudDauce:
Hum… I think it could happen when deleting an harvester dataset?

maudetes:
It is taken care of [in the purge logic](https://github.com/opendatateam/udata/blob/64ae30eed631dcafc9f677b7c6035052005ad8d2/udata/core/dataset/tasks.py#L56)

PR #3341, opened 2025-06-17
Do not harvest spatial with invalid values

ThibaudDauce:
Examples values that are currently harvested (from [this dataset](https://demo.data.gouv.fr/fr/datasets/corail-rouge-sur-la-cote-vermeille-mediterranee/)):
```
"geom": {
    "coordinates": [[[[NaN, NaN], [NaN, NaN], [NaN, NaN], [NaN, NaN], [NaN, NaN]]]], "type": "MultiPolygon"
}
```

### DIFF
--- a/udata/core/dataset/rdf.py
+++ b/udata/core/dataset/rdf.py
@@ -8,7 +8,6 @@ import logging
 from datetime import date, datetime
 from typing import Optional
 
-import geojson as geojson_lib
 from dateutil.parser import parse as parse_dt
 from flask import current_app
 from geomet import wkt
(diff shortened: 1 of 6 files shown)


udata/core/dataset/rdf.py
```diff
@@ -525,12 +526,18 @@ def spatial_from_rdf(graph):
...
+    try:
+        geojson_lib.loads(geojson_lib.dumps(geom))
+    except (ValueError, TypeError) as err:
+        log.warning(f"Invalid GeoJSON data: {err}.")
+        return None
+
+    spatial_coverage = SpatialCoverage(geom=geom)
 
     try:
         spatial_coverage.clean()
```
maudetes, line 536: Unsure if it's a good idea, but could we have handled it in `.clean`?

ThibaudDauce: Done in e78a05103c152f6dfa8979d7d8566bdf0fd74205

ThibaudDauce:
Yes, you're right, done in b13db7e2e39ab3cc2261340b35b7bc5326e3bd8c

udata/migrations/2025-06-18-clean-spatial-coverages.py
```diff
@@ -0,0 +1,25 @@
...
+from udata.core.dataset.models import Dataset
+
+log = logging.getLogger(__name__)
+
+
+def migrate(db):
+    datasets = Dataset.objects(spatial__geom__exists=True)
+    count = Dataset.objects(spatial__geom__exists=True).count()
+
+    with click.progressbar(datasets, length=count) as datasets:
```
maudetes, line 18: Can't we do?
```suggestion

    with click.progressbar(datasets, length=datasets.count()) as datasets:
```

maudetes (approved):
:ok_hand:

udata/migrations/2025-06-18-clean-spatial-coverages.py
ThibaudDauce: I think it's slower, no?

PR #3343, opened 2025-06-18
feat(harvest): add missing EU update frequencies

streino:
Fix https://github.com/ecolabdata/ecospheres/issues/458

- Add missing frequencies from EU vocabulary.
- Preserve existing data.gouv frequency names for backward compatibility.
- Do not attempt to map unknown vocabulary namespaces anymore, to avoid semantic mismatches.
- Refactor frequency field to a type-safe StrEnum.


udata/core/dataset/constants.py
```diff
@@ -4,42 +4,62 @@
...
-        ("quarterly", _("Quarterly")),  # freq:quarterly
-        ("threeTimesAYear", _("Three times a year")),  # freq:threeTimesAYear
-        ("semiannual", _("Biannual")),  # freq:semiannual
-        ("annual", _("Annual")),  # freq:annual
-        ("biennial", _("Biennial")),  # freq:biennial
-        ("triennial", _("Triennial")),  # freq:triennial
-        ("quinquennial", _("Quinquennial")),  # N/A
-        ("irregular", _("Irregular")),  # freq:irregular
+    [
+        # FIXME: Add ALL EU terms (from 30yrs to 1mn)?
```
streino: Need to decide on this.

udata/core/dataset/constants.py
```diff
@@ -4,42 +4,62 @@
...
+        # FIXME: Uppercase to match EU vocab?
```
streino: Need to decide on this. Do we want to stick with EU's uppercasing (enum-like)? Or lowercase to stay consistent with udata API output?

We could also go with an enum instead of a dict here => uppercase for the enum values, but output lowercase in the API response.

udata/core/dataset/constants.py
```diff
@@ -4,42 +4,62 @@
...
+        ("triennial", _("Triennial")),  # =
+        ("quinquennial", _("Quinquennial")),  # =
+        ("decennial", _("Decennial")),  # N/A
     ]
 )
 
-#: Map legacy frequencies to currents
+#: Map legacy (including DC) frequencies to currents
 LEGACY_FREQUENCIES = {
+    # FIXME: Preserve older mappings?
```
streino: I assume we might want to keep older mappings is for API users to still be able to push data with the older vocabulary. Any other use-case for that `LEGACY_FREQUENCIES` mapping?

udata/core/dataset/constants.py
```diff
@@ -4,42 +4,62 @@
...
+    "irregular": "irreg",
+    "punctual": "as_needed",
+    "realtime": "update_cont",
+    "continuous": "update_cont",
+    "fourTimesADay": "cont",  # FIXME: or OTHER?
```
streino: Need to decide on those.

`CONT` is defined as "The item is updated more frequent than daily."
`OTHER` is defined as "The event occurs with another type of regularity (for instance, every leap year)."

udata/core/dataset/rdf.py
```diff
@@ -118,7 +91,10 @@ def temporal_to_rdf(
 def frequency_to_rdf(frequency: str, graph: Optional[Graph] = None) -> Optional[str]:
     if not frequency:
         return
-    return RDF_FREQUENCIES.get(frequency, getattr(FREQ, frequency))
+    # FIXME: Do we want to map to FREQ (if possible) then EUFREQ?
```
streino: The PR currently uses EUFREQ for everything, but maybe we want to prioritize (more "general") mapping to (DC)FREQ if `frequency` exists there?

I'm not sure it's worth making the logic more complex?

udata/core/dataset/rdf.py
```diff
@@ -118,7 +91,10 @@ def temporal_to_rdf(
...
+    # FIXME: Uppercase unless done in UPDATE_FREQUENCIES (see FIXME there)
+    # FIXME: Fallback on FREQ/EUFREQ.$frequency creates invalid URI, ok?
```
streino: This was the case previously too.

udata/core/dataset/rdf.py
```diff
@@ -552,11 +528,12 @@ def frequency_from_rdf(term):
...
     if isinstance(term, URIRef):
-        if EUFREQ in term:
-            return EU_RDF_REQUENCIES.get(term)
-        _, _, freq = namespace_manager.compute_qname(term)
-        if freq.lower() in UPDATE_FREQUENCIES:
-            return freq.lower()
+        freq = namespace_manager.compute_qname(term)[2].lower()
+        if freq in UPDATE_FREQUENCIES:
+            return freq
+        # FIXME: Do we want to restrict to `FREQ in term`?
```
streino: It was previously restricted to EUFREQ.

One reason I see to restricting to known namespaces is to avoid mapping to an homonym with different semantics in another namespace. Most terms shouldn't be a problem, but e.g. `OTHER` could mean different things.

udata/core/dataset/constants.py
```diff
@@ -4,42 +4,62 @@
...
-        ("irregular", _("Irregular")),  # freq:irregular
+    [
+        # FIXME: Add ALL EU terms (from 30yrs to 1mn)?
+        # FIXME: Uppercase to match EU vocab?
+        # FIXME: Drop end-of-line annotations
+        ("unknown", _("Unknown")),  # =
+        ("other", _("Other")),  # N/A
+        ("never", _("Never")),  # N/A
+        ("not_planned", _("Not planned")),  # N/A
+        ("as_needed", _("As needed")),  # punctual
```
streino: Breaking change : API users will need to update to the new vocabulary. Is that OK?

An alternative is to keep the existing terms and only use EU vocab for new terms, but that would mean creating a franken-vocab for the udata API.

udata/migrations/2025-06-17-dc-to-eu-frenquency-vocabulary.py
abulte, line 1: frenquency -> frequency

udata/core/dataset/constants.py
abulte: If we go the breaking way, I'd go with upper case to match EU's style.

udata/core/dataset/constants.py
abulte: If we go the breaking way, I'd had all the terms.

udata/core/dataset/constants.py
abulte: `CONT` looks good

udata/core/dataset/constants.py
maudetes, line 1: Let's not forget the changelog

streino: Yep. Didn't push the commit yet to avoid merge conflicts when sync'ing the branch. I'll push it once the PR is ready for review.

udata/core/dataset/constants.py
streino: @maudetes This is the most important point to decide. WDYT?

udata/core/dataset/constants.py
maudetes: I think Dublin Core will also be used by other harvester since it's been documented for long. For example [this harvester](https://demo.data.gouv.fr/admin/harvesters/666854d2ae020ef75f404615/) or [this one](https://demo.data.gouv.fr/admin/harvesters/66632580d7365f18b2c81902) use the vocabulary `irregular`.

udata/core/dataset/constants.py
maudetes: I'm not confident in breaking the wide used vocabulary on most common terms.
Ex `irreg` vs `irregular` (~4K datasets) or `as_needed` vs `punctual` (~1,9K) is not worth the breaking change?
I think we should be aligned with the EU freqs, but if we have existing terms, could we keep them? Though I know this would be franken-vocab as you say.

streino: Let's keep existing terms then. I agree it's a lot to break.
We can always control the mapping to be full EU when exporting to specific standards, if necessary.

udata/core/dataset/constants.py
streino: [Not going the breaking way](https://github.com/opendatateam/udata/pull/3343#discussion_r2247735818), so let's keep current casing.

udata/core/dataset/constants.py
streino: [Not going the breaking way](https://github.com/opendatateam/udata/pull/3343#discussion_r2247735818), but I'll add them anyway since we have already added 80% of them. @maudetes @agarrone let me know if you think this is too much.

udata/core/dataset/constants.py
streino: I'm tempted to change things like `threeTimesAWeek` to `weekly_3` so we have a more consistent vocabulary overall: `[weekly, weekly_2, weekly_3, weekly_5]` seems better than `[weekly, semiweekly, threeTimesAWeek, weekly_5]`.

udata/core/dataset/rdf.py
streino: Obsolete

udata/core/dataset/rdf.py
```diff
@@ -119,7 +147,7 @@ def temporal_to_rdf(
 def frequency_to_rdf(frequency: str, graph: Optional[Graph] = None) -> Optional[str]:
     if not frequency:
         return
-    return RDF_FREQUENCIES.get(frequency, getattr(FREQ, frequency))
+    return FREQ_UDATA_TO_TERM.get(frequency) or EUFREQ_UDATA_TO_TERM.get(frequency) or frequency
```
streino: This isn't ideal since e.g. `frequency_to_rdf(frequency_from_rdf(EUFREQ.DAILY))` will return `FREQ.daily`.

But short of keeping track of original URIs this is the best we can do.

udata/core/dataset/constants.py
streino, line 1: Other files using `UPDATE_FREQUENCIES` that might need updates:
- https://github.com/opendatateam/udata/blob/master/js/components/form/frequency-field.vue#L65
    Is this still in use? @nicolaskempf57 
    If yes, do we want to expose all the new frequencies? @agarrone @maudetes
- https://github.com/opendatateam/udata/blob/master/specs/udata.json#L76
    I assume this should be updated? @maudetes 
- https://github.com/opendatateam/udata/blob/master/specs/udata.swagger.json
    Looks generated, but can't find how to do it? @maudetes @abulte
- https://github.com/opendatateam/udata/blob/master/specs/mocks/udata-swagger.json
    Same as previous item.

abulte: Swagger should be (re)generated when server starts (or first request).

udata/core/dataset/constants.py
```diff
@@ -2,44 +2,85 @@
...
+        ("30min", _("Every thirty minutes")),  #       EU
+        ("hourly", _("Hourly")),  #                    DC, EU
+        ("bihourly", _("Every two hours")),  #         EU
+        ("trihourly", _("Every three hours")),  #      EU
+        ("12hrs", _("Every twelve hours")),  #         EU
+        ("cont", _("Several times a day")),  #         EU
+        ("daily_3", _("Three times a day")),  #        DC:threeTimesADay, EU
+        ("daily_2", _("Twice a day")),  #              DC:semidaily, EU
+        ("daily", _("Daily")),  #                      DC, EU
+        ("weekly_5", _("Five times a week")),  #       EU
```
streino: I previously argued using EU's `weekly_2`, `weekly_3`, ... pattern was more consistent, but semantics is less obvious than `semiweekly`, `threeTimesAWeek`, ...

Also, `semi*` might already be used quite a bit so it might not be a good idea to break it.

I can revert back to the DC names if you think it's preferable.

udata/core/dataset/constants.py
streino: So, generated but we commit it back?

udata/core/dataset/constants.py
streino: Yeah, no...

> I think we should be aligned with the EU freqs, but if we have existing terms, could we keep them? Though I know this would be franken-vocab as you say.

See also https://github.com/opendatateam/udata/pull/3343#discussion_r2321535576

udata/core/dataset/constants.py
streino: Forgot previous discussions : https://github.com/opendatateam/udata/pull/3343#discussion_r2321543381

=> will revert to existing terms

udata/core/dataset/constants.py
abulte: Oh, it's not the "real" swagger.json. I have no idea then :p

udata/core/dataset/rdf.py
streino:
```
# FIXME: Do we want to map to FREQ (if possible) then EUFREQ?
```

This is what we do now. See https://github.com/opendatateam/udata/pull/3343#discussion_r2321468085.

```
# FIXME: Fallback on FREQ/EUFREQ.$frequency creates invalid URI, ok?
```

Not doing this anymore. If `frequency` doesn't exist in the namespace, we don't generate an URI and instead only output the `frequency` string itself (sans namespace).

I believe this is preferable to the old behavior, and is OK with the spec, which only specifies `dcat:accrualPeriodicity` *should* (not must) use controlled vocabularies : 

> The update schedule for a resource is indicated using [dcterms:accrualPeriodicity](https://www.w3.org/TR/vocab-dcat-3/#Property:dataset_frequency). The value should be taken from a controlled vocabulary such as [Dublin Core Collection Description Frequency Vocabulary](http://www.dublincore.org/specifications/dublin-core/collection-description/frequency/).

I found [examples of string frequencies](https://doi-do.github.io/dcat-us/#json-ld-temporal-handling) in the US application profile, so I assume this is OK.

udata/core/dataset/constants.py
streino: This was a pain to do because we have raw strings scattered all over. Also, tests are brittle because there is no correlation between `UPDATE_FREQUENCIES` and the `*FREQ_TERM_TO_UDATA` mappings in dataset/rdf.py.

I'll try to refactor into an enum to prevent future issues.

udata/core/dataset/constants.py
nicolaskempf57: The `js` folder is for the (old) administration. It is not used anymore.

I think we have to decide what to do with it. Do we want to remove it @maudetes ?
I think we definitely don't want to maintain it.

udata/core/dataset/constants.py
streino: Made another PR with an Enum version, so the diff vs master is easier to see: https://github.com/opendatateam/udata/pull/3414

It consolidates frequency-related properties (id, label, timedelta so far) into a single source of truth and ties RDF mappings to it, so it should be more robust to future vocabulary changes.

It's not perfect though: Typing and mongo ORM fields don't play nicely... so you can still set `Dataset.frequency` to a string where `UpdateFrequency` is expected. I also have a few other typing issues I'll double-check but given the weak typing all around, I'm not sure it counts as a showstopper.

IMO the Enum version isn't much more complex (+ can be simplified in the future with python 3.11's `StrEnum`) and doesn't add many LOC, so given the extra safety it provides, I'd go for it. Might even open a path for other similar vocabulary-based fields?

requirements/install.pip
```diff
@@ -20,6 +20,8 @@ babel==2.17.0
     # via
     #   -r requirements/install.in
     #   flask-babel
+backports.strenum==1.3.1 ; python_version < '3.11'
```
streino: This was manually added because I'm on a 3.11 venv, so pip compile drops backports.strenum.

BTW, setup.py has `python_requires=">=3.8"` but pip compile fails on 3.8 (works from 3.9).

udata/core/dataset/models.py
```diff
@@ -568,7 +567,10 @@ class Dataset(Auditable, WithMetrics, DatasetBadgeMixin, Owned, Linkable, db.Doc
     resources = field(db.ListField(db.EmbeddedDocumentField(Resource)), auditable=False)
 
     private = field(db.BooleanField(default=False))
-    frequency = field(db.StringField(choices=list(UPDATE_FREQUENCIES.keys())))
+
+    # FIXME: Do we want required=True to avoid None?
```
streino: @maudetes WDYT? I don't see a use for None in addition to UNKNOWN, since most places we use frequency we already default to UNKNOWN when missing : 
- API v1: https://github.com/opendatateam/udata/pull/3414/files#diff-70339dabd7e6a7c7ca04a6632989bc7a55aa1e35c09b44e08b40142e239b5cd9L358
- API v2: https://github.com/opendatateam/udata/pull/3414/files#diff-a2ac95ffb00785fc6cd9cd022bb32a7f2cddae25fcbfd636bef4d0c650771a19L158
- DatasetFactory: https://github.com/opendatateam/udata/pull/3414/files#diff-4018c6e109955b41743a2d9fca9f9a562eba16c50d0494a685336f9e7c973bb9L20
- Dataset.frequency_label: https://github.com/opendatateam/udata/pull/3414/files#diff-154a812d394e4af5de85aabf23de04611e6dbe7854b188b48f374a536edf294aL773

Only (important?) exception is DCAT: If frequency is None we don't output `dcat:accrualPeriodicity`. But this will only work for harvested datasets, because datasets created via API will always have a frequency set. Not sure it's worth the trouble.

Simple solution IMO is required=True + default=UNKNOWN (in the mongo field definition), so API and harvest will behave the same way. Otherwise, if we want to leave that field optional (so harvest can leave it null) I'll have to fix the current behavior.

TODO to self (if we're keeping None): I have to double-check but I think None is already being converted to UNKNOWN via a roundtrip to mongo: `StrEnumField.to_python(None)` [will return UNKNOWN](https://github.com/opendatateam/udata/pull/3414/files#diff-2d26a6c3d477470033d7130169ec25f9d0590b523d2b664ca73a49cfc6cbaa61R106), so the next to_mongo will end up writing that value.

udata/migrations/2025-09-04-update-legacy-frequencies.py
streino, line 1: Not sure this is needed since we're keeping the existing vocabulary and only adding terms. Might still be a good idea in case some old legacy terms weren't previous converted?

Also, note to self: check the FIXME below.

udata/core/dataset/constants.py
streino: Going with the enum version :v: Merged here.

requirements/install.pip
streino: Obsolete, master is now >= 3.9

udata/migrations/2025-09-04-update-legacy-frequencies.py
streino: It's not strictly needed since `UpdateFrequency._missing_` will silently map legacy frequencies on the fly when fetching an item from mongo.

Still, we might want to do a full migration so we don't have lingering legacy vocabulary in the database. This will make dropping legacy terms (= `UpdateFrequency._missing_`) safer in the future.

streino: Fixed the migration so legacy terms are updated in mongo.

streino:
@maudetes I'm done with the stuff I could do without your and @agarrone's feedback. Keeping the PR as a draft since it might still require work depending on your answers.

udata/core/dataset/constants.py
maudetes: Yep, it should be removed in https://github.com/opendatateam/udata/pull/3422.

udata/core/dataset/constants.py
```diff
@@ -1,48 +1,155 @@
...
+class UpdateFrequency(StrEnum):
+    """
+    Udata frequency vocabulary
+
+    Based on the following vocabularies:
+    - DC: http://dublincore.org/groups/collections/frequency/
+    - EU: https://publications.europa.eu/en/web/eu-vocabularies/at-dataset/-/resource/dataset/frequency
+    """
+
+    # FIXME(python 3.11.1+): Drop tupple syntax and use auto() where possible:
```
maudetes: Can't we already since we're also using `backports.strenum`?

udata/core/dataset/models.py
maudetes: Let's go for `required=True` and `default = UNKNOWN` as you say.
We'll need to update the migration to set objects with `frequencies__exists=False` to default.

udata/core/dataset/models.py
```diff
@@ -835,33 +835,7 @@ def next_update(self):
...
-            delta = timedelta(days=365 * 2)
-        elif self.frequency == "triennial":
-            delta = timedelta(days=365 * 3)
-        elif self.frequency == "quinquennial":
-            delta = timedelta(days=365 * 5)
-        if delta is None:
-            return
-        else:
-            return self.last_update + delta
+        return self.frequency.next_update(self.last_update) if self.has_frequency else None
```
maudetes, line 836: I guess we could go for `self.frequency.next_update(self.last_update)` only if we don't have None anymore

udata/core/dataset/rdf.py
```diff
@@ -53,54 +53,83 @@
...
+    EUFREQ.NEVER: UpdateFrequency.NEVER,
+    EUFREQ.NOT_PLANNED: UpdateFrequency.NOT_PLANNED,
+    EUFREQ.OTHER: UpdateFrequency.OTHER,
+}
+EUFREQ_ID_TO_UDATA = {
+    namespace_manager.compute_qname(k)[2].lower(): v for k, v in EUFREQ_TERM_TO_UDATA.items()
 }
 
+# Merge order matters: we want FREQ to win over EUFREQ
+UDATA_ID_TO_TERM = {v: k for k, v in {**EUFREQ_TERM_TO_UDATA, **FREQ_TERM_TO_UDATA}.items()}
```
maudetes: Maybe mention the freq part?
```suggestion
UDATA_FREQ_ID_TO_TERM = {v: k for k, v in {**EUFREQ_TERM_TO_UDATA, **FREQ_TERM_TO_UDATA}.items()}
```

udata/core/dataset/rdf.py
```diff
@@ -513,23 +539,19 @@ def spatial_from_rdf(graph):
...
-        if term.toPython().lower() in UPDATE_FREQUENCIES:
-            return term.toPython().lower()
+        term = term.toPython().lower()
+        return FREQ_ID_TO_UDATA.get(term) or EUFREQ_ID_TO_UDATA.get(term)
     if isinstance(term, RdfResource):
         term = term.identifier
     if isinstance(term, URIRef):
-        if EUFREQ in term:
-            return EU_RDF_REQUENCIES.get(term)
-        _, _, freq = namespace_manager.compute_qname(term)
```
maudetes, line 527: I think here we tried to match the `freq`  from compute_qname but we're not trying anymore here.
Not sure it was a valid case though?

udata/harvest/backends/ckan/harvesters.py
```diff
@@ -196,11 +191,13 @@ def inner_process_dataset(self, item: HarvestItem):
...
-                    dataset.frequency = value
                 else:
-                    dataset.extras["ckan:frequency"] = value
-                    log.debug("frequency value not handled: %s", value)
+                    # FIXME(python 3.12+): prefer `if value in UpdateFrequency`
+                    try:
+                        dataset.frequency = UpdateFrequency(value)
+                    except ValueError:
+                        dataset.extras["ckan:frequency"] = value
+                        log.debug("frequency value not handled: %s", value)
```
maudetes, line 199: Do we need to check for a valid value with try/catch if we're using `freq = frequency_from_rdf(value)`?

udata/migrations/2025-09-04-update-legacy-frequencies.py
```diff
@@ -0,0 +1,33 @@
...
+from udata.core.dataset.constants import UpdateFrequency
+from udata.models import Dataset
+
+log = logging.getLogger(__name__)
+
+
+def migrate(db):
+    log.info("Processing datasets.")
+
+    for legacy_value, frequency in UpdateFrequency._LEGACY_FREQUENCIES.items():
```
maudetes, line 16: The only legacy frequencies currently stored in mongo should be `fourTimesADay` and `fourTimesAWeek`, right?
Could we log the number of dataset updated? :)

udata/core/dataset/constants.py
```diff
@@ -1,48 +1,155 @@
...
-        ("irregular", _("Irregular")),  # freq:irregular
-    ]
-)
 
-#: Map legacy frequencies to currents
-LEGACY_FREQUENCIES = {
-    "fortnighly": "biweekly",
-    "biannual": "semiannual",
-    "realtime": "continuous",
+class UpdateFrequency(StrEnum):
```
maudetes, line 10: I find the StrEnum not so trivial to setup, but quite readable, congrats!

udata/core/dataset/models.py
streino: Yes, that's one of the places we won't have to deal with None anymore :v:

udata/core/dataset/rdf.py
streino: Wondered about that as well here: https://github.com/opendatateam/udata/pull/3343/files/01fa1ed3001fdf19f042ab9ade1bb4839a097f8d#r2154829481
IMO we should only map a URI when it's a known one, given how semantics can vary between namespaces.

udata/core/dataset/constants.py
streino: Ok, so won't touch stuff in js.

What about the other files, in `specs`?
> - https://github.com/opendatateam/udata/blob/master/specs/udata.json#L76
>   I assume this should be updated?
> - https://github.com/opendatateam/udata/blob/master/specs/udata.swagger.json
>   Looks generated, but can't find how to do it?
> - https://github.com/opendatateam/udata/blob/master/specs/mocks/udata-swagger.json
>   Same as previous item.

udata/harvest/backends/ckan/harvesters.py
streino: We're in the `else` part here, so when `frequency_from_rdf(value)` returned `None`.
What used to be `elif value in UPDATE_FREQUENCIES` is now a try/catch around `UpdateFrequency(value)`, since can raise on unknown values. Python 3.12 will bring back `elif value in UpdateFrequency` but for now we're stuck with the try/catch.

udata/core/dataset/constants.py
streino: Enums in python could be nicer (they are improving) but the added safety vs plain strings is IMO worth the trouble. 

Glad you find it readable. Hopefully it can serve as an example for future vocabularies.

udata/core/dataset/constants.py
streino: I think we can't. `backports.strenum` [only adds the `enum.StrEnum` class](https://github.com/clbarnes/backports.strenum/tree/main/src/backports/strenum) and relies on the standard python core for all other `enum` features, including `enum.auto`.

ThibaudDauce: I don't mind bumping the minimum Python version to 3.11 (we use it in production), @maudetes was there a reason to still support 3.9?

streino: [3.9 is reaching end of life this month](https://devguide.python.org/versions/#versions), so it might be worth planning a bump :+1:

ThibaudDauce: Done in https://github.com/opendatateam/udata/pull/3463

streino: Cool! 
However if we want to take care of the FIXME above we'll need to make sure we're at least on 3.11<em><b>.1</b></em>.  Should I assume it'll be the case and go ahead with the FIXME change?

ThibaudDauce: In dev we are in python 3.11.2, si I think it the same in production since it's defined the same way in Ansible as `3.11` :-) Maybe wait the confirmation of @maudetes about the bump :-)

streino: https://github.com/opendatateam/udata/pull/3463 was merged, so I take it as a green light :)

udata/core/dataset/constants.py
streino: I updated them, not sure what else to do with them.

udata/core/dataset/models.py
streino: Turns out it's ok and fairly easy to keep None... I got sidetracked by a bug in `StrEnumField` :shrug:. And turns out `StrEnumField` isn't even needed because mongoengine has an `EnumField` that'll work nicely on a `StrEnum` (another sidetrack with the multi-valued enum).

So I'm keeping None and UNKNOWN separate, which means everything will behave exactly as before. *And* we're able to tell if a harvested dataset had frequency=UNKNOWN vs no frequency, so we can reexpose it only when it was previously set :v:

udata/core/dataset/constants.py
```diff
@@ -1,48 +1,134 @@
...
+    """
+    Udata frequency vocabulary
+
+    Based on the following vocabularies:
+    - DC: http://dublincore.org/groups/collections/frequency/
+    - EU: https://publications.europa.eu/en/web/eu-vocabularies/at-dataset/-/resource/dataset/frequency
+    """
+
+    CONTINUOUS = auto(), _("Real time"), None  # DC, EU:UPDATE_CONT
```
streino, line 19: @maudetes @ThibaudDauce Do I need to `inv i18n`? Doing so picks up changes unrelated to this PR. What's the process?

udata/core/dataset/constants.py
maudetes: I don't think this is needed anymore either but thanks

udata/core/dataset/rdf.py
maudetes: I think it's fair. For the context, I know [we've encountered poorly defined URI](https://github.com/datagouv/data.gouv.fr/issues/986) in the past, and it was an acceptable fallback at that time.

udata/core/dataset/constants.py
maudetes: Yes, when in sync with master you can run it. It is not a problem if there are other changes :)

maudetes: Done in https://github.com/opendatateam/udata/pull/3343/commits/dd706dfdebfbe232033de2ae2c49fdc713d39e19

udata/migrations/2025-09-04-update-legacy-frequencies.py
maudetes: On my local dump
```
  │ Updating datasets legacy frequencies:
  │ - realtime -> continuous: 0 updated
  │ - fourTimesADay -> severalTimesADay: 15 updated
  │ - fourTimesAWeek -> other: 12 updated
  │ - fortnighly -> biweekly: 0 updated
  │ - biannual -> semiannual: 0 updated
  │ Completed.
```

PR #3352, opened 2025-06-30
fix(harvest): refactor dates handling

abulte:
Replaces https://github.com/opendatateam/udata/pull/3172/
Fix https://github.com/ecolabdata/ecospheres/issues/52

This modifies harvest dates handling on:
- Resources / Distributions
- Dataservices
- Datasets

When reading from RDF: `dct.issued` is now correctly mapped to `harvest.issued_at`, and `dct.created` to `harvest.created_at`. This mapping does not have any fallback.

When writing to RDF: `dct.created` and `dct.issued` are either mapped from their respective harvest attributes, or fallback to `object.created_at` (alternative: do not expose `dct.created` at all, or do not fallback to `created_at` for `dct.created`).

`dct.modified` is mapped in the same fashion, except for `Dataservice` that has a `metadata_modified_at` that seems well-handled enough to act as a storage for the harvested modified date (if it exists).

`Resource.harvest.created_at` has been renamed to `Resource.harvest.issued_at` for coherence, and a migration written. GeoDCAT-AP does not define a `created` on `Distribution`.

I don't think we need a migration for stock datasets and data services harvest data. Active ones will be updated on the next harvest run. Even the resources migration might be overkill, but I wanted to cleanup the schema. YMMV.


udata/harvest/tests/dcat/catalog.xml
```diff
@@ -21,6 +21,7 @@
         <dcat:keyword>Tag 2</dcat:keyword>
         <dcterms:publisher rdf:resource="http://data.test.org/organizations/1"/>
         <dcterms:issued rdf:datatype="http://www.w3.org/2001/XMLSchema#dateTime">2016-12-14T18:59:02.737480</dcterms:issued>
+        <dcterms:created rdf:datatype="http://www.w3.org/2001/XMLSchema#dateTime">2016-12-12T18:59:02.737480</dcterms:created>
```
streino, line 24: Nitpick: Make the dates (here and below) more clearly different in case the test ever breaks?

udata/core/dataservices/rdf.py
```diff
@@ -127,7 +128,19 @@ def dataservice_to_rdf(dataservice: Dataservice, graph=None):
     d.set(DCT.identifier, Literal(identifier))
     d.set(DCT.title, Literal(dataservice.title))
     d.set(DCT.description, Literal(dataservice.description))
-    d.set(DCT.issued, Literal(dataservice.created_at))
+
+    # created
+    if dataservice.harvest and dataservice.harvest.created_at:
```
streino: IMO we don't want fallbacks for harvested records, to ensure original harvested record == re-exposed record as much as possible.

abulte: Why not. I guess `if dataservice.harvest` is good enough to detect those. Will wait for data.gouv.fr's side on this.

udata/harvest/tests/dcat/catalog.xml
abulte: I think them being different is enough, fixing a test will require verifying the exact value anyway.

streino: Glancing at the dates (beginning and end of pattern) to review the test I first thought they were the same, so I figured there's a slight risk of being side-tracked when diagnosing a failure if one's looking a bit too quickly at the assert message.
It doesn't hurt to make it more obviously different, but there's also only a slight risk of it ever occurring, so whatever works :)

udata/core/dataservices/rdf.py
maudetes: Aggred, let's not fallback on created or issued for harvested datasets if we don't have the info. `dataservice.harvest` is good enought.

I think we can use a fallback on last modified for resources since we actively detect this info.

udata/migrations/2025-06-30-rename-resource-harvest-created-to-issued.py
```diff
@@ -0,0 +1,45 @@
...
+                                            ]
+                                        }
+                                    },
+                                ]
+                            },
+                        }
+                    }
+                }
+            },
+            {"$unset": "resources.harvest.created_at"},
```
maudetes, line 42: Oh the mongodb aggregation pipelines are always so clear and graceful :sweat_smile: 
For real though, would it be possible to use a `$rename` instruction?

udata/core/dataset/models.py
```diff
@@ -81,6 +81,7 @@ def get_json_ld_extra(key, value):
 class HarvestDatasetMetadata(DynamicEmbeddedDocument):
     backend = db.StringField()
     created_at = db.DateTimeField()
+    issued_at = db.DateTimeField()
```
maudetes, line 84: Do we want to update dataset computed property `created_at` to take issued_at into account?
Else we will have our main created_at property being mostly empty since issued is more common right?

udata/core/dataset/models.py
```diff
@@ -94,7 +95,7 @@ class HarvestDatasetMetadata(DynamicEmbeddedDocument):
 
 
 class HarvestResourceMetadata(DynamicEmbeddedDocument):
-    created_at = db.DateTimeField()
+    issued_at = db.DateTimeField()
```
maudetes, line 100: Same for resource `created_at` computed property that should now use `harvest.issued_at`

udata/core/dataservices/rdf.py
abulte: Okay, no fallback anymore for harvested objects.

> I think we can use a fallback on last modified for resources since we actively detect this info.

It feels easier to reason about to use the same pattern everywhere (no fallback for harvested), so I did not implement this.

abulte:
The PR for new fixture payload is here https://github.com/opendatateam/udata-fixtures/pull/10. I upgraded `DEFAULT_FIXTURE_FILE_TAG` to `7.0.0` here, since it seems to be the future target tag?

udata/migrations/2025-06-30-rename-resource-harvest-created-to-issued.py
abulte: I hate those and have no idea what I'm doing 🐶 

It seems like `$rename` can not handle fields in array? YMMV.

udata/core/dataservices/rdf.py
```diff
@@ -119,7 +121,13 @@ def dataservice_to_rdf(dataservice: Dataservice, graph=None):
...
     d.set(DCT.title, Literal(dataservice.title))
     d.set(DCT.description, Literal(dataservice.description))
-    d.set(DCT.issued, Literal(dataservice.created_at))
+
+    # created
+    set_harvested_date(dataservice, d, DCT.created, "created_at", fallback=dataservice.created_at)
+    # issued
+    set_harvested_date(dataservice, d, DCT.issued, "issued_at", fallback=dataservice.created_at)
+    # modified
+    d.set(DCT.modified, Literal(dataservice.metadata_modified_at))
```
ThibaudDauce: Is it possible to copy/paste the PR explanation about why no `modified` fallback vs created / issued?

ThibaudDauce (approved):
I publish the v7.0.0 tag for the fixtures with your change so you may rebase to run the test again :-)

Do not forget the changelog :-)

PR #3360, opened 2025-07-07
Add config entry for ISO-DCAT XSLT URL

ThibaudDauce:


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Current (in progress)
 
+- Add config entry for ISO-DCAT XSLT URL [#3359](https://github.com/opendatateam/udata/pull/3359)
```
streino:
```suggestion
- Add config entry for ISO-DCAT XSLT URL [#3360](https://github.com/opendatateam/udata/pull/3360)
```

udata/harvest/backends/dcat.py
```diff
@@ -355,7 +353,8 @@ def walk_graph(self, url: str, fmt: str) -> Generator[tuple[int, Graph], None, N
         See https://github.com/SEMICeu/iso-19139-to-dcat-ap for more information on the XSLT.
         """
         # Load XSLT
-        xsl = ET.fromstring(self.get(self.XSL_URL).content, parser=SAFE_PARSER)
+        xsl_url = current_app.config.get("HARVEST_ISO19139_XSL_URL")
```
streino:
```suggestion
        xsl_url = current_app.config["HARVEST_ISO19139_XSL_URL"]
```
Either that, or guard against None some other way.

ThibaudDauce: Since we have a default value in `settings.py` I don't think it could be None, having no value in env will get the default value of `settings.py`, no?

streino: Yes, which is why I suggest using `[]` instead of `get`, to make it explicit you expect to always get a value.

ThibaudDauce: Yes sorry didn't understood :-)

PR #3362, opened 2025-07-09
Add name and siret filters for orgs

ThibaudDauce:


udata/core/organization/api.py
```diff
@@ -88,6 +98,10 @@ def parse_filters(organizations, args):
             organizations = organizations.search_text(phrase_query)
         if args.get("badge"):
             organizations = organizations.with_badge(args["badge"])
+        if args.get("name"):
+            organizations = organizations.filter(name=args["name"])
```
maudetes, line 102: I'm slightly confused about the need for the name param when there is `q`?

ThibaudDauce: It's for an exact match not a search, to prevent creating an org with an existing name

PR #3366, opened 2025-07-10
Atomic get_or_create

ThibaudDauce:


udata/mongo/queryset.py
```diff
@@ -46,29 +47,28 @@ def bulk_list(self, ids):
...
+        update_fields = {f"set__{k}": v for k, v in updates.items()}
 
-        Returns a tuple of ``(object, created)``, where ``object`` is
-        the retrieved or created object and ``created`` is a boolean
-        specifying whether a new object was created.
+        existing_doc_before_upsert = self.filter(**query).modify(
+            upsert=True,
+            **query,
+            **update_fields,
+        )
```
maudetes: I'm confused about the `**query` param after the previous `.filter(**query)`?

udata/mongo/queryset.py
```diff
@@ -46,29 +47,28 @@ def bulk_list(self, ids):
...
-    def get_or_create(self, write_concern=None, auto_save=True, *q_objs, **query):
-        """Retrieve unique object or create, if it doesn't exist.
+    def get_or_create(self, **query):
+        """
+        Atomically retrieve or create a document using findAndModify (modify in MongoEngine).
+
+        Returns:
+            tuple: (document, created)
+        """
+        updates = query.pop("updates", {})
```
maudetes, line 59: I find it quite confusing that the `get_or_create` function can set an `updates` param in query

udata/mongo/slug_fields.py
```diff
@@ -199,13 +199,11 @@ def exists(s):
 
         if old_slug:
             # Create a redirect for previous slug
-            slug_follower, created = SlugFollow.objects.get_or_create(
+            SlugFollow.objects.get_or_create(
                 namespace=ns,
                 old_slug=old_slug,
-                auto_save=False,
+                new_slug=slug,
```
maudetes: Shouldn't `new_slug=slug` be in a `updates` dict param?

udata/mongo/queryset.py
ThibaudDauce: Yes it was a confusing point for me too. It seems that `modifiy` doesn't read `filter` params during the creation/update of the existing model, so we need to pass it twice…

udata/mongo/queryset.py
ThibaudDauce: Yes need to clarify this point

udata/mongo/slug_fields.py
ThibaudDauce: Maybe, but we shouldn't have an object matching `old_slug` but not `new_slug` I think…

udata/mongo/queryset.py
ThibaudDauce: I pushed a test on `get_current_site` to clarify what we expect with the `updates` param. Tell me if it's better for you? d57edf54e

udata/mongo/queryset.py
maudetes: Hmm I think testing it locally, the filter seems to apply, ex:
```
Site.objects.filter(**{"id": "local"}).modify(upsert=True, **{"title": "new_title"})
```
-> It modifies only the filtered element. I think `**query` would be used as `**update` according [to the modify definition](https://docs.mongoengine.org/apireference.html#mongoengine.queryset.QuerySet.modify).

udata/mongo/queryset.py
maudetes: Yes, I think it's more clear now. I'm just unsure if we should support updates in the `get_or_create` directly.
I think updating the document subsequently is okay? The most critical part is the uniqueness of the `get` or `create` part?

ThibaudDauce: Yes but it's a little bit inefficient to do another query to update for example in `get_current_site()`. What do you think? I remove the `update` logic?

udata/mongo/queryset.py
ThibaudDauce: Yes but `update` cannot be empty:

```
    def validate_ok_for_update(update: Any) -> None:
        """Validate an update document."""
        validate_list_or_mapping("update", update)
        # Update cannot be {}.
        if not update:
>           raise ValueError("update cannot be empty")
E           ValueError: update cannot be empty

venv/lib/python3.11/site-packages/pymongo/common.py:618: ValueError
```

So I must send `query` too to be sure to have a value…

udata/mongo/slug_fields.py
ThibaudDauce: I don't think `get_or_create` is useful here. A simple `create` should be enough… I don't see with the `unique` constraint how we could have an existing old to new follow created.

udata/mongo/queryset.py
maudetes: As you prefer, I think it's fine this way :)

PR #3368, opened 2025-07-15
Fix datasets with no slug

ThibaudDauce:
In local I have some datasets without slugs. I don't think it's normal but it's nice to not crash on some datasets pages?


ThibaudDauce:
> Do you have examples of datasets without slugs? I [stumbled on this one](https://github.com/datagouv/data.gouv.fr/issues/1814) last week 🕵️

No, just in local for some very big pages' numbers in /datasets :-(

PR #3371, opened 2025-07-16
Fix format strings in Jinja mails

ThibaudDauce:


udata/templates/mail/frequency_reminder.html
```diff
@@ -23,7 +23,7 @@
             <td align="center">
                 {{ mail_button(
                     _('Update the dataset and associated resources'),
-                    cdata_url(f"/admin/datasets/{dataset.slug}/")
+                    cdata_url('/admin/datasets/' ~ dataset.slug ~ '/')
```
maudetes: oh wow, I'm discovering the `~` operator

udata/templates/mail/membership_request.html
```diff
@@ -38,7 +38,7 @@
         <td align="center">
             {{ mail_button(
                 _('See the request'),
-                cdata_url(f"/admin/organizations/{org.slug}/members")
+                cdata_url('/admin/organizations/' ~ org.slug ~ '/members/')
```
maudetes: I think the URL with the org slug doesn't work? Ex: https://demo.data.gouv.fr/fr/admin/organizations/data-gouv-fr/members

ThibaudDauce: Really weird, fixed in 107dd32d0 but we may fix this in cdata too to accept slugs in admin URLs.

udata/templates/mail/frequency_reminder.html
ThibaudDauce: Really sad that Jinja doesn't allow format strings…

maudetes (approved):

PR #3374, opened 2025-07-16
feat(harvest): add support for all GeoDCAT-AP roles

streino:
Fix https://github.com/ecolabdata/ecospheres/issues/651

### DIFF
--- a/udata/core/contact_point/models.py
+++ b/udata/core/contact_point/models.py
@@ -9,13 +9,13 @@ CONTACT_ROLES = {
     "contact": _("Contact"),
     "creator": _("Creator"),
     "publisher": _("Publisher"),
-    "rightsHolder": _("RightsHolder"),
+    "rightsHolder": _("Rights Holder"),
     "custodian": _("Custodian"),
     "distributor": _("Distributor"),
     "originator": _("Originator"),
-    "principalInvestigator": _("PrincipalInvestigator"),
+    "principalInvestigator": _("Principal Investigator"),
     "processor": _("Processor"),
-    "resourceProvider": _("ResourceProvider"),
+    "resourceProvider": _("Resource Provider"),
     "user": _("User"),
 }
 
--- a/udata/translations/udata.pot
+++ b/udata/translations/udata.pot
@@ -8,8 +8,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: i18n@opendata.team\n"
-"POT-Creation-Date: 2025-10-27 16:24+0100\n"
-"PO-Revision-Date: 2025-10-27 16:24+0100\n"
+"POT-Creation-Date: 2025-10-27 16:50+0100\n"
+"PO-Revision-Date: 2025-10-27 16:50+0100\n"
 "Last-Translator: Open Data Team <i18n@opendata.team>\n"
 "Language: en\n"
 "Language-Team: Open Data Team <i18n@opendata.team>\n"
@@ -284,7 +284,7 @@ msgid "Publisher"
 msgstr ""
 
 #: udata/core/contact_point/models.py:12
-msgid "RightsHolder"
+msgid "Rights Holder"
 msgstr ""
 
 #: udata/core/contact_point/models.py:13
@@ -300,7 +300,7 @@ msgid "Originator"
 msgstr ""
 
 #: udata/core/contact_point/models.py:16
-msgid "PrincipalInvestigator"
+msgid "Principal Investigator"
 msgstr ""
 
 #: udata/core/contact_point/models.py:17
@@ -308,7 +308,7 @@ msgid "Processor"
 msgstr ""
 
 #: udata/core/contact_point/models.py:18
-msgid "ResourceProvider"
+msgid "Resource Provider"
 msgstr ""
 
 #: udata/core/contact_point/models.py:19


udata/rdf.py
```diff
@@ -376,14 +392,24 @@ def contact_points_to_rdf(contacts, graph=None):
...
+            if contact.email:
+                node.set(VCARD.hasEmail, URIRef(f"mailto:{contact.email}"))
+            if contact.contact_form:
+                node.set(VCARD.hasUrl, URIRef(contact.contact_form))
+        else:
+            node.set(RDF.type, FOAF.Agent)
+            node.set(FOAF.name, Literal(contact.name))
+            if contact.email:
+                node.set(FOAF.mbox, URIRef(f"mailto:{contact.email}"))
+            # FIXME: contact_form mapping?
```
streino: Since we're already using `VCARD.hasUrl`, we might want to use `FOAF.page` here? Neither is guaranteed to be a contact form though.

udata/core/dataset/rdf.py
```diff
@@ -331,6 +331,7 @@ def dataset_to_rdf(dataset: Dataset, graph: Optional[Graph] = None) -> RdfResour
     if frequency:
         d.set(DCT.accrualPeriodicity, frequency)
 
+    # FIXME: owner should map to rightsHolder instead of publisher?
```
streino: SEMIC maps [ISO-19139 "owner" to DCT.rightsHolder](https://github.com/SEMICeu/iso-19139-to-dcat-ap/blob/main/iso-19139-to-dcat-ap.xsl#L1953). We might want to do the same here?

udata/rdf.py
```diff
@@ -376,14 +392,24 @@ def contact_points_to_rdf(contacts, graph=None):
...
-        node.set(RDF.type, VCARD.Kind)
-        if contact.name:
-            node.set(VCARD.fn, Literal(contact.name))
-        if contact.email:
-            node.set(VCARD.hasEmail, URIRef(f"mailto:{contact.email}"))
-        if contact.contact_form:
-            node.set(VCARD.hasUrl, URIRef(contact.contact_form))
-        yield node, AGENT_ROLE_TO_RDF_PREDICATE.get(contact.role, DCAT.contactPoint)
+        role = AGENT_ROLE_TO_RDF_PREDICATE.get(contact.role, DCAT.contactPoint)
+        # TODO: we could use user_to_rdf/organization_to_rdf if ContactPoint was typed
```
streino: Is this something we want to tackle? Otherwise I'll just drop the TODO.

udata/rdf.py
```diff
@@ -376,14 +392,24 @@ def contact_points_to_rdf(contacts, graph=None):
...
+        if role == DCAT.contactPoint:
```
streino, line 431: This is according to the [GeoDCAT-AP spec](https://semiceu.github.io/GeoDCAT-AP/releases/3.0.0/). Only the contact point is a [`Kind`](https://semiceu.github.io/GeoDCAT-AP/releases/3.0.0/#Kind). Other roles are [`Agent`](https://semiceu.github.io/GeoDCAT-AP/releases/3.0.0/#Agent). Maybe I should add it as a comment in the code...

udata/core/contact_point/models.py
```diff
@@ -9,6 +9,14 @@
     "contact": _("Contact"),
     "creator": _("Creator"),
     "publisher": _("Publisher"),
+    "rightsHolder": _("RightsHolder"),
```
streino: TODO: update i18n

udata/core/dataset/rdf.py
maudetes: I'm unsure if our `owner` vocabulary is actually this one? All we know is the organization published this dataset on our platform so I think publisher seems like a safer move?

udata/rdf.py
```diff
@@ -130,15 +130,31 @@
...
+    "distributor": GEODCAT.distributor,
+    "originator": GEODCAT.originator,
+    "principalInvestigator": GEODCAT.principalInvestigator,
+    "processor": GEODCAT.processor,
+    "resourceProvider": GEODCAT.resourceProvider,
+    "user": GEODCAT.user,
 }
 
 # Map rdf contact point entity to role
 CONTACT_POINT_ENTITY_TO_ROLE = {
```
maudetes, line 152: Isn't this a bijection?
```suggestion
CONTACT_POINT_ENTITY_TO_ROLE = {predicate:role for role, predicate in AGENT_ROLE_TO_RDF_PREDICATE.items()}
```

maudetes (commented):
Thank you! It seems fair!
It would be nice if you a have some real catalog examples to add these in the tests? As is done for example with the different `sig.oreme.rdf` & co test files

udata/core/dataset/rdf.py
streino: You're right it is safer. We may at some point we'll need to make an exception for harvested INSPIRE records, but I'd leave it as a future task, since we know RDF export will need reviewing in that context anyway.

udata/rdf.py
streino: I'd rather handle this as part of another PR since it's somewhat orthogonal.

streino: https://github.com/ecolabdata/ecospheres/issues/744

udata/harvest/tests/test_dcat_backend.py
```diff
@@ -654,6 +654,48 @@ def test_sigoreme_xml_catalog(self, rmock):
...
+        dataset = Dataset.objects.filter(organization=org).first()
+
+        assert dataset is not None
+        assert len(dataset.contact_points) == 3
+
+        assert dataset.contact_points[0].name == "Administrateur de Données"
+        assert dataset.contact_points[0].email == "sig.dreal-ara@developpement-durable.gouv.fr"
+        assert dataset.contact_points[0].role == "contact"
+
+        assert dataset.contact_points[1].name == "Jean-Michel GENIS"
```
streino, line 751: This isn't great 😬 (but broken independent of this PR).

We're only storing (and therefore displaying) the individual's name, even though the original metadata also contains their organization info. 

In this specific case of "rights holder" contact, using the person's name probably also is an incorrect attribution, since it's most likely the org, not the person, who holds the rights.

Same idea for "administrateur de données" higher in the test, but at least in this example the individual info is a title, not a name. So less problematic.

For comparison, GeoNetwork (which has access to all contact info in the original record) displays the org when both individual and org info are available:



Best solution would be to store both individual and org info in the udata contact point model (we'll also need it to reexport to DCAT properly). But if we want to go for a quick fix here, I'd either keep the org name only, or fuse individual+org names.

@maudetes WDYT?

udata/core/contact_point/models.py
streino: @maudetes I get a bunch of diff when running `inv i18n`, including some that's not related to this PR. Is this something I need to do? If yes, do I push everything in this PR, cherry-pick what's mine, or something else?

udata/harvest/tests/test_dcat_backend.py
streino: cc @agarrone

maudetes: Hmm good point, I think getting the orga name seems like a nice move in the case of an individual contact, however we won't know if we are in the case of an individual?
Do you have some statistics on such cases?

streino: I don't see how we can know if the `foaf:Person` (or vcard equivalent) really is an individual or instead represents a function (like "administrateur de données" above). I'd argue this is why we would be safer only displaying the org in the main info panel. We can always display the full metadata somewhere else (provided we properly store it), like in the "informations" tab.

What stats are you talking about? Cases where we have org+person metadata (can try to find that info), or cases where `foaf:Person` is really an individual (won't be able to get that since we can't tell the difference).

streino: Found an example of the issue on current prod : https://www.data.gouv.fr/datasets/espaces-terrestres-france-et-pays-riverains-1

#### data.gouv



#### GeoNetwork
with the details pop-up open

streino: Another example with an entity name: https://www.data.gouv.fr/datasets/diversite-des-habitats-be

Less problematic but shows it might not be a simple either/or choice between org and entity/individual name. Both could be useful.

#### data.gouv



#### GeoNetwork

streino: @agarrone Following our discussion yesterday, I went with the organization name when a contact point has both agent (=individual) and organization names.

So in examples above, we'll have _"Office français de la biodiversité"_ instead of _"Pôle géomatique"_, and _"Agence française pour la biodiversité"_ instead of _"Steven PIEL"_.

I think this is better than only displaying the agent name (which can be an person's name), but if you prefer I could go with displaying both in those cases : _"Office français de la biodiversité / Pôle géomatique"_, or _"Agence française pour la biodiversité / Steven PIEL"_. Though it'll be quite long in some cases, e.g. _"Direction Régionale de l’Environnement de l’Aménagement et du Logement d'Auvergne-Rhône-Alpes (DREAL Auvergne-Rhône-Alpes) / Administrateur de Données"_.

agarrone: Thanks!

I think @maudetes wants to analyse the data to make a more informed decision.

udata/core/contact_point/models.py
streino: Answered in https://github.com/opendatateam/udata/pull/3343#discussion_r2413146628:

> Yes, when in sync with master you can run it. It is not a problem if there are other changes :)

=> TODO juste before merge to be in sync w/ master

udata/harvest/tests/test_dcat_backend.py
streino: After a chat with @maudetes we settled on reverting the change from individual to org name so this PR only adds roles and doesn't touch how they're displayed. That way we can merge.

I'll create another PR for the display change and we'll get some stats to figure out what to do.

udata/core/contact_point/models.py
```diff
@@ -9,6 +9,14 @@
...
+    "custodian": _("Custodian"),
+    "distributor": _("Distributor"),
+    "originator": _("Originator"),
+    "principalInvestigator": _("PrincipalInvestigator"),
```
maudetes: I would add a space in the different contact roles labels?
```suggestion
    "principalInvestigator": _("Principal Investigator"),
```

udata/harvest/tests/test_dcat_backend.py
streino: -> https://github.com/opendatateam/udata/pull/3497

PR #3376, opened 2025-07-18
Add contact point suggest

nicolaskempf57:


ThibaudDauce (commented):
Could you add a test?

ThibaudDauce (approved):

udata/core/organization/api.py
```diff
@@ -305,6 +305,31 @@ def get(self, org):
...
+    @api.marshal_list_with(contact_point_fields)
+    def get(self, org):
+        """Contact points suggest endpoint using mongoDB contains"""
+        args = suggest_parser.parse_args()
+        contact_points = ContactPoint.objects(
+            Q(name__icontains=args["q"])
+            | Q(email__icontains=args["q"])
+            | Q(contact_form__icontains=args["q"])
+        ).owned_by(org)
+        return [contact_point for contact_point in contact_points.limit(args["size"])]
```
maudetes: Can't we return contact_points directly?
```suggestion
        return contact_points.limit(args["size"])
```

nicolaskempf57: > Can't we return contact_points directly?

We can do `list(contact_points.limit(args["size"]))` too but `return contact_points.limit(args["size"])` is not working

PR #3377, opened 2025-07-18
Add last_update as a dataset model property

maudetes:
Fix issue with `last_update` in api v2 not being based on resources.
Example [API v2](https://www.data.gouv.fr/api/2/datasets/plan-cadastral-informatise/) vs [API v1](https://www.data.gouv.fr/api/1/datasets/plan-cadastral-informatise/).
Indeed, when resources were missing (using API v2 for example), the last_update was erroneous.

We compute it in dataset clean method to ensure that resources are fetched.
It also allows us to use it as a mongo sort, ex http://dev.local:7000/api/2/datasets/?sort=-last_update.
It should be less costly than computing it on every api call as was previously the case.

### DIFF
--- a/udata/migrations/2025-07-18-compute-last-update.py
+++ b/udata/migrations/2025-07-18-compute-last-update.py
@@ -1,0 +1,25 @@
+"""
+This migration computes the dataset last_udpate property that is now stored in the model.
+It simply iterates on dataset and save them, triggering the clean() method where the last_update compute ocurs.
+"""
+
+import logging
+
+import click
+
+from udata.core.dataset.models import Dataset
+
+log = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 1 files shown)


udata/core/dataset/models.py
```diff
@@ -595,6 +595,14 @@ class Dataset(Auditable, WithMetrics, DatasetBadgeMixin, Owned, Linkable, db.Doc
         ),
         auditable=False,
     )
+    last_update = field(
+        DateTimeField(
+            verbose_name=_("Last update of the dataset resources"),
+            default=datetime.utcnow,
+            required=True,
```
maudetes, line 602: I've added a default since the field is required (similar to other datetime fields)

PR #3378, opened 2025-07-21
Consolidate CSW harvesting logic

streino:
Fix https://github.com/ecolabdata/ecospheres/issues/688

Refactor to facilitate the switch to Saxon.

Note: I fixed some inconsistencies while refactoring. See my PR comments.


udata/harvest/backends/dcat.py
```diff
@@ -235,98 +233,100 @@ def get_node_from_item(self, graph, item):
...
 class CswDcatBackend(DcatBackend):
     display_name = "CSW-DCAT"
 
-    DCAT_SCHEMA = "http://www.w3.org/ns/dcat#"
+    CSW_REQUEST = """
+    <csw:GetRecords xmlns:csw="http://www.opengis.net/cat/csw/2.0.2"
+                    xmlns:ogc="http://www.opengis.net/ogc"
+                    service="CSW" version="2.0.2" resultType="results"
+                    outputSchema="http://www.w3.org/ns/dcat#"
+                    startPosition="{start}" maxRecords="25">
```
streino: Fixed `maxPosition` -> `maxRecords`.
I didn't set it to 200 since that value was never applied. This seems like a large number compared to the default of 25, so if we want to go with 200 we might need to test it first?

udata/harvest/backends/dcat.py
```diff
@@ -235,98 +233,100 @@ def get_node_from_item(self, graph, item):
...
+      <csw:Query typeNames="csw:Record">
+        <csw:ElementSetName>full</csw:ElementSetName>
+        <ogc:SortBy>
```
streino, line 304: Is there a reason we use a sortby here? 
And if yes, should we add one for the Iso19139 backend?

udata/harvest/backends/dcat.py
```diff
@@ -235,98 +233,100 @@ def get_node_from_item(self, graph, item):
...
     display_name = "CSW-DCAT"
 
-    DCAT_SCHEMA = "http://www.w3.org/ns/dcat#"
+    CSW_REQUEST = """
+    <csw:GetRecords xmlns:csw="http://www.opengis.net/cat/csw/2.0.2"
+                    xmlns:ogc="http://www.opengis.net/ogc"
+                    service="CSW" version="2.0.2" resultType="results"
+                    outputSchema="http://www.w3.org/ns/dcat#"
+                    startPosition="{start}" maxRecords="25">
+      <csw:Query typeNames="csw:Record">
```
streino: Changed `gmd:MD_Metadata` to `csw:Record`, to match the type used in the Iso19139 backend.

The doc I could find isn't very informative. E.g the [OpenDataSoft doc](https://help.opendatasoft.com/apis/csw/#getrecords):

> `typeNames`: Used to indicate which named set the service shall present to the client. It usually defines the level of details present in the result set. Mandatory. Values can be `csw:Record` or `gmd:MD_Metadata`.

I couldn't find any difference in my tests. Both types returned exactly the same responses.

So since we're issuing a CSW request, I figured it makes more sense to me to use a `csw` type? Also, if this were to make a difference, I think it would be for Iso1913 (which used `csw:Record` already), since here we get DCAT contents, which has nothing to do with the `gmd` namespace.

udata/harvest/backends/dcat.py
```diff
@@ -235,98 +233,100 @@ def get_node_from_item(self, graph, item):
...
                 log.error(f"No search results found for {url} on page {page_number}")
-                break
-            for child in search_results:
+                return
+
+            for result in search_results:
                 subgraph = Graph(namespace_manager=namespace_manager)
-                subgraph.parse(data=ET.tostring(child), format=fmt)
+                doc = ET.tostring(self.as_dcat(result))
+                subgraph.parse(data=doc, format=fmt)
```
streino, line 346: In Iso19139 there was an extra test here:
```
if not subgraph.subjects(RDF.type, DCAT.Dataset):
  raise ValueError("Failed to fetch CSW content")
```

I'm not sure what the reasoning was behind this and couldn't find any hints in git blame.

I'm thinking it might have made sense when we were only retrieving Dataset, but now that we fetch more (Dataservices and Series) I think this might actually cause problems?

udata/harvest/backends/dcat.py
```diff
@@ -235,98 +233,100 @@ def get_node_from_item(self, graph, item):
...
+            # A value of 0 means all records have been returned (standard CSW)
+            (next_record == 0)
+            # Misbehaving CSW server returning a next record > matched count
+            or (next_record > matched_count)
+            # No results returned already
+            or (returned_count == 0)
+            # Current next record is lower than previous one
+            or (next_record < start)
+            # Enough items have been harvested already
+            or (self.max_items and len(self.job.items) >= self.max_items)
```
ThibaudDauce: Maybe call `has_reached_max_items` here?

udata/harvest/backends/dcat.py
```diff
@@ -342,94 +342,43 @@ class CswIso19139DcatBackend(DcatBackend):
...
+    """
 
-            response = self.post(
-                url, data=body.format(start=start, schema=self.ISO_SCHEMA), headers=headers
-            )
-            response.raise_for_status()
+    @override
+    def inner_init(self) -> None:
+        super().inner_init()
+        xslt_url = current_app.config["HARVEST_ISO19139_XSL_URL"]
```
streino: I made a mistake when naming the `HARVEST_ISO19139_XSL_URL` config parameter. It really should be `HARVEST_ISO19139_XSLT_URL`. How about I rename it as part of this PR?

ThibaudDauce: Yes you can, just be sure to note it in the changelog so we don't forget to change the env variable during the deployment.

udata/harvest/backends/dcat.py
ThibaudDauce: We could put this in a config variable in the future if needed.

udata/harvest/backends/dcat.py
streino: Yes!
I think it's actually a noop to check that here since we already have a call to `has_reached_max_items` after each yield. But I'm tempted to keep it since it doesn't hurt and it's safer in case the main logic ever changes.

udata/harvest/backends/dcat.py
streino: I know @maudetes was reluctant to add config variables, but I'd actually really like to have that value configurable. Reason is most servers can easily handle pretty high values for this, but GeoIDE struggles beyond something like 10... So we have to set the Iso19139 backend value to 10 just to avoid sinking GeoIDE :/

ThibaudDauce: I was thinking of global config variable but yes we also have harvester level config for this stuff. If you say there is really different behavior for different server it could be worth to have an harvester config for this to speed up the harvesting process.

udata/harvest/backends/dcat.py
streino: Is there a specific way I should mention it in the changelog?

I see occurrences of `:warning:`, `**breaking change**`. Should I use any of those or be less ominous since that config param is a really recent addition?

udata/harvest/backends/dcat.py
streino: I suggest handling it in another PR since it's not really refactoring anymore. Happy to make another PR if you (ie data.gouv side) want to see either solution happen now.

udata/harvest/backends/dcat.py
ThibaudDauce: One of the two is fine, it's just for me or @maudetes to think about it during deployment :-)

udata/harvest/backends/dcat.py
maudetes: This was introduced following [this comment](https://github.com/opendatateam/udata/pull/2800#discussion_r1129053500). I guess we should add it then to the Iso19139 backend?

udata/harvest/backends/dcat.py
maudetes: The 200 value was suggested [in this comment](https://github.com/opendatateam/udata/pull/2800#discussion_r1129053500).
I was reluctant to add too much custom params for harvesters since it was an issue we had with ODS harvesters previously, asking more reviewing care when validating. I think this param in particular wouldn't be too much of an issue however.
For the moment though, it seems the page size works, even for GeoIDE, right?

udata/harvest/backends/dcat.py
maudetes: Same, I think it came from the original query suggestion in [the issue](https://github.com/datagouv/data.gouv.fr/issues/913).
Indeed, it doesn't seem to have much impact?

udata/harvest/backends/dcat.py
streino: I can imagine this is used to deterministically fetch items, so we get the same list each time if we reach max items (modulo changes in source item list)?
If this is the case, I suppose we should also add it to Iso19139, though this might impact perfs on large/slow catalogs.

udata/harvest/backends/dcat.py
```diff
@@ -235,98 +233,100 @@ def get_node_from_item(self, graph, item):
...
-            tree = ET.fromstring(
-                self.post(
-                    url, data=body.format(start=start, schema=self.DCAT_SCHEMA), headers=headers
-                ).content,
-                parser=SAFE_PARSER,
-            )
+    def as_dcat(self, tree: ET._Element) -> ET._Element:
+        """
+        Return the input tree as a DCAT tree.
+        For CswDcatBackend this method is a noop.
```
maudetes: I'm not familiar with the word `noop` if you want to reword :p

udata/harvest/backends/dcat.py
maudetes: I think the idea was to recognize cases where we would get some empty content I think (and not an error, since there are some poor implementations out there) for example. I couldn't find logs of this happening though.

maudetes (commented):
Nice! Thank you for the needed mutualisation! :rocket: 

It's always quite tricky to check for any side effects, I think we could try deploying it on demo for example and check for any discrepancy in jobs before/after?

udata/harvest/backends/dcat.py
streino: Thanks for the reference! And it explains the sortBy :v:

I would be wary of pushing 200 now since it was never effectively applied. I can test it as part of another PR though.

Unfortunately I think even 25 might be too much for GeoIDE, especially if we have several harvesters hitting GeoIDE sub-catalogs at the same time. A quick web+claude search gives a default `maxRecords=10` (though [some server implementations will set different values](https://github.com/geonetwork/core-geonetwork/issues/7995)). I'd keep that 10 value, pending more tests.

So, for now I suggest we set 25 for the CswDcat backend and 10 for the CswIso backend, and I'll do more tests/tuning later.

udata/harvest/backends/dcat.py
streino: Weird, I didn't have your comment when I posted mine :/ Anyway, yes, let's add it to the Iso backend.

udata/harvest/backends/dcat.py
streino: I'll double-check the mentioned ticket

udata/harvest/backends/dcat.py
streino: Ah yes, I do remember something like that...
I'll keep it but extending it with Series + Dataservices + comment to keep it in sync with `CSW_REQUEST`. Can't hurt.

streino:
> Nice! Thank you for the needed mutualisation! 🚀
> 
> It's always quite tricky to check for any side effects, I think we could try deploying it on demo for example and check for any discrepancy in jobs before/after?

I've stared those tests on local deployments. I'm at the stage of actually checking discrepancies. Do you have tooling for that?

If you do deploy on demo I'll rerun my (upcoming) tests on that env as well. Ceinture et bretelles ;)

maudetes:
> I've stared those tests on local deployments. I'm at the stage of actually checking discrepancies. Do you have tooling for that?

I don't have one specific tooling. I think I would go with comparing on master VS this branch on real harvest runs (DCAT, CSW-DCAT & CSW-ISO-DCAT harvesters):
- raw DCAT graphs diff
- jobs items number & status

udata/harvest/backends/dcat.py
streino: Turns out GeoIDE default is 25... So scratch what I said, we can use 25 for both backends.

udata/harvest/backends/dcat.py
streino: BTW, why don't we restrict on types in `CswDcatBackend.CSW_REQUEST`? Anything other than Dataset/DataService will be ignored further down, no?

streino: @maudetes

maudetes: Yes, seems fair!

udata/harvest/backends/dcat.py
maudetes: I think it's cause my comment was pending and not already published, sorry! :p

udata/harvest/backends/dcat.py
streino: Checked the mentioned ticket and couldn't find any explanation about why `gmd:MD_Metadata` vs `csw:Record`.
So I'll rely on my ongoing large-scale tests (testing all "csw-dcat" and "csw-iso-19139" harvesters currently deployed on data.gouv.fr) to settle the question.

udata/harvest/backends/dcat.py
```diff
@@ -235,104 +233,159 @@ def get_node_from_item(self, graph, item):
...
+    # [2] GeoNetwork - CSW test requests
+    #     https://github.com/geonetwork/core-geonetwork/tree/3.10.4/web/src/main/webapp/xml/csw/test
+    # [3] Udata - Support csw dcat harvest
+    #     https://github.com/opendatateam/udata/pull/2800#discussion_r1129053500
+    # [4] GeoNetwork - GetRecords ignores namespaces for Filter/SortBy fields
+    #     https://github.com/geonetwork/core-geonetwork/blob/3.10.4/csw-server/src/main/java/org/fao/geonet/kernel/csw/services/getrecords/FieldMapper.java#L92
+    # [5] GeoNetwork - GetRecords ignores `typeNames`
+    #     https://github.com/geonetwork/core-geonetwork/blob/3.10.4/csw-server/src/main/java/org/fao/geonet/kernel/csw/services/getrecords/CatalogSearcher.java#L194
+    CSW_REQUEST: ClassVar[str] = """
+    <csw:GetRecords xmlns:apiso="http://www.opengis.net/cat/csw/apiso/1.0"
```
maudetes, line 274: Why the added `apiso` namespace?

udata/harvest/backends/dcat.py
```diff
@@ -235,104 +233,159 @@ def get_node_from_item(self, graph, item):
...
+            if not search_results:
                 log.error(f"No search results found for {url} on page {page_number}")
-                break
-            for child in search_results:
+                return
+
+            for result in search_results:
                 subgraph = Graph(namespace_manager=namespace_manager)
-                subgraph.parse(data=ET.tostring(child), format=fmt)
+                doc = ET.tostring(self.as_dcat(result))
```
maudetes, line 344: Wouldn't it be better to transform the entire graph once instead of multiple times for each search result?

streino:
> With that pipeline I do find some differences in some harvests. Items that get harvested with one version and not the other (or vice versa). Looking into it.

I think I figured out most of those.

First, 10 out of 14 harvests tested have 0 diff :heavy_check_mark:

```
OK    csw-dcat        Catalogue du CRAIG (DCAT)
OK    csw-dcat        Géo2France geo2france_data
OK    csw-dcat        Ifremer - Sextant - Opendata
DIFF  csw-dcat        Moisssonneur SI Eau
OK    csw-dcat        Open Data Région Hauts-de-France
OK    csw-dcat        opendata Atmo Nouvelle Aquitaine
OK    csw-iso-19139   Catalogue BRGM - HVD
DIFF  csw-iso-19139   DREAL Pays de la Loire via SIGLOIRE
DIFF  csw-iso-19139   Direction Régionale de l’Environnement de l’Aménagement et du Logement d'Auvergne-Rhône-Alpes
OK    csw-iso-19139   Géo-IDE DDT 38
OK    csw-iso-19139   Géo-IDE DREAL Provence-Alpes-Côte d'Azur (GeoIDE)
DIFF  csw-iso-19139   Ressources IGN en open data sur le catalogue de la Géoplateforme
OK    csw-iso-19139   Sigena
OK    csw-iso-19139   open-datara-ddt38 (GeoIDE)
```

Details of the 4 harvests with diff below.

---

### Important

"**item**" here means `HarvestItem`. To compare harvest jobs' items lists across instances of udata we can't rely on internal IDs, so the above mongo pipeline uses `remote_id` as a grouping key. This means catalog records that have the same `remote_id` will be grouped under the same key and what we compare is lists of records found under each key. Items with no `remote_id` are grouped under the "MISSING" key.

If several elements are found under the same key, I call them "**dupes**". In theory `remote_id` (the "data ID", as opposed to the "metadata ID") should always be present (no MISSING key), and unless several catalog records describe the same data, the data IDs should be unique. In practice, we find records with missing data ID, and some different records (different metadata ID) with the same data ID, often because a record was duplicated (creates a new metadata ID) but the data ID wasn't updated.

Example of a comparison:
:yellow_circle: = diff, :green_circle:/:red_circle: = insertion/deletion, :large_blue_circle: = key, :white_circle: = ignored field



---

### Diffs found

#### `remote_id` found in REF, not found in PR

Two such cases in "Moisssonneur SI Eau" (csw-dcat): [sieau:2b5ab7c3-f0d3-4653-9fd5-cfb133a914fb](https://data.ofb.fr/catalogue/srv/eng/catalog.search#/metadata/2b5ab7c3-f0d3-4653-9fd5-cfb133a914fb), [sieau:7db4d17b-0f5f-4312-b1a5-8ec93cc9ab72](https://data.ofb.fr/catalogue/srv/eng/catalog.search#/metadata/7db4d17b-0f5f-4312-b1a5-8ec93cc9ab72).

Both have type = "model". REF didn't filter on record type, and PR does, but doesn't include "model.

@maudetes WDYT? I don't know that type but IMO it's close-enough to a dataset that it makes sense to include it.

We had the same issue with type = "nonGeographicDataset", which I consequently added to the filter.


#### `remote_id` not found in REF, found in PR

Example: [datara:c2941fee-2b8a-4d5a-beda-ba7dedc06c7e
](https://catalogue.open-datara.fr/geonetwork/srv/eng/catalog.search#/metadata/c2941fee-2b8a-4d5a-beda-ba7dedc06c7e)

The three csw-iso-19139 catalogs with DIFF have such occurrences.

The specific diffs I checked so far seem legit. Not sure why they're not harvested by REF. Still investigating, but PR seems better on that one.


#### `remote_id` has dupes in REF, no dupe in PR

Examples: [sigloire:022d9983-1f0a-4193-94b6-a7a90541704a](https://catalogue.sigloire.fr/geonetwork/srv/eng/catalog.search#/metadata/022d9983-1f0a-4193-94b6-a7a90541704a), [ign:IGNF_CONTOURS-IRIS](https://data.geopf.fr/csw/ressources_o?namespace=xmlns:csw=http://www.opengis.net/cat/csw/2.0.2&service=CSW&version=2.0.2&outputFormat=application/xml&request=GetRecordById&resultType=results&outputSchema=http://www.isotc211.org/2005/gmd&typeNames=gmd:MD_Metadata&Id=IGNF_CONTOURS-IRIS)

I can only find one occurrence of each record in the catalogues.
 
For "sigloire", on REF, the 2 dupes show up in page 5 and page 34. Other such dupes in sigloire also show up with a large gap in page numbers.
However the "ign" dupes show up on page 2 and 3.

One likely possibility is that csw-iso didn't have a `SortBy` in REF and therefore could display the same record twice while paging, but the "ign" case seems very unlucky in that case.


#### `remote_id` has no dupe in REF, dupes in PR

Example: [datara:86d9e6f3-0dad-40de-b528-f5a516f61a50](https://catalogue.open-datara.fr/geonetwork/srv/eng/catalog.search#/metadata/86d9e6f3-0dad-40de-b528-f5a516f61a50) and [datara:92d4bca0-d6cf-4045-8f76-095d0a552a5b](https://catalogue.open-datara.fr/geonetwork/srv/eng/catalog.search#/metadata/92d4bca0-d6cf-4045-8f76-095d0a552a5b).

Those are items with the same `remote_id` (and `dct_identifier` but it's not available in `HarvestItem`) and different `remote_url` (based on metadata ID, i.e. `fileIdentifier` in ISO).

This is because the ISO backend now individually converts each ISO item to DCAT (so each one has it's own entire `rdf:RDF` graph), whereas before we converted the whole CSW page which could contain several items (so all elements in the page shared a single `rdf:RDF` graph, and possibly stepped on each other's toes).

REF:
```
<!-- single rdf:RDF -->
  <rdf:Description rdf:about="https://catalogue.open-datara.fr/geonetwork/srv/92d4bca0-d6cf-4045-8f76-095d0a552a5b">
    <foaf:isPrimaryTopicOf rdf:nodeID="Nd2ce46c32f3a4f7d95fac01fb083f3ba"/>
    <foaf:isPrimaryTopicOf rdf:nodeID="Ndb95e3f4ed45493598b3a659c44a66b0"/>
    <rdf:type rdf:resource="http://www.w3.org/ns/dcat#Dataset"/>
```
=> Only one `Dataset` with two primary topics "merging" the two records => Only processed once by harvester (which primary topic wins is up to the rdflib parsing, hopefully at least it's deterministic).

PR:
```
<!-- first rdf:RDF -->
  <rdf:Description rdf:about="https://catalogue.open-datara.fr/geonetwork/srv/92d4bca0-d6cf-4045-8f76-095d0a552a5b">
    <foaf:isPrimaryTopicOf rdf:nodeID="Nd4d2a9d7ca854214824ee56774aceb94"/>
    <rdf:type rdf:resource="http://www.w3.org/ns/dcat#Dataset"/>

<!-- second rdf:RDF -->
  <rdf:Description rdf:about="https://catalogue.open-datara.fr/geonetwork/srv/92d4bca0-d6cf-4045-8f76-095d0a552a5b">
    <foaf:isPrimaryTopicOf rdf:nodeID="N80ec9b5d3d3a4ecca19cb136c18b8e50"/>
    <rdf:type rdf:resource="http://www.w3.org/ns/dcat#Dataset"/>
```
=> One `Dataset` for each record => Each one is independently processed by harvester => dupes (but handled by udata, so one udata `dataset` in the end).

I'd argue PR's graph-per-record approach is better than REF's graph-per-page, because: 
- PR makes it consistent between CswIso and CswDcat: CswDcat has always been graph-per-record because CSW outputs each DCAT item as a separate `rdf:RDF`.
- PR CswIso treats items on two different pages and items on the same page exactly the same way: independent graphs in all cases.
- PR CswIso always leaves disambiguation to udata, whereas on REF we have no control over it: it's up to the SEMIC converter and/or rdflib, and in the case of the converter, I'm not sure it was designed to handle the dupes corner-cases.

---

I'll finish looking into the remaining questions, and then do a dataset diff to make sure the actual result of harvesting it the same.

udata/harvest/backends/dcat.py
streino: It's in the [CSW spec](https://portal.ogc.org/files/80534), section 8.2.2.1.1 page 72, and in their example in Annex G.1 (citing rev2 here, but original spec has the same mention):

> If the typeNames attribute of a query equals to ‘gmd:MD_Metadata’ (‘gmd’ representing the ‘http://www.isotc211.org/2005/gmd’ namespace) any queryable that is part of the associated filter must be represented by a qualified name with a prefix (e.g. ‘apiso’), representing the ‘http://www.opengis.net/cat/csw/apiso/1.0’ namespace. This is true for both application profile queryables as well as for the OGC common core queryables (which are mapped to the gmd metadata schema then).
>
> Case sensitivity is as follows: For the common queryables use the same case as defined in the base specification (e.g. ‘apiso:title’), for the additional queryables use the cases as defined in this profile (tables 9-14), e.g. ‘apiso:RevisionDate’. 

I figured since most servers we've dealt with so far ignore the namespace (I tested on GeoNetwork and Geo-IDE, and looked at the GeoNetwork and pycsw code), we might as well put it in case some server is more strict with the spec (found some Oracle doc giving examples with `apiso` for instance), or if GeoNetwork ever act on their [note](https://github.com/geonetwork/core-geonetwork/blob/3.10.4/csw-server/src/main/java/org/fao/geonet/kernel/csw/services/getrecords/FieldMapper.java#L92).

udata/harvest/backends/dcat.py
streino: `search_results` contains a list of `rdf:RDF` (csw-dcat) or `gmd:MD_Metadata` (csw-iso) elements. 

We could either: 
1) Parse the full CSW response at once (before we XPath the elements to retrieve `search_results`) once, which is what I assume you suggest here.
2) Parse each `result` element (either `rdf:RDF` or `gmd:MD_Metadata`) element independently.

In both cases we'll parse the same amount of string, just all at once (1) or in bits (2).

On master (REF), we do (1) for [csw-iso](https://github.com/opendatateam/udata/blob/master/udata/harvest/backends/dcat.py#L402), and (2) for [csw-dcat](https://github.com/opendatateam/udata/blob/master/udata/harvest/backends/dcat.py#L302).

I went for (2) in the refactoring because:
- The CSW response returns a list of independent records. In csw-dcat, each record is returned into its own `rdf:RDF` top element (containing all its relationships). Same for csw-iso with `gmd:MD_Metadata`. 
  - In csw-iso it worked because the SEMIC ISO-DCAT converter will take an XML with multiple `gmd:MD_Metadata` and output a single `rdf:RDF` graph (though I'm not sure it was intended and/or fully tested for that usecase). 
  - In csw-dcat, we'd have to somehow merge the list of `rdf:RDF` items. We can do that either by parsing each and merging the graphs (defeats the purpose), or manipulating the XML tree to cut the `rdf:RDF` layer and fuse the children together under a single `rdf:RDF` root, but we'll end up with duplicate declarations of some related elements that are tied to different records (e.g. same organization). I think this is the biggest argument against going the graph-per-page route.
- It seams easier to reason per-item than per-graph, since the whole process is per-item. We parse the graph but right after we select each Dataset or Dataservice, etc. So might as well go per-item. Caveat: there is a per-page level that becomes useless, but it's not much of an overhead.
- No significant difference in perf. And parsing isn't a perf issue anyway.

See also my [test results comment](https://github.com/opendatateam/udata/pull/3378#issuecomment-3132797294) under section "remote_id has no dupe in REF, dupes in PR". I think per-item makes it less risky because we're not depending on ISO-DCAT or rdflib logic and the per-item RDF graphs we preserve in `HarvestItem` will be closer to the truth.

udata/harvest/backends/dcat.py
streino: I did find an explanation. Added as doc comment here: https://github.com/opendatateam/udata/pull/3378/files#diff-ffb68741d786c85e85cbc06802bc6afc1e23539c113a37b7138ee03037e702c9R245

streino:
I re-harvested the 3 catalogs that had weird diffs and it's all gone. PR was right the first time around, and REF caught up. The only remaining diff is the "datara" conflict on `remote_id`, but that one makes sense.

A re-re-harvest remained stable, so there must have been something wrong on those catalogs with my initial test run. 

I figure if `harvest_job.items` and `datasets` are consistent we don't need to diff the RDF graphs (which are a pain to diff since we changed page sizes). But WDYT @maudetes?

udata/harvest/backends/dcat.py
```diff
@@ -235,104 +233,159 @@ def get_node_from_item(self, graph, item):
...
+    # [5] GeoNetwork - GetRecords ignores `typeNames`
+    #     https://github.com/geonetwork/core-geonetwork/blob/3.10.4/csw-server/src/main/java/org/fao/geonet/kernel/csw/services/getrecords/CatalogSearcher.java#L194
+    CSW_REQUEST: ClassVar[str] = """
+    <csw:GetRecords xmlns:apiso="http://www.opengis.net/cat/csw/apiso/1.0"
+                    xmlns:csw="http://www.opengis.net/cat/csw/2.0.2"
+                    xmlns:ogc="http://www.opengis.net/ogc"
+                    service="CSW" version="2.0.2" outputFormat="application/xml"
+                    resultType="results" startPosition="{start}" maxRecords="25"
+                    outputSchema="http://www.w3.org/ns/dcat#">
+      <csw:Query typeNames="gmd:MD_Metadata">
```
streino, line 280: Doc will need updating since it mentions the query: https://doc.data.gouv.fr/moissonnage/dcat/#geonetwork-v4
Is this something I can do (and in that case, where do I find it), or will you have to do it?

streino:
Oops, forgot to push my latest commits (my tests ran on it).

udata/harvest/backends/dcat.py
```diff
@@ -235,104 +233,165 @@ def get_node_from_item(self, graph, item):
...
+                    xmlns:csw="http://www.opengis.net/cat/csw/2.0.2"
+                    xmlns:ogc="http://www.opengis.net/ogc"
+                    service="CSW" version="2.0.2" outputFormat="application/xml"
+                    resultType="results" startPosition="{start}" maxRecords="25"
+                    outputSchema="{output_schema}">
+      <csw:Query typeNames="gmd:MD_Metadata">
+        <csw:ElementSetName>full</csw:ElementSetName>
+        <csw:Constraint version="1.1.0">
+          <ogc:Filter>
+            <ogc:Or>
```
streino, line 284: We might want to also add the "model" type.

See https://github.com/opendatateam/udata/pull/3378#issuecomment-3132797294 for details:

> **remote_id found in REF, not found in PR**
> 
> Two such cases in "Moisssonneur SI Eau" (csw-dcat): [sieau:2b5ab7c3-f0d3-4653-9fd5-cfb133a914fb](https://data.ofb.fr/catalogue/srv/eng/catalog.search#/metadata/2b5ab7c3-f0d3-4653-9fd5-cfb133a914fb), [sieau:7db4d17b-0f5f-4312-b1a5-8ec93cc9ab72](https://data.ofb.fr/catalogue/srv/eng/catalog.search#/metadata/7db4d17b-0f5f-4312-b1a5-8ec93cc9ab72).
> 
> Both have type = "model". REF didn't filter on record type, and PR does, but doesn't include "model", so they were ignored.
> 
> @maudetes WDYT? I don't know that type but IMO it's close-enough to a dataset that it makes sense to include it.

maudetes: Agreed, we can add the `model` type if it get converted to Dataset by the GeoDCAT-AP conversion anyway

udata/harvest/backends/dcat.py
maudetes: I think I can do it [here](https://guides.data.gouv.fr/guide-data.gouv.fr/moissonnage/moissonnage-des-plateformes-geographiques) (we really should deactivate doc.data.gouv.fr)

udata/harvest/backends/dcat.py
streino: Hmmm, you raise a good point. 

It's converted to Dataset when requesting DCAT directly (csw-dcat), so this means the *old* GN DCAT converter did it. But it's not covered by the SEMIC converter, and I can't find anything related to models in the new GN converter (not declared in [dcat](https://github.com/geonetwork/core-geonetwork/blob/19865e608a528abb92ac9c7ff798e422767a54b6/schemas/iso19115-3.2018/src/main/plugin/iso19115-3.2018/formatter/dcat/dcat-variables.xsl#L75) nor in [geodcat-ap](https://github.com/geonetwork/core-geonetwork/blob/19865e608a528abb92ac9c7ff798e422767a54b6/schemas/iso19115-3.2018/src/main/plugin/iso19115-3.2018/formatter/eu-geodcat-ap/eu-geodcat-ap-variables.xsl#L22)).

So maybe it's not such a good idea, since it won't be handled the same way between csw-dcat and csw-iso, nor will it even be consistent within csw-dcat depending on the source GN version. Or if we're ok with the inconsistencies, we add it now and we'll be future proof?

maudetes: Hmm then if it's not mentioned in the new versions, I wouldn't add it for now. Let's add it explicitly only when it makes sense.

streino: Works for me!

udata/harvest/backends/dcat.py
```diff
@@ -235,104 +233,165 @@ def get_node_from_item(self, graph, item):
...
+    The parsing of items is then the same as for the DcatBackend.
+    """
 
-        if any(break_conditions):
-            return None
-        else:
-            return next_record
+    display_name = "CSW-DCAT"
 
+    # CSW_REQUEST is based on:
```
maudetes, line 250: Thank you for this nice documentation!

udata/harvest/backends/dcat.py
maudetes: Thank you for the clear explanation, I'm convinced! :+1:

PR #3379, opened 2025-07-21
Fix duplicate contact point creation

ThibaudDauce:


udata/tests/api/test_contact_points.py
```diff
@@ -75,6 +78,53 @@ def test_contact_point_api_create_email_or_contact_form(self, api):
...
+        assert ContactPoint.objects.count() == 1
+
+        contact_point = ContactPoint.objects.first()
+        assert contact_point.owner is None
+        assert contact_point.organization.id == org_a.id
+
+        org_b = OrganizationFactory(members=[])
+        data["organization"] = str(org_b.id)
+        response = api.post(url_for("api.contact_points"), data=data)
+        assert400(response)
```
maudetes, line 118: Does it make sense it's a 400 here?

udata/core/contact_point/api.py
```diff
@@ -31,9 +33,9 @@ class ContactPointsListAPI(API):
     def post(self):
         """Creates a contact point"""
         form = api.validate(ContactPointForm)
-        contact_point = form.save()
+        contact_point, created = ContactPoint.objects.get_or_create(**form.data)
 
-        return contact_point, 201
+        return contact_point, 201 if created else 200
```
maudetes, line 38: Wouldn't it be more obvious to fail?

udata/tests/api/test_contact_points.py
ThibaudDauce: It's the behavior of the other forms about this. So if we change the 400 to 403 it'll change for all forms

udata/core/contact_point/api.py
ThibaudDauce: I think returning the existing contact point will be easier for the frontend to catch and use the existing contact point?

maudetes: I think it would be better to show an error, ex : `this contact point already exists`.
And optionally show the existing contact point to the user for selection

udata/migrations/2025-07-22-fix-duplicate-contact-points.py
```diff
@@ -0,0 +1,53 @@
...
+import logging
+
+from udata.core.contact_point.models import ContactPoint
+from udata.core.dataservices.models import Dataservice
+from udata.core.dataset.models import Dataset
+
+log = logging.getLogger(__name__)
+
+
+def migrate(db):
```
maudetes, line 15: I think you can add a small log on this migration on the number of contact points removed :)

PR #3380, opened 2025-07-21
Change CSW harvesters XML processor from lxml to saxonche

streino:
Replaces https://github.com/opendatateam/udata/pull/3250.

Only switch the parser. Will update to the new SEMIC XSLT in another PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Current (in progress)
 
+- Consolidate CSW harvesting logic [#3378](https://github.com/opendatateam/udata/pull/3378)
 - Remove placeholder Jinja filter [#3372](https://github.com/opendatateam/udata/pull/3372/)
 - Add `self_web_url` for discussion [#3375](https://github.com/opendatateam/udata/pull/3375)
 - Fix dataset schemas endpoint when no resource [#3373](https://github.com/opendatateam/udata/pull/3373)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Current (in progress)
 
-- Consolidate CSW harvesting logic [#3378](https://github.com/opendatateam/udata/pull/3378)
+- Consolidate CSW harvesting logic. [#3378](https://github.com/opendatateam/udata/pull/3378)
+  - :warning: Config setting `HARVEST_ISO19139_XSL_URL` is renamed to `HARVEST_ISO19139_XSLT_URL`.
 - Remove placeholder Jinja filter [#3372](https://github.com/opendatateam/udata/pull/3372/)
 - Add `self_web_url` for discussion [#3375](https://github.com/opendatateam/udata/pull/3375)
 - Fix dataset schemas endpoint when no resource [#3373](https://github.com/opendatateam/udata/pull/3373)
--- a/udata/harvest/backends/dcat.py
+++ b/udata/harvest/backends/dcat.py
@@ -1,6 +1,6 @@
 import logging
 from datetime import date
-from typing import Generator
+from typing import ClassVar, Generator
 
 import lxml.etree as ET
 from flask import current_app
@@ -237,7 +237,7 @@ class DcatBackend(BaseBackend):
 class CswDcatBackend(DcatBackend):
     display_name = "CSW-DCAT"
 
-    CSW_REQUEST = """
+    CSW_REQUEST: ClassVar[str] = """
     <csw:GetRecords xmlns:csw="http://www.opengis.net/cat/csw/2.0.2"
                     xmlns:ogc="http://www.opengis.net/ogc"
                     service="CSW" version="2.0.2" resultType="results"
@@ -342,7 +342,7 @@ class CswIso19139DcatBackend(CswDcatBackend):
         ),
     )
 
-    CSW_REQUEST = """
+    CSW_REQUEST: ClassVar[str] = """
     <csw:GetRecords xmlns:csw="http://www.opengis.net/cat/csw/2.0.2"
                     xmlns:ogc="http://www.opengis.net/ogc"
                     service="CSW" version="2.0.2" resultType="results"
--- a/udata/harvest/backends/dcat.py
+++ b/udata/harvest/backends/dcat.py
@@ -6,6 +6,7 @@ import lxml.etree as ET
 from flask import current_app
 from rdflib import Graph
 from rdflib.namespace import RDF
+from typing_extensions import override
 
 from udata.core.dataservices.rdf import dataservice_from_rdf
 from udata.core.dataset.rdf import dataset_from_rdf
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Current (in progress)
 
+- Consolidate CSW harvesting logic. [#3378](https://github.com/opendatateam/udata/pull/3378)
+  - :warning: Config setting `HARVEST_ISO19139_XSL_URL` is renamed to `HARVEST_ISO19139_XSLT_URL`.
```
maudetes: I think you'll need to update the changelog with this PR

udata/harvest/backends/dcat.py
```diff
@@ -235,98 +232,109 @@ def get_node_from_item(self, graph, item):
...
+            text = response.text
+            tree = self.saxon_proc.parse_xml(xml_text=text)
+            self.xpath_proc.set_context(xdm_item=tree)
+
+            # Using * namespace so we don't have to enumerate ows versions
+            if self.xpath_proc.evaluate("/*:ExceptionReport"):
+                raise ValueError(f"Failed to query CSW:\n{text}")
+
+            if r := self.xpath_proc.evaluate("/csw:GetRecordsResponse/csw:SearchResults"):
+                search_results = r.head
```
maudetes, line 366: I'm unsure about the `r.head` here

udata/harvest/backends/dcat.py
```diff
@@ -235,98 +232,109 @@ def get_node_from_item(self, graph, item):
...
+      </csw:Query>
+    </csw:GetRecords>
+    """
+
+    @override
+    def inner_init(self) -> None:
+        self.saxon_proc = PySaxonProcessor(license=False)
+        self.saxon_proc.set_configuration_property(
+            "http://saxon.sf.net/feature/strip-whitespace", "all"
+        )
```
maudetes, line 341: Why this configuration? Do we need to add a configuration to make sure we won't resolve external entities?
For reference, we've had a security issue [with the default lxml parser](https://github.com/opendatateam/udata/pull/3205).

udata/harvest/backends/dcat.py
```diff
@@ -235,98 +232,109 @@ def get_node_from_item(self, graph, item):
...
+        self.xpath_proc = self.saxon_proc.new_xpath_processor()
+        self.xpath_proc.declare_namespace("csw", CSW_NAMESPACE)
```
maudetes, line 343: Do we need to declare a namespace here?

CHANGELOG.md
streino: Yep, waiting for stuff to stabilize before I tackle it.

udata/harvest/backends/dcat.py
streino: Functions like `evaluate()` return a [`PyXdmValue`](https://www.saxonica.com/saxon-c/doc12/html/saxonc.html#PyXdmValue), which is a list of [`PyXdmItem`](https://www.saxonica.com/saxon-c/doc12/html/saxonc.html#PyXdmItem). Even if the xpath only matched once. Same a lxml where `r = find("//xpath")` returns a list and you need to get `r[0]`.

`PyXdmValue.head` will return the first `PyXdmItem` of the list it it exists.

It gets a bit confusing because `PyXdmItem` is a subclass of `PyXdmValue`, and `PyXdmItem.head` returns itself. Also, functions can take `PyXdmItem` but mostly return `PyXdmValue`. So you end up calling `head` after every function call to get those `PyXdmItem` back.

It's clearly more fluid in lxml.

udata/harvest/backends/dcat.py
streino: Saxon is very thorough and will return nodes (in the XML sense == `PyXdmItem`) for *everything* it parses, including anonymous text nodes that capture the whitespace *between* XML elements. Took me a while to figure it out...

`strip-whitespace` strips that whitespace inter-node spacing so the "whitepace nodes" aren't created. Otherwise, we'd have to filter them out from [`search_results.children`](https://github.com/opendatateam/udata/pull/3380/files#diff-ffb68741d786c85e85cbc06802bc6afc1e23539c113a37b7138ee03037e702c9R292). The default value for the option is "ignorable" and it still produced the whitespace nodes, so I went with "all". If I understand [the doc](https://www.saxonica.com/html/documentation12/configuration/config-features.html) correctly, "ignorable" not working might be related to something missing from the schema. But this is beyond my knowledge of the XML ecosystem.

I'll double-check, but from my understanding it should never touch the spacing within actual text elements.

lxml hides all this but it looks like you can access the inter-node spacing via the [`tail` property](https://stackoverflow.com/questions/25755556/setting-element-content-with-lxml-removes-trailing-whitespaces).

streino: Will do about external entities. I though I'd done it at some point but I don't remember why I removed it.

udata/harvest/backends/dcat.py
streino: `xpath_proc` is going to [process a CSW response](https://github.com/opendatateam/udata/pull/3380/files#diff-ffb68741d786c85e85cbc06802bc6afc1e23539c113a37b7138ee03037e702c9R286), so I guess we need it?

maudetes: Ok, I thought it would use the namespace fed by the parsed XML response

streino: You're right, it should... I blindly copied what was [already there](https://github.com/opendatateam/udata/pull/3380/files#diff-ffb68741d786c85e85cbc06802bc6afc1e23539c113a37b7138ee03037e702c9L302), but I'll test without the namespace.

streino: Tried removing it, but it is indeed needed by the xpath processor :
```
  File "/var/home/p-ecospheres/workspace/datagouv/udata.saxon/udata/harvest/backends/dcat.py", line 348, in walk_graph
    if r := self.xpath_proc.evaluate("/csw:GetRecordsResponse/csw:SearchResults"):
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/saxonc/saxonc.pyx", line 3356, in saxonche.PyXPathProcessor.evaluate
saxonc.PySaxonApiError: XPST0081: Namespace prefix 'csw' has not been declared. Line number: -1
```

streino:
Ran the same diff tests as the ones for https://github.com/opendatateam/udata/pull/3378.
Zero diff on both job.items and datasets :tada:

udata/harvest/backends/dcat.py
```diff
@@ -329,19 +342,23 @@ def walk_graph(self, url: str, fmt: str) -> Generator[tuple[int, Graph], None, N
...
-            content = response.content
-            tree = ET.fromstring(content, parser=self.xml_parser)
-            if tree.tag == "{" + OWS_NAMESPACE + "}ExceptionReport":
-                raise ValueError(f"Failed to query CSW:\n{content}")
+            text = response.text
+            tree = self.saxon_proc.parse_xml(xml_text=text)
+            self.xpath_proc.set_context(xdm_item=tree)
 
-            search_results = tree.find("csw:SearchResults", {"csw": CSW_NAMESPACE})
-            if not search_results:
```
maudetes, line 350: For your information, following the https://github.com/opendatateam/udata/pull/3378/ PR, we now have this warning, that should disappear with this PR.
```
/srv/dev/lib/python3.11/site-packages/udata/harvest/backends/dcat.py:350: FutureWarning: Truth-testing of elements was a source of confusion and will always return True in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
  if not search_results:
```

maudetes (approved):
Thank you for the additional security configs! :clap: 

Seems to work nicely and indeed, testing it out for any diff does not yield anything, looks like we're iso :+1:

PR #3382, opened 2025-07-22
Fix not returning 404 for harvest sources

ThibaudDauce:


udata/harvest/tasks.py
```diff
@@ -1,20 +1,21 @@
...
-from .models import HarvestJob, HarvestSource
+from .models import HarvestJob
 
 log = get_logger(__name__)
 
 
 @job("harvest", route="low.harvest")
-def harvest(self, ident):
-    log.info('Launching harvest job for source "%s"', ident)
+def harvest(self, ident_or_source):
```
maudetes: I think this is always an `ident` here?

udata/harvest/actions.py
```diff
@@ -44,9 +44,11 @@ def list_sources(owner=None, deleted=False):
     return list(sources)
 
 
-def get_source(ident):
+def get_source(ident_or_source: HarvestSource | str) -> HarvestSource:
```
maudetes: I don't understand why calling `get_source` if we already have a source.
I don't see why we wouldn't know if we're in the context of one or th other and I think we should enforce to make it clear.

ThibaudDauce: To avoid repeting the same condition everywhere. These functions are called from tasks (where we have only IDs) and API (where we already have the full object available)

udata/harvest/tasks.py
ThibaudDauce: yes you're right, fixed in f2f3b57c6

udata/harvest/actions.py
maudetes: I don't think it's the case anymore since you've added the routing converter?

ThibaudDauce: It's the case in tasks that don't use the routing converter?

maudetes: I mean, we use get_source in tasks because we always have a string, whereas we don't use it in API since we now have a source object.
I don't think `get_source` should get called with a source object

ThibaudDauce: I don't understand why it bother you so much but I've done the (much bigger) change to always pass a Source object and change all the calling sites. Not sure if you had a different implementation in mind…

maudetes: Sorry for the bother! I've just re-added the `get_source` util in https://github.com/opendatateam/udata/pull/3382/commits/d7eb36e88f712c5c46ed1ed481cda3900a7dcc0e.

PR #3384, opened 2025-07-22
Add search to discussions

ThibaudDauce:


udata/core/discussions/api.py
nicolaskempf57, line 1: Do we want new tests for the q argument ?

ThibaudDauce: Done in ac62d9588

PR #3385, opened 2025-07-23
Migrate udata-ckan

ThibaudDauce:


requirements/develop.pip
```diff
@@ -314,6 +314,11 @@ httpretty==1.1.4
     # via
     #   -c requirements/test.pip
     #   -r requirements/test.in
+humanfriendly==10.0
```
maudetes, line 317: As discussed, in a second iteration, we may want to get rid of this lib :)

udata/harvest/backends/ckan/translations/udata-ckan.pot
maudetes, line 1: Do we want to have translations in this file instead of the general translation file?

PR #3386, opened 2025-07-23
Migrate udata-metrics

ThibaudDauce:


udata/core/metrics/helpers.py
```diff
@@ -68,6 +69,15 @@ def get_metrics_for_model(
         return [{} for _ in range(len(metrics_labels))]
 
 
+def get_download_url(model: str, id: Union[str, ObjectId, None]) -> str:
+    api_namespace = model + "s" if model != "site" else model
+    base_url = f"{current_app.config['METRICS_API']}/{api_namespace}/data/csv/"
```
maudetes, line 74: Can you add `METRICS_API = None` in settings.py ? :)

maudetes (approved):
:broom:

udata/core/metrics/helpers.py
ThibaudDauce: It's already done :-)

PR #3387, opened 2025-07-23
Add edito pages

ThibaudDauce:


udata/core/pages/models.py
```diff
@@ -0,0 +1,102 @@
...
+        sortable="created",
+        auditable=False,
+    )
+    updated_at = field(
+        db.DateTimeField(
+            verbose_name=_("Last modification date"), default=datetime.utcnow, required=True
+        ),
+        readonly=True,
+        auditable=False,
+    )
```
maudetes: Don't you want to inherit `Datetimed` class? I would use the same vocabulary at least

udata/core/pages/models.py
```diff
@@ -0,0 +1,102 @@
...
+from udata.core.owned import Owned
+from udata.core.reuse.models import Reuse
+from udata.i18n import lazy_gettext as _
+from udata.models import db
+
+
+@generate_fields()
+class Bloc(db.EmbeddedDocument):
+    id = field(db.AutoUUIDField(primary_key=True))
+    type = field(db.StringField(required=True))
```
maudetes: Could we have a `class` instead of having a `type` property? I guess having a similar interface as the [discussions subject](https://guides.data.gouv.fr/guide-data.gouv.fr/readme-1/reference/discussions#post-discussions) (or [the upcoming element topic](https://github.com/opendatateam/udata/pull/3318)) would be nice? :)

udata/core/site/api.py
```diff
@@ -45,6 +54,22 @@ def get(self):
         """Site-wide variables"""
         return current_site
 
+    @api.doc(id="get_site")
+    @api.marshal_with(site_fields)
+    def patch(self):
```
maudetes, line 40: Do we want to merge this with the SiteSettings and `home_datasets` and `home_reuses` dedicated endpoints instead?

udata/core/pages/models.py
```diff
@@ -0,0 +1,102 @@
...
+    main_link_url = field(db.StringField())
+    main_link_title = field(db.StringField())
+
+    links = field(db.EmbeddedDocumentListField(LinkInBloc))
+
+
+@generate_fields()
+class Page(Auditable, Owned, db.Document):
+    blocs = field(
+        db.EmbeddedDocumentListField(Bloc),
```
maudetes, line 89: I find the `EmbeddedDocumentListField(Bloc)` elegant! :eyes:

udata/api_fields.py
```diff
@@ -147,9 +163,30 @@ def constructor_read(**kwargs):
         #     2. `__additional_field_info__` of the inner field
         #     3. `__additional_field_info__` of the parent
         inner_info: dict = getattr(field.field, "__additional_field_info__", {})
-        field_read, field_write = convert_db_to_field(
-            f"{key}.inner", field.field, {**info, **inner_info, **info.get("inner_field_info", {})}
-        )
+        generic = info.get("generic", {})
```
maudetes: At first sight, I personally find the `generic` logic hard to understand. I'm not sure if we could have something transparent however?

udata/core/pages/api.py
```diff
@@ -0,0 +1,46 @@
...
+        return page
+
+    @api.secure
+    @api.doc("update_page")
+    @api.expect(Page.__write_fields__)
+    @api.marshal_with(Page.__read_fields__)
+    def put(self, page: Page):
+        page = patch(Page(), request)
+        if not page.owner and not page.organization:
+            page.owner = current_user._get_current_object()
```
maudetes: I'm confused about this logic on a put?

maudetes (commented):
So cool! :clap:

Well done for this implementation. I'm not sure everything is clear to me (as you'll see in comments), but I find the approach very nice :)

udata/core/pages/models.py
ThibaudDauce: Yes, always forget this mixin exists :-)

udata/core/pages/models.py
ThibaudDauce: Yes it's better!

udata/core/site/api.py
ThibaudDauce: Yes, I don't think `home_datasets` and `home_reuses` are used anymore, so this new page system will replaced it. I quickly created this endpoint but I could use the new api_fields system to patch the `Site` object it will be simpler too.

udata/api_fields.py
ThibaudDauce: Maybe if we use a `class` attribute instead of a `type` we could just find the class without having to have a lookup table `type` to `class`… But it could be dangerous to allow to have any class, so we may need to have a list of authorized classes… Or maybe check the inheritance is enough?

udata/core/pages/api.py
ThibaudDauce: Yes it may not be necessary, copied from the `POST`. But I'm not sure why it's necessary in the `POST` too. I think it could be handle but the `Owned` class since we always want to do that in all classes using these attributes…

maudetes: I thought the same on the `POST` too :p But I saw you did this on the reuse POST API too and didn't check if there was a reason for it or not :p

udata/api_fields.py
maudetes: I think it would be ideal indeed! I think relying on the inheritance would be the nicest if doable

udata/core/pages/models.py
ThibaudDauce: Done in a20897116

udata/api_fields.py
ThibaudDauce: Done in 62561e602

udata/core/pages/api.py
```diff
@@ -0,0 +1,46 @@
...
+common_doc = {"params": {"page": "The page ID or slug"}}
+
+
+@ns.route("/", endpoint="pages")
+class PagesAPI(API):
+    @api.secure
+    @api.doc("create_page", responses={400: "Validation error"})
+    @api.expect(Page.__write_fields__)
+    @api.marshal_with(Page.__read_fields__, code=201)
+    def post(self):
```
maudetes, line 29: I think a `get` route would be nice here to help us debugging and easily get all the existing pages.

maudetes, line 29: I guess it may come from cdata logic, but playing on https://demo.data.gouv.fr/admin/site/edito/datasets, reuses & API, I ended up with 18 object counts in the end.
It seemed that multiple times, it created new objects instead of saving it to the existing one.

nicolaskempf57: Do we want to restrict it to admins ? This way, we have time to work on `private` or other fields for the organizations customization

nicolaskempf57: We may get performance issues with this route because it is fetching all refs even when this is deeply nested like :
pages with blocs with reuses with datasets with organization

udata/core/pages/api.py
nicolaskempf57: It is necessary for the `POST` endpoint because our validation allow organization and owner to be empty at the same time.
For this one, it is not necessary because this is not really a PUT, we are patching the existing page so there is always an owner or an organization.

udata/core/pages/api.py
maudetes: Yes, we can restrict it to admins to get started.
Regarding the performance issues, we could tackle it using reference fields, ex `dataset_ref_fields` to return a minimal payload. At least, we should make sure not to return the dataset resources for example.

maudetes: Done in https://github.com/opendatateam/udata/pull/3387/commits/8fe49224321133ee8180a65e85cad9d032d2a372

udata/core/site/models.py
```diff
@@ -19,14 +20,18 @@ class SiteSettings(db.EmbeddedDocument):
     home_reuses = db.ListField(db.ReferenceField(Reuse))
```
maudetes, line 20: Should we remove the SiteSettings while we're at it? Or do it later on?

ThibaudDauce: We'll do it in another PR if it's ok for you?

PR #3391, opened 2025-07-28
Compute org aggregate metrics only once at the end of harvest

streino:
This came up during local harvesting tests on a docker setup: 

- _py-spy `udata harvest run` on `master`: duration 10+ minutes (98% of which is `get_stock_metrics`)._

    pyspy-reference

- _py-spy `udata harvest run` on `test/aggregate-metrics`: duration 22 seconds (a large part is still `get_stock_metrics`, but at least it's only seconds now)._

    pyspy-test-aggregate-metrics

The perf hit might be less important in prod with a real mongo cluster, but it's still quite wasteful. It could also explain at least part of the recent issues we've had with harvesting on demo.

I leave the PR as draft because you might want to address the issue differently. This is just a quick minimal patch to evaluate the impact. IMO aggregate statistics could be run as an independent daily batch but it'd be a larger change. Might be worth merging something quick to mitigate the impact.


udata/harvest/backends/dcat.py
```diff
@@ -68,6 +68,10 @@ def extract_graph(source, target, node, specs):
 class DcatBackend(BaseBackend):
     display_name = "DCAT"
 
+    def __init__(self, *args, **kwargs):
```
streino, line 69: Or I can use the `inner_init()` method I'm introducing in the [CSW refactor PR](https://github.com/opendatateam/udata/pull/3378/files#diff-d16bd4e81540b081ea7786e5f3e67961a92b35d6064d71c1c5f6dd075ac89a93R157).
Whichever you prefer, and I'll adapt the other PR accordingly.

udata/harvest/backends/dcat.py
```diff
@@ -214,7 +221,10 @@ def inner_process_dataset(self, item: HarvestItem, page_number: int, page: Graph
 
         dataset = self.get_dataset(item.remote_id)
         remote_url_prefix = self.get_extra_config_value("remote_url_prefix")
-        return dataset_from_rdf(page, dataset, node=node, remote_url_prefix=remote_url_prefix)
+        dataset = dataset_from_rdf(page, dataset, node=node, remote_url_prefix=remote_url_prefix)
+        if dataset.organization:
+            self.organizations_to_update.add(dataset.organization)
```
ThibaudDauce, line 231:
```suggestion
            dataset.organization.do_not_compute_metrics = true
            self.organizations_to_update.add(dataset.organization)
```

udata/core/organization/models.py
```diff
@@ -303,19 +303,20 @@ def count_members(self):
...
-        self.metrics["datasets_by_months"] = get_stock_metrics(
-            Dataset.objects(organization=self).visible(), date_label="created_at_internal"
-        )
-        self.metrics["datasets_followers_by_months"] = get_stock_metrics(
-            Follow.objects(following__in=Dataset.objects(organization=self)), date_label="since"
-        )
-        self.metrics["datasets_reuses_by_months"] = get_stock_metrics(
-            Reuse.objects(datasets__in=Dataset.objects(organization=self)).visible()
-        )
+        if compute_aggregates:
```
ThibaudDauce:
```suggestion
        if not self.do_not_compute_metrics:
```

maudetes:
> I'm afraid that we are not covering all signals in tests and missing some **kwargs… Is it possible to do that with an attribute on the org instead of an attribute? I'm not sure if it's better but it may be safer… What do you think @maudetes ?

I was thinking that a more general fix would be to check for visibility-related fields modification to know if we should or not compute `count_datasets` in `update_datasets_metrics`.
This may be related to the work in [this PR](https://github.com/opendatateam/udata/pull/3335).

udata/harvest/backends/dcat.py
streino: I'm worried we're introducing a state that will need to be managed afterward. We should at least reset it at the end of the run. And what if the harvester aborts before state is reset?

ThibaudDauce: It's not a big deal since `dataset` is a local variable, no? It will be refetch from the database in another attempt.

streino: Ah right, not something you map in the model obviously :)

udata/harvest/backends/dcat.py
streino: Will do the other way. `__init__` is simpler.

udata/harvest/backends/dcat.py
```diff
@@ -76,6 +80,9 @@ def inner_harvest(self):
             self.process_one_datasets_page(page_number, page)
             pages.append((page_number, page))
 
+        for org in self.organizations_to_update:
+            org.count_datasets()
```
maudetes: Are we sure that the `org` has `compute_aggregate_metrics=True` here? Shouldn't we set it before calling `org.count_datasets`?

streino: It's set to True in [`Organization.__init__`](https://github.com/opendatateam/udata/pull/3391/files#diff-a19a9cbf1d8470b63726d645e31a91d271e758249d0554d0c423ebc201a0a764R193). Not too nice for a one-trick variable, but since this workaround should go away I figured it's ok. The alternative is to use `hasattr` (not a fan myself).

maudetes: Yes, sorry I mean in the context of harvest, we never set it back to True, which we probably should in this `for` loop?
I think we're still using the organization object that has been loaded in memory.

streino: Right :facepalm:

PR #3392, opened 2025-07-28
Add missing slashes

ThibaudDauce:


udata/core/dataservices/api.py
```diff
@@ -210,7 +210,7 @@ def delete(self, dataservice, dataset):
         return None, 204
 
 
-@ns.route("/<dataservice:dataservice>/rdf", endpoint="dataservice_rdf", doc=common_doc)
+@ns.route("/<dataservice:dataservice>/rdf/", endpoint="dataservice_rdf", doc=common_doc)
```
maudetes: I would not add a trailing slash on rdf routes since they use content negociation and redirect to files (ie, rdf.xml rdf.jsonld, etc.)

udata/core/dataset/api.py
```diff
@@ -419,7 +419,7 @@ def delete(self, dataset):
         return dataset
 
 
-@ns.route("/<dataset:dataset>/rdf", endpoint="dataset_rdf", doc=common_doc)
+@ns.route("/<dataset:dataset>/rdf/", endpoint="dataset_rdf", doc=common_doc)
```
maudetes: same

udata/core/dataset/api.py
```diff
@@ -476,7 +476,7 @@ def delete(self, dataset, badge_kind):
         return badges_api.remove(dataset, badge_kind)
 
 
-@ns.route("/r/<uuid:id>", endpoint="resource_redirect")
+@ns.route("/r/<uuid:id>/", endpoint="resource_redirect")
```
maudetes: hmm the main permalink (with locale, ex https://www.data.gouv.fr/fr/datasets/r/7c3a854b-7344-4c68-b1f9-9d651b4ca823) does not have a trailing slash.
Not sure what's cleaner here since 1. it points probably towards a file and 2. it would imply modifying the new permalink (again)?

udata/core/organization/api.py
```diff
@@ -229,7 +229,7 @@ def get(self, org):
         return csv.stream(adapter, "{0}-datasets-resources".format(org.slug))
 
 
-@ns.route("/<org:org>/catalog", endpoint="organization_rdf", doc=common_doc)
+@ns.route("/<org:org>/catalog/", endpoint="organization_rdf", doc=common_doc)
```
maudetes: Same as rdf

udata/core/site/api.py
```diff
@@ -100,7 +100,7 @@ def get(self, format):
         return redirect(url)
 
 
-@api.route("/site/catalog", endpoint="site_rdf_catalog")
+@api.route("/site/catalog/", endpoint="site_rdf_catalog")
```
maudetes: Same as rdf

udata/core/discussions/api.py
```diff
@@ -247,7 +247,7 @@ def delete(self, id):
         return "", 204
 
 
-@ns.route("/<id>/comments/<int:cidx>/spam", endpoint="discussion_comment_spam")
+@ns.route("/<id>/comments/<int:cidx>/spam/", endpoint="discussion_comment_spam")
```
maudetes, line 250: This may be related to the issue that we've had at some point about marking spam comments as non-spam?

udata/core/dataservices/api.py
ThibaudDauce: Done in 22527b493

udata/core/dataset/api.py
ThibaudDauce: Done in 22527b493

udata/core/dataset/api.py
ThibaudDauce: Removed in 22527b493

udata/core/site/api.py
ThibaudDauce: Done in 22527b493

udata/core/organization/api.py
ThibaudDauce: DOne in 22527b493

udata/core/discussions/api.py
ThibaudDauce: Not sure…

PR #3394, opened 2025-07-29
test(harvest): add snapshot testing

ThibaudDauce:
Right now some keys from the diff check because of minors changes but we could improve this in the future to check for more things.

### DIFF
--- a/udata/harvest/tests/test_snapshots.py
+++ b/udata/harvest/tests/test_snapshots.py
@@ -14,6 +14,7 @@ from udata.core.dataset.models import Dataset
 from udata.harvest import actions
 from udata.harvest.models import HarvestJob
 from udata.harvest.tests.factories import HarvestSourceFactory
+from udata.tests.api import PytestOnlyAPITestCase
 
 SNAPSHOTS_DIR = join(dirname(__file__), "snapshots")
 
@@ -42,9 +43,8 @@ harvester_configs = [
 ]
 
 
-@pytest.mark.usefixtures("clean_db")
-@pytest.mark.options(PLUGINS=["dcat", "csw-iso-19139", "ckan"])
-class SnapshotsTest:
+@pytest.mark.options(HARVESTER_BACKENDS=["dcat", "csw-iso-19139", "ckan"])
+class SnapshotsTest(PytestOnlyAPITestCase):
     @pytest.mark.parametrize("harvester_conf", harvester_configs)
     def test_all(self, harvester_conf):
         os.makedirs(SNAPSHOTS_DIR, exist_ok=True)


udata/harvest/tests/test_snapshots.py
```diff
@@ -0,0 +1,127 @@
...
+            "requests_history": m.history,
+            "job": HarvestJob.objects[0].to_dict(),
+            "datasets": [d.to_dict() for d in Dataset.objects],
+            "dataservices": [d.to_dict() for d in Dataservice.objects],
+        }
+
+        if refresh:
+            json.dump(new_data, open(data_path, "w"), indent=2, default=str)
+        else:
+            diff = DeepDiff(
```
streino: How well does it handle insert/delete/~~swaps~~ in the items list? In [my attempts at comparing](https://github.com/opendatateam/udata/pull/3378#issuecomment-3126767369) I had to group by `items.remote_id` to get something useful. I see options in `DeepDiff` like `group_by` or `ignore_order` that might be useful/needed?

udata/harvest/tests/test_snapshots.py
```diff
@@ -0,0 +1,127 @@
...
+        harvester = HarvestSourceFactory(
+            backend=harvester_conf["backend"], url=harvester_conf["url"]
+        )
+
+        data = {}
+        data_path = join(
+            SNAPSHOTS_DIR,
+            f"{harvester.backend}-{harvester.url.replace('://', '_').replace('/', '_')}.json",
+        )
+        refresh = not isfile(data_path)
```
streino: I wonder if it would be better to have separate paths for refresh/not refresh, since most of what's below applies on only one or the other path?

Also, since most checks are `if not refresh` (ie `if not not isfile` :grimacing:), I'd invert the variable.

udata/harvest/tests/test_snapshots.py
streino, line 1: Might want to ignore timed out jobs to make sure you're comparing actual full runs?

ThibaudDauce: On the second run there is no timeout possible since it only call the mock API?

streino: I'm thinking you don't want to save a run if it was incomplete, nor do you want to compare if the new (fresh) run is incomplete.

I might have misunderstood the intent, but except on the first run when you have no previous data, you'll be comparing a fresh run to an older run, right?

ThibaudDauce: No I'll compare the older run (so a successful one since it's done once on local machine, I could add an assert but the developer wants to check the generated file to see if everything worked), to the new run (but all external requests are mocked, so there is no possibility of failure here)

udata/harvest/tests/test_snapshots.py
ThibaudDauce: `ignore_order` is required yes :-)

udata/harvest/tests/test_snapshots.py
streino: Ah ok, so it's more static than what I assumed. NP then!

maudetes (commented):
Nice approach!

I find the tests hard to read because of the mix of writing the snapshots and testing.
Could we add a separated `cli` to build the snapshots?

I think it could also replace the existing contents stored in `udata/harvest/tests/dcat`.

PR #3399, opened 2025-07-31
Ignore Dataset.accessService when processing DataService

streino:
`Dataset.accessService` are anonymous nodes within a `Dataset`. They don't have an identifier and therefore are ignored when processing `DataServices`. It's more efficient to skip them, and avoids reporting a bunch of spurious skipped items.

### DIFF
--- a/udata/harvest/backends/dcat.py
+++ b/udata/harvest/backends/dcat.py
@@ -202,7 +202,10 @@ class DcatBackend(BaseBackend):
         )
 
     def process_one_dataservices_page(self, page_number: int, page: Graph):
+        access_services = {o for _, _, o in page.triples((None, DCAT.accessService, None))}
         for node in page.subjects(RDF.type, DCAT.DataService):
+            if node in access_services:
+                continue
             remote_id = page.value(node, DCT.identifier)
             self.process_dataservice(remote_id, page_number=page_number, page=page, node=node)
 


udata/harvest/backends/dcat.py
```diff
@@ -202,7 +202,10 @@ def is_dataset_external_to_this_page(self, page: Graph, node) -> bool:
         )
 
     def process_one_dataservices_page(self, page_number: int, page: Graph):
+        access_services = {o for _, _, o in page.triples((None, DCAT.accessService, None))}
         for node in page.subjects(RDF.type, DCAT.DataService):
+            if node in access_services:
```
maudetes, line 207: I think it's not related to this PR because we were not supporting this either but just to make sure: we're not linking Dataservice to Datasets if they are only linked by `Distribution>accessService` and not `servesDataset`?

streino: I could only find code for handling datasets from services (= `servesDataset`), but not services from dataset (= `accessService`). From dataset, `resource_from_rdf` directly looks at (and stores) `accessURL`, but nothing about `accessService`.

We have an [Ecospheres ticket about this](https://github.com/ecolabdata/ecospheres/issues/626), which I hope to tackle early fall. But yes, not related to this PR.

PR #3401, opened 2025-08-19
Optimize csv organization export

maudetes:
Prevent huge increase in memory when computing organization resources downloads following its introduction in https://github.com/opendatateam/udata/pull/2973.

This lead to workers (and celery entirely?) crashing periodically. See [the related infra ticket](https://gitlab.com/etalab/data.gouv.fr/infra/-/issues/280).

`export-csv` task for organization model also goes from ~10min to ~1min30 locally.

The expected differences are :
- the columns order change (`downloads` is now before the other metrics columns)
- we only sum download metrics for visible datasets
- we base the count on `resources_downloads`, that is fetched from metric-api (instead of iterating on every resource metrics)


Pierlou (approved):
Thanks for the improvements 🙏

PR #3412, opened 2025-09-03
Emit harvest activities

maudetes:
Use a service user account to set as actor using the setting `HARVEST_ACTIVITY_USER_ID`.

Improve the `changed_fields`  logic used for signals and add more exhaustive tests on activity and auditable functionalities.

## Note
The `document._get_changed_fields()` raises false positive (ex: tags slugified only at clean time, datetime stored as date (see [failing tests](https://github.com/opendatateam/udata/pull/3412/files#diff-abcb39b023219b48c18f7cd304bdd18ee53d8741077787cf17cd93b15e8ff0e7R144-R158))).
Indeed, the internal `_changed_fields` is set when setting an attribute to a "different" value that would be normalized at clean time and stored exactly the same in the mongo document.

In order to prevent emitting false positive update activity, we reload the document  and manually compare the document value with the previous value (stored in `_previous_changed_fields`).

*Deployed on demo on 2025/09/12 with `HARVEST_ACTIVITY_USER_ID = 6749894c92a7225d9bc51da7`. Activités à suivre [ici](https://demo.data.gouv.fr/api/1/activity/?user=6749894c92a7225d9bc51da7)*


udata/harvest/backends/base.py
```diff
@@ -168,6 +168,16 @@ def harvest(self):
         self.job = factory(status="initialized", started=datetime.utcnow(), source=self.source)
 
         before_harvest_job.send(self)
+        # Set harvest_activity_user_id on global context during the run
+        if current_app.config["HARVEST_ACTIVITY_USER_ID"]:
+            try:
+                # Try to fetch the existing harvest activity user
+                User.objects.get(id=current_app.config["HARVEST_ACTIVITY_USER_ID"])
+                g.harvest_activity_user_id = current_app.config["HARVEST_ACTIVITY_USER_ID"]
```
ThibaudDauce: Why not storing the user inside `g` to avoid getting it again in activity code?

maudetes: True! I guess it's cause I've added the check of the user existence in `backends/base.py` in a second iteration only.

Stored the user directly in https://github.com/opendatateam/udata/pull/3412/commits/1b3c857c13b0b59ed93e96d848b0d17b14f61ebd.

PR #3413, opened 2025-09-04
refactor: migrate to pyproject.toml

bolinocroustibat:
Building has been tested successfully.

Question: do we want to have separate locked dependencies files for 100% reproductible installs?

If yes, we could use:

- the very new PEP standard `pylock.toml` for lock files ([specs](https://packaging.python.org/en/latest/specifications/pylock-toml/), [PEP751](https://peps.python.org/pep-0751/)) 
    - 👍
        - future-proof
        - can be managed with native python tool `pip` with `python3 -m pip lock pyproject.toml`
        - supports `extras` and `optional-dependencies`
    - 👎
        - still experimental support
        - not broadly adopted yet by package managers

- the old `requirements.txt` as lock files
    - 👍
        - very retrocompatible
        - can be managed with native python tool `pip`
        - can be managed by the previous tool `pip-compile` from `pip-tools` which was the tool previously used in this project
        - we can reuse the previous *.pip from before this PR
    - 👎
        - not future-proof
        - can be difficult to import with modern package managers
        - needs one file per optional group, bloats the repo
        - bad dependency tree management

- the [Poetry](https://python-poetry.org) lock file `poetry.lock`
    - 👍
        - Poetry is the same tool and file we use in other projects (Hydra) in the team
    - 👎
        - forces contributors to use Poetry 

- the [uv](https://docs.astral.sh/uv/) lock file `uv.lock`
    - 👍
        - uv has become the de-facto standard for modern Python packaging
        - easy to convert to standard `pylock.toml` for future-proofing
        - extra and optional groups supported
    - 👎
        - forces the contributors to use uv
        - Also, `uv` can optionnaly generate `pylock.toml` files as lock files, so might as well commit the standard `pylock.toml` if the team and/or contributors use uv

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Current (in progress)
 
 - fix(harvest): refactor dates handling [#3352](https://github.com/opendatateam/udata/pull/3352)
-
+- migrate to pyproject.toml [#3413](https://github.com/opendatateam/udata/pull/3413)
 
 ## 11.0.1 (2025-09-15)
 


.circleci/config.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - checkout
       - run:
           name: Compute Python dependencies key
-          command: cat requirements/*.pip > python.deps
+          command: cat pyproject.toml > python.deps
```
ThibaudDauce: We could use pyproject.toml since it's a single file now?

.circleci/config.yml
```diff
@@ -62,7 +61,7 @@ jobs:
           # ruff format --check . : check code formatting without fixing (to fix, remove --check)
           command: |
             source venv/bin/activate
-            ruff check .
+            ruff check --select I .
```
ThibaudDauce: Why this change? As I understand it it check only import orders? Is there an `extend-select` like in pyproject.toml?

.circleci/config.yml
```diff
@@ -62,7 +61,7 @@ jobs:
           # ruff format --check . : check code formatting without fixing (to fix, remove --check)
```
ThibaudDauce, line 58: Ce commentaire semble outdated?

docker-compose.test.yml
```diff
@@ -1,4 +1,3 @@
-version: "3"
```
ThibaudDauce, line 1: Why remove this?

docs/getting-started.md
```diff
@@ -242,7 +241,7 @@ Then, you can install pip requirements.
 
 ```bash
 cd udata-front
-pip install -e . -r requirements/test.pip -r requirements/develop.pip
+pip install -e ".[test,dev]"
```
ThibaudDauce, line 244: Is it possible to automaticly install test when installing dev? I don't see why we need to have the dev environnement without test packages?

ThibaudDauce: And here we're talking about `udata-front` which doesn't use pyproject.toml, no? Not sure its important because udtaa-front will be remove soon.

docs/testing-code.md
```diff
@@ -41,7 +41,7 @@ $ pytest udata/tests/api/test_me_api.py -k test_get_profile
 To launch tests with coverage report:
 
 ```shell
-$ pip install requirements/report.pip
+$ pip install -e ".[report]"
```
ThibaudDauce, line 44: I'm not sure it's important to have so many groups. Is it really costly to always have report dependencies in dev? Maybe a little bit in CI?

pyproject.toml
```diff
@@ -0,0 +1,260 @@
+[build-system]
+requires = ["setuptools>=45", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "udata"
+version = "10.9.1.dev"
```
ThibaudDauce, line 7: Does it work with bumpr to create new release?

pyproject.toml
streino, line 1: You might want to integrate https://github.com/opendatateam/udata/pull/3411 at some point, instead of keeping a separate file.

.circleci/config.yml
bolinocroustibat: `select` and `extend-select`are not very clear indeed, it's inherited from `black` if I'm correct.
- select: Replaces the rules or sets the rules explicitly
- extend-select: Adds new rules to what's configured by select

This is a bit more clear than the official doc: https://github.com/astral-sh/ruff/discussions/10753

Indeed it should be `ruff check --extend-select I .`, I'll change that

docker-compose.test.yml
bolinocroustibat: It's obsolete and raises warning messages.

https://docs.docker.com/reference/compose-file/version-and-name/

docs/testing-code.md
bolinocroustibat: I agree with this, however this PR aims to reproduce how it was before. As I'm not very familiar with this repo, I suggest we reduce the number of groups in another PR.

.circleci/config.yml
bolinocroustibat: Why is that? Since there can be confusion between linting and formatting, and between `check --fix` and `format --check`, it seems that it might be useful to keep this

.circleci/config.yml
bolinocroustibat: The idea is to allow for future flexibility like combining multiple dependency files if needed, or lock files for later.
But that might decrease the readability and maintainability for now, so let's remove it

docs/getting-started.md
bolinocroustibat: Same as next comment, I agree with this, however this PR aims to reproduce how it was before. As I'm not very familiar with this repo, I'm not confortable changing this in the same PR, as it might have consequences I don't know about, so I suggest we do that in a separate PR!

pyproject.toml
bolinocroustibat: Yes bumpr/bumpx work with pyproject.toml (it is used in other datagouv projects like Hydra)

pyproject.toml
bolinocroustibat: That would be great indeed! Since you know pyright nd the pyright PR, would you mind doing it in this same PR? I'm not familiar with pyright

streino: Done here: https://github.com/opendatateam/udata/pull/3426

.circleci/config.yml
ThibaudDauce: It's weird that these comments are not just above the bash line :-) but ok

docs/getting-started.md
maudetes: udata-front hasn't been migrated to pyproject though (and shouldn't be)

PR #3414, opened 2025-09-04
Frequencies as StrEnum

streino:
Alternative to https://github.com/opendatateam/udata/pull/3343 (shared code except the last few commits on this branch). Set to target `master` so the full diff is clearer, but would merge into the initial PR if we decide to go with the the Enum version here.

### DIFF
--- a/udata/core/dataset/models.py
+++ b/udata/core/dataset/models.py
@@ -845,7 +845,7 @@ class Dataset(Auditable, WithMetrics, DatasetBadgeMixin, Owned, Linkable, db.Doc
             # We may have up to one day delay due to harvesting for example
             quality["update_fulfilled_in_time"] = (next_update - datetime.utcnow()).days >= -1
         elif self.has_frequency and self.frequency.delta is None:
-            # Unbonded frequencies can't be estimated, so we always consider the
+            # Unbounded frequencies can't be estimated, so we always consider the
             # update_fulfilled_in_time quality criterion to be true.
             quality["update_fulfilled_in_time"] = True
 


requirements/install.pip
```diff
@@ -20,6 +20,8 @@ babel==2.17.0
     # via
     #   -r requirements/install.in
     #   flask-babel
+backports.strenum==1.3.1 ; python_version < '3.11'
```
streino, line 23: This was manually added because I'm on a 3.11 venv, so pip compile drops backports.strenum.

BTW, setup.py has `python_requires=">=3.8"` but pip compile fails on 3.8 (works from 3.9).

udata/core/dataset/constants.py
```diff
@@ -1,48 +1,159 @@
...
-        ("irregular", _("Irregular")),  # freq:irregular
-    ]
-)
 
-#: Map legacy frequencies to currents
-LEGACY_FREQUENCIES = {
-    "fortnighly": "biweekly",
-    "biannual": "semiannual",
-    "realtime": "continuous",
+class UpdateFrequency(StrEnum):
```
streino, line 15: If we go with this proposal we might want to extract the class to a separate file.

udata/core/dataset/models.py
```diff
@@ -568,7 +567,10 @@ class Dataset(Auditable, WithMetrics, DatasetBadgeMixin, Owned, Linkable, db.Doc
     resources = field(db.ListField(db.EmbeddedDocumentField(Resource)), auditable=False)
 
     private = field(db.BooleanField(default=False))
-    frequency = field(db.StringField(choices=list(UPDATE_FREQUENCIES.keys())))
+
+    # FIXME: Do we want required=True to avoid None?
```
streino, line 571: @maudetes WDYT? I don't see a use for None in addition to UNKNOWN, since most places we use frequency we already default to UNKNOWN when missing : 
- API v1: https://github.com/opendatateam/udata/pull/3414/files#diff-70339dabd7e6a7c7ca04a6632989bc7a55aa1e35c09b44e08b40142e239b5cd9L358
- API v2: https://github.com/opendatateam/udata/pull/3414/files#diff-a2ac95ffb00785fc6cd9cd022bb32a7f2cddae25fcbfd636bef4d0c650771a19L158
- DatasetFactory: https://github.com/opendatateam/udata/pull/3414/files#diff-4018c6e109955b41743a2d9fca9f9a562eba16c50d0494a685336f9e7c973bb9L20
- Dataset.frequency_label: https://github.com/opendatateam/udata/pull/3414/files#diff-154a812d394e4af5de85aabf23de04611e6dbe7854b188b48f374a536edf294aL773

Only (important?) exception is DCAT: If frequency is None we don't output `dcat:accrualPeriodicity`. But this will only work for harvested datasets, because datasets created via API should always have a frequency set. Not sure it's worth the trouble?

Note to self: I have to double-check but I think in this PR None is already being converted to UNKNOWN via a roundtrip to mongo: StrEnumField.to_python(None) [will return UNKNOWN](https://github.com/opendatateam/udata/pull/3414/files#diff-2d26a6c3d477470033d7130169ec25f9d0590b523d2b664ca73a49cfc6cbaa61R106), so the next to_mongo will end up writing that value.

Simple solution IMO is required=True + default=UNKNOWN. If we want to leave that field as optional I'll have to fix the current behavior.

udata/core/dataset/models.py
```diff
@@ -870,9 +844,9 @@ def quality(self):
             # Allow for being one day late on update.
             # We may have up to one day delay due to harvesting for example
             quality["update_fulfilled_in_time"] = (next_update - datetime.utcnow()).days >= -1
-        elif self.frequency in ["continuous", "irregular", "punctual"]:
-            # For these frequencies, we don't expect regular updates or can't quantify them.
-            # Thus we consider the update_fulfilled_in_time quality criterion to be true.
+        elif self.has_frequency and self.frequency.delta is None:
+            # Unbonded frequencies can't be estimated, so we always consider the
```
streino:
```suggestion
            # Unbounded frequencies can't be estimated, so we always consider the
```

udata/migrations/2025-09-04-update-legacy-frequencies.py
streino, line 1: Not sure this is needed since we're keeping the existing vocabulary and only adding terms. Might still be a good idea in case some old legacy terms weren't previous converted?

udata/migrations/2025-09-04-update-legacy-frequencies.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def migrate(db):
+    log.info("Processing datasets.")
+
+    datasets = Dataset.objects().no_cache().timeout(False)
+    for dataset in datasets:
+        # Mapping of legacy values occurs implicitely when dataset.frequency
+        # is deserialized from mongo, so we just have to save the new value.
+        # FIXME: No need to save if it didn't change.
```
streino, line 19: 😬 -- haven't tested, will do

abulte (commented):
Although a bit convoluted, due to python version mostly, I think it's pretty neat. I'd go with this version for the sake of maintainability of the enums vs "random" strings in the future.

udata/core/dataset/constants.py
streino: I'll leave it here for now

PR #3415, opened 2025-09-05
Add new badges

nicolaskempf57:
Part of https://github.com/datagouv/data.gouv.fr/issues/1828


udata/commands/fixtures.py
```diff
@@ -51,7 +51,6 @@
         "last_update",
         "last_modified",
         "license",
-        "badges",
```
ThibaudDauce, line 54: Is there a reason to add back this? (or was there a reason for not having badges in fixtures?)

nicolaskempf57: I think it was removed because we didn't use dataset badges before. I'm adding it otherwise the e2e won't have badges filled.

udata/commands/fixtures.py
maudetes, line 1: Don't forget the changelog ;)

udata/tests/apiv2/test_datasets.py
```diff
@@ -119,6 +119,21 @@ def test_search_dataset_tags(self):
         assert len(data) == 1
         assert data[0]["id"] == str(tag_dataset_1.id)
 
+    def test_search_datasetbadges(self):
```
maudetes:
```suggestion
    def test_search_dataset_badges(self):
```

PR #3416, opened 2025-09-08
fix: PUT Topic.elements should replace instead of append

abulte:
Follows https://github.com/opendatateam/udata/pull/3318


maudetes (approved):
Thank you for the fix! I think you can mention this PR in the existing changelog entry?

PR #3417, opened 2025-09-09
Improve topic api perfs

maudetes:
Follows https://github.com/opendatateam/udata/pull/3318

Use `count()` instead of `len()` on ElementTopic queryset total.

On a huge topic (ex http://dev.local:7000/api/2/topics/65e9aa6cb5c809c30c70ee02/), we go from ~4sec to ~0.25sec


abulte (approved):
🙏 💪

PR #3418, opened 2025-09-09
fix: topic element form validation and data flow

abulte:
Might fix this error when using topic API v2 PUT on Topic.

```
'csrf_token' = ['The CSRF token is missing.']
```

Also fix the data flow in Element form, not inheriting from parent form.

### DIFF
--- a/udata/core/topic/forms.py
+++ b/udata/core/topic/forms.py
@@ -68,7 +68,9 @@ class TopicForm(ModelForm):
 
         # Create elements and associate them with the topic
         for element_data in elements_data or []:
-            element_form = TopicElementForm(data=element_data)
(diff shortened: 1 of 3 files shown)


udata/core/topic/forms.py
```diff
@@ -68,7 +68,9 @@ def save(self, commit=True, **kwargs):
 
         # Create elements and associate them with the topic
         for element_data in elements_data or []:
-            element_form = TopicElementForm(data=element_data)
+            # Create element form with only its own data, not inheriting from parent
+            element_form = TopicElementForm()
```
maudetes:
```suggestion
            element_form = TopicElementForm(meta={"csrf": False})
```

PR #3348, opened 2025-06-25
feat(security)!: enable SPA type security views (login, register) for cdata 

ThibaudDauce:
- Add `SECURITY_SPA_ON_SAME_DOMAIN` to `True` if your frontend is on the same domain as the backend.
- Change the redirects URLs in ProConnect:
  - From https://www.data.gouv.fr/proconnect/auth to https://www.data.gouv.fr/api/1/proconnect/auth
  - From https://www.data.gouv.fr/proconnect/logout to https://www.data.gouv.fr/api/1/proconnect/logout
- Part of https://github.com/datagouv/data.gouv.fr/issues/1717

PR #3422, opened 2025-09-11
refactor!: remove VueJS old admin and all assets

ThibaudDauce:
- [x] Dossiers frontend : js/, less/, specs/, node_modules/
- [x] Configuration NPM/Webpack : package.json, webpack.config.*, karma.conf.js, .eslintrc.yml, .nvmrc, babel.cfg
- [x] Assets statiques compilés : udata/static/* (remplacé par dossier vide)
- [x] Tâches Python frontend : assets_build, widgets_build, oembed_build, *_watch, jstest
- [x] CI/CD : Job assets et dépendances dans .circleci/config.yml

### DIFF
--- a/udata/templates/macros/forms.html
+++ b/udata/templates/macros/forms.html
@@ -1,0 +1,87 @@
+{% macro render(form, exclude=None) %}
+{% set excluded = exclude or [] %}
+{% for f in form %}
+    {% if f in excluded %}
+    {% elif f.type in ('CSRFTokenField', 'HiddenField') %}
+        {{ f() }}
+    {% else %}
+        {{ field(f) }}
+    {% endif %}
+{% endfor %}
+{% endmacro %}
+
+{% macro field(f, sizes='sm-2,md-3') %}
+<div class="form-group{% if f.errors %} has-error{% endif %}">
+    <label for="{{ f.id }}"
+        class="control-label{% if f.flags.required %} required{% endif %}">
+        {{ f.label.text }}
+    </label>
+    <div class="field-wrapper">
+        {% if f.type == 'RadioField' %}
+        {{ radio(f) }}
+        {% elif f.type == 'BooleanField' %}
+        {{ checkbox(f) }}
+        {% else %}
+        {{ f(class_='form-control') }}
+        {% endif %}
+        {% for error in f.errors %}
+        <label for="{{ f.id }}" class="help-block">{{ error }}</label>
+        {% endfor %}
+    </div>
+</div>
+{% endmacro %}
+
+{% macro radio(field) %}
+{% if field.stacked %}
+{{ radio_stacked(field) }}
+{% else %}
+{{ radio_inline(field) }}
+{% endif %}
+{% endmacro %}
+
+{% macro radio_inline(field) %}
+{% for value, label, selected, _ in field.iter_choices() %}
+<label class="radio-inline">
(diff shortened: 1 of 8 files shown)


udata/static/.gitkeep
maudetes, line 1: Why do we need to keep this directory?

udata/assets.py
maudetes, line 1: Seems like you did not keep this file actually? :stuck_out_tongue_closed_eyes:

udata/templates/raw.html
maudetes, line 1: I think we can't merge this before https://github.com/opendatateam/udata/pull/3348/, as it breaks the currently used security pages.

.circleci/config.yml
```diff
@@ -265,14 +217,9 @@ workflows:
           filters:
             tags:
               only: /v[0-9]+(\.[0-9]+)*/
-      - assets:
-          filters:
-            tags:
-              only: /v[0-9]+(\.[0-9]+)*/
       - dist:
```
maudetes: We should make sure the dist produced is working properly when installed as a standalone.

.eslintignore
maudetes, line 1: I've taken the liberty to remove some more stuff in https://github.com/opendatateam/udata/pull/3422/commits/c897bb24f25b2d5d928aaf639adf4c8d411db5f5 as I was testing what seemed required or not.
Let me know if you think I should have kept some things

udata/commands/static.py
maudetes, line 1: It isn't used anymore, right? Tried to remove it in https://github.com/opendatateam/udata/pull/3422/commits/6b433a7939c6fb5bfc686c76651cc34f143bc1ba

maudetes (commented):
Such a nice cleanup! Thank you for taking a go at it! I have minor questions/contributions :)
image

udata/static/.gitkeep
ThibaudDauce: Is it used for storing files in local dev?

udata/templates/raw.html
ThibaudDauce: Good catch!

udata/static/.gitkeep
maudetes: Hmm, I think it was used for `udata/commands/static.py`, but we don't require it anymore? I'm unsure I'm missing something though

udata/templates/admin.html
maudetes, line 1: We additionally should remove the `udata/admin/views.py` (the entire folder actually?)

ThibaudDauce: Done in 399c226cd

udata/templates/raw.html
maudetes: The security views return a 500: http://localhost:7000/login/.
Should we keep the templates around for a while, or should we remove the views entirely?

.circleci/config.yml
maudetes: Seems to be the case :) Played around with a clean env and the latest udata dist built in CI without issues.

babel.cfg
maudetes, line 1: I think `babel.cfg` is used in `i18n` tasks for `inv i18n`

udata/templates/raw.html
ThibaudDauce: Added back in 1001d9019, will think about it and see what we do…

babel.cfg
ThibaudDauce: Restore in 90837771454bbe6d26a878148ba664894bfee4d9 Thanks!

PR #3424, opened 2025-09-15
fix: PUT topic without elements

abulte:


udata/tests/apiv2/test_topics.py
```diff
@@ -295,6 +295,33 @@ def test_topic_api_update_with_elements(self):
         topic.reload()
         self.assertEqual(len(topic.elements), initial_length + 1)
 
+    def test_topic_api_update_without_elements(self):
+        """It should update a topic without affecting existing elements when elements field is not provided"""
+        user = self.login()
+        topic = TopicWithElementsFactory(owner=user)
+        initial_elements_count = len(topic.elements)
```
ThibaudDauce: You could use `len(initial_element_ids)` instead of having an extra variable?

ThibaudDauce (approved):
Do not forget changelog :-) Thanks for the quick fix

PR #3425, opened 2025-09-15
Datasets access types

ThibaudDauce:


udata/core/dataset/models.py
```diff
@@ -713,6 +717,10 @@ def clean(self):
 
         self.quality_cached = self.compute_quality()
 
+        check_only_one_condition_per_role(self.access_audiences)
+        if self.access_type and self.access_type != ACCESS_TYPE_OPEN:
+            self.license = None
```
maudetes, line 691: I think restricted access can still have licences though? I wouldn't enforce this logic in the model

udata/migrations/2025-01-05-dataservices-fields-changes.py
maudetes, line 1: This migration shouldn't re-run, right? We're just making sure it would still work?

udata/utils.py
```diff
@@ -65,7 +65,11 @@ def filter_changed_fields(document, previous, changed_fields: list[str]):
     document.reload()
     filtered_changed_fields = []
     for field in changed_fields:
-        previous_value = previous[field]
+        # Sometimes, we nullify a field in the clean method (for exemple the `license` when we change
+        # the `access_type` of a dataset). This field is then not present in `previous`. Returning `None`
```
maudetes, line 70: Shouldn't it be present in `previous` in this case? We do have previous value for it in the case of nullify.

udata/core/access_type/constants.py
```diff
@@ -0,0 +1,48 @@
...
+ACCESS_AUDIENCE_NO = "no"
+ACCESS_AUDIENCE_UNDER_CONDITIONS = "under_condition"
+ACCESS_AUDIENCE_CONDITIONS = [
+    ACCESS_AUDIENCE_YES,
+    ACCESS_AUDIENCE_NO,
+    ACCESS_AUDIENCE_UNDER_CONDITIONS,
+]
+
+# INSPIRE Directive Article 13(1) limitation reason categories
+INSPIRE_13_A_PUBLIC_AUTHORITIES = "confidentialityOfProceedingsOfPublicAuthorities"
```
maudetes: I think we're using snake case mostly all round for constants, I think it would make sense to keep using it

udata/core/access_type/models.py
```diff
@@ -0,0 +1,42 @@
...
+    roles = set(e["role"] for e in access_audiences)
+    if len(roles) != len(access_audiences):
+        raise FieldValidationError(
+            _("You can only set one condition for a given access audience role"),
+            field="access_audiences",
+        )
+
+
+class WithAccessType:
+    access_type = field(
```
maudetes, line 31: I was wondering if we should make access_type a nested document with many properties or use a mixin to add all properties to the main document as you've done here. Do you think it's better like this?

udata/core/access_type/utils.py
```diff
@@ -0,0 +1,61 @@
...
+    INSPIRE_13_D_COMMERCIAL_CONFIDENTIALITY,
+    INSPIRE_13_E_INTELLECTUAL_PROPERTY,
+    INSPIRE_13_F_PERSONAL_DATA,
+    INSPIRE_13_G_VOLUNTARY_SUPPLIER,
+    INSPIRE_13_H_ENVIRONMENTAL_PROTECTION,
+)
+from udata.i18n import lazy_gettext as _
+
+
+def get_inspire_limitation_labels():
```
maudetes: We could use the StrEnum approach as done in https://github.com/opendatateam/udata/pull/3343/

udata/core/access_type/utils.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    Returns a dictionary of INSPIRE limitation category labels.
```
maudetes: Do you think we need some juridical review on these labels?

udata/core/dataset/api_fields.py
```diff
@@ -395,6 +401,11 @@
         "license": fields.String(
             attribute="license.id", default=DEFAULT_LICENSE["id"], description="The dataset license"
         ),
+        "access_type": fields.String(allow_null=True),
+        "access_audiences": fields.Nested(AccessAudience.__read_fields__),
```
maudetes: I think it is a list of `AccessAudience`? And somehow, it shows an `id: null` in api/1/datasets/ output

udata/core/access_type/api.py
```diff
@@ -0,0 +1,18 @@
...
+
+blueprint = Blueprint("access_type", __name__, url_prefix="/api/1/access_type")
+ns = api.namespace("access_type", "Access type related operations")
+
+
+@ns.route("/reason_categories/", endpoint="reason_categories")
+class ReasonCategoriesAPI(API):
+    @api.doc("reason_categories")
+    def get(self):
+        """List all INSPIRE limitation reason categories"""
```
maudetes: I wouldn't mention INSPIRE here, seems it isn't known by all and may be extended with other categories
```suggestion
        """List all limitation reason categories"""
```

maudetes (approved):
:rocket: nice!

udata/migrations/2025-01-05-dataservices-fields-changes.py
ThibaudDauce: Yes but we never know with migrations if someone migrate from a very old version

udata/core/access_type/models.py
ThibaudDauce: Since for `Dataservice`  it's already inside the main document I don't want to change this and do a migration… And having embed document is not really SQL compliant so not a big fan to link more stuff to MongoDB (even if it's not planned to switch off Mongo in the near future)

udata/core/access_type/utils.py
ThibaudDauce: I don't think `StrEnum` works well with multiple language… The `_()` function is only called once with the default language at the server startup, no?

udata/core/access_type/utils.py
ThibaudDauce: Right now there is no juridical implication of our system. When we start mapping to/from INSPIRE, maybe?

udata/utils.py
ThibaudDauce: The problem here (as I understood it) is that we modify the document in another `clean` (when we remove the license) so the `field` is not in `previous`. Even if we remove the `license  = None`,  I think we can keep it to avoid crashing in some cases… `previous` is a random `dict`, we don't have strong guarantee that all `field` will be present inside, no?

udata/core/dataset/models.py
ThibaudDauce: Ping @agarrone ?

udata/core/access_type/constants.py
ThibaudDauce: Done in 47dd85fad

udata/core/dataset/api_fields.py
ThibaudDauce: Where do you see `id:null`, I don't see it… 
image

udata/core/access_type/utils.py
streino: Good point... I have no idea. If this turns out to be an issue, you could wrap `_()` in a callable and resolve it when the corresponding `StrEnum` property is accessed.

ThibaudDauce: Done in f7af0ba3840280e203a2a626841cacf2ebfee5d2 I think we shouldn't put the translated labels inside the enum and just add properties for it…

udata/core/access_type/constants.py
```diff
@@ -0,0 +1,85 @@
+from enum import StrEnum
+
+from udata.i18n import lazy_gettext as _
+
+
+class AccessType(StrEnum):
+    OPEN = "open"
+    OPEN_WITH_ACCOUNT = "open_with_account"
+    RESTRICTED = "restricted"
```
streino: When value is just the lowercased ID, you can use auto() instead of the string value:

```suggestion
class AccessType(StrEnum):
    OPEN = auto()
    OPEN_WITH_ACCOUNT = auto()
    RESTRICTED = auto()
```

udata/core/access_type/utils.py
streino: This means having two different locations to keep in sync. I think having it in the actual enum definition makes it less error-prone.

udata/core/access_type/utils.py
streino: As long as "INSPIRE" doesn't show anywhere but in the code that might be OK, but once you start calling something "INSPIRE" IMO you'll need to stick to the official labels.

udata/core/access_type/utils.py
ThibaudDauce: Not really because it's inside the enum class. And I'm not sure about having a tuple as a value, it seems you've need really weird Python stuff to make it work with StrEnum, no?

```
    def __new__(cls, id: str, label: LazyString, delta: timedelta | None):
        # Set _value_ so the enum value-based lookup depends only on the id field.
        # See https://docs.python.org/3/howto/enum.html#when-to-use-new-vs-init
        obj = str.__new__(cls, id)
        obj._value_ = id
        obj._label = label  # type: ignore[misc]
        obj._delta = delta  # type: ignore[misc]
        return obj
```

ThibaudDauce: Maybe using https://typing.python.org/en/latest/guides/unreachable.html#assert-never-and-exhaustiveness-checking it could be the best of both worlds?

udata/core/access_type/models.py
```diff
@@ -0,0 +1,44 @@
...
+        db.StringField(choices=[e.value for e in AccessType], default=AccessType.OPEN.value),
```
streino:
```suggestion
        db.EnumField(AccessType, default=AccessType.OPEN),
```

udata/core/access_type/utils.py
ThibaudDauce: I've add assert_never() in 465e6f507

udata/core/dataset/forms.py
```diff
@@ -158,6 +172,19 @@ class DatasetForm(ModelForm):
         description=_("A short description of the dataset."),
     )
     license = fields.ModelSelectField(_("License"), model=License, allow_blank=True)
+    access_type = fields.SelectField(
+        choices=[(e.value, e.value) for e in AccessType],
+        default=AccessType.OPEN.value,
```
streino: `StrEnum` behaves as `str` (inherits from it) so you shouldn't need the value in places taking a `str`

```suggestion
        default=AccessType.OPEN,
```

udata/core/dataset/forms.py
```diff
@@ -158,6 +172,19 @@ class DatasetForm(ModelForm):
...
+        validators=[validators.optional()],
+    )
+    access_audiences = fields.NestedModelList(AccessAudienceForm)
+    authorization_request_url = fields.StringField(_("Authorization request URL"))
+    access_type_reason_category = fields.SelectField(
+        _("Access type reason category"),
+        choices=[(e.value, e.value) for e in InspireLimitationCategory],
```
streino: ?

```suggestion
        choices=[(e.value, e.label) for e in InspireLimitationCategory],
```

udata/core/access_type/utils.py
streino: > Not really because it's inside the enum class. And I'm not sure about having a tuple as a value, it seems you've need really weird Python stuff to make it work with StrEnum, no?
> 
> ```
>     def __new__(cls, id: str, label: LazyString, delta: timedelta | None):
>         # Set _value_ so the enum value-based lookup depends only on the id field.
>         # See https://docs.python.org/3/howto/enum.html#when-to-use-new-vs-init
>         obj = str.__new__(cls, id)
>         obj._value_ = id
>         obj._label = label  # type: ignore[misc]
>         obj._delta = delta  # type: ignore[misc]
>         return obj
> ```

I agree it's pretty ugly, but it's not weird in the sense that : 
- It's The Way :tm: suggested in the python doc.
- It's not so different from the usual `__init__` surcharge, except it's on `__new__` so you actually use `obj` and not `self`.
- Accessing parent's "private" `_value_` member as usual.
- Everything that's not `_value_` is just extra members. You can actually directly use "public" members if you don't need to wrap them with `@property` accessors like I needed.

So you would have:
```
    def __new__(cls, id: str, label: LazyString):
        obj = str.__new__(cls, id)
        obj._value_ = id
        obj.label = label
        return obj
```

Not so different from:
```
    def __init__(self, id: str, label: LazyString):
        super(str, self).__init__(id)
        self._value_ = id  # for init it'd usually be `super(Enum, self).__init__(id)`
        self.label = label

Comment (streino) [reply udata/core/access_type/utils.py]: `assert_never` works too though!
I wonder which way is (or will be) better supported by tooling, so you can get early feedback on errors.

Comment (ThibaudDauce) [reply udata/core/access_type/utils.py]: I'm looking into translation right now and I see that `_()` is using `LazyString` / `lazy_gettext`. "Lazy strings will not be evaluated until they are actually used.", so I think it's ok to use it at startup

File: udata/api_fields.py [hunk 13]
```diff
@@ -93,12 +93,14 @@ def convert_db_to_field(key, field, info) -> tuple[Callable | None, Callable | N
         # TODO: this is currently never used. We may remove it if the auto-conversion
         # is always good enough.
         return info.get("convert_to"), info.get("convert_to")
-    elif isinstance(field, mongo_fields.StringField):
+    elif isinstance(field, mongo_fields.StringField) or isinstance(field, mongo_fields.EnumField):
```
streino, line 96: Could it become problematic if we ever use an `EnumField` for something other than `str`, like `int`? If yes, I'd at least put an assert or similar to be reminded we need to implement support here.

udata/core/dataset/api_fields.py
maudetes: On a dataset payload
image
By the way, `access_audiences` shows as a dict instead of a list of dict in the swagger
```
  "access_audiences": {
    "condition": "yes",
    "role": "local_authority_and_administration"
  },
```

udata/utils.py
maudetes: Mmm ok I see! Could we do
```
if field not in previous:
    continue
```
instead of comparing with None then?

udata/core/dataset/api_fields.py
ThibaudDauce: Fixed id: null in 79f676c27

udata/utils.py
ThibaudDauce: Done in 5dce8d104

udata/core/dataset/api_fields.py
ThibaudDauce: For the list, don't understand how it was working… Fixed in 9a4c6afcb

PR #3433, opened 2025-09-15
Harvest model records in csw-dcat

maudetes:
Following https://github.com/opendatateam/udata/pull/3378, they were ignored from the CSW request when we want these records to be harvested as datasets (in particular these [two](https://www.data.gouv.fr/datasets/donnees-aux-sites-hydrometriques/) [hvd](https://www.data.gouv.fr/datasets/donnees-aux-stations-hydrometriques/) datasets).

It was identified by @streino at [that time](https://github.com/opendatateam/udata/pull/3378#issuecomment-3132797294), but seemed like I didn't reply to it
> remote_id found in REF, not found in PR
Two such cases in "Moisssonneur SI Eau" (csw-dcat): [sieau:2b5ab7c3-f0d3-4653-9fd5-cfb133a914fb](https://data.ofb.fr/catalogue/srv/eng/catalog.search#/metadata/2b5ab7c3-f0d3-4653-9fd5-cfb133a914fb), [sieau:7db4d17b-0f5f-4312-b1a5-8ec93cc9ab72](https://data.ofb.fr/catalogue/srv/eng/catalog.search#/metadata/7db4d17b-0f5f-4312-b1a5-8ec93cc9ab72).
Both have type = "model". REF didn't filter on record type, and PR does, but doesn't include "model", so they were ignored.
@maudetes WDYT? I don't know that type but IMO it's close-enough to a dataset that it makes sense to include it.
We had the same issue with type = "nonGeographicDataset", which I consequently added to the filter.


udata/harvest/backends/dcat.py
```diff
@@ -309,6 +309,10 @@ class CswDcatBackend(DcatBackend):
                 <ogc:PropertyName>apiso:type</ogc:PropertyName>
                 <ogc:Literal>service</ogc:Literal>
               </ogc:PropertyIsEqualTo>
+              <ogc:PropertyIsEqualTo>
```
streino, line 312: Can you move it up to keep the list sorted? 🙏

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Current (in progress)
 
+- Harvest model records in csw-dcat [#3433](https://github.com/opendatateam/udata/pull/3433)
```
streino, line 5: Might want to add "and csw-iso-19139"?

PR #3434, opened 2025-09-16
Fix dist version

ThibaudDauce:
I replaced the manual versioning logic (setup.py + egg\_info -b) with setuptools\_scm.
The issue was that after moving to pyproject.toml we had a hardcoded `version = "10.9.1.dev"`, so every build generated the same dist file and uploads to PyPI failed.
With setuptools\_scm the version is no longer stored in `__init__.py` or pyproject.toml, but automatically derived from Git tags.
setuptools\_scm detects if the commit is exactly on a tag (`X.Y.Z`) or after it (e.g. `X.Y.Z.devN+g<hash>`).
The CI no longer needs to calculate or inject BUILDNO, builds are done with `python -m build`.
Tags created by bumpr keep the `vX.Y.Z` format, setuptools\_scm strips the `v` for the PyPI version.
The bumpr config no longer writes to `__init__.py`, it only handles commits, tags, changelog, and files like README.

`__version__` was removed from the code; if we ever need it at runtime, it can be dynamically retrieved via `from importlib.metadata import version; version("udata")`.

The `.devN` suffix is now based on the number of commits since the last tag, rather than the global build number, which may produce smaller dev numbers than before. In this PR we have the `+g<hash>` but it was removed in https://github.com/opendatateam/udata/pull/3435 because Pypi refuse the hash version when sending build from master. Now branches can have the same .dev42 but they are only distributed by a CircleCI artifact URL so it's not a problem.

The computed `RELEASE_VERSION` is written to `$BASH_ENV`, so it’s automatically available in all subsequent steps without needing a separate `version.sh` or manual sourcing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,6 +131,8 @@ jobs:
       - image: udata/circleci:py<< pipeline.parameters.python-version >>
     steps:
       - checkout
+      - attach_workspace:
+          at: .
       - run:
           name: Install build tools
           command: |


bolinocroustibat (approved):
Well done, thanks for the quick fix.

To simplify in the future we could also use [uv](https://github.com/astral-sh/uv) as the package manager and to [bump version and publish on PyPi](https://docs.astral.sh/uv/guides/package/#updating-your-version), instead of bumpr/bumpx.

- To bump the version (will also update it in pyproject.toml): `uv version $VERSION` and/or `uv version --bump`

- To publish on PyPi: `uv publish --username $PYPI_USER --password $PYPI_PASSWORD` or `uv publish --token $PYPI_TOKEN`

Should we plan a PR later to do that?

PR #3435, opened 2025-09-16
Do not put commit hash in Pypi version

ThibaudDauce:
Following https://github.com/opendatateam/udata/pull/3434

The .devN suffix no longer includes the commit hash (+g<hash>) to be compatible with PyPI.
It now only counts the number of commits since the last tag. Branches may produce the same .devN, but since only master releases are published to PyPI, there’s no conflict.
Branch builds are only distributed via CircleCI artifact URLs, so uniqueness isn’t required.


maudetes (approved):
:pray:

PR #3423, opened 2025-09-11
feat(topics): batch reindex on topic_elements_delete

abulte:
This aims at preventing the error below when deleting all elements from a huge Topic (10k+). It reindexes elements with 500 items batches instead of launching one task per element.

```
Traceback (most recent call last):
  File "/srv/demo/lib/python3.11/site-packages/flask/app.py", line 1523, in full_dispatch_request
  File "/srv/demo/lib/python3.11/site-packages/flask/app.py", line 1509, in dispatch_request
  File "/srv/demo/lib/python3.11/site-packages/udata/api/__init__.py", line 117, in wrapper
  File "/srv/demo/lib/python3.11/site-packages/flask_restx/api.py", line 402, in wrapper
  File "/srv/demo/lib/python3.11/site-packages/flask/views.py", line 84, in view
    return current_app.ensure_sync(self.dispatch_request)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/srv/demo/lib/python3.11/site-packages/flask_restx/resource.py", line 41, in dispatch_request
    resp = meth(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/srv/demo/lib/python3.11/site-packages/udata/api/__init__.py", line 91, in wrapper
  File "/srv/demo/lib/python3.11/site-packages/udata/core/topic/apiv2.py", line 160, in delete
    topic.elements.delete()
  File "/srv/demo/lib/python3.11/site-packages/mongoengine/queryset/base.py", line 474, in delete
  File "/srv/demo/lib/python3.11/site-packages/mongoengine/document.py", line 693, in delete
  File "/srv/demo/lib/python3.11/site-packages/blinker/base.py", line 263, in send
  File "/srv/demo/lib/python3.11/site-packages/blinker/base.py", line 263, in <listcomp>
  File "/srv/demo/lib/python3.11/site-packages/udata/core/topic/models.py", line 44, in post_delete
  File "/srv/demo/lib/python3.11/site-packages/celery/app/task.py", line 444, in delay
  File "/srv/demo/lib/python3.11/site-packages/sentry_sdk/integrations/celery/__init__.py", line 289, in apply_async
  File "/srv/demo/lib/python3.11/site-packages/celery/app/task.py", line 594, in apply_async
[...]

PR #3439, opened 2025-09-17
feat: add topic element activities

abulte:


udata/core/topic/apiv2.py
```diff
@@ -157,6 +154,7 @@ def delete(self, topic):
         if not TopicEditPermission(topic).can():
             apiv2.abort(403, "Forbidden")
 
+        # TODO: should we ignore activity creation here?
```
abulte: This will create a _lot_ of deleted activities (and probably crash since it is synchronous) if we empty a universe Topic. I'm not sure we should introduce an exception here to prevent that, WDYT?

Emptying a universe crashes anyway 😬 #3423

udata/core/topic/apiv2.py
```diff
@@ -130,16 +130,13 @@ def post(self, topic):
...
                 element.save()
                 elements.append(element)
 
         if errors:
             apiv2.abort(400, errors=errors)
 
-        for element in elements:
-            element.topic = topic
-            element.save()
-
```
maudetes, line 142: I'm confused, we used to abort if there were errors and save the `topic` field only after.
Seems like we're changing the behavior here but what was the reason before?

udata/core/topic/apiv2.py
maudetes: Hmm... I think it can be dealt at the same time as https://github.com/opendatateam/udata/pull/3423 indeed?

udata/core/topic/apiv2.py
abulte: I _think_ this is a left over from the first architecture where the TopicElements were embedded in the Topic. I also _think_ I had a test for that and none fail. I don't see any good reason to keep it if the element validation works, as the tests seem to validate.

udata/core/topic/apiv2.py
abulte: I changed the TODO wording.

udata/core/topic/apiv2.py
maudetes: :crossed_fingers:

PR #3440, opened 2025-09-22
Convert posts to new fields v2

ThibaudDauce:


udata/core/post/models.py
```diff
@@ -16,27 +18,84 @@ def published(self):
         return self(published__ne=None).order_by("-published")
 
 
+@generate_fields(
+    searchable=True,
+    additional_sorts=[
+        {"key": "modified", "value": "last_modified"},
+    ],
```
maudetes, line 26: I got a 400 with sort `created_at` when accessing [from admin](http://dev.local:3000/admin/site/posts) with the following call http://dev.local:7000/api/1/posts/?with_drafts=true&sort=-created_at&q&page_size=20&page=1&lang=fr

ThibaudDauce: Fixed in 48fce286c (Post are using `created_at` as sort key but the default is `created` in the mixin `Datetimed` :-( )

nicolaskempf57 (commented):
The cdata post publishing form isn't working anymore. To make it work, the default `credit_url` should be `null` otherwise the validation fail because an empty string isn't a valid URL.

PR #3441, opened 2025-09-22
Add type and doc to field()

ThibaudDauce:


udata/api_fields.py
```diff
@@ -1,39 +1,24 @@
...
+- `@generate_fields`: Decorator that adds API field generation to document classes
+- `field()`: Universal function to add metadata to fields and methods
 
-- default_filterable_field: which field in this document should be the default filter, eg when filtering by Badge, you're actually filtering on `Badge.kind`
-- searchable: boolean, if True, the document can be full-text searched using MongoEngine text search
-- additional_sorts: add more sorts than the already available ones based on fields (see below). Eg, sort by metrics.
-- additional_filters: filter on a field of a field (aka "join"), eg filter on `Reuse__organization__badge=PUBLIC_SERVICE`.
+The `@generate_fields` decorator parameters:
+- default_filterable_field: Default field for filtering (e.g., Badge.kind)
+- searchable: Enables full-text search via MongoEngine
```
maudetes: Maybe we could me more precise?
```suggestion
- searchable: Enables full-text search with param `q` via MongoEngine on indexed text fields
```

udata/api_fields.py
```diff
@@ -558,22 +543,92 @@ def apply_pagination(base_query) -> DBPaginator:
...
-def function_field(**info) -> Callable:
-    def inner(func):
-        func.__additional_field_info__ = info
-        return func
-
-    return inner
-
+def field(
+    inner=None,
+    *,  # Force named arguments after this
```
maudetes: I don't think we should add this? We don't want args, right?

udata/api_fields.py
```diff
@@ -558,22 +543,92 @@ def apply_pagination(base_query) -> DBPaginator:
...
+    nested_fields: dict[str, Any] | None = None,
+    inner_field_info: dict[str, Any] | None = None,
+    size: int | None = None,
+    is_thumbnail: bool | None = None,
+    href: Callable | None = None,
+    generic: bool | None = None,
+    generic_key: str | None = None,
+    convert_to: Callable | None = None,
+    allow_null: bool | None = None,
+    **kwargs: Any,  # Accept any additional parameters
```
maudetes: Do we want to support `**kwargs`? don't we want to be exhaustive, so that all params are documented here?

udata/api_fields.py
```diff
@@ -558,22 +543,92 @@ def apply_pagination(base_query) -> DBPaginator:
...
-    def inner(func):
-        func.__additional_field_info__ = info
-        return func
-
-    return inner
-
+def field(
+    inner=None,
+    *,  # Force named arguments after this
+    sortable: bool | str | None = None,
```
maudetes, line 548: Do we want to set the default at `None` everywhere instead of false for boolean?

udata/api_fields.py
ThibaudDauce: Yes, removed in 90cc9796f

udata/api_fields.py
ThibaudDauce: I want to be allow to send arguments to the Flask RestX without specifying them. I changed the comment in https://github.com/opendatateam/udata/commit/deb77a858f2f0faf6b8b3ae3b42d3235ffa5dd03

udata/api_fields.py
ThibaudDauce: Yes it's better but I started making the change but sometimes we do stuff like: `{**info, **info.get("thumbnail_info", {}), "is_thumbnail": True, "attribute": key},` where we fallback from one info to another. And in this case there is a real difference between no value and value at `False` (because value at `False` means the user specify this case and so we don't override it). Maybe I could some test in another PR to check for these cases and verify removing the `None` do not break this kind of stuff…

PR #3443, opened 2025-09-23
feat(topics): search by elements content

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/618

Extend Topic's search to TopicElements content.

Performance is acceptable locally with a 25k elements Topic.


udata/core/topic/parsers.py
```diff
@@ -71,7 +73,20 @@ def parse_filters(topics, args):
             # This allows the search_text method to tokenise with an AND
             # between tokens whereas an OR is used without it.
             phrase_query = " ".join([f'"{elem}"' for elem in args["q"].split(" ")])
-            topics = topics.search_text(phrase_query)
+
+            # Search topics by their own content
+            topic_text_filter = db.Q(__raw__={"$text": {"$search": phrase_query}})
+
+            # Find topics that have elements matching the search
+            matching_elements = TopicElement.objects.search_text(phrase_query)
```
maudetes, line 81: It doesn't search on the nested `TopicElement.element` title & description right?

abulte: Nope. Not yet? ;-) It would be nice but I'm not sure we can do that efficiently without Elastic.

PR #3444, opened 2025-09-23
feat(topic): add Dataservice to supported elements

abulte:
Related https://github.com/ecolabdata/ecospheres/issues/535 https://github.com/ecolabdata/ecospheres/issues/524

Add Dataservice capability to [`Topic.elements`](https://github.com/opendatateam/udata/pull/3318/).

This introduces `standalone_filters` to the `api_fields` logic in order to declare liberal filters (e.g on something that is not a model field). `additional_filters` has been renamed to `nested_filters` in order to better separate the fields' additional filters and the new standalone filters.

I did not add the topic filter to api v2. It's currently quite limited (no dataset filter for example), so it probably can wait.

### DIFF
--- a/udata/tests/apiv2/test_topics.py
+++ b/udata/tests/apiv2/test_topics.py
@@ -614,7 +614,7 @@ class TopicElementsAPITest(APITestCase):
         # Verify response payload contains the created elements
         response_data = response.json
         assert isinstance(response_data, list)
-        assert len(response_data) == 3
+        assert len(response_data) == 4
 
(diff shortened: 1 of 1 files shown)


ThibaudDauce (approved):
I don't know why the test was failing… Can you confirm everything is ok for you for merging?

abulte:
@ThibaudDauce makes sense that test would be failing since #3446 has been merged, we have one more element in our test case. LGTM, thanks!

PR #3446, opened 2025-09-24
feat(topics): serialize elements in topic_elements_create response

abulte:
Respond to `topic_elements_create` with created elements rather than the full Topic.


maudetes (approved):
Indeed, it's more straightforward as a user :pray:

PR #3449, opened 2025-09-26
Keep existing attached datasets to a dataservice at harvest time

maudetes:
Close https://github.com/datagouv/data.gouv.fr/issues/1865

Do not overwrite the list of datasets attached to a dataservice but only update it with new ones

### DIFF
--- a/udata/core/dataservices/rdf.py
+++ b/udata/core/dataservices/rdf.py
@@ -32,7 +32,7 @@ def dataservice_from_rdf(
     remote_url_prefix: str | None = None,
 ) -> Dataservice:
     """
-    Create or update a dataset from a RDF/DCAT graph
+    Create or update a dataservice from a RDF/DCAT graph
     """
     if node is None:  # Assume first match is the only match
         node = graph.value(predicate=RDF.type, object=DCAT.DataService)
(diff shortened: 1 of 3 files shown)


udata/core/dataservices/rdf.py
```diff
@@ -71,11 +70,9 @@ def dataservice_from_rdf(
                 None,
             )
 
-        if dataset is not None:
-            datasets.append(dataset.id)
-
-    if datasets:
-        dataservice.datasets = datasets
+        # We append the dataset to the list of the current attached ones if not already attached
+        if dataset is not None and dataset not in dataservice.datasets:
```
ThibaudDauce, line 74: `dataset not in dataservice.datasets` is it working? Could you make sur we have a test on this because I'm always septical about comparing objects like this :) (we may need to compare IDs?)

maudetes: I've added an example in https://github.com/opendatateam/udata/pull/3449/commits/8a339fa2b0a40b80541a68582757e659d7be817b? Was it what you expect?

ThibaudDauce: Yes it seems to work (but don't understand how it react to a change in the title of a dataset after the first harvest for exemple?)

ThibaudDauce (approved):

udata/core/dataservices/rdf.py
maudetes: The comparison is based on the id in [mongoengine BaseDocument](https://github.com/MongoEngine/mongoengine/blob/370460d78f9e9193d9a8883459e36738cd4697f4/mongoengine/base/document.py#L291)

PR #3454, opened 2025-10-02
Cleanup dependencies

ThibaudDauce:


bolinocroustibat:
> > Most of the dependencies are probably transitive? The required ones [were in `install.in`](https://github.com/opendatateam/udata/pull/3413/files#diff-9a7a7630ef132897dc00c238763d3b8c40ef6e459ec1aa22089bbb82fe4df2dc) while [the lock deps were in `install.pip`](https://github.com/opendatateam/udata/pull/3413/files#diff-6ec864a8a424750278a7b20e821157cd7505dc5651e2a193eaa185a0afe5228b). I guess in #3413, the install.pip was used for dependencies to stay ISO.
> > Since we don't have a lock file (yet), I'm not sure it's a good idea to remove those? It may be better to have a fixed version than no lock on the transitive deps?
> 
> I'm putting this PR in draft, waiting the addition of `uv.lock` from @bolinocroustibat

Added uv.lock and updated pyproject.toml and README in https://github.com/opendatateam/udata/pull/3454/commits/0cb71be3dcaa02d73a7aafd6bc762419059d8498

PR #3455, opened 2025-10-02
ci: do not publish on tags

ThibaudDauce:
With `setuptools_scm` (introduced in https://github.com/opendatateam/udata/pull/3434), the Pypi CI from the `master` branch publish with the tag on the current commit (before it always published with a suffix `.devx`).

Since the `11.1.1` and `11.1.1.dev0` are the same we don't need to publish from `master` and from the tag. And since publishing from `master` do the correct thing, we removed the CI on tags.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -236,6 +236,8 @@ jobs:
 workflows:
   build-publish-deploy:
     jobs:
+      - python
+      - assets
       - dist:
           requires:
             - python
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -236,31 +236,17 @@ jobs:
 workflows:
   build-publish-deploy:
     jobs:
-      - python:
```
maudetes, line 239: I think removing the job from the jobs list break the CI right?

ThibaudDauce: You're right :-) Fixed in a0b3be9c8dfcbf0f8b4e76fad850efd45a50e0a8

bolinocroustibat (approved):
Nice simplification, thanks. If we are satisfied with this workflow, we should probably use setuptools_scm in other Python repos.

PR #3457, opened 2025-10-02
feat: add standalone 404 page

ThibaudDauce:
Right row we route all URLs to udata and only some special ones to cdata. https://demo.data.gouv.fr/xxx shows us the default Flask 404.

In the near future, we want to switch the way we route the HTTP requests by defaulting to cdata and just routing /api/* to udata so a random user should never see this new 404 page (so it's not a problem to have a simpler design than the one from cdata).

- [x] Create 404 page
- [ ] Return the 404 page while `abort(404)` inside the API routes (for exemple for `/datasets/r/xxx`)
- [ ] Check if the user wants HTML or JSON to return a JSON response to the API users. (`/datasets/r/xxx` is one of the very few routes, that a user can be redirected to in a browser, maybe with the CSV downloads?)

### DIFF
--- a/udata/app.py
+++ b/udata/app.py
@@ -7,7 +7,7 @@ from os.path import abspath, dirname, exists, isfile, join
 
 import bson
 from flask import Blueprint as BaseBlueprint
-from flask import Flask, abort, g, json, make_response, send_from_directory
+from flask import Flask, abort, g, json, make_response, render_template, send_from_directory
 from flask_caching import Cache
 from flask_wtf.csrf import CSRFProtect
 from speaklater import is_lazy_string
@@ -225,7 +225,18 @@ def register_extensions(app):
     mail.init_app(app)
     search.init_app(app)
     sentry.init_app(app)
+    register_error_handlers(app)
     return app
+
+
+def register_error_handlers(app):
+    """Register error handlers for the application"""
+
+    @app.errorhandler(404)
+    def page_not_found(e):
+        from udata.uris import homepage_url
+
+        return render_template("404.html", homepage_url=homepage_url()), 404
 
 
 def register_features(app):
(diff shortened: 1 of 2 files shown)


udata/uris.py
```diff
@@ -69,7 +69,13 @@ def config_for(value, key):
 
 
 def homepage_url(**kwargs) -> str:
-    return cdata_url("/", **kwargs) or url_for("api.site", **kwargs)
+    # Some tests were crashing not finding the route for api.site
```
maudetes, line 72: We don't want to keep this part right?

udata/templates/404.html
```diff
@@ -0,0 +1,43 @@
...
+        a {
+            color: #000;
+            text-decoration: underline;
+        }
+    </style>
+</head>
+<body>
+    <div class="container">
+        <h1>404</h1>
+        <p>{{ _('The page you are looking for does not exist.') }}</p>
```
maudetes, line 39: You can extract translations

udata/app.py
```diff
@@ -225,9 +236,47 @@ def register_extensions(app):
     mail.init_app(app)
     search.init_app(app)
     sentry.init_app(app)
+    app.after_request(return_404_html_if_requested)
+    app.register_error_handler(NotFound, page_not_found)
     return app
 
 
+def return_404_html_if_requested(response):
```
maudetes, line 244: Weird that you had to go with this error handler, but seems to work as expected.
Not sure if it helps, but in the case of RDF endpoints, we also use `return make_response()` directly to bypass the JSON marshalling

ThibaudDauce: Yes but it's impossible to do return make_response because during error handling Flask RestX is reading a key from the `dict` returned :-(

udata/uris.py
ThibaudDauce: I don't see a way to prevent this from crashing without having this…

udata/templates/404.html
ThibaudDauce: Done in b91bc7205

PR #3459, opened 2025-10-02
build: use uv

bolinocroustibat:
Use [uv](https://docs.astral.sh/uv/) as package manager and lockfile, locally and in CI.
Use [uv](https://docs.astral.sh/uv/) instead of Twine in the CI to publish on PyPi.

For information:
- `uv add <package>` install package and updates pyproject.toml and uv.lock
- `uv pip install<package>` install package but DOESN'T update pyproject.toml and uv.lock (reproduces pip behaviour in the project venv), so better for temporary installs in the CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ jobs:
           name: Install Python dependencies
           command: |
             pip install uv
-            uv sync --locked --extra dev --extra test
+            uv sync --frozen --extra dev --extra test
       - save_cache:
           key: py3-cache-v13-{{ arch }}-{{ checksum "pyproject.toml" }}-{{ checksum "uv.lock" }}
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
       - restore_cache:
           keys:
             # 1. Exact match of dependencies
-            - py3-cache-v12-{{ arch }}-{{ checksum "pyproject.toml" }}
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -30,52 +30,47 @@ jobs:
...
+            - py3-cache-v13-{{ arch }}-
       - run:
           name: Install Python dependencies
           command: |
-            python -m venv venv
-            source venv/bin/activate
-            pip install --upgrade pip
-            pip install -e ".[dev]"
+            pip install uv
+            uv sync --extra dev --extra test
```
ThibaudDauce: I don't think we should do `uv sync` in the CI. It should be only on local machine to update the `uv.lock`. We want the CI to test the exact versions from `uv.lock` and not update to new available versions. But I don't see a way to download locked dependencies with `uv`. I only see `uv run --locked …` to run Pytest/publish with the locked dependencies.

bolinocroustibat: You're totally right, that's a mistake. I think we can use uv pip install instead.

ThibaudDauce: But it doesn't seems to be available right now… https://github.com/astral-sh/uv/issues/15945 Maybe skip the additional step (which just download the dependencies) and just run the tests with `--locked` (which should download the correct dependencies from the `uv.lock` before running the test)

bolinocroustibat: Even better, uv sync --locked

bolinocroustibat: After a few try/fail, I think what we want is `uv sync --frozen`.

ThibaudDauce (approved):

PR #3463, opened 2025-10-07
feat!: bump minimum Python version to 3.11

ThibaudDauce:
Python 3.9 is reaching end of life at the end of october https://devguide.python.org/versions/#versions

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -10,7 +10,7 @@ We’ll use the following repositories:
 # Check the system requirements
 
 !!! info
-    Be aware that udata now requires Python **>3.9,<=3.11** to work.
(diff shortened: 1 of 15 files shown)


ThibaudDauce:
@maudetes I think we can start to use 3.13 in tests, dev and demo too :-)

bolinocroustibat (approved):
Yay!

If we move to Python >= 3.11, I suggest we remove all the ugly typing imports i.e. `from typing import Dict, List, Optional, Union`

- `Dict` becomes native `dict`
- `List` becomes native `list`
- `Optional[sometype]` becomes `sometype | None`
- `Union[sometype, someothertype]` becomes `sometype | someothertype`

PR #3465, opened 2025-10-07
feat(topic): improve perf in dataservice topic filter

maudetes:
Increasing perfs from way too long to ~50ms locally on a request such as https://demo.data.gouv.fr/api/1/dataservices/?page_size=20&topic=65e9aa6cb5c809c30c70ee02&page=1

### DIFF
--- a/udata/core/dataservices/models.py
+++ b/udata/core/dataservices/models.py
@@ -135,7 +135,7 @@ def check_only_one_condition_per_role(access_audiences, **_kwargs):
 
 
 def filter_by_topic(base_query, filter_value):
-    from udata.core.topic.models import Topic
+    from udata.core.topic.models import Topic, TopicElement
 
     try:
         topic = Topic.objects.get(id=filter_value)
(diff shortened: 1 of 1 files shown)


udata/core/dataservices/models.py
```diff
@@ -144,9 +144,8 @@ def filter_by_topic(base_query, filter_value):
     else:
         return base_query.filter(
             id__in=[
-                elt.element.id
-                for elt in topic.elements
-                if elt.element.__class__.__name__ == "Dataservice"
+                element.element.id
+                for element in topic.elements.filter(__raw__={"element._cls": "Dataservice"})
```
ThibaudDauce, line 148: Could you try? I can't try it because I don't have any topics in my database :-)
```suggestion
                topic.elements.filter(__raw__={"element._cls": "Dataservice"}).values_list(
                    "element.id", flat=True
                )
```

abulte:
Oh I see, the filter on class is the trick 👍

udata/core/dataservices/models.py
maudetes: `flat=True` doesn't exist in `values_list`.
Trying with `values_list("element.id")` returns a `Cannot perform join in mongoDB: element__id`, same with `.pk`.

ThibaudDauce: Too bad :-) Weird that there is no way to just return a sub field only (to avoid getting all the objects)

maudetes: Yes, I think this would be the way, but does not work with a join and somehow the `id` is considered as a join?

PR #3466, opened 2025-10-07
feat: Add remote URL prefix option to csw-dcat harvester

streino:
Fix https://github.com/ecolabdata/udata/tree/feat/csw-harvester-remote-url-prefix

Draft because of https://github.com/ecolabdata/ecospheres/issues/733#issuecomment-3376565668.


maudetes (approved):
Clean, thank you!

PR #3467, opened 2025-10-07
refactor(mongo): remove flask_mongoengine

ThibaudDauce:
Since `flask-storage` still require `flask-mongoengine`, it's still present in the dependencies. See https://github.com/etalab/flask-storage/pull/12


udata/flask_mongoengine/connection.py
maudetes, line 1: Could you add maybe a `__init__.py` with a comment on the inspiration for this module?

udata/mongo/engine.py
```diff
@@ -64,4 +65,3 @@ def resolve_model(self, model):
 
 
 db = UDataMongoEngine()
-session_interface = MongoEngineSessionInterface(db)
```
maudetes, line 67: Was it used anywhere?

ThibaudDauce: No, didn't seems to have any reference of this variable anywhere…

ThibaudDauce: From https://tedboy.github.io/flask/interface_api.session_interface.html

```
app = Flask(__name__)
app.session_interface = MySessionInterface()
```

I don't think we do that…

udata/flask_mongoengine/connection.py
ThibaudDauce: Done in af396b449

PR #3471, opened 2025-10-09
feat(badge): add HVD and generic update badge job

maudetes:
Related to https://github.com/datagouv/data.gouv.fr/issues/1878

Add generic update badge logic.

Add specific update jobs for HVD badges for dataset & dataservices:
* Update HVD badges to candidate datasets/dataservices, based on the hvd tag.
* Only to objects owned by certified and public service organizations are candidate to have a HVD badge.
* Remove the HVD badge if the object has lost its tag.
* We're not adding badge at harvest time cause we also need to deal with non-harvested objects.

## Notes

* Is it a good idea to distinguish tag and badge?
    * It allows a different action between adding a tag freely and :sparkles: _having a badge_ :sparkles:. But it adds some indirection.
* We could add mattermost notification, partially in replacement of the ones done [by this DAG](https://github.com/datagouv/datagouvfr_data_pipelines/tree/main/dgv/monitoring/hvd) -> it would require adding a check on dataset deletion

### DIFF
--- a/udata/core/badges/tests/test_tasks.py
+++ b/udata/core/badges/tests/test_tasks.py
@@ -8,46 +8,48 @@ from udata.core.dataset.factories import DatasetFactory
 from udata.core.dataset.models import Dataset
 from udata.core.organization.constants import CERTIFIED, PUBLIC_SERVICE
 from udata.core.organization.factories import OrganizationFactory
+from udata.tests.api import DBTestCase
 
 
-def test_update_badges(app, rmock):
-    """
-    Test update_badges run the appropriate badge update jobs.
-    In particular, test that the two following registered jobs run and work as expected:
-    - update_dataset_hvd_badge
-    - update_dataservice_hvd_badge
-    """
-    org = OrganizationFactory()
-    org.add_badge(PUBLIC_SERVICE)
-    org.add_badge(CERTIFIED)
+class BadgeTasksTest(DBTestCase):
+    def test_update_badges(self):
+        """
+        Test update_badges run the appropriate badge update jobs.
+        In particular, test that the two following registered jobs run and work as expected:
+        - update_dataset_hvd_badge
+        - update_dataservice_hvd_badge
+        """
+        org = OrganizationFactory()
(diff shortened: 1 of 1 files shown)


udata/core/badges/models.py
```diff
@@ -57,7 +57,7 @@ def add_badge(self, kind):
             msg = "Unknown badge type for {model}: {kind}"
             raise db.ValidationError(msg.format(model=self.__class__.__name__, kind=kind))
         badge = self._fields["badges"].field.document_type(kind=kind)
-        if current_user.is_authenticated:
+        if current_user and current_user.is_authenticated:
```
ThibaudDauce, line 60: `current_user` is never `None`?

udata/core/dataservices/models.py
```diff
@@ -89,6 +85,20 @@ def filter_by_dataset_pagination(self, datasets: list[Dataset], page: int):
...
+        raise db.ValidationError("Unknown badge type")
+
+
+class DataserviceBadge(Badge):
+    kind = db.StringField(required=True, validation=validate_badge)
+
+
+class DataserviceBadgeMixin(BadgeMixin):
+    badges = field(BadgesList(DataserviceBadge), **BadgeMixin.default_badges_list_params)
+    __badges__ = BADGES
```
ThibaudDauce, line 99: Why having a mixin for that?

udata/core/badges/models.py
maudetes: It is None if we are in the context of a job for example :p

udata/core/dataservices/models.py
maudetes: It's the logic we're having at the moment on existing models that use badges (`Dataset`, `Reuse,` etc.).
Agreed that it's weird, but it's already the result [of this refactoring](https://github.com/opendatateam/udata/pull/3155) and I didn't want to try a new refactoring in this PR.

udata/core/badges/models.py
ThibaudDauce: Oh yes, the difference between job and HTTP, didn't know that :-)

PR #3479, opened 2025-10-13
feat(dataset): if harvested, use the harvested dataset modified_at date

maudetes:
Related to https://github.com/ecolabdata/ecospheres/issues/810

[Example dataset](https://www.data.gouv.fr/datasets/balisage-maritime-aides-a-la-navigation) where a legit harvested modified_at date exists.

## Current status study
See [datasets_date_export.csv](https://github.com/user-attachments/files/23098684/datasets_date_export.csv), an export of datasets with current last_update more recent than dataset harvest `modified_at`.

### CKAN

Currently, [we store `metadata_modified` in CKAN harvester](https://github.com/opendatateam/udata/blob/9fe20fb1fa5134a883fe2982b40eb0dc420fd210/udata/harvest/backends/ckan/harvesters.py#L161).
I don't think we have a `data_modified` value at the dataset level in CKAN payload.
See example [dataset](https://www.data.gouv.fr/datasets/dechetteries-de-la-cagtd/), [original record](https://www.datasud.fr/explorer/fr/jeux-de-donnees/dechetteries-de-la-cagtd/info) and [ckan payload](https://www.datasud.fr/fr/indexer/service/ckan/api/3/action/package_search?q=D%C3%A9chetteries%20Communaut%C3%A9%20d%27Agglom%C3%A9ration%20Gap%20-%20Tallard%20-%20Durance) with last modified on 2025-08-04 not exposed at the dataset level.
:heavy_check_mark: We stop storing the `metadata_modified` info in our `dataset.harvest.modified_at` for CKAN harvester (https://github.com/opendatateam/udata/pull/3479/commits/23d279c1042f90f275d4d623e593b394c297cf91)

### CSW-ISO-19139

Many CSW-ISO-1939 datasets don't have a harvest date at the resource level, but only at the dataset level.
In the case of GeoIDE, there is no modified_at detected by hydra at the resource level either. It is due because hydra can't crawl these (they are in exception). Without any additional information, we use the internal modification date, that is based on the date of creation of the resource on the platform, ie. the first day the dataset has been harvested on the platform.
:heavy_check_mark: The harvested `modified_at` is more relevant in studied examples.

### Diverse

There are some cases where hydra detects resource modification more recent than the harvest `modified_at`.
In most cases, it's due to one of the resource being a link to a web site for example and not the data itself.

:heavy_check_mark: Again, in this case, the harvested `modified_at` makes sense in those cases.

### DIFF
--- a/udata/core/dataset/models.py
+++ b/udata/core/dataset/models.py
@@ -780,7 +780,7 @@ class Dataset(Auditable, WithMetrics, DatasetBadgeMixin, Owned, Linkable, db.Doc
 
     def compute_last_update(self):
         """
-        If harvested, use the harvested dataset modified_at date.
+        If dataset is harvested and its metadata contains a modified_at date, use it.
         Else, use the more recent date we would have at the resource level (harvest, modified).
         Default to dataset last_modified if no resource.
         Resources should be fetched when calling this method.


maudetes:
@streino for your information, I've made a short study on cases where the current `last_update` was more recent than  `dataset.harvest.modified_at` to check for impact of this PR.
It seems mostly legit but maybe you also want to take a look.

udata/core/dataset/models.py
```diff
@@ -780,10 +780,13 @@ def last_modified(self):
 
     def compute_last_update(self):
         """
-        Use the more recent date we would have on resources (harvest, modified).
+        If harvested, use the harvested dataset modified_at date.
```
streino:
```suggestion
        If dataset is harvested and its metadata contains a modified_at date, use it.
```

udata/harvest/backends/ckan/harvesters.py
```diff
@@ -158,7 +158,6 @@ def inner_process_dataset(self, item: HarvestItem):
         dataset.tags = [t["name"] for t in data["tags"] if t["name"]]
 
         dataset.harvest.created_at = data["metadata_created"]
```
streino, line 160: Should the same reasoning be applied to `metadata_created`? `created_at` is about the *data* in other standards.

udata/harvest/backends/ckan/harvesters.py
```diff
@@ -158,7 +158,6 @@ def inner_process_dataset(self, item: HarvestItem):
...
-        dataset.harvest.modified_at = data["metadata_modified"]
```
streino, line 161: We could preserve "metadata_modified" in `harvest.metadata_modified_at`. The equivalent in other standards would be:
- ISO-19139: top-level `gmd:dateStamp`.
- DCAT: CatalogRecord's `dct:modified`.

Maybe even do the same for "metadata_created":
- ISO-19139: no equivalent.
- DCAT: CatalogRecord's `dct:issued` (date the record was first *listed* on the catalog, so not exactly *created* but equivalent semantics).

Dates...


I'd say future PR? (maybe when we take a look at all the metadata-related info)

maudetes: I would rather not add more dates at the moment. If need, we'll add specific metadata dates in a second iteration.

udata/harvest/backends/ckan/harvesters.py
maudetes: I think that in the case of `created_at`, it's better to keep the metadata creation date than have nothing, we don't have hydra or other hints to get a creation date.
It would lead to show the date of the dataset creation on data.gouv.fr. It seems like a valid heuristic for now.

streino: Both datagouv dataset creation and metadata creation are wrong, but if that's the choice, I agree metadata creation might be less wrong. 

Still, IMO it'd be better to leave it blank than put an incorrect date. Less room for misunderstanding. This in turn would support adding "metadata" dates (in another iteration) so we still can show the information if it's available; and let the user apply whatever heuristic that works for them.

My reasoning is people who don't care too much about any of this will see a bunch of dates and approximate anyway, thus we end up with the same result no matter what solution we pick. But people who need precise information will have the possibility to find out, whereas if datagouv approximates for them they'll get incorrect information, possibly not even being aware of the issue.

maudetes: Agreed with the points raised, however I think the creation dates consideration would be a different issue and are out of scope for this PR.
I would rather not degrade user experience without a specific study.

PR #3482, opened 2025-10-14
refactor: remove avatars entrypoint and remove robohash and adorable providers

ThibaudDauce:
The goal is to remove the entrypoint system to simplify loading udata. We could add back robohash and adorable if needed without this complex system (just with a condition on the `AVATAR_PROVIDER` setting)

Also working towards removing deprecated `pkg_resources` by removing `entrypoints.py`.

### DIFF
--- a/udata/commands/info.py
+++ b/udata/commands/info.py
@@ -42,7 +42,5 @@ def plugins():
     plugins = current_app.config["PLUGINS"]
     for name, description in entrypoints.ENTRYPOINTS.items():
         echo("{0} ({1})".format(white(description), name))
-        else:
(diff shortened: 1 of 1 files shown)


udata/core/avatars/api.py
maudetes, line 1: Don't we want to keep this into features?

ThibaudDauce: I don't see the distinction between `features` and `core`. I think it's a useless separation (why `identicon`, `notifications`, `territories` and `transfer` are features and `badges` or `metrics` are `core`?)

PR #3483, opened 2025-10-14
refactor!: remove linkchecker

ThibaudDauce:
The goal is to remove the entrypoint system. I don't think we use this linkchecker anymore (it's disabled in our settings), hydra is doing this job now? We maybe need to check the `ResourceMixin@ExtrasField` (I just copy/paste but it could be nice in another PR to add hydra types here?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,7 +45,6 @@ pip-log.txt
 .cache
 .coverage
 .pytest_cache
-.tox
 cover
 nosetests.xml
 coverage.xml
--- a/docs/extending.md
+++ b/docs/extending.md
@@ -111,46 +111,6 @@ If you only need to expose migrations, just provide an empty `models` module.
 
 This module entrypoint allows to register new asynchronous tasks and schedulable jobs.
 
-### Previews (`udata.preview`)
-
-A class entrypoint for preview providers.
-
-These plugins should extend `udata.core.dataset.preview.PreviewPlugin`.
-
-*Example:*
-
-```python
-from udata.core.dataset.preview import PreviewPlugin
-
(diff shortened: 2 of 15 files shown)


udata/core/dataset/models.py
```diff
@@ -369,7 +368,15 @@ class ResourceMixin(object):
     mime = db.StringField()
     filesize = db.IntField()  # `size` is a reserved keyword for mongoengine.
     fs_filename = db.StringField()
-    extras = db.ExtrasField()
+    extras = db.ExtrasField(
+        {
+            "check:available": db.BooleanField,
+            "check:count-availability": db.IntField,
```
maudetes: This isn't used
```suggestion
```

udata/core/dataset/models.py
```diff
@@ -369,7 +368,15 @@ class ResourceMixin(object):
...
+            "check:status": db.IntField,
+            "check:url": db.StringField,
```
maudetes:
```suggestion
```

udata/core/dataset/models.py
```diff
@@ -369,7 +368,15 @@ class ResourceMixin(object):
     mime = db.StringField()
     filesize = db.IntField()  # `size` is a reserved keyword for mongoengine.
     fs_filename = db.StringField()
-    extras = db.ExtrasField()
+    extras = db.ExtrasField(
```
maudetes, line 371: Should we keep the register approach on top?

ThibaudDauce: It's used in tests… Not sure if we want to remove the test or do the test differently… Maybe in another PR?

PR #3487, opened 2025-10-14
fix(tests): add some configuration variable for tests requiring it.

ThibaudDauce:
A few independant commits, you can read them one by one. Some of the tests were failing due to my `udata.cfg`, I don't think tests should expect default values are set, they should set the values needed.

- [x] Rename `MIN_TAG_LENGTH` constant to `TAG_MIN_LENGTH` to be the same as the config value
- [x] Add some `@pytest.mark.options(TAG_MIN_LENGTH=3, TAG_MAX_LENGTH=10)` because with `TAG_MIN_LENGTH=1` in `udata.cfg` it fails.
- [x] Add some missing `SCHEMA_CATALOG_URL` mock (we could also set it to `None` to avoid having to mock but it would test less things?)
- [x] Add some `CDATA_BASE_URL` to avoid comparing cdata URLs with API ones when configured locally
- [x] Rewrote suggest formats API tests because they were wrong? The API code and the test didn't do the same things (the tests seem to think it needed a resource with some format to get it from the suggest but it's not the case, it's from a config…)

### DIFF
--- a/udata/harvest/tests/test_dcat_backend.py
+++ b/udata/harvest/tests/test_dcat_backend.py
@@ -189,10 +189,8 @@ class DcatBackendTest:
             == "https://data.paris2024.org/api/explore/v2.1/console"
         )
 
-    @pytest.mark.options(SCHEMA_CATALOG_URL="https://example.com/schemas")
     def test_harvest_dataservices_keep_attached_associated_datasets(self, rmock):
         """It should update the existing list of dataservice.datasets and not overwrite existing ones"""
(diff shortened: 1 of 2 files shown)


ThibaudDauce:
> I think we could set the default values in `settings.py::Testing` instead of using `pytest.mark.options` for default values in tests
> 
> ```
> class Testing(object):
>     """Sane values for testing. Should be applied as override"""
> ```

I think `udata.cfg` override default Testing values (maybe it shouldn't…) but can we merge this fix so all tests pass on my machine? And today I'm looking into testing to simplify all this because we have 5 ways of doing testing right now, each with a similar but slightly different setup…

PR #3488, opened 2025-10-14
refactor: remove views entrypoint

ThibaudDauce:
Not used since the switch to https://github.com/datagouv/cdata/. You can use the generic `udata.plugins` to add routes in an `init_app()`

Also working towards removing deprecated `pkg_resources` by removing `entrypoints.py`.


udata/frontend/__init__.py
```diff
@@ -101,32 +98,12 @@ def __new__(cls, base, *args, **kwargs):
...
-        app.register_blueprint(blueprint)
-
-
-VIEWS = ["core.storages"]
-
-
-def init_app(app, views=None):
-    views = views or VIEWS
+def init_app(app):
+    from udata.core.storages.views import blueprint as storage_blueprint
```
maudetes, line 102: Do we need those views by the way?

ThibaudDauce: Not sure, I need to think about how we handle storage. I don't have a clear picture but it could be nice to understand this part before moving to S3 :-)

PR #3490, opened 2025-10-14
test: simplifications

ThibaudDauce:
Why do this?

Because today we have a lot of systems in places to do tests:
- Some test are pure functions with fixtures
- Some tests are class without inheritance (pure Pytest only tests)
- Some tests inherit from `APITestCase` with or without `modules = []` (`modules = []` means load API routes…)
- Some tests use `mark.frontend` to load API routes (yes it's the same thing as `modules = []`
- Some tests use `pytest.mark.usefixtures("clean_db")` but other inherit from `DBTestMixin` or sometimes you just do `def setUp(self, api, clean_db)` and it forward the `clean_db` fixture…
- Some tests use `TestClient` from the fixture auto injected `app` or `api` or maybe use `self.get()` which do the same thing but not really… 
- Some tests use `assert x == 1` from simple Pytest but others do `self.assertEqual(x, 1)` from `unittest.TestCase`
- But if you inherit from `unittest.TestCase` you don't have access to fixture injection anymore, so you need to use `self.get()` instead of accepting `api` or `app` from the fixture…
- Some tests do `with api.user()` or `self.login()` or `with self.api_user()`, or `client.login()` or `api.login()` or the standalone function `login_user()`
- We also have 6 different `create_app()` in tests…

The long term goal is to have only pure tests without `unittest.TestCase` since it's recommanded by Pytest https://docs.pytest.org/en/latest/how-to/unittest.html#unittest-testcase

The short term goal is to remove the mess by having two hierarchy:
- `unittest.TestCase` -> `DBTestCase` (only DB) -> `APITestCase` (DB + API routes)
- `object` -> `PytestOnlyDBTestCase` -> `PytestOnlyAPITestCase`

These classes allows us to remove all manual calls to fixtures like `clean_db` and reduce the number of mixins and copy/paste code.

This PR could fix the problems in https://github.com/opendatateam/udata/pull/3488

---

- [x] remove the old syntax with `qs=` for query string parameters
- [x] remove unused `get_json_ld`
- [x] remove `modules = []` concept, load API routes with `*APITestCase`
- [x] remove weird fixture checking for `modules = []`, now `*TestCase` is here to create the app and `*APITestCase` just load the routes afterwards
- [x] remove `@pytest.mark.frontend`, use `*APITestCase` to load API routes instead of this marker
- [x] remove `FrontTestCase` it was the same as `APITestCase`
- [x] remove `WebTestMixin` it was the same as `APITestCase`
- [x] remove `raw_db` fixture (only used in a few places to just get the raw connection, can use `*DBTestCase` and get the connection)
- [x] remove `clean_db` fixture, use `*DBTestCase` to reset database
- [x] remove useless `create_app()`
- [x] simplify `get_settings()` but we should do more…

### DIFF
--- a/udata/tests/api/test_datasets_api.py
+++ b/udata/tests/api/test_datasets_api.py
@@ -1953,7 +1953,7 @@ class DatasetResourceAPITest(APITestCase):
 
     @pytest.mark.options(ALLOWED_RESOURCES_EXTENSIONS=["txt", "html", "kml", "kml-1", "qml", "xml"])
     def test_suggest_formats_api(self):
-        response = self.get(url_for("api.suggest_formats", q="km", size= 5))
+        response = self.get(url_for("api.suggest_formats", q="km", size=5))
         self.assert200(response)
 
         self.assertEqual(len(response.json), 2)


udata/tests/__init__.py
```diff
@@ -1,83 +1,65 @@
...
-        return self.create_app()
-
-    def create_app(self):
+    def get_settings(self, request):
         """
-        Here for compatibility legacy test classes
+        This seems really complicated for what it's doing. I think we want to create
+        an app with the default setting being `settings.Testing` and some overrides
+        from the "options" markers. But to do that `settings.Testing` should inherit from
+        `settings.Default`?
```
maudetes, line 21: I also find this logic confusing and I'm unsure how it works at the moment.
I don't think `settings.Testing` should inherit from `settings.Default`, as it is passed as override to the default ones.

PR #3491, opened 2025-10-15
fix(api): "None is not a valid UpdateFrequency" on dataset update

streino:
400 on PUT, raised when updating a dataset with no frequency set in mongo (exposed as "unknown" at the API level, but null in the model).

### DIFF
--- a/udata/core/dataservices/api.py
+++ b/udata/core/dataservices/api.py
@@ -12,10 +12,10 @@ from udata.auth import admin_permission
 from udata.core.dataservices.constants import DATASERVICE_ACCESS_TYPE_RESTRICTED
 from udata.core.dataset.models import Dataset
 from udata.core.followers.api import FollowAPI
-from udata.core.site.models import current_site
 from udata.frontend.markdown import md
 from udata.i18n import gettext as _
 from udata.rdf import RDF_EXTENSIONS, graph_response, negociate_content
+from udata.utils import get_rss_feed_list
 
 from .models import Dataservice
 from .rdf import dataservice_to_rdf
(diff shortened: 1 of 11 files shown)


udata/core/dataset/forms.py
```diff
@@ -116,6 +116,8 @@ class CommunityResourceForm(BaseResourceForm):
 
 
 def unmarshal_frequency(form, field):
+    if field.data is None:
+        return
     # We don't need to worry about invalid field.data being fed to UpdateFrequency here,
     # since the API will already have ensured incoming data matches the field definition,
     # which in our case is an enum of valid UpdateFrequency values.
```
maudetes, line 123: Should `UpdateFrequency(None)` return None instead?

streino: Nope, that's what actually caused the 400. We ended up [keeping `Dataset.frequency` as optional](https://github.com/opendatateam/udata/pull/3343#discussion_r2411602725), which means we need to handle the case where frequency doesn't exist at all, and therefore is not mappable to an `UpdateFrequency` value.

streino: Sorry, misread your question...

I don't see how we could get `UpdateFrequency(None)` to return `None`: `UpdateFrequency` is a `StrEnum`, so values must be of type `str`, and so we can't declare an `UpdateFrequency.NONE` entry with a value of `None`. We could map `UpdateFrequency.NONE` to some arbitrary string value that we'd treat as if it were `None`, but I think it's just easier to use `None` directly and treat `frequency` as `UpdateFrequency | None`, instead of trying to somehow embed the `Optional` trait inside `UpdateFrequency`. We'd still have to differentiate `NONE`-meaning-`None` vs "real" values anyway.

For the record, I also thought of not explicitly mapping `None` to an `UpdateFrequency` value, and going through `Enum._missing_(None)` to handle that special case, but we can't have `_missing_(None)` return `None`, because `None` is used to signal that `_missing_()` failed to resolve its argument, and so callers (within `Enum`) will raise an exception if they get `None` back. So we still couldn't have `UpdateFrequency(None)` return `None` with that approach.

maudetes: Thank you for the detailed explanation

PR #3493, opened 2025-10-20
fix(dataservice): unindex archived dataservices

maudetes:
Archived dataservices are currently indexed in the search engine.
They are not taken into account in site metrics, leading to confusing numbers on https://www.data.gouv.fr/dataservices/search
image

Fix dataservices visibility and deduplicate visibility logics for datasets, reuses and dataservices.


bolinocroustibat (approved):
Much more consistent and understandable, thank you!

PR #3495, opened 2025-10-27
refactor: simplify plugins entrypoint

ThibaudDauce:

If plugins have a `udata.plugins` entrypoint they will always be loaded. If you want to disable a plugin, just uninstall it.

The `udata.plugins` replaces the `models` (removed in https://github.com/opendatateam/udata/pull/3496) and `metrics` (removed in https://github.com/opendatateam/udata/pull/3500).

### DIFF
--- a/udata/core/dataservices/models.py
+++ b/udata/core/dataservices/models.py
@@ -130,7 +130,7 @@ def filter_by_topic(base_query, filter_value):
     try:
         topic = Topic.objects.get(id=filter_value)
     except Topic.DoesNotExist:
-        pass
+        return base_query
     else:
         return base_query.filter(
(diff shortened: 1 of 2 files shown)


docs/extending.md
```diff
@@ -111,13 +111,6 @@ If you only need to expose migrations, just provide an empty `models` module.
 
 This module entrypoint allows to register new asynchronous tasks and schedulable jobs.
 
-### Generic plugins (`udata.plugins`)
```
maudetes, line 94: Shouldn't we keep this doc?

ThibaudDauce: Added back in e04bee8bb

PR #3496, opened 2025-10-27
refactor: remove models entrypoint and simplify migrations core code

ThibaudDauce:

If you want to register some models extras you can use `udata.plugins` instead.

- [x] Remove `udata.models` plugin system in migrations
- [x] Rewrote tests without mocking
- [x] Test that all the migrations run and pass on the default fixture database (not perfect but can catch problems when PR add migrations with a syntax error for example…)
- [x] Switch from `log.*` to `echo` because otherwise the prints didn't show up in `result.output` in tests…
- [x] Remove `MigrationFormatter` it seems to do nothing?
- [x] Remove `rollback` support, it wasn't use
- [x] Remove `state` system, it was only (not) used in rollback

### DIFF
--- a/udata/models/__init__.py
+++ b/udata/models/__init__.py
@@ -1,3 +1,5 @@
+from udata.mongo import *  # noqa
+
 # Load all core models and mixins
 from udata.core.spatial.models import *  # noqa
 from udata.core.metrics.models import *  # noqa


udata/db/migrations.py
maudetes, line 1: Why not keeping it in `__init__.py`?

udata/commands/db.py
maudetes, line 1: I think you haven't pushed the models entrypoint removal?
There is still `udata.models` in `entrypoints.py`.

udata/db/migrations.py
ThibaudDauce: Because it forces to do a custom check on the `migrations` folder to exclude the `__init__`. If there is only migrations in the migrations folder there is no more the need of some filtering code

ThibaudDauce: And I don't like putting too much code in `__init__` since it's loaded everytime a children is loaded it can lead to circular dependencies (even if it's not the case for this one)

udata/commands/db.py
ThibaudDauce: Weird, removed in 3a3f18a2a

udata/db/migrations.py
maudetes: Yes, makes sense, thanks!

PR #3497, opened 2025-10-27
 feat(harvest): combine agent and organization names when contact point has both

streino:
Fix https://github.com/ecolabdata/ecospheres/issues/843

Context : https://github.com/opendatateam/udata/pull/3374#discussion_r2263423257

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+## 14.4.0 (2025-12-16)
+
+- chore: adapt CI and docs to main branch ([#3559](https://github.com/opendatateam/udata/pull/3559))
+- chore: move pyproject optional-dependencies to dependency-groups ([#3560](https://github.com/opendatateam/udata/pull/3560))
+- chore: remove trailing slash in cdata url ([#3565](https://github.com/opendatateam/udata/pull/3565))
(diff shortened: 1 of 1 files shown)


udata/rdf.py
```diff
@@ -360,6 +361,18 @@ def themes_from_rdf(rdf):
...
+def contact_point_name(agent_name: str | None, org_name: str | None) -> str:
+    match (agent_name, org_name):
+        case (None, None):
+            return ""
+        case (agent_name, None):
+            return agent_name
+        case (None, org_name):
+            return org_name
+        case (agent_name, org_name):
+            return f"{agent_name} ({org_name})"
```
ThibaudDauce: Is it simpler?
```suggestion
    if agent_name and org_name:
        return f"{agent_name} ({org_name})"
    return agent_name or org_name or ""
```

streino: Much simpler indeed!

streino:
LGTM!

https://demo.data.gouv.fr/datasets/atlas-des-fonds-meubles-du-plateau-continental-du-golfe-de-gascogne-cartes-biosedimentaires-source-chasse-c-glemarec-m-1976-echelle-1-100-000-a-1-500-000-1


https://demo.data.gouv.fr/datasets/inventaire-regional-du-patrimoine-geologique-dauvergne-rhone-alpes-1

PR #3499, opened 2025-10-28
feat(harvest): report duplicate remote ids as an harvesting error

streino:
Fix https://github.com/ecolabdata/ecospheres/issues/842

### DIFF
--- a/udata/harvest/tests/test_base_backend.py
+++ b/udata/harvest/tests/test_base_backend.py
@@ -483,7 +483,6 @@ class BaseBackendTest(PytestOnlyDBTestCase):
             else:
                 assert job.status == "failed"
                 assert job.remote_id in job.errors[0].message
-                assert job.remote_url in job.errors[0].message
 
 
 class BaseBackendValidateTest(PytestOnlyDBTestCase):


ThibaudDauce:
For the newlines, I think you can try replacing:
```
          <div
            v-if="error.details"
            class="text-mention-grey bg-gray-some p-2 text-xs/4"
          >
            {{ error.details }}
          </div>
```

with:
```
          <pre
            v-if="error.details"
            class="text-mention-grey bg-gray-some p-2 text-xs/4"
          >
            {{ error.details }}
          </pre>
```

streino:
Change discussed above done : 
- Dupe detection on the fly, so first item with a dupe id is OK, and following ones are errors.
- Dupe items aren't saved, and therefore not indexed.
- Error message shows problematic `remote_id`, and `remote_url` if available.
- No more overall error (so need to worry about formatting), since duplicate items are ignored and errors are reported at the item level.

streino:
Ready for review! @maudetes or @ThibaudDauce 

Just ran into another case of dupes where this feature would have prevented a lot of head scratching. Can't wait to see it deployed :grin: :pray:

udata/harvest/models.py
```diff
@@ -66,6 +66,7 @@ class HarvestLog(db.EmbeddedDocument):
 
 class HarvestItem(db.EmbeddedDocument):
     remote_id = db.StringField()
+    remote_url = db.StringField()
```
maudetes, line 69: :sparkles:

udata/harvest/backends/base.py
```diff
@@ -325,6 +336,16 @@ def process_dataservice(self, remote_id: str, **kwargs) -> bool:
             item.ended = datetime.utcnow()
             self.save_job()
 
+    def ensure_unique_remote_id(self, item):
+        if item.remote_id in self.remote_ids:
+            raise HarvestValidationError(
+                f"Identifier '{item.remote_id}'"
+                + (f" for {item.remote_url}" if item.remote_url else "")
```
maudetes: If you're okay with that, I would say we should not mention the `remote_url` in the error, but instead expose the newly stored `remote_url` as a new column in the items table.

streino: Yes definitely!

streino: You beat me to it :grin: :pray:

streino: Implemented in: 
- https://github.com/opendatateam/udata/pull/3538
- https://github.com/datagouv/cdata/pull/779

PR #3502, opened 2025-10-29
refactor: simplify getting udata version

ThibaudDauce:
Working towards removing deprecated `pkg_resources`.


maudetes (commented):
Don't we want to update the current `package_version` function and use it instead of a specific udata one?

ThibaudDauce:
We could but `package_version` was a filter Jinja and I don't find the wording very clear about what we talk about vs `get_udata_version`. We never used it for anything other than udata so no point to having a generic param too I think…

PR #3504, opened 2025-10-29
refactor!: simplify harvesters entrypoint

ThibaudDauce:

- The `udata.harvesters` entrypoints are automaticly loaded (without the need of `PLUGINS` settings)
- Need to move the harvesters from `PLUGINS` to `HARVESTER_BACKENDS` to enable them.
- Need to add explicit glob with `*`, like `csw*` to enable all backends starting with `csw` (before it was an implicit `startswith`)
- Need to add explicit `name` to the backends' classes (before it was the name of the entrypoint)

Example of new setting:

```
HARVESTER_BACKENDS = [
  'ckan',
  'dkan',
  'dcat',
  'csw*',
  'maaf',
]
```

- [x] Added `name` to all internal harvesters backends
- [x] Removed `actions.list_backends()`, use `backends.get_enabled_backends()` since it's not really an action.
- [x] Added `HARVESTER_BACKENDS` setting to replace `PLUGINS` for enabling/disabling harvesters backends
- [x] Removed broken `/api/1/harvest/job_status/` endpoint

### DIFF
--- a/udata/app.py
+++ b/udata/app.py
@@ -20,6 +20,7 @@ from flask import (
 )
 from flask_caching import Cache
 from flask_wtf.csrf import CSRFProtect
+from importlib_metadata import entry_points
 from speaklater import is_lazy_string
 from werkzeug.exceptions import NotFound
 from werkzeug.middleware.proxy_fix import ProxyFix
@@ -200,12 +201,21 @@ def create_app(config="udata.settings.Defaults", override=None, init_logging=ini
 def standalone(app):
     """Factory for an all in one application"""
     from udata import api, core, frontend
+    from udata.features import notifications
 
     core.init_app(app)
     frontend.init_app(app)
     api.init_app(app)
+    notifications.init_app(app)
 
-    register_features(app)
+    eps = entry_points(group="udata.plugins")
+    for ep in eps:
+        plugin_module = ep.load()
+
+        if hasattr(plugin_module, "init_app"):
+            plugin_module.init_app(app)
+        else:
+            log.error(f"Plugin {ep.name} ({ep.value}) doesn't expose an `init_app()` function.")
 
     return app
 
@@ -278,12 +288,3 @@ def page_not_found(e: NotFound):
         return render_template("404.html", homepage_url=homepage_url()), 404
 
     return jsonify({"error": e.description, "status": 404}), 404
-
-
-def register_features(app):
-    from udata.features import notifications
-
-    notifications.init_app(app)
-
-    for ep in entrypoints.get_enabled("udata.plugins", app).values():
-        ep.init_app(app)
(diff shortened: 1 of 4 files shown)


udata/settings.py
```diff
@@ -176,6 +176,7 @@ class Defaults(object):
     UDATA_INSTANCE_NAME = "udata"
 
     PLUGINS = []
+    HARVESTERS_BACKENDS = []
```
maudetes:
```suggestion
    HARVESTER_BACKENDS = []
```

udata/harvest/backends/__init__.py
```diff
@@ -1,17 +1,33 @@
...
+    return any(fnmatch(backend.name, g) for g in current_app.config["HARVESTERS_BACKENDS"])
+
+
+def get_enabled_backends() -> dict[str, type[BaseBackend]]:
+    return {
+        name: backend for name, backend in get_all_backends().items() if is_backend_enabled(backend)
+    }
+
+
+def get_enabled_backends_ids() -> list[str]:
```
maudetes: I think we could use a lambda directly, ie `enum=lambda: list(get_enabled_backends().keys())`?

udata/harvest/api.py
```diff
@@ -457,15 +456,7 @@ def get(self):
...
             key=lambda b: b["label"],
         )
-
-
-@ns.route("/job_status/", endpoint="havest_job_status")
-class ListHarvesterAPI(API):
-    @api.doc(model=[str])
-    def get(self):
-        """List all available harvesters"""
-        return actions.list_backends()
```
maudetes, line 471: What was that ahah? :eyes:

udata/harvest/commands.py
```diff
@@ -89,9 +90,10 @@ def sources(scheduled=False):
 @grp.command()
 def backends():
     """List available backends"""
-    log.info("Available backends:")
-    for backend in actions.list_backends():
-        log.info("%s (%s)", backend.name, backend.display_name or backend.name)
+    print("Available backends:")
```
maudetes, line 93: Could we keep `log.info`?

maudetes (approved):
:rocket:

udata/settings.py
ThibaudDauce: Done in 027464557

udata/harvest/backends/__init__.py
ThibaudDauce: Yes! Much better :-) In 821db33bc

udata/harvest/commands.py
ThibaudDauce: Why?

ThibaudDauce: And `log.info` is not testable (and I don't think it's wanted here for this CLI)

maudetes: I would rather keep a proper logging with levels by default?

ThibaudDauce: We need to talk about this because I don't think we have the same point of view on logging vs printing :-)

PR #3506, opened 2025-10-30
feat: add dataservices to reuses

ThibaudDauce:
- Working on https://github.com/datagouv/data.gouv.fr/issues/1895


udata/core/reuse/api.py
```diff
@@ -227,14 +228,39 @@ def post(self, reuse):
             dataset = Dataset.objects.get_or_404(id=id_or_404(request.json["id"]))
         except Dataset.DoesNotExist:
             msg = "Dataset {0} does not exists".format(request.json["id"])
-            api.abort(404, msg)
+            return api.abort(404, msg)
```
maudetes, line 231: Why the need to `return`?

maudetes (approved):
Clean! Thank you! :pray:

udata/core/reuse/api.py
ThibaudDauce: Without the `return` static analyzers don't understand that `dataset` is present. It's a small trick to avoid some static errors (but not necessary since `api.abort()` stop the execution of the function anyway)

PR #3507, opened 2025-11-04
feat: add email to user search

ThibaudDauce:
- Fix https://github.com/datagouv/data.gouv.fr/issues/1891


udata/core/user/models.py
```diff
@@ -103,7 +103,16 @@ class User(WithMetrics, UserMixin, Linkable, db.Document):
     on_delete = Signal()
 
     meta = {
-        "indexes": ["$slug", "-created_at", "slug", "apikey"],
+        "indexes": [
```
maudetes, line 105: Testing it out user search locally with my own email address [in admin](http://localhost:3000/admin/site/users), it matches too many users, mine not even returned on the first page.
We should remove the `sort=-created` in cdata query to make sure to use `$text_score` scoring to return the best matching user at the top.

ThibaudDauce: Done in https://github.com/datagouv/cdata/pull/768

maudetes (approved):
:pray:

PR #3509, opened 2025-11-04
feat: archive and purge csv catalog

maudetes:
Close https://github.com/datagouv/data.gouv.fr/issues/532.


maudetes:
> Nice, that's quite light! Is it called with `replace=True` in the CRON job?

Yep, it's here: https://github.com/opendatateam/udata/pull/3509/files#diff-0ba90c9a5db29bc2ef0536194b2c1de96761e094cd99954cfd32d4cd4fcf9c16R194

PR #3511, opened 2025-11-05
build: try using conventional commit in Crowdin PR

ThibaudDauce:


crowdin.yml
```diff
@@ -1,4 +1,6 @@
-commit_message": '[ci skip]'
+pull_request_title: "feat: update translations"
+
+commit_message": "[ci skip]"
```
maudetes, line 3: It's not used, right?

ThibaudDauce: I'm not sure :-D

PR #3512, opened 2025-11-06
feat: add reuse filter on dataservices

ThibaudDauce:
This got removed with a bad merge I think :-(


udata/core/dataservices/models.py
```diff
@@ -140,11 +140,23 @@ def filter_by_topic(base_query, filter_value):
...
+        return base_query.filter(id__in=[dataservice.id for dataservice in reuse.dataservices])
+
+
 @generate_fields(
     searchable=True,
     nested_filters={"organization_badge": "organization.badges"},
     standalone_filters=[
-        {"key": "topic", "constraints": "objectid", "query": filter_by_topic, "type": str}
+        {"key": "topic", "constraints": "objectid", "query": filter_by_topic, "type": str},
+        {"key": "reuse", "constraints": "objectid", "query": filter_by_reuse, "type": str},
```
maudetes: Seems like constraints [should be a list](https://github.com/opendatateam/udata/blob/master/udata/api_fields.py#L520)?

udata/core/dataservices/models.py
```diff
@@ -140,11 +140,23 @@ def filter_by_topic(base_query, filter_value):
         )
 
 
+def filter_by_reuse(base_query, filter_value):
+    from udata.core.reuse.models import Reuse
+
+    try:
+        reuse = Reuse.objects.get(id=filter_value)
+    except Reuse.DoesNotExist:
+        pass
```
maudetes: I guess we should either raise an error or return base_query (it's the same for topics).
Else `filter_by_reuse` returns None, leading to a 500 `AttributeError: 'NoneType' object has no attribute 'paginate'`.

udata/core/dataservices/models.py
ThibaudDauce: Fixed in 03888370b (there was another catch somewhere that returns a 400 with a message on this error… weird…)

udata/core/dataservices/models.py
ThibaudDauce: Good catch! Thanks! Fixed in 03888370b

PR #3524, opened 2025-11-18
fix: deleting and purging page's blocs elements (reuses, datasets and dataservices)

ThibaudDauce:


udata/core/dataservices/tasks.py
```diff
@@ -28,6 +29,12 @@ def purge_dataservices(self):
         Transfer.objects(subject=dataservice).delete()
         # Remove dataservices references in Topics
         TopicElement.objects(element=dataservice).update(element=None)
+        # Remove dataservices in pages (mongoengine doesn't support updating a field in a generic embed)
+        Page._get_collection().update_many(
+            {"blocs.dataservices": dataservice.id},
+            {"$pull": {"blocs.$[b].dataservices": dataservice.id}},
+            array_filters=[{"b.dataservices": dataservice.id}],
+        )
```
maudetes, line 37: I don't think you need to use the raw collection, you could do
```
Page.objects(blocs__dataservices=dataservice.id).update(pull__blocs__S__dataservices=dataservice.id)
```

ThibaudDauce: I tried this and it doesn't work… Have you manage to make it work?

maudetes: Hmm yep, I retested it at the moment
```python
from udata.models import Page
from udata.core.dataservices.models import Dataservice
dataservice = Dataservice.objects(id="672cf668fcc8065be6e66f52").first()
Page.objects(blocs__dataservices=dataservice.id).update(pull__blocs__S__dataservices=dataservice.id)
>>> 1
```

PR #3525, opened 2025-11-18
refactor: standardization of test clients, CLI runner, mocker…

ThibaudDauce:
I encounter a few bugs with the way we are testing / authenticating in API tests during the upgrade of Flask / Flask-Security. This PR goal is to merge when we have multiple ways of doing the same things to better understand how requests, auth, cli calls are made to the application. I switch from fixture to parent functions in inheritance because it's way easier to understand what the function is doing by ctrl-clic than magical fixtures…

- [x] Remove `TestClient` (replaced by functions inside `APITestCaseMixin`)
- [x] Remove `ApiClient` (replaced by the same functions inside `APITestCaseMixin`)
- [x] Remove `client` and `api` fixtures (use functions inside `APITestCaseMixin`)
- [x] Switch to `with app.test_request_context(): login_user(user)` for activities tests requiring authentication since it's not really an API test…
- [x] Use `mocker` instead of `monkeypatch` (there was only one usage of `monkeypatch` in the tests)
- [x] Remove `TemplateRecorder` since there is no more templates
- [x] Remove unused `httpretty` fixture
- [x] Remove `SitemapClient` since the sitemap is in cdata

### DIFF
--- a/udata/core/dataservices/tasks.py
+++ b/udata/core/dataservices/tasks.py
@@ -6,6 +6,7 @@ from udata.core.constants import HVD
 from udata.core.dataservices.models import Dataservice
 from udata.core.organization.constants import CERTIFIED, PUBLIC_SERVICE
 from udata.core.organization.models import Organization
+from udata.core.pages.models import Page
 from udata.core.topic.models import TopicElement
 from udata.harvest.models import HarvestJob
 from udata.models import Discussion, Follow, Transfer
@@ -28,6 +29,12 @@ def purge_dataservices(self):
         Transfer.objects(subject=dataservice).delete()
         # Remove dataservices references in Topics
         TopicElement.objects(element=dataservice).update(element=None)
+        # Remove dataservices in pages (mongoengine doesn't support updating a field in a generic embed)
+        Page._get_collection().update_many(
+            {"blocs.dataservices": dataservice.id},
+            {"$pull": {"blocs.$[b].dataservices": dataservice.id}},
+            array_filters=[{"b.dataservices": dataservice.id}],
+        )
         # Remove dataservice
         dataservice.delete()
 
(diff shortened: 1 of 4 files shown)


udata/harvest/tests/test_dcat_backend.py
```diff
@@ -872,24 +872,30 @@ def test_unable_to_detect_format(self, rmock):
...
                 "dataset": [],
             },
         )
         rmock.head(url, headers={"Content-Type": "application/json"})
+
         org = OrganizationFactory()
         source = HarvestSourceFactory(backend="dcat", url=url, organization=org)
+
+        # The test just checks that the replacement mechanism exists and can be patched
+        # We don't actually test URL replacement here since it would require mocking urllib
```
maudetes, line 894: If we want to test that something actually happened, we could check for the graph stored in `job.data`?

udata/tests/api/test_auth_api.py
```diff
@@ -67,34 +67,34 @@ def oauth(app, request):
...
 
-        response = api.post(url_for("api.fake"))
+        response = self.post(url_for("api.fake"))
 
         assert200(response)
         assert response.content_type == "application/json"
         assert response.json == {"success": True}
 
-    def test_header_auth(self, api):
+    def test_header_auth(self):
```
maudetes, line 88: Does this test actually fail without the correct api key?

udata/tests/dataset/test_dataset_model.py
```diff
@@ -326,9 +326,11 @@ def test_dataset_fetch_exclude_resource(self):
 
         assert dataset_without_resources.resources_len == 0
 
-    def test_dataset_activities(self, api, mocker):
+    def test_dataset_activities(self, app, mocker):
         # A user must be authenticated for activities to be emitted
-        user = api.login()
```
maudetes, line 331: I'm slightly confused, wouldn't we want a `self.login` here as well? Even, if we may need to use a `PytestOnlyAPITestCase`?

maudetes (approved):
Thank you for the clean! :pray: I'm not completely sure about the login logic.

Additionally, I think typing fixtures may be enough ,ie using `def my_test(app: UDataApp)` to get autocompletion.
But indeed, pytest fixtures may be magical and I'm fine with moving more things at the class level for readability!

PR #3530, opened 2025-11-19
fix: do not send security mail on demo

ThibaudDauce:


udata/auth/__init__.py
```diff
@@ -36,6 +38,20 @@ def __init__(self, *needs):
...
+        self,
+        template: str,
+        subject: str,
+        recipient: str,
+        sender: t.Union[str, tuple],
+        body: str,
+        html: t.Optional[str],
+        **kwargs: t.Any,
+    ) -> None:
+        return None
```
maudetes, line 56: Don't we wanna `log.debug`?

ThibaudDauce: Yes! Done in 807218463

PR #3535, opened 2025-11-20
fix: logo in mail

ThibaudDauce:
- [x] Use logo in PNG instead of SVG
- [x] Change data.gouv blue
- [x] Show 200 chars instead of 50 for user content
- [x] Put user content in italic


maudetes (approved):
Thank you for the quickfix

PR #3538, opened 2025-11-24
feat(harvest): expose remote_url at the item level

maudetes:


udata/harvest/api.py
```diff
@@ -55,6 +55,9 @@
     "HarvestItem",
     {
         "remote_id": fields.String(description="The item remote ID to process", required=True),
+        "remote_url": fields.String(
+            description="The item remote url (if available)", required=True
```
streino: I'm not sure about `required=True`. And if it is, what about the "if available" in the description?

maudetes: Indeed, removed in https://github.com/opendatateam/udata/pull/3538/commits/87bd3078a21b42be731322431f23159fcdcfe1c1.

PR #3539, opened 2025-11-24
feat: dismiss reports

ThibaudDauce:


udata/core/reports/models.py
```diff
@@ -12,7 +12,20 @@
 from .constants import REPORT_REASONS_CHOICES, REPORTABLE_MODELS
 
 
-@generate_fields()
+def filter_by_status(base_query, filter_value):
+    if filter_value == "ongoing":
+        return base_query.filter(dismissed_at=None, subject__exists=True)
+    elif filter_value == "done":
+        return base_query.filter(dismissed_at__ne=None)
```
maudetes: Maybe we should add these "ongoing", "done" logic as a `UDataQuerySet` for reusability?

udata/core/reports/models.py
```diff
@@ -12,7 +12,20 @@
 from .constants import REPORT_REASONS_CHOICES, REPORTABLE_MODELS
 
 
-@generate_fields()
+def filter_by_status(base_query, filter_value):
```
maudetes: Could you add a test on this?

ThibaudDauce: Yes. Done in 387b890ed

udata/core/reports/models.py
ThibaudDauce: Done in ec670f12c

udata/core/reports/models.py
```diff
@@ -11,8 +11,38 @@
...
+REPORT_STATUS_CHOICES = [REPORT_STATUS_ONGOING, REPORT_STATUS_DONE]
 
-@generate_fields()
+
+class ReportQuerySet(db.BaseQuerySet):
+    def ongoing(self):
+        return self.filter(dismissed_at=None, subject_deleted_at=None)
+
+    def done(self):
+        return self.filter(Q(dismissed_at__ne=None) | Q(subject_deleted_at__ne=None))
```
maudetes: I think that on top of having `done`, we could have a separate filter for dismissed vs deleted?

udata/core/reports/models.py
```diff
@@ -11,8 +11,38 @@
 
 from .constants import REPORT_REASONS_CHOICES, REPORTABLE_MODELS
 
+REPORT_STATUS_ONGOING = "ongoing"
+REPORT_STATUS_DONE = "done"
```
maudetes: Do we want to go with handle/unhandle wording for the filters here?

udata/core/reports/api.py
```diff
@@ -42,6 +42,22 @@ class ReportAPI(API):
     def get(self, report):
         return report
 
+    @api.doc("update_report", responses={400: "Validation error"})
+    @api.secure(admin_permission)
+    @api.expect(Report.__write_fields__)
+    @api.marshal_with(Report.__read_fields__, code=200)
+    def patch(self, report):
+        is_dismiss = (
```
maudetes: If for some reason we want to "undismiss", we should also unset `dismissed_by`?

udata/core/reports/models.py
ThibaudDauce: Can we add it when we need it?

udata/core/reports/models.py
ThibaudDauce: Done in 67768c8ff

ThibaudDauce: Could it be a boolean?

udata/core/reports/api.py
ThibaudDauce: Support added in 1f7a08047

udata/core/reports/models.py
maudetes: Yep, when needed!

udata/core/reports/models.py
maudetes: hmm True

ThibaudDauce: Switch to bool in 2b1e85f45

PR #3542, opened 2025-11-26
fix: validating dict in boolean harvest filter

ThibaudDauce:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/273692/?alert_rule_id=10&alert_timestamp=1764115213721&alert_type=email&environment=demo.data.gouv.fr&notification_uuid=44880ccd-2a42-443d-8e76-2e6e5edefe74&project=12&referrer=alert_email

Can be tested with https://www.datasud.fr/fr/indexer/service/ckan/ 

`value = extra["value"]` can already be a `dict` in Ckan, so `json.loads() is failing`. The change is the `boolean` function is not useful I think with the second change in spatial Ckan management but it's better anyway to check for TypeError for `int()` parsing failing.


udata/harvest/filters.py
```diff
@@ -17,16 +17,22 @@ def boolean(value):
 
     try:
```
maudetes: Wouldn't it be more explicit to have something as
```
    if not (isinstance(value, str) or isinstance(value, bool) or isinstance(value, int)):
        raise Invalid(f"Unable to parse boolean '{value}' of type {type(value)}")
```

ThibaudDauce: If we explicitly do each types, it could be simpler but we need to manage the string "1" or "0". Right now we support setting "42" as `str` and it converts to `True`… Not sure if it's something we want?

ThibaudDauce: What do you think of 55b8b349c?

maudetes: Yep, seems nice!

PR #3548, opened 2025-12-01
fix: invalid dataset.harvest.uri

ThibaudDauce:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/273971/?alert_rule_id=10&alert_timestamp=1764517223302&alert_type=email&environment=data.gouv.fr&notification_uuid=1148d298-3e41-4bdb-b542-e45e0e1ec12b&project=12&referrer=alert_email

https://errors.data.gouv.fr/organizations/sentry/issues/274240/?alert_rule_id=10&alert_timestamp=1764814875616&alert_type=email&environment=data.gouv.fr&notification_uuid=4ef4a817-e42b-4078-973c-af744c2e2ef9&project=12&referrer=alert_email

https://www.data.gouv.fr/api/1/site/catalog.trig is crashing with 500.

Not sure if we want to do a migration to remove these URI?

### DIFF
--- a/udata/core/dataset/rdf.py
+++ b/udata/core/dataset/rdf.py
@@ -276,11 +276,17 @@ def resource_to_rdf(
 
 
 def is_valid_rdf_uri(uri: str) -> bool:
-    """Check if a URI is valid for RDF serialization (no spaces or invalid chars)"""
-    # Spaces and other whitespace are not allowed in URIs
+    """Check if a URI is valid for RDF serialization.
+
+    Uses RDFLib's own validation by attempting to serialize the URI.
+    """
     if not uri:
         return False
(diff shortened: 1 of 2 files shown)


udata/core/dataset/rdf.py
```diff
@@ -275,10 +275,27 @@ def resource_to_rdf(
     return r
 
 
+def is_valid_rdf_uri(uri: str) -> bool:
```
maudetes, line 278: Hmm I wonder if we should merge it with `udata.uris.validate`?

maudetes (commented):
I wonder if we should keep storing it but not expose it instead?
I think the issue occurs on other fields than harvest.uri, but all kinf of URIs (email, resource url etc.)

PR #3549, opened 2025-12-01
fix: change email for already taken email

ThibaudDauce:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/273985/?alert_rule_id=10&alert_timestamp=1764576202302&alert_type=email&environment=data.gouv.fr&notification_uuid=87c0b62b-79f2-4680-9bb8-8bfb28ce7308&project=12&referrer=alert_email

### DIFF
--- a/docs/adapting-settings.md
+++ b/docs/adapting-settings.md
@@ -452,6 +452,16 @@ The background color used by the internal provider
 
 The padding (in percent) used by the internal provider
 
+## Notifications configuration
+
+These settings allow you to customize the notification feature.
+
+### DAYS_AFTER_NOTIFICATION_EXPIRED
+
+**default**: 90
+
+The delay of days between an handled notification and its deletion.
+
 ## Posts configuration
 
 Theses settings allow you to customize the post feature.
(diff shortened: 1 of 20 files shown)


udata/auth/views.py
```diff
@@ -84,6 +84,11 @@ def confirm_change_email(token):
     if flash:
         return redirect(homepage_url(flash=flash, flash_data=flash_data))
 
+    # Check if the new email is already taken by another user
+    existing_user = _datastore.find_user(email=new_email)
+    if existing_user and existing_user.id != user.id:
+        return redirect(homepage_url(flash="change_email_already_taken"))
```
maudetes, line 90: See the issue [here](https://github.com/datagouv/data.gouv.fr/issues/1478) about generic response.
I think we should show a success response but send a mail in the background if `SECURITY_RETURN_GENERIC_RESPONSES=True`

ThibaudDauce: I don't think it's a problem here, because we are in the confirm endpoint so the user is already clicking a link from the new email address. So it's confirmed he is the owner, no?

ThibaudDauce: It could be a problem if we don't require to confirm email but since it's not the case currently in production, do we want to do something?

maudetes: Hmm true, seems fine for now!

PR #3550, opened 2025-12-02
feat: expose harvest source permissions

ThibaudDauce:


udata/harvest/models.py
```diff
@@ -173,6 +173,20 @@ def schedule(self):
...
+    @property
+    def permissions(self):
+        from udata.auth import admin_permission
+        from udata.core.dataset.permissions import OwnablePermission
+
+        return {
+            "edit": OwnablePermission(self),
+            "delete": OwnablePermission(self),
+            "run": OwnablePermission(self),
+            "preview": OwnablePermission(self),
```
maudetes: Shouldn't it be something like `EditOrganizationPermission` for all? Except maybe for preview

ThibaudDauce: It's a copy from the current permissions in the API in production, we could change it later if we want?

maudetes: I mentioned it while we're at it. It seems weird that an editor can delete but can't create

ThibaudDauce: Yes it's weird, done in 25a1d4459fd64b7cf9abd9773358e056aa3fcc8a

PR #3553, opened 2025-12-03
fix: Sentry exception log when wrong coordinates

ThibaudDauce:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/274185/?alert_rule_id=10&alert_timestamp=1764773902993&alert_type=email&environment=data.gouv.fr&notification_uuid=3bb23108-c921-48ea-a177-717e43324dd2&project=12&referrer=alert_email

- Use log.warning instead of log.exception → not captured by Sentry
- Include the invalid value in the log message for debugging
- Use ValidationError instead of ValueError for better semantics


udata/core/spatial/forms.py
```diff
@@ -64,8 +64,8 @@ def process_formdata(self, valuelist):
                     self.data = geojson.GeoJSON.to_instance(value)
             except Exception:
                 self.data = None
-                log.exception("Unable to parse GeoJSON")
-                raise ValueError(self.gettext("Not a valid GeoJSON"))
+                log.warning("Unable to parse GeoJSON: %s", value)
```
maudetes:
```suggestion
                log.warning(f"Unable to parse GeoJSON: {value}")
```

PR #3555, opened 2025-12-04
feat: add UUID to discussions' messages

ThibaudDauce:
In https://github.com/opendatateam/udata/pull/3546 or in https://github.com/datagouv/cdata/pull/781#discussion_r2568266786 we want to target a message inside a discussion. Discussions' messages are the only one we target with the index, so we add an UUID in this PR to be the same as Resource for exemple and in a next PR we'll change the index indexing to a UUID indexing to delete / update / report…


udata/migrations/2025-12-04-add-uuid-to-discussion-messages.py
```diff
@@ -0,0 +1,39 @@
...
+            for message in messages:
+                if "id" not in message:
+                    message["id"] = uuid.uuid4()
+                    updated = True
+                    total_messages_updated += 1
+
+            if updated:
+                collection.update_one(
+                    {"_id": discussion["_id"]}, {"$set": {"discussion": messages}}
+                )
```
maudetes: I think you could use `discussion._mark_as_changed('discussion')` directly, it would auto populate directly the id field.

maudetes (approved):
:partying_face:

udata/migrations/2025-12-04-add-uuid-to-discussion-messages.py
ThibaudDauce: Didn't know this function, thanks! Done in 5d1dce0cf

PR #3556, opened 2025-12-04
refactor: remove SpamInfo in profit of Report

ThibaudDauce:


udata/migrations/2025-11-27-migrate-spaminfo-to-reports.py
maudetes, line 1: I still have an error after this migration
```
mongoengine.errors.FieldDoesNotExist: The fields "{'spam'}" do not exist on the document "Message"
```

udata/core/reports/models.py
```diff
@@ -90,6 +105,67 @@ class Report(db.Document):
     def self_api_url(self):
         return url_for("api.report", report=self, _external=True)
 
+    def get_target(self):
```
maudetes: Are the following functions still used?

udata/core/spam/models.py
maudetes, line 1: I don't understand if we could have a more straightforward approach now using dedicated reports?
Since we're not mixing the spam info in the model itself, do we still need to have weird signals connect/disconnect.

ThibaudDauce: Removed in 671351581 (move logic from clean to post_save to be similar to Activity)

udata/migrations/2025-11-27-migrate-spaminfo-to-reports.py
ThibaudDauce: Fixed in ca2af3901047970e266a075e4def82e9cd6830a4

udata/core/reports/models.py
ThibaudDauce: Removed in e5ab6d2485277cba4ce1ff608a619e5ede20f94a

PR #3557, opened 2025-12-04
fix(rdf): escape invalid URIRef in RDF graph

maudetes:
Fix https://github.com/datagouv/data.gouv.fr/issues/1437

Fix https://errors.data.gouv.fr/organizations/sentry/issues/273971 and https://errors.data.gouv.fr/organizations/sentry/issues/274240

Some invalid uri could exist in the graph and they can't be serialized in N3/Turtle.
We use urllib.parse.quote to escape these at best for invalid (as in `rdflib.term._is_valid_uri`) URIRef.

Alternative to https://github.com/opendatateam/udata/pull/3548 that focus on graph serialization only and that supports all URIRef.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ parameters:
     default: "3.11"
   publish-branch:
     type: string
-    default: "master"
+    default: "main"
     description: "Branch to publish to PyPi and trigger the Gitlab CI/CD pipeline when pushed to"
   deploy-env:
     type: string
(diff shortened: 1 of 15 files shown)


udata/rdf.py
```diff
@@ -568,6 +570,25 @@ def paginate_catalog(catalog, graph, datasets, _format, rdf_catalog_endpoint, **
...
+    Some invalid uri could exist in the graph and they can't be serialized in N3/Turtle.
+    We use a urllib.parse.quote to escape these at best for invalid URIRef.
+    """
+    escaped_graph = Graph()
+    for s, p, o in graph:
+        try:
+            if isinstance(o, URIRef):
+                if not _is_valid_uri(str(o)):
+                    encoded_uri = quote(str(o), safe=":/?#[]@!$&'()*+,;=")
+                    o = URIRef(encoded_uri)
```
ThibaudDauce: Claude tells me that the URIRef can be in the subject too. He even tells me that `dataset.harvest.uri` is used in the subject of the graph, maybe add a test on `dataset.harvest.uri` to see if he is correct?

maudetes: Indeed, I was thinking user URIs could only be as object but it's not the case since we use the `dataset.harvest.uri` as the dataset node id.
Fix in https://github.com/opendatateam/udata/pull/3557/commits/fc5517c0d4f3176f7fba8dfecb9dc03f65344b35.

ThibaudDauce (approved):

PR #3560, opened 2025-12-05
chore: move pyproject optional-dependencies to dependency-groups

streino:
`optional-dependencies` (PEP 621) are published to pypi as "extras". For dev-only dependencies it is recommended to instead use the more recent `dependency-groups` (PEP 735).

Sources:
- https://pydevtools.com/handbook/explanation/what-are-optional-dependencies-and-dependency-groups/
- https://docs.astral.sh/uv/concepts/projects/dependencies/
- https://github.com/astral-sh/uv/issues/9011

Works locally with `uv sync --all-groups`. Not sure if there's anything else udata-specific I need to check for this PR? @bolinocroustibat

### DIFF
--- a/udata/core/dataset/tasks.py
+++ b/udata/core/dataset/tasks.py
@@ -1,4 +1,5 @@
 import collections
+import gzip
 import os
 from datetime import date, datetime
 from tempfile import NamedTemporaryFile
@@ -173,7 +174,12 @@ def export_csv_for_model(model, dataset, replace: bool = False):
             dataset.save()
         # remove previous catalog if exists and replace is True
         if replace and fs_filename_to_remove:
-            storages.resources.delete(fs_filename_to_remove)
+            try:
+                storages.resources.delete(fs_filename_to_remove)
+            except FileNotFoundError:
+                log.error(
+                    f"File not found while deleting resource #{resource.id} ({fs_filename_to_remove}) in export_csv_for_model cleanup"
+                )
         return resource
     finally:
         csvfile.close()
@@ -217,8 +223,8 @@ def export_csv(self, model=None):
             with storages.resources.open(resource.fs_filename, "rb") as f:
                 store_bytes(
                     bucket=current_app.config["EXPORT_CSV_ARCHIVE_S3_BUCKET"],
-                    filename=f"{current_app.config['EXPORT_CSV_ARCHIVE_S3_FILENAME_PREFIX']}{resource.title}",
-                    bytes=f.read(),
+                    filename=f"{current_app.config['EXPORT_CSV_ARCHIVE_S3_FILENAME_PREFIX']}{resource.title}.gz",
+                    bytes=gzip.compress(f.read()),
                 )
 
 


streino:
> If we use `dependency-groups` instead of `optional-dependencies`, does `uv sync --extra dev` become `uv sync --dev` ? Should we update the documentation about it while we're at it?

According to the [uv doc](https://docs.astral.sh/uv/concepts/projects/dependencies/#development-dependencies), `uv sync` includes `dev` by default:

> The `dev` group is special-cased; there are `--dev`, `--only-dev`, and `--no-dev` flags to toggle inclusion or exclusion of its dependencies. See `--no-default-groups` to disable all default groups instead. Additionally, the `dev` group is [synced by default](https://docs.astral.sh/uv/concepts/projects/dependencies/#default-groups).

If you want to sync other dependency groups you'll have to `uv sync --group xxx` or `uv sync --all-groups`.

We could also make `dev` include all other dev-related dependency groups by default, eg:
```
[dependency-groups]
dev = [
  {include-group = "lint"},
  {include-group = "test"}
]
lint = [
  "ruff"
]
test = [
  "pytest"
]
```
That way everything would be installed locally with a simple `uv sync`.

The CI pipeline is something I can fix or you have to change it on your end? And if that's something I can fix, where do I find it? Same questions about the doc.

bolinocroustibat:
> According to the [uv doc](https://docs.astral.sh/uv/concepts/projects/dependencies/#development-dependencies), `uv sync` includes `dev` by default:
> 
> > The `dev` group is special-cased; there are `--dev`, `--only-dev`, and `--no-dev` flags to toggle inclusion or exclusion of its dependencies. See `--no-default-groups` to disable all default groups instead. Additionally, the `dev` group is [synced by default](https://docs.astral.sh/uv/concepts/projects/dependencies/#default-groups).
> 
> If you want to sync other dependency groups you'll have to `uv sync --group xxx` or `uv sync --all-groups`.
> 
> We could also make `dev` include all other dev-related dependency groups by default, eg:
> 
> ```
> [dependency-groups]
> dev = [
>   {include-group = "lint"},
>   {include-group = "test"}
> ]
> lint = [
>   "ruff"
> ]
> test = [
>   "pytest"
> ]
> ```
> 
> That way everything would be installed locally with a simple `uv sync`.

That could be nice but probably a bit overkill to have several sub-groups for most of our projects - except udata indeed.

> 
> The CI pipeline is something I can fix or you have to change it on your end? And if that's something I can fix, where do I find it? Same questions about the doc.

The CI pipeline can be fixed here, build and publishing is done in this repo's CI (while deploying is done from the published release in another repo). You probably would only have to remove the `--extra dev` in the CI if I understand correctly.

It would be great not install dev dependencies to build the production package by the way (using `uv sync --no-dev`). It seems to be the case now, isn't it? Maybe we could create an issue about that?

For the doc, I have no idea how it works, maybe @ThibaudDauce can help? I'm not sure the doc is up-to-date with the new lock file.

streino:
> That could be nice but probably a bit overkill to have several sub-groups for most of our projects - except udata indeed.

If other projects put everything in `dev` directly then there's nothing else to do. But since udata already has several dev-related groups, it might be better to "fold" them all in `dev` so they get installed by default. It'll save a few keystrokes and occasional head-scratches (I had to open `pyproject.toml` to figure out why my checkout couldn't find `pytest` anymore).

"Folding" means either:

1. Directly moving all the dependencies from `test`, `doc`, ... in the `dev` group:
    ```
    [dependency-groups]
    dev = [
        "flask-shell-ipython >= 0.5.3, < 1.0",
        "invoke >= 2.2.0, < 3.0",
        "pip-tools >= 7.4.1, < 8.0",
        "pre-commit >= 4.2.0, < 5.0",
        "ruff >= 0.11.0, < 1.0",
        # stuff from test
        "faker >= 37.0.2, < 38.0",
        ...
        # stuff from doc
        "mkdocs >= 1.6.1, < 2.0",
        # stuff from report
        ...
    ]
    ```

2. Or, if you want to keep things organized the way they already are, adding `include-group` entries for each of those sub-groups:
    ```
    [dependency-groups]
    dev = [
        "flask-shell-ipython >= 0.5.3, < 1.0",
        "invoke >= 2.2.0, < 3.0",
        ...,
        {include-group = "test"},
        {include-group = "doc"},
        {include-group = "report"},
    ]
    test = [
        "faker >= 37.0.2, < 38.0",
        ...
    ]
    doc = ...
    report = ...
    ```

> The CI pipeline can be fixed here, build and publishing is done in this repo's CI (while deploying is done from the published release in another repo). You probably would only have to remove the `--extra dev` in the CI if I understand correctly.

```
uv sync --frozen --extra dev --extra test
```

becomes either (depending on what we decide about "folding" above):

```
# no folding
uv sync --frozen --group test

# with "folding"
uv sync --frozen
```

_Edit:_ We could even exclude the superfluous groups (but `uv` is so fast it barely matters?):
```
# with "folding"
uv sync --frozen --no-group doc --no-group report
```
 
> It would be great not install dev dependencies to build the production package by the way (using `uv sync --no-dev`). It seems to be the case now, isn't it? Maybe we could create an issue about that?

I'm not sure what the `dist` job does. There's no explicit `uv sync`, so I'm guessing it'll depend on what `uv run inv pydist` does. Unless you have a way to run those locally we're going to have to fix the `python` (test) job for circleci to reach `dist` and show us what's going on.

PR #3565, opened 2025-12-09
chore: remove trailing slash in cdata url

maudetes:
Related to https://github.com/datagouv/data.gouv.fr/issues/1896

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -120,7 +120,7 @@ udata = "udata.tests.plugin"
 [dependency-groups]
 dev = [
     "coverage[toml] >= 7.7.0, < 8.0",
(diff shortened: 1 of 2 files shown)


udata/uris.py
```diff
@@ -91,8 +91,6 @@ def cdata_url(uri: str, **kwargs) -> str | None:
         uri += f"/{append.lstrip('/')}"
 
     url = urljoin(base_url, uri)
-    if not url.endswith("/"):
```
ThibaudDauce, line 94: Do we want to do the inverse? If there is a trailing slash: remove it?

maudetes: We already have `uri = uri.rstrip("/")` just before

PR #3566, opened 2025-12-10
feat: replace `private` by `published_at` for `Dataset`

### LINKED ISSUE #4240: NewDatagouvDatasetsJob : gérer les JDDs publiés en privé initialement
Ce job est en charge d'identifier les JDDs publiés récemment sur data.gouv.fr susceptibles d'avoir leur place sur notre plateforme. Ce job n'identifie pas les JDDs pertinents dans le cas où un JDD est créé en privé puis passé en public.

## Plus de détails

`created_at_internal` [dans l'API](https://www.data.gouv.fr/api/1/datasets/bagnoles-de-lorne/) donne une date technique de création. Dans ce cas le producteur a créé le JDD vendredi matin en privé, lundi on regarde ce qui a été publié vendredi, samedi, dimanche mais le JDD n'était pas remonté car en privé.

Il passe en public courant de journée le lundi mais ne remonte pas dans le job du mardi chez nous car sa date de création est le vendredi.

## Fix

Adapter le code ? Pas certain que l'on puisse. En [échangeant avec data.gouv.fr](https://mattermost.incubateur.net/betagouv/pl/hhf8ggmqt7fx9d6jocmbtewqxo) peut-être avoir une date de publication d'un JDD (qui correspond à la création si JDD public dès le début ou lors du passage de privé à public)

cc @ptitfred qui a travaillé sur ce job il y a peu, pour ta curiosité

ThibaudDauce:
Fix https://github.com/etalab/transport-site/issues/4240

Need to be done for `Reuse` and `Dataservice` too in other PRs. `cdata` is working without any change, but could be updated to use and send the new `published_at` field instead of `private`.

### DIFF
--- a/udata/commands/fixtures.py
+++ b/udata/commands/fixtures.py
@@ -54,6 +54,7 @@ UNWANTED_KEYS: dict[str, list[str]] = {
         "spatial",
         "quality",
         "permissions",
+        "private",  # Computed from published_at, not a real field
     ],
     "resource": ["latest", "preview_url", "last_modified"],
     "organization": ["class", "page", "uri", "logo_thumbnail"],
--- a/udata/core/dataset/api.py
+++ b/udata/core/dataset/api.py
@@ -308,7 +308,9 @@ class DatasetListAPI(API):
     def get(self):
         """List or search all datasets"""
         args = dataset_parser.parse()
-        datasets = Dataset.objects.visible_by_user(current_user)
+        datasets = Dataset.objects.visible_by_user(
+            current_user, Q(published_at__ne=None, archived=None, deleted=None)
+        )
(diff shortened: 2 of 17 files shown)


streino:
How (if) will this be articulated with harvested records? ISO-19139 at least has a "publication date" for the metadata (not processed by udata at this point). I could see `dataset.published_at` being the target for this metadata date if we're ever going to map it, with consequences like:
- `dataset.published_at` could possibly be anterior to  `dataset.created_at`.
- (un)publication on the source catalog would control publication on udata.
- ...?

Not saying this *needs* to be considered at this point nor whether that potential tie is good or bad. Just raising the question, and I can possibly help dig other potential consequences if needed.

ThibaudDauce:
Let's not do that right now.

PR #3567, opened 2025-12-10
fix: activities on private object not visible by owner

ThibaudDauce:
> :wave: je vois que l'API ne renvoie les activités d'un objet private que pour les super-admin https://github.com/opendatateam/udata/blob/5313356a9ad9c94bbc1cb8ae7615cd4f35debd82/udata/core/activity/api.py#L112-L117
>
> Est-ce que ce serait envisageable d'introduire un check de type visible_by_user plutôt ?

:warning: need fixing if merged before or after https://github.com/opendatateam/udata/pull/3566


abulte (approved):

udata/core/owned.py
```diff
@@ -110,6 +111,33 @@ class Owned(object):
...
+        Check if this object is visible to the given user.
+        An object is visible if:
+        - It's not private, OR
+        - The user is a sysadmin, OR
+        - The user owns it or is a member of the owning organization
+
+        Models using this method must have a `private` attribute.
+        """
+        if not getattr(self, "private", False):
+            return True
```
maudetes: I think it may be confusing with the `is_visible` computed properties that exist on multiple objects and that also takes other properties into account (`deleted`, `archived`, etc.), not behaving the same as this.

maudetes (approved):
Thank you for taking a go a it!

I think the existing activity logic filtering quite terrible :cry:, but it should probably be investigated in a different iteration

udata/core/owned.py
ThibaudDauce: What do you think of ea1cf239d?

nicolaskempf57 (approved):

PR #3570, opened 2025-12-11
feat: add MembershipRequest from org

ThibaudDauce:
Close https://github.com/datagouv/data.gouv.fr/issues/1652

### DIFF
--- a/udata/core/organization/mails.py
+++ b/udata/core/organization/mails.py
@@ -191,7 +191,7 @@ def membership_invitation(
     if user_exists:
         url = cdata_url("/admin/me/profile")
     else:
-        url = cdata_url("/register?next=/admin/me/profile")
+        url = cdata_url("/register", next="/admin/me/profile", email=invitation.email)
 
     paragraphs.append(MailCTA(_("View and respond to invitation"), url))
     return MailMessage(


udata/core/organization/tasks.py
```diff
@@ -71,15 +71,18 @@ def notify_membership_response(org_id, request_id):
         mails.membership_refused(org).send(request.user)
```
maudetes, line 71: Following the membership_refused, I get an error
```
  File "udata/udata/mail.py", line 119, in send_mail
    to = recipient if isinstance(recipient, str) else recipient.email
                                                      ^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'email'
```

udata/core/organization/mails.py
```diff
@@ -142,3 +125,23 @@ def badge_added_association(org: Organization) -> MailMessage:
...
+            _(
+                "You have been invited to join the organization %(org)s.",
+                org=org,
+            )
+        ),
+    ]
+    if invitation.comment:
+        paragraphs.append(LabelledContent(_("Message:"), invitation.comment))
+    paragraphs.append(
+        MailCTA(_("View and respond to invitation"), cdata_url("/admin/me/org_invitations"))
```
maudetes: I'm unsure about this URL, it doesn't exist in cdata?

ThibaudDauce: Fixed!

udata/core/organization/tasks.py
ThibaudDauce: It's for an email invitation? Will try this. I have no problem when inviting an existing user…

maudetes: I don't reproduce it anymore

udata/core/organization/mails.py
```diff
@@ -142,3 +125,68 @@ def badge_added_association(org: Organization) -> MailMessage:
...
+        ParagraphWithLinks(
+            _(
+                "You have been invited to join the organization %(org)s.",
+                org=org,
+            )
+        ),
+    ]
+    if invitation.comment:
+        paragraphs.append(LabelledContent(_("Message:"), invitation.comment))
+    paragraphs.append(MailCTA(_("View and respond to invitation"), cdata_url("/admin/me/profile")))
```
maudetes: It could be nice to link to  `/register?next=/admin/me/profile` if the user doesn't already exists?

ThibaudDauce: Done in 1683a1545 (not sure the `next` works in register yet…) I've added `email` to the link too for pre-fill 070e6ff6a

maudetes: True, I don't think it would work since we need confirmation anyway!
When adding email as a param, we should add a rule to ignore this param in matomo (I've added a rule to ignore an `email` param).

ThibaudDauce: Good catch for Matomo! Didn't think of this…

maudetes (approved):

PR #3603, opened 2026-01-06
feat: add post kind of news or page

ThibaudDauce:


udata/core/post/models.py
```diff
@@ -100,6 +100,12 @@ class Post(db.Datetimed, Linkable, db.Document):
         description="HTML or markdown body type",
     )
 
+    kind = field(
```
maudetes, line 103: Do we want to filter on `kind=news` for `api/1/posts/recent.atom` (and `api/1/posts/` ?) ?

ThibaudDauce: Done in 552300029 for RSS. I think for the main API it should be done client side

ThibaudDauce: Done in https://github.com/datagouv/cdata/pull/865

maudetes (approved):

PR #3695, opened 2026-03-17
feat: migrate topics to api fields

ThibaudDauce:
Part of https://github.com/datagouv/data.gouv.fr/issues/1868


abulte:
@ThibaudDauce maybe resolve conflicts before I review? 🙏

ThibaudDauce:
> @ThibaudDauce maybe resolve conflicts before I review? 🙏

Done

udata/core/topic/apiv2.py
```diff
@@ -206,8 +269,8 @@ def delete(self, topic, element_id):
 
     @apiv2.secure
     @apiv2.doc("topic_element_update")
-    @apiv2.expect(element_fields)
-    @apiv2.marshal_with(element_fields)
+    @apiv2.expect(TopicElement.__read_fields__)
```
abulte:
```suggestion
    @apiv2.expect(TopicElement.__write_fields__)
```

udata/core/topic/apiv2.py
```diff
@@ -75,8 +106,28 @@ def get(self):
...
+        if not topic.owner and not topic.organization:
+            topic.owner = current_user._get_current_object()
+
+        topic.save()
+
+        if elements_data is not None:
+            for element_data in elements_data:
+                element = patch(TopicElement(), element_data)
+                element.topic = topic
+                element.save()
```
abulte, line 133: We might want to reuse the validate-then-save pattern https://github.com/opendatateam/udata/pull/3695/changes#diff-479230526cf856f3f5ac586257924b5a455a19fcaea8cdfdae47fa1d7a007d14L139 (here and in Topic update)? This would allow for atomic operation on Topic. I don't think it was possible with the previous Form pattern, and seems doable here.

udata/core/topic/models.py
```diff
@@ -35,9 +53,14 @@ class TopicElement(Auditable, Document):
...
-    # Made optional to allow proper form handling with commit=False
-    topic = field(ReferenceField("Topic", required=False))
+    element = field(
+        GenericReferenceField(choices=["Dataset", "Reuse", "Dataservice"]),
+        nested_fields=api.model_reference,
+        allow_null=True,
+        checks=[check_title_or_element_required],
+    )
+    # Not exposed in the API (not wrapped with field()), only used internally.
+    topic = ReferenceField("Topic", required=False)
```
abulte: `required=False` still needed? Was a hack for Form handling. If still needed, might warrant a comment as to why.

udata/core/topic/apiv2.py
abulte: Ok I've seen your tests of this explicit behaviour. I guess doing it properly now is more complicated than I thought?

abulte (commented):
🙏

abulte:
For the record, e2e tested with:
- local udata and ecologie.data.gouv.fr site: Topic manipulation through frontend
- local udata-front-kit-universe and udata: universe feed

udata/core/topic/apiv2.py
ThibaudDauce: Thanks, I looked into this. Turns out making it properly atomic is trickier than it looks.

A simple `patch-all` then `save-all` split isn't enough: `patch()` only catches the explicit `FieldValidationError`s and malformed reference payloads. Full MongoEngine validation (`required=True`, `GenericReferenceField` `choices`, `clean()`) only runs at `.save()`. So a valid-at-patch payload can still blow up mid-save and leave a partial state.

The natural next step is `doc.validate()` before `doc.save()`, which does run the full validation (and `clean()`). But `save()` internally calls `validate(clean=True)` too, so `clean()` ends up running twice. In our case that's a problem because `Owned.clean()` and `Auditable.clean()` have side effects (mutate `owner`/`organization`, fetch the previous values from DB to drive the activity/change signals). Running them twice means a stale `_previous_owner` and duplicate queries. Using `validate(clean=False)` avoids the double run but silently drops clean-based checks like `SpatialCoverage.clean()` rejecting `zones` + `geom` together.

udata/core/topic/models.py
ThibaudDauce: Removed in [4f2985b](https://github.com/opendatateam/udata/pull/3695/commits/4f2985b9a97db1b5a054b15ff80a0c96861e3b6c)

udata/core/topic/apiv2.py
ThibaudDauce: Done in [4f2985b](https://github.com/opendatateam/udata/pull/3695/commits/4f2985b9a97db1b5a054b15ff80a0c96861e3b6c)

abulte (approved):
Thanks! 😎

PR #3698, opened 2026-03-18
fix: shadowing translation function

ThibaudDauce:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/279254/?environment=data.gouv.fr&project=12&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=1h&stream_index=0

Happening on https://www.data.gouv.fr/tf-validate?next=https://www.data.gouv.fr?flash%3Dconnected but not sure why we don't get a cdata response on this route…

The route is `validate-two-factor` in cdata and in the infra routing…

### DIFF
--- a/udata/templates/macros/forms.html
+++ b/udata/templates/macros/forms.html
@@ -40,7 +40,7 @@
 {% endmacro %}
 
 {% macro radio_inline(field) %}
-{% for value, label, selected, _ in field.iter_choices() %}
+{% for value, label, selected, _attrs in field.iter_choices() %}
 <label class="radio-inline">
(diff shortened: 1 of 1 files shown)


udata/templates/macros/forms.html
maudetes: I think I still have an issue with `_(label)`? Should it be `label`if label is a LazyProxy?

ThibaudDauce: yes done in 00f7ae66c (it was working because non existing strings are returned as is)

maudetes (approved):

PR #3710, opened 2026-03-31
feat: migrate organizations to new API fields

ThibaudDauce:


udata/core/metrics/models.py
```diff
@@ -10,6 +10,7 @@ class WithMetrics(object):
         DictField(),
         readonly=True,
         auditable=False,
+        attribute=lambda o: o.get_metrics(),
```
maudetes, line 13: Why didn't we need that before?

udata/core/user/models.py
```diff
@@ -4,6 +4,7 @@
 from itertools import chain
 
 from blinker import Signal
+from flask import request as flask_request
```
maudetes: I think we're using `request` in many places in the code.
I would rather use the same denomination everywhere, thus not renaming it in this PR?
```suggestion
from flask import request
```

udata/core/organization/models.py
```diff
@@ -103,23 +130,27 @@ class MembershipRequest(EmbeddedDocument):
...
-    kind = StringField(choices=list(REQUEST_TYPES), default="request")
-    email = StringField()  # For inviting non-registered users by email
-    created_by = ReferenceField("User")  # Admin who created the invitation
-    role = StringField(choices=list(ORG_ROLES), default=DEFAULT_ROLE)
+    kind = field(
+        StringField(choices=list(REQUEST_TYPES), default="request"),
+        readonly=True,
+    )
+    email = field(StringField(), readonly=True)
+    created_by = field(ReferenceField("User"), readonly=True)
```
maudetes: If `created_by` is None, (ex for [older request memberships on datagouv orga](http://dev.local:7000/api/1/organizations/646b7187b50b2a93b1ae3d45/?lang=fr)) it fails with a 500.
Same for `handled_by` on a pending membership request ([exemple on Insee](http://dev.local:7000/api/1/organizations/534fff81a3a7292c64a77e5c/?lang=fr))?
Shouldn't we deduce `allow_null` from the inner field `required=True` property?

```suggestion
    created_by = field(ReferenceField("User"), allow_null=True, readonly=True)
```

udata/core/organization/api.py
```diff
@@ -63,6 +54,31 @@
 SUGGEST_SORTING = "-metrics.followers"
 
 
+def resolve_assignment_subjects(raw_assignments, org):
```
maudetes, line 58: Thank you for extracting this part

udata/core/metrics/models.py
ThibaudDauce: It was a bug before, Reuse.metrics returned all the metrics in the database without taken into account `__metrics_keys__`

udata/core/user/models.py
ThibaudDauce: Done in 67446eee7

udata/core/organization/models.py
ThibaudDauce: Done in 67446eee7

maudetes: My point was also that I'm unsure why we need to specify `allow_null` when we can deduce this information with `required=True`?

maudetes (approved):

udata/core/organization/models.py
ThibaudDauce: Good idea, I think they are missing in other fields so I will do another PR to check for this.

PR #3712, opened 2026-03-31
fix: filter with multiple tags

ThibaudDauce:
The changes on the `fix_tags_filters` branch add support for filtering by multiple tags (`?tag=a&tag=b`). Our contribution on top:

**Regression test** (`test_site_csv_exports.py`): switching from `multi_to_dict` to `ReuseSearch.as_request_parser()` for the `/api/1/site/reuses.csv` endpoint silently dropped the `dataset` filter, because `ReuseSearch.filters` doesn't declare it while `ReuseApiParser.parse_filters` supports it.

**Fix** (`site/api.py`): replaced `ReuseApiParser.parse_filters` with `Reuse.apply_sort_filters()` — the same pattern already used by the dataservices CSV endpoint. `apply_sort_filters` discovers its filters automatically from model field definitions, so there's no risk of desynchronization between two separately maintained filter lists. As a bonus, the generic parser always returns a list for tag values (via `action="append"`), avoiding the `multi_to_dict` issue of returning either a string or a list depending on the number of values.

---

### Cleanup to do

This bug reveals a broader issue: two independent filtering systems doing the same thing.

**The generic system** (`api_fields.py` / `@generate_fields`): declares filters once on model fields via `filterable={}`, auto-generates the parser and `apply_sort_filters()`. Used by v2 endpoints and v1 list endpoints.

**The manual `ApiParser`s** (`ReuseApiParser`, `DatasetApiParser`, `DataserviceApiParser`, `OrgApiParser`): redeclare the same filters by hand in `parse_filters` methods with chains of `if args.get("xxx")`. Used by `mongo_search()` (fallback when the search service is down) and CSV exports.

Both systems support the same filters for the same models, but they're not synchronized — adding a filter in one requires remembering to add it in the other. That's the root cause of the bug we just fixed.

**What remains to align:**
- **CSV exports**: `datasets.csv`, `resources.csv` and `organizations.csv` still use `DatasetApiParser.parse_filters` / `OrgApiParser.parse_filters` + `multi_to_dict`. They should use `Model.apply_sort_filters()` like dataservices and reuses do now.
- **`mongo_search()`**: each Search class (`ReuseSearch`, `DataserviceSearch`, `DatasetSearch`, `OrgSearch`) has a `mongo_search()` that calls `ApiParser.parse_filters` + sort + pagination. `apply_sort_filters` already does all of that. `mongo_search` could be rewritten to use it, which would make the `ApiParser`s removable.
- **The `ApiParser`s themselves**: once CSV exports and `mongo_search` are migrated, `ReuseApiParser.parse_filters`, `DataserviceApiParser.parse_filters`, `DatasetApiParser.parse_filters` and `OrgApiParser.parse_filters` become dead code and can be deleted.


ThibaudDauce:
@maudetes Could you link into this?

nicolaskempf57 (approved):

PR #3716, opened 2026-04-08
fix: orphan users in reports

ThibaudDauce:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/274123/?environment=demo.data.gouv.fr&project=12&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=1h&stream_index=1 but not sure how it could have happened… We should never delete users?

Last page of moderation is crashing in demo


ThibaudDauce:
Fixed directly in database

PR #3717, opened 2026-04-08
feat: add spam detection to all objects

ThibaudDauce:


udata/core/user/models.py
```diff
@@ -147,6 +148,9 @@ class User(WithMetrics, UserMixin, Linkable, Document):
 
     verbose_name = _("account")
 
+    def fields_to_check_for_spam(self):
+        return {"about": self.about, "website": self.website}
```
maudetes, line 196: Maybe add first and last name as well?

I think we could want to have custom detection for users.
For example, having a website in `self.about` field could raise a report.

maudetes (commented):
I think we may want to add a logic to exclude contents published by certified organizations.

udata/core/user/models.py
ThibaudDauce: Could we add this in another PR?

ThibaudDauce:
> I think we may want to add a logic to exclude contents published by certified organizations.

Done in [d18abd0](https://github.com/opendatateam/udata/pull/3717/commits/d18abd0a88cca9ce29d970fad5a32c4663a501c3)

udata/core/dataservices/models.py
```diff
@@ -212,6 +214,12 @@ class Dataservice(
 
     verbose_name = _("dataservice")
 
+    def fields_to_check_for_spam(self):
+        return {"title": self.title, "description": self.description}
+
+    def spam_is_whitelisted(self) -> bool:
+        return self.organization and self.organization.certified
```
maudetes, line 221: Could we imagine having `spam_is_whitelisted` on `Owned` class?

maudetes (approved):

udata/core/dataservices/models.py
ThibaudDauce: We could but inheritance is tricky with Owned and SpamMixim… Not sure how to do that. We could extract this to a function on Owned?

maudetes: I think it's about import order, if you import Owned before SpamMixin, it would work right

ThibaudDauce: Not sure we want to depend from Spam in Owned… It's seems a little bit weird…

maudetes: Hmm I meant 
```
class Dataset(
    Auditable,
    WithMetrics,
    WithAccessType,
    DatasetBadgeMixin,
    Owned,
    SpamMixin
    Linkable,
    Document[DatasetQuerySet],
):
```

ThibaudDauce: Yes but we need to have a `spam_is_whitelisted` inside the `Owned` class so `Owned` is "aware" of the `SpamMixin` interface…

PR #3720, opened 2026-04-13
fix(harvest): missing new (Geo)DCAT-AP mapping for dct:provenance and dct:*rights

### LINKED ISSUE #1018: Régression csw-dcat + GeoDCAT-AP sur provenance et rights
Contexte : https://github.com/ecolabdata/ecospheres/issues/958#issuecomment-4225641249

> | Métadonnée | DCAT (prod) | | GeoDCAT-AP (demo) | | Notes |
> |-|-|-|-|-|-|
> | Fiche datagouv | [*lien*](https://www.data.gouv.fr/datasets/qualite-physicochimie-des-cours-deau) | | [*lien*](https://demo.data.gouv.fr/datasets/qualite-physicochimie-des-cours-deau) | | |
> | `extra.provenance` | `{"provenance": ["Données issues ..."]}` | 🟢 | `{"provenance": [ "Nf0fd7e6c1b7e4b898233dc6f6eb7d28e"]}` | 🔴 | Nœud anonyme non géré |

streino:
Fix https://github.com/ecolabdata/ecospheres/issues/1018

`dct:provenance` and `dct:*rights` used to contain a `rdfs:label`, but the GeoDCAT-AP 3 spec mentions that `dct:description` can be used ([doc provenance](https://semiceu.github.io/GeoDCAT-AP/releases/3.1.0/#lineage---lineage), [doc rights](https://semiceu.github.io/GeoDCAT-AP/releases/3.1.0/#conditions-for-access-and-use-and-limitations-on-public-access-use-limitation-and-access-other-constraints)), and the [SEMIC (main branch)](https://github.com/SEMICeu/iso-19139-to-dcat-ap/blob/main/iso-19139-to-dcat-ap.xsl#L825) and [GeoNetwork](https://github.com/geonetwork/core-geonetwork/blob/2f2f187039ba07a66a14446f4ff1eed075125dae/schemas/iso19115-3.2018/src/main/plugin/iso19115-3.2018/formatter/eu-dcat-ap/eu-dcat-ap-core.xsl#L149) converters now use that spec.

It is not clear where DCAT stands, and whether GeoDCAT-AP defines `dct:description` as a replacement or an addition to `rdfs:label`. Properties of `dct:*Statements` are not clearly defined so we can only rely on examples provided in the spec.

So we add resolution of `dct:description` but keep `rdfs:label` as well. If an element contains both properties, we take the `rdfs:label` as the more likely "title" of the element.

Even though the recent specs only give URI examples for licenses, we also apply the logic above to `dct:license` to preserve the already implemented coverage of free-text cases.

### DIFF
--- a/udata/rdf.py
+++ b/udata/rdf.py
@@ -229,37 +229,39 @@ CONTEXT = {
 }
 
 
-def serialize_value(value, parse_label=False):
+def serialize_value(value, unwrap: list[URIRef] | None = None):
     """
     If the value is a URIRef or a Literal, return it as a string.
     If the value is a RdfResource:
-        - Return the label of the RdfResource if any and `parse_label`,
-        - or the identifier of the RdfResource.
+        - if `unwrap` is set, look for children of the RdfResource in the order they are listed in
+          `unwrap`, and return the value of the first matching element (if any),
+        - otherwise return the identifier of the RdfResource.
     """
     if isinstance(value, (URIRef, Literal)):
         return value.toPython()
     elif isinstance(value, RdfResource):
-        if parse_label and (rdfs_label := rdf_value(value, RDFS.label)):
-            return rdfs_label
+        for uriref in unwrap or []:
+            if val := rdf_value(value, uriref):
+                return val
         return value.identifier.toPython()
 
 
-def rdf_unique_values(resource, predicate, parse_label=False) -> set[str]:
+def rdf_unique_values(resource, predicate, unwrap: list[URIRef] | None = None) -> set[str]:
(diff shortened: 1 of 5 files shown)


udata/rdf.py
```diff
@@ -229,37 +229,39 @@ def want_rdf():
 }
 
 
-def serialize_value(value, parse_label=False):
+def serialize_value(value, unwrap: list[URIRef] | None = None):
```
maudetes, line 232: I find the approach clean, nice 🎉

maudetes (approved):

PR #3724, opened 2026-04-14
fix(harvest): invalid remote URL on record identifiers with "urn:uuid" codespace

### LINKED ISSUE #1019: URL source invalide lorsque l'identifiant MD contient un codespace
Contexte : https://github.com/ecolabdata/ecospheres/issues/958#issuecomment-4225641249

> | Métadonnée | DCAT (prod) | | GeoDCAT-AP (demo) | | Notes |
> |-|-|-|-|-|-|
> | Fiche datagouv | [*lien*](https://www.data.gouv.fr/datasets/qualite-physicochimie-des-cours-deau) | | [*lien*](https://demo.data.gouv.fr/datasets/qualite-physicochimie-des-cours-deau) | | |
> | `harvest.remote_url` | *absent* | 🔴 | https://data.ofb.fr/catalogue/data-eaufrance/fre/catalog.search#/metadata/urn:uuid:0689e153-fe7b-44f4-a599-0f7f60918e43 | 🟠 | URL invalide causé par ajout du codespace |

L'URL valide ne contient pas le codespace : https://data.ofb.fr/catalogue/data-eaufrance/fre/catalog.search#/metadata/0689e153-fe7b-44f4-a599-0f7f60918e43

XML :
```xml
<mdb:metadataIdentifier>
  <mcc:MD_Identifier>
    <mcc:code>
      <gco:CharacterString>0689e153-fe7b-44f4-a599-0f7f60918e43</gco:CharacterString>
    </mcc:code>
    <mcc:codeSpace>
      <gco:CharacterString>urn:uuid</gco:CharacterString>
    </mcc:codeSpace>
  </mcc:MD_Identifier>
</mdb:metadataIdentifier>
```

streino:
Fix https://github.com/ecolabdata/ecospheres/issues/1019

### DIFF
--- a/udata/rdf.py
+++ b/udata/rdf.py
@@ -500,10 +500,34 @@ def remote_url_from_rdf(rdf: RdfResource, graph: Graph, remote_url_prefix: str |
     Otherwise, use DCAT.landingPage if found and uri validation succeeds.
     In this latter case, use RDF identifier as fallback if uri validation succeeds.
     """
-    if remote_url_prefix and (
-        primary_topic_identifier := primary_topic_identifier_from_rdf(graph, rdf)
-    ):
-        return f"{remote_url_prefix.rstrip('/')}/{primary_topic_identifier}"
+    if remote_url_prefix and (identifier := primary_topic_identifier_from_rdf(graph, rdf)):
+        # Some ISO-19115-3 GeoNetwork templates set a codespace for the record IDs (e.g. the
+        # geodata-multilingual.xml template,), i.e. they set:
+        # - mdb:metadataIdentifier/mcc:MD_Identifier/mcc:code to the record ID, and
+        # - mdb:metadataIdentifier/mcc:MD_Identifier/mcc:codeSpace to "urn:uuid".
+        # It is not clear why only some templates set a codespace, git blame didn't yield any clue.
+        #
+        # That codespace ends up collapsed in the record dct:identifier, since dct:identifier has no
+        # separate codespace. This means dct:identifier differs (by the codespace) from the record
+        # ID that GeoNetwork uses internally, and we need to extract that original record ID from
+        # dct:identifier to build a valid remote_url.
+        #
+        # It is impractical to reliably infer the codespace from dct:identifier since parsing of
+        # URNs is namespace dependent. So we restrict the heuristic to the "uuid" namespace, which
+        # is both clearly specified (RFC-4122) and used in the wild by GeoNetwork (e.g. OFB,
+        # Sextant, and likely most future ISO-19115-3 catalogs given it's in the templates).
+        #
+        # Note that the test on datatype being xsd:anyURI is important, because some ISO-19139
+        # catalogs (e.g. Geo2France) are using "urn:...:" prefixes in their gmd:fileIdentifier.
+        # ISO-19139 having no concept of codespace (for fileIdentifier), that prefix is treated as
+        # as a regular part of the gco:CharacterString record ID, meaning:
+        # - dct:identifier is typed xsd:string, and
+        # - the prefix is part of the record URL, so it must stay in remote_url.
+        uuid_prefix = "urn:uuid:"
+        if identifier.datatype == XSD.anyURI and identifier.value.startswith(uuid_prefix):
+            offset = len(uuid_prefix)
+            identifier = identifier.value[offset:]
+        return f"{remote_url_prefix.rstrip('/')}/{identifier}"
 
     landing_page = url_from_rdf(rdf, DCAT.landingPage)
     uri = rdf.identifier.toPython()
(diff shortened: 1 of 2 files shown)


streino:
@maudetes I'd like your input before polishing the PR :pray: 

Doesn't feel too good to handle a corner case, but I don't see any practical/robust way to handle all URN namespaces. I'd say "urn:uuid" is standard enough to warrant an extra `if`?

Note that there's nothing GeoNetwork specific here: other software, or even GeoNetwork users designing their own record templates, can legitimately use codespaces in ISO-19115-3, and since IDs are often UUIDs, that makes "urn:uuid" one of the most likely candidates.

maudetes:
Thank you for the study.
I'm wondering if we can't recommend to not use REMOTE_PREFIX_URL and use either a dcat:landingPage or dataset identifier instead in this case?

If we don't have a proper URI, then we'll need to go with this PR indeed

streino:
I took another look at the suggested alternatives.

**`dcat:landingPage`**

Given the current state things, I don't see how we could use `dcat:landingPage`, for the following reasons:

- The [current definition for `dcat:landingPage` in GeoDCAT-AP](https://semiceu.github.io/GeoDCAT-AP/releases/3.1.0/#resource-locator---on-line-resource), clearly states that any resource with unknown `gmd:function` should be a `dcat:landingPage`, meaning we'll get a bunch of them in ISO records converted to DCAT.
  

- There is nothing in INSPIRE that suggests putting the record URL in a distribution. So we would have no spec support pushing this to the catalogs.

- Having the first resource listed on every record page pointing to themselves would look awkward in the source catalogs. So we'd make the source catalogs look worse (and all downstream uses of the metadata). Hard to sell to the catalog managers. Also feels like an even more arbitrary hack than what we're doing with the source prefix.

This is still moving grounds though, both on the spec side (SEMIC pilot is supposed to have another look at it) and possibly on the GeoNetwork side (since they're not currently aligned with the spec). So it might become possible at some point in the future, but for now I wouldn't go that way.

**Resource identifier**

Two reasons this might be tricky, but not impossible:

- The INSPIRE TG only *recommends* (not even *strongly recommends*) making the resource identifier resolvable in [TG Recommendation 1.3: metadata/2.0/rec/datasets-and-series/resolvable-uid](https://github.com/INSPIRE-MIF/technical-guidelines/blob/main/metadata/metadata-iso19139/metadata-iso19139.adoc#3121-unique-resource-identifier):

  > It is recommended to make the unique resource identifier resolvable into a document providing information about described data set or data sets series. In the case of HTTP URIs[[25](https://github.com/INSPIRE-MIF/technical-guidelines/blob/main/metadata/metadata-iso19139/metadata-iso19139.adoc#_footnotedef_25)], the public DNS and the usual HTTP URI resolving mechanisms should be used for implementing this resolvability. For other types of URIs, a resolving service should be provided implementing similar functionality.
  >
  > This document retrieved as the result of the resolving process may be, but is not required to be, the metadata description of the described resource. It is also recommended that viewing of the provided document does not require user authentication, authorisation or specialised viewing tools beyond a typical web browser.

  That second paragraph even weakens the recommendation a bit more, not requiring the URL to be the metadata URL in case the URI is resolvable.

- [Resource identifier cardinality is `1..n`](https://github.com/INSPIRE-MIF/technical-guidelines/blob/main/metadata/metadata-iso19139/metadata-iso19139.adoc#c25-unique-resource-identifier). This means we'd need to enforce some logic to be able to pick the right one as the source URL. Although this might also be an advantage because catalogs wouldn't have to replace their IDs, only add new ones.

We might actually be able to make a recommendation to set/add the metadata source URL as a resource identifier. Ideally we'd want to push this at the CNIG or even INSPIRE TG guidelines, but we could start with our own [ISO-DCAT recommendation](https://github.com/ecolabdata/ecospheres/wiki/Recommandations-ISO-DCAT). 

If we want to go that way:
- I'll have to check the impact of having multiple resource identifiers in GeoNetwork and SEMIC converter.
- I'd still go with the present PR as a temporary workaround, until we get the catalogs to update their identifier strategies. I'm worried this is going to take a long time...

streino:
> * I'll have to check the impact of having multiple resource identifiers in GeoNetwork and SEMIC converter.

SEMIC: OK. All identifiers end up as `dct:identifier`, preserving the original order.

GeoNetwork: Doable but clumsy:
- Generation: GeoNetwork "generate ID" helper button will create non-resolvable URIs. So it'll require explicit action from the record editor.
- Edition: Must use the "INSPIRE" or "Complete" views to be able to add another ID. The "Simple" view doesn't let you add extra identifiers fields.
- Viewing: All IDs are showns in a comma-separated list under "Identificateur de ressource unique". Singular might trip users seeing more than one ID?

streino:
Conclusion from my two previous comments : 

I don't see how we could go with `dcat:landingPage` at this point, and I'd be careful impacting the ressource identifier.

It still might be worth aiming for one of these solutions, but we're talking long term: waiting for GeoDCAT-AP to reconsider their mapping of `landingPage`, or design a practical recommendation for resolvable identifiers. 

So short term I'd go with the (simple) patch in this PR, so we have the option to use csw-dcat+GeoDCAT-AP on GeoNetwork ISO-19115-3 catalogs and still have working source URLs.

maudetes (approved):
Thank you for the patch, as well as the investigation on alternatives ✨ 

I'll let you relevant issues if we want to one day remove this patch.

PR #3726, opened 2026-04-16
feat: expose callbacks_count for reports

ThibaudDauce:
Require to add a warning on the moderation dashboard if a report is blocking callbacks. And we don't want to expose internals (even for super-admins)

### DIFF
--- a/udata/api_fields.py
+++ b/udata/api_fields.py
@@ -26,6 +26,7 @@ For field-specific metadata, see the `field()` function documentation.
 """
 
 import functools
+import typing
 from datetime import datetime
 from typing import Any, Callable, Iterable, TypedDict, TypeVar, Unpack, overload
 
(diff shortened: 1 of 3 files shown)


maudetes:
> Require to add a warning on the moderation dashboard if a report is blocking callbacks. And we don't want to expose internals (even for super-admins)

I'm not sure we should expose the entire args and kwargs payload, but having the callback name could be useful, no?

ThibaudDauce:
> > Require to add a warning on the moderation dashboard if a report is blocking callbacks. And we don't want to expose internals (even for super-admins)
> 
> I'm not sure we should expose the entire args and kwargs payload, but having the callback name could be useful, no?

Not sure, it's technical and I don't think it adds a lot of information (I was even going for a boolean at first…).

ThibaudDauce:
The names are `signal_new`, `signal_close` and `signal_comment`, so not really user friendly?

maudetes:
> The names are `signal_new`, `signal_close` and `signal_comment`, so not really user friendly?

I think as a super-user I would prefer having this information available in the api, but it doesn't matter much.

maudetes (approved):

PR #3733, opened 2026-04-20
feat: better user commands output

maudetes:


udata/core/user/commands.py
```diff
@@ -105,17 +107,21 @@ def rotate_password(email):
...
     """
     user: User = datastore.find_user(email=email)
+    if not user:
+        exit_with_error(f"User {email} does not exist")
     user.password_rotation_demanded = datetime.now(UTC)
     user.save()
     # Reset ongoing sessions by uniquifier
     datastore.set_uniquifier(user)
-    success(f"Password rotated for user {email}")
+    success(f"Password rotated for user {email}. Last login at: {user.current_login_at}")
```
ThibaudDauce: NIT: maybe check for `None`? 

```suggestion
    success(f"Password rotated for user {email}. Last login at: {user.current_login_at or 'never'}")
```

ThibaudDauce (approved):
You could do `set_admin()` too?

PR #3734, opened 2026-04-20
feat: add harvest config in csv export

maudetes:


udata/harvest/tests/test_csv_adapter.py
```diff
@@ -0,0 +1,30 @@
+import json
+
+from udata.models import Dataset  # noqa
```
ThibaudDauce, line 3: Why do you need to import Dataset?

Pierlou (approved):
Looks good to me, we don't have anything touchy in the config?

ThibaudDauce (approved):
Do we have non-JSON-natively-serializable (ex : datetime, ObjectId) in configs to test for the `default=str` in json.dumps?

udata/harvest/tests/test_csv_adapter.py
maudetes: I added it for circular import error, but doesn't seem needed anymore. Removed in https://github.com/opendatateam/udata/pull/3734/commits/39279168502ef9d0e0730f094c0ea19141fba3ca

PR #3735, opened 2026-04-21
Attach a DOI to a dataset

Pierlou:
See https://github.com/datagouv/data.gouv.fr/issues/1983

Mints a DOI for a dataset, stores it, and pushes the metadata back to DataCite when the title or the organization changes.

The two open questions are settled:
- the DOI is a readonly `doi` field on the dataset, not an extra
- no retraction: a dataset carrying a DOI can no longer be deleted, only archived (in the API and in `purge_datasets`)

Minting is `POST /datasets/{id}/doi`, sysadmin only. The citation block and the "request a DOI" button are not here.

Checked against the sandbox instead of guessed: `PUT /dois/{doi}` upserts, so creation is idempotent and the 422 special case is gone, and a PUT without `event` keeps the DOI findable. The recorded URL is the permalink, not the slug, which follows the title.

New settings to set on deploy:
- `DOI_PREFIX`
- `DOI_REPO_USER`
- `DOI_REPO_PASSWORD`
- `DOI_PLATFORM_URI`, `https://api.test.datacite.org` outside production


geoffreyaldebert:
👏  pour moi pas de rétractation

PR #3738, opened 2026-04-23
feat(search): configurable facets size

abulte:
Related https://github.com/datagouv/data.gouv.fr/issues/1432 
Related https://github.com/datagouv/cdata/pull/1054

This lets the consumer use a query parameter `?facet_size__{facet_name}={value}` to request more bucket values for a facet from ElasticSearch.

There's a configurable `MAX_FACET_SIZE` that defaults to 500. Increasing the facet size leads to more resource usage in ES, this a defensive value that needs to be discussed.

Not all facets can be configured, it's in a `configurable_size_facets[]` list on the adapter. Only `organization_id_with_name` is in there for now. Listing the facts was needed to generate a proper query arg parser, and I took the opportunity to limit the facets that can be configured. We could include more facets by default. We could also refactor to auto-list all the supported facets, but it's a bigger rework ⬇️.

⬆️ This rework is proposed in https://github.com/ecolabdata/udata/pull/10, I can merge that one into this one if it makes sense.

For ecologie.data.gouv.fr, this would let use `facet_size__organization_id_with_name=200` for the filter from https://github.com/datagouv/cdata/pull/1054 to have all the values from our universe. It would also need to be wired up in cdata's `GlobalSearch`.

Reference: https://www.elastic.co/docs/reference/aggregations/search-aggregations-bucket-terms-aggregation#search-aggregations-bucket-terms-aggregation-size

### DIFF
--- a/udata/core/dataservices/search.py
+++ b/udata/core/dataservices/search.py
@@ -83,6 +83,7 @@ class DataserviceSearch(ModelSearchAdapter):
     model = Dataservice
     service_class = DataserviceService
     consumer_class = DataserviceConsumer
+    configurable_size_facets = ["organization_id_with_name"]
 
     sorts = {"created": "created_at", "views": "views", "followers": "followers"}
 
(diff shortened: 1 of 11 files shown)


abulte:
@ThibaudDauce can't assign you as reviewer from a fork... Can you have a look? 🙏

abulte:
Idea: use SOLR DSL for url queries https://solr.apache.org/guide/solr/latest/query-guide/faceting.html (thanks @streino)

PR #3739, opened 2026-04-23
chore: add dependabot config for weekly updates

maudetes:
For now, we only have security updates ([example](https://github.com/opendatateam/udata/pull/3736)) with renovate.

Adds a weekly dependabot update for deps.

I don't have strong opinion on the two tools, but if we should use one over the other, I'm all ears.


maudetes:
Closing in favor of [renovate updates](https://github.com/opendatateam/udata/pull/3740).

PR #3750, opened 2026-04-29
refactor: ping on tchap

Pierlou:


udata/notifications/tchap.py
```diff
@@ -0,0 +1,50 @@
...
+
+    r = requests.post(
+        url,
+        headers={
+            "content-type": "application/json",
+            "authorization": f"Bearer {token}",
+        },
+        json={
+            "msgtype": "m.text",
+            "body": "_",
```
ThibaudDauce, line 44: Could we add the markdown as body too? Claude tells me it's used in push notification (maybe?)

ThibaudDauce (approved):

udata/notifications/tchap.py
Pierlou: We could, my understanding is that if there is a `formatted_body`, it supplants the `body`, and the notification part is managed by `m.mentions` (which here always pings the whole room, but we can have [something more custom](https://github.com/datagouv/datagouvfr_data_pipelines/blob/abb683db67e1e2867ba03ec6cfc52b1a9c5c2193/utils/tchap.py#L47) if needed). The current layout of messages sent from Airflow looks good to me, I am not sure what having a `body` would improve

ThibaudDauce: Clause says it's for the mobile push notification. But we can improve that later if we encounter a `_` somewhere

PR #3753, opened 2026-05-04
feat: new tabular explore edito bloc

ThibaudDauce:
Waiting to code the cdata part.

And we're not sure we'll use this for ouverture.data.gouv.fr, so it's not anymore a priority for the moment.


nicolaskempf57 (approved):

udata/core/edito_blocs/models.py
```diff
@@ -109,6 +115,13 @@ class MarkdownBloc(Bloc):
     )
 
 
+@generate_fields()
+class ExploreBloc(Bloc):
+    title = field(StringField())
+    subtitle = field(StringField())
+    resource_id = field(UUIDField(required=True, binary=False))
```
maudetes, line 122: We'll need to add a purge task in case or resource disparition?

PR #3759, opened 2026-05-06
feat: harvest API refactor

ThibaudDauce:
- Require https://github.com/datagouv/cdata/pull/1074

## Summary

- Migrate harvest API models to the model-driven `@generate_fields` system (replaces manual flask-restx models for `HarvestError`, `HarvestLog`, `HarvestItem`, `HarvestSourceValidation`, `HarvestJob`, `HarvestSource`)
- Replace `HarvestSourceValidationForm` (WTForms) with `patch()` + `required_if(state=VALIDATION_REFUSED)` on the validation endpoint
- Extract `source_permissions_fields` into `udata/harvest/api_fields.py` to break a circular import between `models.py` and `api.py`

## New endpoint

- Add `GET /harvest/job/<ident>/items/` as a paginated subresource for harvest job items
- On `GET /harvest/job/<ident>/`, `items` is now exposed as a link object `{rel, href, type, total, by_status, by_type}` with counters per status and per kind (dataset / dataservice), instead of inlining the full list

## API Fields

- Add an `href_extra` parameter on `field()` to merge extra keys (e.g. counters) into the `href` payload generated for list-style fields
- Validate `StringField` `choices` inside `patch()` so invalid enum values raise 400 even on embedded documents that are patched but never saved (with a comment documenting the duplication of MongoEngine's logic and the underlying `clean()` mixing pure validation with stateful side-effects — to be removed once those are split)

## Bug fixes

- Fix pagination URLs (`next_page` / `previous_page`) on `/harvest/job/<id>/items/` — handler used to return a plain dict, now returns `DBPaginator(Pagination(...))`
- Fix security leak: `HarvestError.details` was exposed to everyone because the lambda checked truthiness of the `admin_permission` object instead of `.can()`
- Restore `@cached_property` on `HarvestSource.last_job` (was accidentally demoted to `@property`, causing one Mongo query per access)
- Reject unknown `state` values on the validation endpoint (previously silently routed to `reject_source`)

## Breaking API change

- `HarvestJob.source` is now serialized as a `{id, class}` reference object (auto-generated from `HarvestSource.__ref_fields__`) instead of a bare ID string. To document in the release notes.

## Cleanups

- Drop redundant `job_items_page_fields` model — use the auto-generated `HarvestItem.__page_fields__`
- Replace duplicated `jobs_parser` / `items_parser` with a shared `api.page_parser()`
- Remove redundant inline import of `admin_permission` and a redundant `# Details are reserved to super-admins` comment


udata/harvest/api.py
```diff
@@ -3,173 +3,74 @@
...
-            allow_null=True,
-            readonly=True,
-            description="Who performed the validation",
-        ),
-        "on": fields.ISODateTime(
-            readonly=True, description="Date date on which validation was performed"
-        ),
-        "comment": fields.String(
-            description="A comment about the validation. Required on rejection"
+            preview_dataservice_fields, description="The processed dataset", allow_null=True
```
nicolaskempf57:
```suggestion
            preview_dataservice_fields, description="The processed dataservice", allow_null=True
```

udata/harvest/api.py
```diff
@@ -219,67 +120,27 @@
...
 source_parser.add_argument(
     "owner", type=str, location="args", help="The organization or user ID to filter on"
 )
 source_parser.add_argument(
-    "deleted", type=bool, location="args", default=False, help="Include sources flaggued as deleted"
+    "deleted",
+    type=bool,
+    location="args",
+    default=False,
+    help="Include sources flaggued as deleted",
```
nicolaskempf57:
```suggestion
    help="Include sources flagged as deleted",
```

nicolaskempf57 (approved):

ThibaudDauce:
@abulte do you know if some people use this API? Are you ok with the breaking changes?

abulte:
@ThibaudDauce I don't know anybody that does but let's still ask @streino, he's the most likely to those days. Maybe drop a message to transport.data.gouv.fr to be sure?

streino:
I'm not aware of anyone using that API either.

udata/harvest/tests/test_api.py
```diff
@@ -463,6 +464,30 @@ def test_reject_source(self):
         assert source.validation.comment == "Not valid"
         assert source.validation.by == user
 
+    def test_validate_source_rejects_unknown_state(self):
```
maudetes, line 476: Thank you for all the new tests, it helps to get an overview of the changes 👌

udata/harvest/api.py
```diff
@@ -3,173 +3,74 @@
...
     HarvestSource,
+    HarvestSourceValidation,
 )
 
 ns = api.namespace("harvest", "Harvest related operations")
 
 
-error_fields = api.model(
-    "HarvestError",
+# Manual preview-specific models: previews don't have a stable ID/URL since
```
maudetes, line 26: Should we add a dummy ID in any case for preview items?

It would lead to a 404 if users followed links but it would make sense since we're in a preview?
We wouldn't need a dedicated logic?

udata/harvest/api.py
```diff
@@ -289,6 +150,9 @@ def get(self):
         if not args["deleted"]:
             sources = sources.visible()
 
+        # `owner` is overloaded here: it matches sources owned by the user
+        # OR by the organization, so we cannot reuse the auto-generated
+        # `owner` filter from the Owned mixin (which only matches `owner=`).
```
maudetes: Since we're breaking the API, maybe we could align this filter with other owned models (adding an `organization` filter)?

udata/harvest/models.py
```diff
@@ -181,34 +277,54 @@ def schedule(self):
...
+        description="The job initialization errors",
+    )
+    # Items are exposed as a paginated subresource link with status/type counters,
+    # to avoid loading thousands of dereferenced items for every job in a list.
+    items = field(
+        ListField(EmbeddedDocumentField(HarvestItem)),
+        readonly=True,
+        href=lambda o: url_for("api.harvest_job_items", ident=o.id),
+        href_extra=lambda o: {
+            "by_status": {
```
maudetes: Could we store these as "metrics" in the job document during finalization?
It would prevent going through all items at query time for each status and also be more udata-like?

udata/harvest/models.py
maudetes, line 1: I had trouble reproducing the poor performances locally.
I deployed this branch on dev and tested performances on [this endpoint](https://dev.data.gouv.fr/api/1/harvest/source/66632580d7365f18b2c81902/jobs/?page=1&page_size=20&lang=fr), even though they vary greatly.

Perfs went from ~11s to ~7s with this PR 🎉

Seeing the actually small payload we return on this route, we could improve perfs even more, but directly when loading the document from mongo.
* Running `HarvestJob.objects(source="66632580d7365f18b2c81902").paginate(1, 20)` on dev takes already ~7.1 seconds. Adding an `.exclude("data")` goes down to ~2.8 seconds.
* Turning datasets `ReferenceField` to `LazyReferenceField`, we are now at  ~1 second. We could also exclude `items.dataset` in specific context if switching to `LazyReferenceField` degrades performances elsewhere.

Command to run:
```
import timeit
timeit.timeit('HarvestJob.objects(source="66632580d7365f18b2c81902").paginate(1, 20)', number=10, setup="from udata.harvest.models import HarvestJob")
```

udata/harvest/models.py
```diff
@@ -181,34 +277,54 @@ def schedule(self):
...
+        readonly=True,
+        description="The job status",
+    )
+    errors = field(
+        ListField(EmbeddedDocumentField(HarvestError)),
+        readonly=True,
+        description="The job initialization errors",
+    )
+    # Items are exposed as a paginated subresource link with status/type counters,
+    # to avoid loading thousands of dereferenced items for every job in a list.
```
maudetes: In my understanding, items would load reference at the moment. We should either use `LazyReferenceField` or exclude the field or use `no_dereference()` if we don't want to?

udata/harvest/api.py
ThibaudDauce: Done in [1d03f96](https://github.com/opendatateam/udata/pull/3759/commits/1d03f966ea9e496a28ea49a2de88f5ae2f80060a)

udata/harvest/models.py
ThibaudDauce: Done in [1d03f96](https://github.com/opendatateam/udata/pull/3759/commits/1d03f966ea9e496a28ea49a2de88f5ae2f80060a)

udata/harvest/models.py
ThibaudDauce: `LazyReferenceField` raises new issues, but switching to `metrics` style is working

udata/harvest/models.py
ThibaudDauce: `LazyReferenceField` raises new issues on other endpoints… :-( Not sure why…

udata/harvest/api.py
ThibaudDauce: I liked the idea to having an empty URL if the dataset doesn't really exists but as you wish…

udata/harvest/models.py
maudetes: How about adding `exclude("data")` in the routes that don't need it?

ThibaudDauce: Done in 138ca93a8

maudetes: Ah sorry, I hadn't seen you had already excluded `data` as well in https://github.com/opendatateam/udata/pull/3759/commits/1d03f966ea9e496a28ea49a2de88f5ae2f80060a.

maudetes (approved):
Looks good 👌

As discussed, we could:
- compute metrics as the job run (and not only in the finalize step)
- add a migration to compute previous metrics (may be too expensive)

PR #3765, opened 2026-05-11
feat: allow external URL and name for discussions

### LINKED ISSUE #263: Notification à l'ouverture d'une discussion sur une collection
Lorsque un user commente un bouquet, l'owner du bouquet ne reçoit pas de notification.
Est de possible, à l'image d'un commentaire sur un jdd sur dgfr, que le owner d'un bouquet  reçoive une notification par mail quant à la publication d'un commentaire sur un bouquet. SDans le cas où le bouquet est owned par une orga, alors tous les users de l'orga reçoivent la notif
lorsque un nouveau commentaire est fait dans la discussion, l'ensemble des personnes ayant aussi commenté recoivent une notification

ThibaudDauce:
- Replace https://github.com/opendatateam/udata/pull/3072
- Fix https://github.com/ecolabdata/ecospheres/issues/263
- Require https://github.com/datagouv/cdata/pull/1110

### DIFF
--- a/udata/core/discussions/constants.py
+++ b/udata/core/discussions/constants.py
@@ -2,7 +2,13 @@ from udata.core.dataservices.models import Dataservice
 from udata.core.dataset.models import Dataset
 from udata.core.post.models import Post
 from udata.core.reuse.models import Reuse
+from udata.core.topic.models import Topic
 
 COMMENT_SIZE_LIMIT = 50000
 
-NOTIFY_DISCUSSION_SUBJECTS = (Dataset, Reuse, Post, Dataservice)
+NOTIFY_DISCUSSION_SUBJECTS = (Dataset, Reuse, Post, Dataservice, Topic)
+
+# Subjects with no canonical detail page on the udata frontend. For these,
+# notifications are only sent when `extras.notification.external_url` is
+# provided (e.g. a Topic shown on a third-party platform like Ecosphères).
+NOTIFY_REQUIRES_EXTERNAL_URL = (Topic,)
--- a/udata/core/discussions/constants.py
+++ b/udata/core/discussions/constants.py
@@ -7,8 +7,3 @@ from udata.core.topic.models import Topic
 COMMENT_SIZE_LIMIT = 50000
 
 NOTIFY_DISCUSSION_SUBJECTS = (Dataset, Reuse, Post, Dataservice, Topic)
-
-# Subjects with no canonical detail page on the udata frontend. For these,
-# notifications are only sent when `extras.notification.external_url` is
-# provided (e.g. a Topic shown on a third-party platform like Ecosphères).
-NOTIFY_REQUIRES_EXTERNAL_URL = (Topic,)
--- a/udata/core/discussions/models.py
+++ b/udata/core/discussions/models.py
@@ -1,9 +1,12 @@
+import fnmatch
 import logging
 from datetime import UTC, datetime
+from urllib.parse import urlparse
 
-from flask import url_for
+from flask import current_app, url_for
 from flask_login import current_user
 from mongoengine import EmbeddedDocument
+from mongoengine.errors import ValidationError
 from mongoengine.fields import (
     DateTimeField,
     EmbeddedDocumentField,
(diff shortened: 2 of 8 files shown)


udata/core/discussions/models.py
```diff
@@ -90,6 +93,73 @@ def spam_report_message(self, breadcrumb):
         return message
 
 
+def is_valid_notification_external_url(url) -> bool:
```
abulte, line 97: Could this be handled by a specialised lib? I'm always afraid of missing something when validating those manually.

abulte (commented):
Great feature 😎

udata/core/discussions/models.py
ThibaudDauce: The closest fit is **`ada-url`** (Python bindings to the Ada parser used by Node.js/Cloudflare) — WHATWG-conformant, so URLs are parsed the way browsers and mail clients resolve them. The `\@` case I handled manually wouldn't even exist: `ada-url` would already give the final hostname the click resolves to.

For the allow-list itself there's no real library — Django has `url_has_allowed_host_and_scheme` for the safe-redirect use case, but it's coupled to Django and doesn't support wildcards.

I'd rather keep `urllib` for now: the documented attack vectors are tested, and pulling in a C++ dependency for one validator feels disproportionate. Happy to switch to `ada-url` if we end up validating URLs in more places.

udata/core/discussions/models.py
```diff
@@ -189,6 +259,48 @@ def self_web_url(self, **kwargs):
...
+
+        Returns the value of `extras.notification.model_name` when it passes
+        the per-class allow-list check, otherwise the subject's `verbose_name`.
+        """
+        meta_name = self._notification_meta().get("model_name")
+        if is_valid_notification_model_name(self.subject, meta_name):
+            return meta_name
+        return self.subject.verbose_name
+
+    def notification_url(self):
```
abulte, line 239: Could be a property, like `notification_subject_type`

abulte (approved):
Just a minor comment, looks good.

Although it's not what I had done in https://github.com/opendatateam/udata/pull/3072, I now wonder if the external notification URL should be handled by the subject object (eg in Topic.extras)? It would make it simpler for the Topic use case, but it would be less generic.

ThibaudDauce:
> Just a minor comment, looks good.
> 
> Although it's not what I had done in #3072, I now wonder if the external notification URL should be handled by the subject object (eg in Topic.extras)? It would make it simpler for the Topic use case, but it would be less generic.

Waiting the @maudetes opinion on this :-)

ThibaudDauce:
> Just a minor comment, looks good.
> 
> Although it's not what I had done in #3072, I now wonder if the external notification URL should be handled by the subject object (eg in Topic.extras)? It would make it simpler for the Topic use case, but it would be less generic.

After a little though, having it on the subject will broke the transport feature. Since a transport dataset could be discussed on data.gouv and on transport.data.gouv (it only works on topics since they are only accessible from ecologie?)

abulte:
> > Just a minor comment, looks good.
> > Although it's not what I had done in #3072, I now wonder if the external notification URL should be handled by the subject object (eg in Topic.extras)? It would make it simpler for the Topic use case, but it would be less generic.
> 
> After a little though, having it on the subject will broke the transport feature. Since a transport dataset could be discussed on data.gouv and on transport.data.gouv (it only works on topics since they are only accessible from ecologie?)

Yeah, that's what I had in mind behind "less generic", it would only work for Topics.

udata/settings.py
```diff
@@ -508,6 +508,21 @@ class Defaults(object):
     # Notifications are deleted after being handled for 90 days
     DAYS_AFTER_NOTIFICATION_EXPIRED = 90
 
+    # Discussion settings
+    ###########################################################################
+    # Allow-list of domains accepted as `extras.notification.external_url`
+    # on a discussion. Supports fnmatch wildcards (e.g. `*.data.gouv.fr`).
+    # The accept-list prevents arbitrary external links from being injected
+    # in notification emails.
+    DISCUSSION_ALLOWED_EXTERNAL_DOMAINS = ["*.data.gouv.fr"]
```
maudetes: I would keep it empty as a default value in settings.py
```suggestion
    DISCUSSION_ALLOWED_EXTERNAL_DOMAINS = []
```

udata/core/discussions/constants.py
```diff
@@ -2,7 +2,13 @@
...
 
 COMMENT_SIZE_LIMIT = 50000
 
-NOTIFY_DISCUSSION_SUBJECTS = (Dataset, Reuse, Post, Dataservice)
+NOTIFY_DISCUSSION_SUBJECTS = (Dataset, Reuse, Post, Dataservice, Topic)
+
+# Subjects with no canonical detail page on the udata frontend. For these,
+# notifications are only sent when `extras.notification.external_url` is
+# provided (e.g. a Topic shown on a third-party platform like Ecosphères).
+NOTIFY_REQUIRES_EXTERNAL_URL = (Topic,)
```
maudetes: How complicated would it be to have a minimal topic page on cdata?

I think it would prevent some safeguards that we now have to implement for almost non-existant cases (since someone posted a discussion, the object must have an existing external url).

udata/core/topic/models.py
```diff
@@ -98,6 +98,8 @@ def post_delete(cls, sender, document, **kwargs):
 
 @generate_fields()
 class Topic(Datetimed, Auditable, Linkable, Document[OwnedQuerySet], Owned):
+    verbose_name = _("topic")
```
maudetes: I think we could translate topic as `collection` (the wording used by ecologie.data.gouv.fr) in udata.
In my understanding, it would be generic enough to be used in different use case.

For example, we could have `collections thématiques` in [our own admin frontend](https://www.data.gouv.fr/admin/site/topics).

We'll probably need to rediscuss this with the product later on, but it seems like a safe bet for a first iteration.

maudetes (commented):
Could we split this PR with the `external_url` first and a second one for the model name?

I'm not as convinced about the model name override yet (should it be at the discussion level? what about the locale translations?) and we may want to wait for a product feedback on this before moving on. It would also remove the need for additional `DiscussionCreateForm.validate`, etc.

In my opinin, the main blocking point are the `external_url` as well as the activation of notifications on topic discussions.

udata/settings.py
ThibaudDauce: Done in [11ca285](https://github.com/opendatateam/udata/pull/3765/commits/11ca285ffbf487c51c3cbd8eecfc61eee00c8326)

udata/core/topic/models.py
ThibaudDauce: Done in [11ca285](https://github.com/opendatateam/udata/pull/3765/commits/11ca285ffbf487c51c3cbd8eecfc61eee00c8326)

udata/core/discussions/constants.py
ThibaudDauce: Done in [11ca285](https://github.com/opendatateam/udata/pull/3765/commits/11ca285ffbf487c51c3cbd8eecfc61eee00c8326)

udata/core/discussions/models.py
maudetes: Don't we want to use `uris.validate` that is used elsewhere in the code?

udata/core/discussions/models.py
```diff
@@ -189,6 +242,32 @@ def self_web_url(self, **kwargs):
...
+        Returns the allow-listed `extras.notification.external_url` when present
+        and valid (e.g. a Topic displayed on a third-party platform such as
+        Ecosphères), otherwise the subject's canonical page via `url_for()`.
+
+        Defense in depth: re-validate the stored `external_url` at read-time.
+        Data written before the validator existed, raw DB writes (imports,
+        migrations), or a config change tightening the allow-list could all
+        leave an unsafe URL in storage.
+        """
+        meta_url = self._notification_meta().get("external_url")
```
maudetes: I'm unsure about the need for a `_notification_meta` wrapper? How about
```
notification_meta = (self.extras or {}).get("notification") or {}
meta_url = notification_meta.get("external_url")
```

udata/core/discussions/models.py
```diff
@@ -189,6 +242,32 @@ def self_web_url(self, **kwargs):
...
+        if is_valid_notification_external_url(meta_url):
+            return f"{meta_url}#discussion-{self.id}"
```
maudetes, line 256: This is a cdata & udata-front-kit pattern. We could imagine allowing a template to format automatically? Not mandatory at all.

maudetes (approved):
🎉

udata/core/discussions/models.py
ThibaudDauce: Done in ce6ba91c8

udata/core/discussions/models.py
ThibaudDauce: We could improve this as needed yes!

PR #3771, opened 2026-05-15
feat(harvest): report conflicting record ownership

### LINKED ISSUE #1017: Détecter les interférences entre moissonneurs
Exemple de la fiche https://www.data.gouv.fr/datasets/ventes-de-pesticides-par-departement/ :
- La fiche appartient à l'organisation [Système d'Information sur l'Eau](https://www.data.gouv.fr/organizations/systeme-d-information-sur-l-eau), dont l'unique [Moissonneur SI eau](https://www.data.gouv.fr/admin/harvesters/66632580d7365f18b2c81902/configuration) est configuré en `csw-dcat`.
- Pourtant, les informations de moissonnage de la fiche indiquent qu'elle provient du [Moissonneur OFB](https://www.data.gouv.fr/admin/harvesters/684af36aeb4fcb5ce4f438bc/configuration) configuré en `csw-iso-19139`.

Il serait utile de détecter et signaler ces incohérences pour les corriger plus efficacement.

streino:
Fix https://github.com/ecolabdata/ecospheres/issues/1017

### DIFF
--- a/udata/harvest/backends/base.py
+++ b/udata/harvest/backends/base.py
@@ -478,6 +478,7 @@ class BaseBackend(object):
             ).first()
 
         if dataset:
+            self.ensure_unique_ownership(dataset)
             return dataset
 
         if self.source.organization:
@@ -502,6 +503,7 @@ class BaseBackend(object):
         ).first()
 
         if dataservice:
+            self.ensure_unique_ownership(dataservice)
             return dataservice
 
         if self.source.organization:
@@ -510,6 +512,19 @@ class BaseBackend(object):
             return Dataservice(owner=self.source.owner)
 
         return Dataservice()
+
+    def ensure_unique_ownership(self, item):
+        other_owner = None
+        if item.organization:
+            if item.organization != self.source.organization:
+                other_owner = item.organization
+        elif item.owner:
+            if item.owner != self.source.owner:
+                other_owner = item.owner
+        if other_owner:
+            raise HarvestValidationError(
+                f"Item has another owner: {other_owner.page() or other_owner.id}"
+            )
 
     def validate(self, data, schema):
         """Perform a data validation against a given schema.
--- a/udata/harvest/backends/base.py
+++ b/udata/harvest/backends/base.py
@@ -514,17 +514,26 @@ class BaseBackend(object):
         return Dataservice()
 
     def ensure_unique_ownership(self, item):
+        """Raise if item already belongs to some other owner.
+
+        Ressources (datasets, services, ...) must have universally unique
+        identifiers, but some catalogs fail to enforce it. Cases seen
+        in the wild:
+        - Copy-pasting record metadata without changing the identifier.
+        - Using the table name of the originating data as identifier, and
+          generating several datasets out of the same table.
+        - "TODO", "A REMPLIR", etc. in the identifier field.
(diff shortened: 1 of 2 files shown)


udata/harvest/tests/test_base_backend.py
```diff
@@ -516,6 +517,84 @@ def test_duplicate_remote_ids(self):
...
+            )
+        )
+        job2 = backend2.harvest()
+        assert job2.status == "done-errors"
+        assert len(job2.items) == 2
+        for item in job2.items:
+            if not item.remote_id.endswith("-repeat"):
+                assert item.status == "done"
+            else:
+                print(item.remote_id)
```
ThibaudDauce:
```suggestion
```

udata/harvest/backends/base.py
```diff
@@ -511,6 +513,19 @@ def get_dataservice(self, remote_id):
 
         return Dataservice()
 
+    def ensure_unique_ownership(self, item):
```
ThibaudDauce, line 516: Could you add a comment to explain a little bit why? Could we have a type on `item` (not sure it's possible)

udata/harvest/backends/base.py
```diff
@@ -511,6 +513,19 @@ def get_dataservice(self, remote_id):
...
+        other_owner = None
+        if item.organization:
+            if item.organization != self.source.organization:
+                other_owner = item.organization
+        elif item.owner:
+            if item.owner != self.source.owner:
+                other_owner = item.owner
```
ThibaudDauce: Maybe better?

```suggestion
        if item.organization and item.organization != self.source.organization:
            other_owner = item.organization
        elif item.owner and item.owner != self.source.owner:
            other_owner = item.owner
        else:
            return
```

udata/harvest/backends/base.py
```diff
@@ -511,6 +513,19 @@ def get_dataservice(self, remote_id):
 
         return Dataservice()
 
+    def ensure_unique_ownership(self, item):
+        other_owner = None
+        if item.organization:
+            if item.organization != self.source.organization:
+                other_owner = item.organization
+        elif item.owner:
```
ThibaudDauce: Doesn't seems ta have a test for owner conflict? (only org)

ThibaudDauce (approved):

udata/harvest/backends/base.py
streino: I went back and forth between the two versions. Behavior isn't exactly the same but both work here, so let's go with the version you suggest, it does look slightly better.

ThibaudDauce: What's the difference? With the `return` in the `else` I don't see it…

streino: In the suggested version we'll evaluate `item.owner and ...` if `item.organization and item.organization == self.source.organization`. Given the org/owner constraints (can't have both) this will behave the same way, but the other version is stricter in its test semantics.

streino: The stricter semantics isn't necessarily better btw. If we had org+owner, we'd probably want to detect conflict in both, but would have to change the logic anyway, i.e. not `elif` and probably append to `other_owner`. Anyway, we'd have to update either version :shrug:

PR #3773, opened 2026-05-18
feat(harvest): harvest spatial resolution

### LINKED ISSUE #1075: Moissonner la résolution spatiale
État des lieux : https://github.com/ecolabdata/ecospheres/issues/954

streino:
Fix https://github.com/ecolabdata/ecospheres/issues/1075


streino:
@maudetes Two questions on this draft, about how far I should go: 
- Went with the minimal support for INSPIRE datasets but there are other representations of spatial resolution in GeoDCAT-AP. Do you want to go with the full coverage while I'm at it?
- Storing a basic string representation in `extra["spatial_resolution"]` for ecologie.datagouv. Do you instead want to update the udata model (we'll want a typed object instead of the string repr in that case)?

streino:
Going with the minimal implementation. We can expand later if necessary.

udata/core/dataset/rdf.py
```diff
@@ -596,6 +606,69 @@ def spatial_from_rdf(graph):
...
+        for value in resource.objects(DCAT.spatialResolutionInMeters)
+        if isinstance(value, Literal)
+    ]:
+        resolution = max(resolutions_in_meters)
+        return f"{resolution} {DistanceUom.METER.symbol}"
+
+    if resolutions_as_distance := [
+        (value.toPython(), unit)
+        for obj in resource.objects(DQV.hasQualityMeasurement)
+        if obj.value(DQV.isMeasurementOf).identifier == GEODCAT.spatialResolutionAsDistance
```
ThibaudDauce: `.value(DQV.isMeasurementOf).identifier` could crash if no predicate? It will crash the all dataset, maybe isolate the `spatial_resolution_from_rdf` to avoid discarding a dataset for a spatial problem?

udata/core/dataset/rdf.py
```diff
@@ -596,6 +606,69 @@ def spatial_from_rdf(graph):
...
+        resolution = max(resolutions_as_distance, key=lambda x: x[1].in_meters(x[0]))
+        return f"{resolution[0]} {resolution[1].symbol}"
+
+    if resolutions_as_scale := [
+        value.toPython()
+        for obj in resource.objects(DQV.hasQualityMeasurement)
+        if obj.value(DQV.isMeasurementOf).identifier == GEODCAT.spatialResolutionAsScale
+        and isinstance(value := obj.value(DQV.value), Literal)
+    ]:
+        resolution = max(resolutions_as_scale)
```
ThibaudDauce: Is it a `min` for the fraction scale `1/25000` is more precise than `1/50000`?

udata/core/dataset/rdf.py
```diff
@@ -596,6 +606,69 @@ def spatial_from_rdf(graph):
...
+        return f"{resolution} {DistanceUom.METER.symbol}"
+
+    if resolutions_as_distance := [
+        (value.toPython(), unit)
+        for obj in resource.objects(DQV.hasQualityMeasurement)
+        if obj.value(DQV.isMeasurementOf).identifier == GEODCAT.spatialResolutionAsDistance
+        and (unit := QUDT_TO_UDATA.get(obj.value(SDMXA.unitMeasure).identifier))
+        and isinstance(value := obj.value(DQV.value), Literal)
+    ]:
+        resolution = max(resolutions_as_distance, key=lambda x: x[1].in_meters(x[0]))
```
ThibaudDauce: `float * Decimal` raise a `TypeError` if distance is a `xsd:double`?

ThibaudDauce (changes requested):
Thanks! Could you add tests for all these cases to be sure it's working as intended?

streino:
> Thanks! Could you add tests for all these cases to be sure it's working as intended?

Do you mean tests for non-implemented cases? 

Otherwise all implemented cases should be tested:

```
# ISO-19139 (INSPIRE) supports the following two representations:
# - Resolution distance [0..n]: `//gmd:MD_Resolution/gmd:distance`.
# - Equivalent scale [0..n]: `//gmd:MD_Resolution/gmd:equivalentScale`.
# 
# GeoDCAT-AP supports the above two:
# - Distance [0..n]: `dcat:spatialResolutionInMeters`.
# - Scale [0..n]: `dqv:hasQualityMeasurement` with `dqv:isMeasurementOf` set to
#   `geodcatap:spatialResolutionAsScale`.
```

Distance (in meters) => `test_parse_spatial_resolution_in_meters`
Scale => `test_parse_spatial_resolution_as_scale`

```
# plus:
# - Other distance units [0..n]: `dqv:isMeasurementOf` set to either:
#   - `geodcatap:spatialResolutionAsDistance` (configurable UOM, somewhat redundant with
#     `dcat:spatialResolutionInMeters`),
```

Distance (in other UOM) => `test_parse_spatial_resolution_as_distance`

```
#   - `geodcatap:spatialResolutionAsAngularDistance` (angles not in ISO-19139),
```

Not implemented => not tested

```
#   - `geodcatap:spatialResolutionAsVerticalDistance` (?).
```

Not implemented => not tested

```
# - Free text [0..1]: `geodcatap:spatialResolutionAsText`.
```

Not implemented => not tested

```
# We implement the following minimalist heuristic:
# - Only support INSPIRE representations (no practical examples for others so far).
```

Cf list above

```
# - Return an acceptable textual representation instead of a typed object (which would require
#   storing a type+value tuple).
```

All tests exercise the output representation

```
# - If several representations are present, we pick distance over scale (arbitrary).
```

=> `test_parse_spatial_resolution_conflicting`

```
# - If cardinality > 1, we use the largest value we find. We don't go for the smallest to avoid
#   giving a false sense of precision.
```

=> `*_interval` tests

udata/core/dataset/rdf.py
streino: `resolutions_as_scale` contains decimals, which are then converted to `Fraction`. 
`1/25000` (`scale` = 0.00004) is less precise than `1/50000` (`scale` = 0.00002), so we want `max(scale)` to get the lowest precision.

streino:
> > Thanks! Could you add tests for all these cases to be sure it's working as intended?
> 
> Do you mean tests for non-implemented cases?

Wait, you probably meant the cases you commented on. I'll add those.

udata/core/dataset/rdf.py
streino: Tested in `test_parse_spatial_resolution_as_scale_interval`

udata/core/dataset/rdf.py
streino: Tested in `test_parse_spatial_resolution_*_missing_*`

udata/core/dataset/rdf.py
streino: Tested in `test_parse_spatial_resolution_*_datatype`

udata/core/dataset/rdf.py
```diff
@@ -596,6 +606,76 @@ def spatial_from_rdf(graph):
...
+        ]:
+            resolution = max(resolutions_as_scale)
+            # str(Fraction) should look ok as we're only dealing with geographical scales (1/x with
+            # a fairly limited set of "round" denominator values), which hopefully survive the
+            # decimal-to-fraction conversion reasonably well.
+            # max_denominator is picked to be larger than existing scales, but still small enough
+            # that we limit the risk of non-Decimal floats causing crazy fractions:
+            # Fraction(0.00004) would otherwise return 5902958103587057/147573952589676412928
+            # instead of 1/25000.
+            return str(Fraction(resolution).limit_denominator(max_denominator=100_000_000))
```
ThibaudDauce, line 673: From Claude, but it seems to be real (he tried a test with an absurd billion digit integer and it hanged…)

> **DoS: an attacker-controlled literal can hang the harvest worker.**
> `Fraction(Decimal(...))` builds the *exact* ratio, so it materializes `10**exponent`. A 13-char literal like `1E-1000000000` (xsd:decimal) forces a billion-digit integer → minutes of CPU, hundreds of MB. `max_denominator` doesn't help: it runs *after* the `Fraction` is built, and the `try/except` doesn't either since the cost is paid during the computation. We should reject absurd exponents before converting:
>
> ```suggestion
>             resolution = max(resolutions_as_scale)
>             if isinstance(resolution, Decimal) and abs(resolution.adjusted()) > 10:
>                 return None
> ```

udata/core/dataset/rdf.py
ThibaudDauce: Hum… Claude tells me and I think he's right? Except if we don't talk about the same thing…

- A scale of 1:25,000 = 1 cm → 250 m: more detail → more precise.
- A scale of 1:50,000 = 1 cm → 500 m: less detail → less precise.

streino: Woah right, and I actually checked this but mixed it up. Anyway, will fix obviously, sorry :/

ThibaudDauce: Haha I read it 3 times to be sure too :-D

udata/core/dataset/rdf.py
streino: Good catch! But `rdflib` itself is vulnerable to the same attack :grimacing:  `Literal(x, datatype=XSD.decimal)` will call `Decimal(x)` in its constructor, so we're too late here.

If we want to prevent this attack, we need to either sanitize the input passed to `rdflib.Graph.parse` or monkey-patch `Literal` (likely `rdflib.term._toPythonMapping`) to inject a safe `Decimal` constructor. 
Sanitizing over multiple input formats doesn't seem realistic, so we're left with monkey-patching :grimacing:

In any case, I suggest we scope this out of the current PR.

ThibaudDauce (approved):

PR #3777, opened 2026-05-19
feat: migrate users to api fields

ThibaudDauce:
### Breaking changes restants (intrinsèques à la migration form → patch)

- **Format des erreurs de validation** : `errors["field"] = "message"` (string) au lieu de `["message"]` (liste). Hérité du système `FieldValidationError`/`mongoengine.ValidationError` — identique aux PR organizations, posts, etc.
- Aucun test existant ne teste ce format, donc invisible côté tests.


udata/core/user/models.py
```diff
@@ -101,25 +115,37 @@ class User(SpamMixin, WithMetrics, UserMixin, Linkable, Document):
     slug = field(
         SlugField(max_length=255, required=True, populate_from="fullname"),
         auditable=False,
+        readonly=True,
         show_as_ref=True,
     )
     email = field(
         StringField(max_length=255, required=True, unique=True),
-        show_as_ref=True,
         attribute=_visible_email,
```
nicolaskempf57: Do we want to add `if current_user_is_admin_or_self() else None` to _visible_email ? It was previously done in api_fields so now we will always at least return an obfuscated email

ThibaudDauce: Done in 73be3fc8c

nicolaskempf57 (approved):

PR #3779, opened 2026-05-20
fix: harvest previews bugs

ThibaudDauce:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/292144 by creating a local `ObjectId` in preview


maudetes (approved):
Thank you for the fix and tests 👌

PR #3781, opened 2026-05-20
feat: improve cors

### LINKED ISSUE #2017: Nettoyage des CORS
Mettre à jour la politique de CORS.
Voir [discussions sur tchap](https://matrix.to/#/!FMXpGlfBgjvMumrnPQ:agent.dinum.tchap.gouv.fr/$DWmeLQN5OecfDFMm0x8vl_YxoPIREBFxat0jnVUy1ZA?via=agent.dinum.tchap.gouv.fr).

ThibaudDauce:
## Harden CORS and session cookie configuration

Fix https://github.com/datagouv/data.gouv.fr/issues/2017

### Context

The CORS layer reflected **any** request `Origin` back in `Access-Control-Allow-Origin` while also sending `Access-Control-Allow-Credentials: true`. Combined with the API accepting the Flask-Login session cookie (not just Bearer tokens), this is a configuration we'd rather not rely on.

In practice this was **not exploitable**: the session cookie behaves as `SameSite=Lax` in current browsers, so it's never sent on cross-site subrequests — Chrome and Firefox both block the credentialed cross-origin request before it reaches the API. The change below is **defense in depth**, so the protection no longer depends solely on browser cookie behaviour.

### Changes

- **`cors.py`**: credentials are now only granted to **trusted origins**. The trusted set is derived from `CDATA_BASE_URL` (so the existing front-end keeps working out of the box) plus an optional `CORS_ALLOWED_ORIGINS` list. Any other origin still gets anonymous cross-origin access via `Access-Control-Allow-Origin: *` (the public API stays readable by third parties) but never receives `Allow-Credentials`. The origin/credentials logic, previously duplicated between the actual-request and preflight handlers, is now factored into `set_allow_origin_headers()`.
- **`settings.py`**: `SESSION_COOKIE_SAMESITE` set to `"Lax"` (was `None`, i.e. no attribute emitted — relying on the browser default). This is safe for data.gouv since cdata shares udata's registrable domain (same-site requests still carry the cookie); a comment documents how to relax it to `None` if a deployment puts the front-end on a different registrable domain. Added `CORS_ALLOWED_ORIGINS = []`.

### Tests

Added 3 cases in `test_cors.py`: an untrusted origin never receives credentials (only `*`), a trusted origin does (both via `CDATA_BASE_URL` and via `CORS_ALLOWED_ORIGINS`), and the same on preflight requests. Existing CORS tests still pass (7 passed).

### Not in scope

Requiring an explicit Bearer token instead of the session cookie for write operations would be a stronger model but is an architectural change that would break cdata's cookie-based auth — left for a separate discussion.

### DIFF
--- a/udata/settings.py
+++ b/udata/settings.py
@@ -75,7 +75,18 @@ class Defaults(object):
     # Flask security settings
 
     SESSION_COOKIE_SECURE = True
-    SESSION_COOKIE_SAMESITE = None  # Can be set to 'Lax' or 'Strict'. See https://flask.palletsprojects.com/en/2.3.x/security/#security-cookie
+    # 'Lax' prevents the session cookie from being sent on cross-site subrequests
+    # (the CSRF/CORS exfiltration vector), while still flowing on same-site
+    # requests from the front-end (cdata shares udata's registrable domain) and
+    # on top-level navigations. Can be tightened to 'Strict' or, if the front-end
+    # lives on a different registrable domain, set to 'None' (requires Secure).
+    # See https://flask.palletsprojects.com/en/2.3.x/security/#security-cookie
+    SESSION_COOKIE_SAMESITE = "Lax"
+
+    # Origins allowed to make *credentialed* cross-origin requests to the API.
+    # Other origins still get anonymous (`Access-Control-Allow-Origin: *`) access.
+    # The `CDATA_BASE_URL` origin is always allowed (see `udata.cors`).
+    CORS_ALLOWED_ORIGINS = []
     SECURITY_USE_REGISTER_V2 = True
 
     # Flask-Security-Too settings
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,5 @@
 # Changelog
 
-## Current (in progress)
-
-- fix(security): only allow credentialed CORS requests from trusted origins and set the session cookie to `SameSite=Lax`
-
 ## 16.5.0 (2026-05-18)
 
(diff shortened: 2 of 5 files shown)


udata/settings.py
```diff
@@ -75,7 +75,18 @@ class Defaults(object):
...
+    # requests from the front-end (cdata shares udata's registrable domain) and
+    # on top-level navigations. Can be tightened to 'Strict' or, if the front-end
+    # lives on a different registrable domain, set to 'None' (requires Secure).
+    # See https://flask.palletsprojects.com/en/2.3.x/security/#security-cookie
+    SESSION_COOKIE_SAMESITE = "Lax"
+
+    # Origins allowed to make *credentialed* cross-origin requests to the API.
+    # Other origins still get anonymous (`Access-Control-Allow-Origin: *`) access.
+    # The `CDATA_BASE_URL` origin is always allowed (see `udata.cors`).
+    CORS_ALLOWED_ORIGINS = []
```
maudetes, line 89: In my understanding, in our case, we won't need additional CORS_ALLOWED_ORIGINS, right?
The other sites that may want to connect with datagouv api use OAuth2 with `Authorization: Bearer token`.

maudetes (approved):

udata/settings.py
ThibaudDauce: Yes that's right, it's to be more generic for users, but we could also remove this? `CDATA_BASE_URL` is enought for us, even if we separate the API from the frontend in the future

maudetes: Yep! It's okay to keep it, I just wanted to make sure I understood it clearly

PR #3782, opened 2026-05-20
fix(harvest): normalize tag values in TagListField to match harvester input

Samuelfaure:
## What

  A single dcat:keyword that's empty, punctuation-only, too short, or too long
  crashes the whole document save during harvest:

  ValidationError (Dataset:None) (Tag "df" must be between 3 and 96 characters
  long.: ['tags'])

  The harvest aborts the record, losing the dataset and all its resources because of
   one bad tag.

  ## Why this happens

  The 3–96 char tag rule is enforced at three layers, with three different
  behaviors:

  | Layer | Behavior on bad tag |
  |---|---|
  | WTForms UI (`TagField.pre_validate`) | Per-field error shown to the user |
  | CKAN harvester intake (`normalize_tag` in the schema) | **Silently normalized**
  — slug + drop short + truncate long |
  | MongoDB field `validate()` (defensive backstop) | Raises — aborts the whole
  document save |

  CKAN harvests pre-normalize their tags, so the backstop never fires in practice.
  **DCAT (and other harvesters) never had an equivalent intake-normalization step**,
   so dirty keywords reach the backstop and crash the record.

  Separately, `TagListField.clean` does run on the deserialization path (`to_python`
   / `to_mongo`), but **attribute assignment** (which is exactly what every
  harvester does: `dataset.tags = themes_from_rdf(d)`) bypasses it. So validate sees
   the raw, un-normalized list.

  ## What the fix does

  - `TagListField.clean` now applies `udata.tags.normalize` (slug + drop short +
  truncate long) instead of a bare `slugify`, matching what the CKAN intake schema
  already does.
  - `TagListField.validate` calls `clean()` first, so the validate path is
  consistent with the persistence path regardless of how the value was set.

  Net effect: every harvester (DCAT, MAAF, CKAN, future ones) and every assignment
  path silently normalizes tags the same way. The Mongo-level error becomes a true
  backstop that shouldn't fire in normal use — which is what it always seemed to
  have been intended as.

### DIFF
--- a/udata/mongo/taglist_field.py
+++ b/udata/mongo/taglist_field.py
@@ -1,5 +1,4 @@
 from mongoengine.fields import ListField, StringField
-from slugify import slugify
 
 from udata import tags
 from udata.i18n import lazy_gettext as _
@@ -20,7 +19,7 @@ class TagListField(ListField):
             return []
 
     def clean(self, value):
-        return sorted(list(set([slugify(v, to_lower=True) for v in value])))
+        return sorted({n for n in (tags.normalize(v) for v in value) if n})
 
     def to_python(self, value):
         return super(TagListField, self).to_python(self.clean(value))
@@ -29,6 +28,8 @@ class TagListField(ListField):
         return super(TagListField, self).to_mongo(self.clean(value))
 
     def validate(self, values):
+        if values is not None:
+            values[:] = self.clean(values)
         super(TagListField, self).validate(values)
 
         for tag in values:
--- a/udata/tests/dataset/test_dataset_rdf.py
+++ b/udata/tests/dataset/test_dataset_rdf.py
@@ -809,6 +809,28 @@ class RdfToDatasetTest(PytestOnlyDBTestCase):
         assert isinstance(dataset, Dataset)
         assert "literal-tag" in dataset.tags
 
+    @pytest.mark.options(TAG_MIN_LENGTH=3, TAG_MAX_LENGTH=10)
+    def test_dirty_keyword_does_not_abort_validation(self, app):
+        """Regression test: a dirty dcat:keyword (punctuation-only, too short or
+        too long) must be normalized away instead of aborting the whole document.
+
+        Harvesters set keywords by attribute (`dataset.tags = themes_from_rdf(d)`),
+        which bypasses TagListField.clean(). Before the fix, a single bad keyword
+        raised at the Mongo-level validator, losing the dataset and its resources.
+        """
+        node = BNode()
+        g = Graph()
+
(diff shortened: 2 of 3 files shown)


Samuelfaure:
## Test plan

  - [x] New `TagListFieldCleanTest` covers: empty/punctuation/too-short drops,
  too-long truncation, valid preservation, and the harvester flow (attribute
  assignment + `validate()`)
  - [x] All existing tag and api-fields tests still pass

udata/mongo/taglist_field.py
```diff
@@ -20,7 +19,7 @@ def from_input(input):
             return []
 
     def clean(self, value):
-        return sorted(list(set([slugify(v, to_lower=True) for v in value])))
+        return sorted({n for n in (tags.normalize(v) for v in value) if n})
```
Samuelfaure, line 22: Before:
clean() only lowercased + slugified. A tag like "df" (already a valid slug) would survive unchanged, then crash the Mongo-level validator's 3 ≤ len ≤ 96 rule. A
  keyword like "&" would slugify to "" and crash differently. A 200-char keyword would also crash.

 After: 
  - tags.normalize (udata/tags.py:13) is the project's canonical "make this value the validator will accept": slug + drop too-short + truncate too-long. The PR's
  argument hinges on the fact that the CKAN harvester already routes tags through this exact function on intake (udata/harvest/backends/ckan/schemas/ckan.py:37). Putting it
  in the field generalizes that pattern to every harvester and every assignment path.
  - The if n filter drops the empty-string outputs that normalize produces for too-short inputs, so the post-clean list is guaranteed to satisfy the validator. set keeps
  dedup, sorted keeps deterministic order.

udata/mongo/taglist_field.py
```diff
@@ -29,6 +28,8 @@ def to_mongo(self, value):
         return super(TagListField, self).to_mongo(self.clean(value))
 
     def validate(self, values):
+        if values is not None:
+            values[:] = self.clean(values)
```
Samuelfaure, line 32: Why it's needed at all: Mongoengine only calls clean() on the deserialization paths (to_python, to_mongo). Direct attribute assignment doesn't go through either, and
  every harvester writes tags by attribute assignment. So pre-fix, clean() was effectively dead code for the harvester flow — only the form/UI path benefited.

ThibaudDauce (approved):
Thanks!

I've added a regression test for this problem.

For information, in production we have:
```
udata__tag_min_length: 0
udata__tag_max_length: 192
```

And locally I use `TAG_MIN_LENGTH=1` in my `udata.cfg`

Samuelfaure:
Very good! Ready to merge then

PR #3783, opened 2026-05-21
fix(dataservices): parse swagger and scope to bouquet fiche for search

Samuelfaure:
OpenAPI documents (up to 1.5 MB for API Entreprise) were truncated at 100 KB before reaching ES, hiding terms like "transfert de siège" that live deep in the spec. Worse, all ~30 fiches sharing a bouquet swagger got the same indexed text, so a hit on any term matched every fiche.

~~Parse the OpenAPI spec server-side, walk operations × schemas with $ref resolution, and for bouquet fiches keep only paths whose operation summary matches the fiche title (ported from cdata's openapi-bouquet.ts). Cap raised to 3 MB; fall back to raw text if parse fails.~~

fix https://github.com/datagouv/data.gouv.fr/issues/1905

Edit: weird CI bug, fixing

PR #3784, opened 2026-05-21
diag: probe stdin tty state at task module import

Samuelfaure:
## Why

Investigating a CI-only crash on PR #3783: invoke v3's `handle_stdin` thread calls `tcgetpgrp(stdin)` which returns ENOTTY on CircleCI when stdin reports as a TTY but isn't a session controller. That PR hits the crash 100% of the time; every other branch passes 100%. This branch off main carries **only** a one-shot stderr probe printing `fd 0` properties so we can compare against the failing PR's run.

## What we already saw on PR #3783

```
[stdin-diag] sys.stdin.isatty()=True os.isatty(0)=True getpgrp()=99 tcgetpgrp(0) FAILS: [Errno 25] Inappropriate ioctl for device fd0 mode=0o20620
```

If this PR shows different values (e.g. `isatty=False`), it confirms runner-allocation asymmetry — same CircleCI config, different stdin allocation per build.
If it shows the same `isatty=True`, then either (a) the crash is timing/race-sensitive and most PRs get lucky, or (b) something in PR #3783's code path actually does trigger the bug deterministically.

## Plan

- Wait for CircleCI to finish (or crash early).
- Read the diagnostic from the build log.
- Throw this branch away.

**Do not merge.**

### DIFF
--- a/tasks/__init__.py
+++ b/tasks/__init__.py
@@ -1,3 +1,5 @@
+import os
+import sys
 from datetime import datetime
 from os.path import join
 from sys import exit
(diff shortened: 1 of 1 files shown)


Samuelfaure:
Got the data we needed. Side-by-side:

|  | PR #3784 (this PR, off main) | PR #3783 (swagger PR) |
|---|---|---|
| isatty(stdin) | True | True |
| fd0 mode | 0o20620 (char dev) | 0o20620 (char dev, same) |
| getpgrp() | 664 | 99 |
| **tcgetpgrp(0)** | **664 (succeeds)** | **FAILS: ENOTTY** |

Both runs use the same TTY-flagged character device. But on this PR the process IS the foreground process group of its controlling terminal (tcgetpgrp returns own pgrp), while on PR #3783 the process has no controlling terminal at all (tcgetpgrp fails with ENOTTY).

Root-cause hypothesis: PR #3783 has a unique `pyproject.toml` checksum, so its CI cache key misses. The fallback path forces `uv sync --frozen` to do real work, and uv's subprocess management severs the shell's controlling-tty linkage. invoke v3's `handle_stdin` thread then crashes on `tcgetpgrp`.

The `in_stream=False` workaround on PR #3783 is correct and minimal.

PR #3751, opened 2026-04-29
fix: publiccode.yml validation errors

bfabio:
Fix validation errors and add a GitHub Actions workflow to catch future ones.

PR #3785, opened 2026-05-21
chore: update publiccode

maudetes:
Alternative to https://github.com/opendatateam/udata/pull/3751

Add automatic version update using tag-version.sh script

### DIFF
--- a/tag-version.sh
+++ b/tag-version.sh
@@ -279,10 +279,13 @@ RELEASE_NOTES="$SORTED_COMMITS"
 if [ "$DRY_RUN" = true ]; then
     echo "Would update CHANGELOG.md with:"
     echo "$NEW_ENTRY"
+    echo "Would update publiccode.yml:"
+    echo "  softwareVersion: v$VERSION"
+    echo "  releaseDate: '$DATE'"
     if [ "$EDIT_CHANGELOG" = true ]; then
         echo "Would open \$EDITOR to edit CHANGELOG.md before committing"
     fi
-    echo "Would run: git add CHANGELOG.md"
+    echo "Would run: git add CHANGELOG.md publiccode.yml"
     echo "Would run: git commit -m \"Bump version $VERSION\""
     echo "Would run: git tag -a \"v$VERSION\" -m \"Version $VERSION\""
     echo "Would run: git push origin HEAD v$VERSION"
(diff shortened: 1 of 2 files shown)


tag-version.sh
```diff
@@ -305,17 +308,26 @@ fi
 
 echo "CHANGELOG.md updated with commits from $LAST_TAG to HEAD"
 
+# Update publiccode.yml
+if [ -f "publiccode.yml" ]; then
+    sed -i "s/^softwareVersion: .*/softwareVersion: v$VERSION/" publiccode.yml
+    sed -i "s/^releaseDate: .*/releaseDate: '$DATE'/" publiccode.yml
```
ThibaudDauce: Update publiccode.yml in tag-version.sh by piping sed to a temp file and mv-ing it back, avoiding the non-portable sed -i (GNU-only) that would break on macOS/BSD.

```suggestion
    sed -e "s/^softwareVersion: .*/softwareVersion: v$VERSION/" \
          -e "s/^releaseDate: .*/releaseDate: '$DATE'/" \
          publiccode.yml > publiccode.yml.tmp
      mv publiccode.yml.tmp publiccode.yml
```

tag-version.sh
```diff
@@ -305,17 +308,26 @@ fi
 
 echo "CHANGELOG.md updated with commits from $LAST_TAG to HEAD"
 
+# Update publiccode.yml
+if [ -f "publiccode.yml" ]; then
```
ThibaudDauce: Not sure why this condition is useful? Since the file exists in Git it should never be absent? (and if it is we should update our script?).

The line below `git add CHANGELOG.md publiccode.yml` will crash anyway if the file is absent.

ThibaudDauce (approved):

tag-version.sh
maudetes: Agreed. We also had this check for `CHANGELOG.md`, but it makes slightly more sense since it would get created if not already existing (even if it exists in the repo anyway).
Removed in https://github.com/opendatateam/udata/pull/3785/commits/a1dc82aa7180baedc9401a802c2b8afff30f7279

tag-version.sh
maudetes: Done in https://github.com/opendatateam/udata/pull/3785/commits/a1dc82aa7180baedc9401a802c2b8afff30f7279

PR #3791, opened 2026-05-26
perf: improve edito blocs perfs

ThibaudDauce:
**Fix N+1 dereferencing when serializing blocs**

Fetching a blocs page (e.g. `GET /posts/<slug>/`) issued one MongoDB query per card to dereference its `organization`/`owner` — hundreds of sequential queries on real pages (the "élections municipales" post hit ~100), which dominated the ~2s response time.

`prefetch_blocs_references()` now collects every reference across the (accordion-nested) bloc tree, reloads each type in a single flat `select_related` query (batching the org/owner lookups), and injects the resolved documents back so marshalling issues no further query. Bloc types and their reference fields are detected by introspection (`ListField(ReferenceField)`), so new bloc types are handled automatically.

Wired into both the Post detail and Site (`datasets_blocs`/`reuses_blocs`/`dataservices_blocs`) endpoints. The prefetch is mask-aware: bloc fields excluded from the response (e.g. the default `/site/` payload) are skipped, so it never adds latency for data that isn't serialized.

Measured on a representative page: **53 → 7 queries** (Post detail), and the masked `/site/` response stays at baseline.

Regression tests lock the query count for: Post detail (no N+1), the `/posts` list (masked blocs not dereferenced), `/site/` default (masked blocs not prefetched), and `/site/` with `X-Fields` (references batched).

One test made locally on a page similar to https://www.data.gouv.fr/posts/jeux-de-donnees-liees-aux-elections-municipales
| Metric                           | Without fix (`main`) | With fix (branch) | Improvement |
| -------------------------------- | -------------------- | ----------------- | ----------- |
| **Live HTTP time** (median)      | **119 ms**           | **45 ms**         | **~2.6×**   |
| min / max                        | 113 / 249 ms         | 43 / 115 ms       |             |
| **MongoDB queries** (in-process) | **109**              | **10**            | **~11×**    |


maudetes (commented):
Could we separate the two aspects to see where the perfs come from?
* strict .exclude 
* mongo prefetch queries
My instinct would be that `exclude` mask matters way more than the number of mongo queries

ThibaudDauce:
@maudetes 

You can't really separate the two: to `.exclude` the heavy fields you have to run the dataset query yourself and inject the result back, because I don't see a simple way to make MongoEngine apply a projection during its internal dereference — its lazy deref always loads the full document. That walk + load + inject is the whole machinery of this PR, and once it's there, batching the lookups (`id__in` + `select_related`) is basically free on top. So the prefetch is really the prerequisite for the `.exclude`, not an independent knob.

That said, I measured each contribution on a local repro of the real page (76 datasets / 1203 resources, 16 reuses, 4 dataservices, 30 orgs):

|                                     | response time | queries |
| ----------------------------------- | ------------: | ------: |
| `main`                              | 289 ms        | 115     |
| `.exclude` only (forced N+1 loads)  | 95 ms         | 193     |
| fewer queries only (full documents) | 107 ms        | 8       |
| both (this PR)                      | 42 ms         | 7       |

You're right that `.exclude` carries most of the win — and the query count isn't the bottleneck (the `.exclude`-only run is faster than `main` while issuing *more* queries). The real cost is CPU: deserializing the 1203 embedded `resources` into Python objects. But since `.exclude` can't be done without the prefetch scaffolding anyway, keeping both in one pass is the simplest correct option.

nicolaskempf57 (approved):

PR #3796, opened 2026-05-27
feat(harvest): update licences and rights support for GeoDCAT-AP 3

### LINKED ISSUE #3787: Update licences and rights harvesting for GeoDCAT-AP 3
From `geodcat-ap-2.0.0` vs `main` branches comparison results in https://github.com/ecolabdata/ecospheres/issues/275#issuecomment-4315124466 :

> 2928: CHANGE (non-URI rights only) `dct:accessRights/dct:RightsStatement` -> `dct:rights/dct:RightsStatement`
> => **TODO[datagouv]**: Add rights to `infer_dataset_access_rights`.

streino:
Fix https://github.com/opendatateam/udata/issues/3787


udata/core/dataset/rdf.py
```diff
@@ -665,71 +669,67 @@ def title_from_rdf(resource: RdfResource, url: str | None = None, format: str |
...
-            return dataset_access_rights, AccessType.RESTRICTED, access_right_categories.pop()
+    # Try to match access rights to known inspire access rights limitation categories
+    country = current_app.config["DEFAULT_COUNTRY_CODE"]
+    categories = set(
+        InspireLimitationCategory.get_category_from_localized_label(access_right, country)
+        for access_right in access_rights
+    )
+    categories.discard(None)
+    if len(categories) != 1:
+        # FIXME: len() > 1 should also return RESTRICTED, or raise
```
streino, line 725: Not related to the SEMIC update, but should we handle that case?

udata/core/dataset/rdf.py
```diff
@@ -896,40 +896,48 @@ def dataset_from_rdf(
...
     for additionnal in d.objects(DCT.hasPart):
         resource_from_rdf(additionnal, dataset, is_additionnal=True)
 
-    access_rights, access_type, access_right_category = infer_dataset_access_rights(
-        d, resources_access_rights
-    )
-    if access_rights:
-        add_dcat_extra(dataset, "accessRights", access_rights)
+    dataset_rights = rights_from_rdf(d)
+    if not dataset_rights and len(resources_rights) == 1:
```
streino, line 912: I'm generalizing the logic we already had for bubbling up `accessRights` from resource extras to dataset extras (when all resources have the same `accessRights`).

Since with SEMIC main [almost everything ends up in `rights`](https://github.com/ecolabdata/udata/pull/11/changes#r3268712634), it seems necessary to do the same for `rights`.

udata/core/dataset/rdf.py
```diff
@@ -896,40 +896,48 @@ def dataset_from_rdf(
...
     if access_type:
         dataset.access_type = access_type
-    if access_right_category:
-        dataset.access_type_reason_category = access_right_category
+    if inspire_category:
+        dataset.access_type_reason_category = inspire_category
 
-    default_license = dataset.license or License.default()
     dataset_licenses = licenses_from_rdf(d)
+    if not dataset_licenses and len(resources_licenses) == 1:
```
streino, line 929: Same logic for bubbling up `license` to dataset extras, as we already do for `accessRights`.

It's a bit redundant with `dataset.license`, but it might still be relevant to have the info somewhere when `License.guess` doesn't match?

udata/core/dataset/rdf.py
streino, line 1: SEMIC main applies the following logic:
- `otherConstraints` in URI form => `license` or `accessRights` (depending on whether `otherConstraints` is attached to `useConstraints` or `accessConstraints`).
- Everything else (non-URI `otherConstraints` and `useLimitation`) => `rights`.

Note that the DCAT and DCAT-AP specs allow free text in `license` and `accessRights`. But [GeoDCAT-AP](https://semiceu.github.io/GeoDCAT-AP/releases/3.1.0/#conditions-for-access-and-use-and-limitations-on-public-access-use-limitation-and-access-other-constraints) seems ambiguous to me, as the spec uses DCAT-AP properties and states that `license` and `accessRights` *should* (not *must*) use URIs, but concludes that only URIs are represented by `license` or `accessRights`, with the rest going into `rights`.

So, since DCAT and DCAT-AP allow free text, and few ISO catalogues (converted by SEMIC) use URIs, **we'll end up with everything (licenses, access rights and others) lumped together in `rights` in most cases**, as is the case in the test here.

And in the case of ISO catalogues, the info won't be found in the same DCAT property, depending on whether a catalogue uses a URI or not for a license or access right.

Even if the GeoDCAT-AP spec aligned with DCAT/DCAT-AP, [I don't know whether the SEMIC converter will follow suit](https://github.com/SEMICeu/iso-19139-to-dcat-ap/issues/68), as that would require detecting what is genuinely a `license` or `accessRights` in free text, to differentiate it from other `rights`.

On the datagouv side, having everything in `rights` doesn't prevent detecting `dataset.license` and `dataset.access_type`, since we rely on everything available (including `rights`) to infer these properties.

But ecologie.datagouv displays the `license`/`rights`/`accessRights` stored as extras. And at some point datagouv may want to expose all of them in its DCAT output?

Since we already have the functions to detect licenses and access rights from `rights`, I'm thinking we could use them further upstream in `rights_from_rdf` to attempt to correctly re-categorise the `license` and `accessRights` "hidden" in free text.

This would allow better categorisation of what we store as extras, and less variation between constraints represented as URIs and those in free text.

Any argument against reworking what we store as extras?

udata/tests/dataset/test_dataset_rdf.py
```diff
@@ -1739,125 +1770,109 @@ def test_licenses_from_rdf(self, xml: str, expected_licenses: set[str]):
...
-        g.add((node, RDF.type, DCAT.Dataset))
-        g.add((node, DCT.title, Literal("Test dataset")))
-        g.add(
-            (
-                node,
-                DCT.accessRights,
-                Literal(
-                    "L124-5-II-1 du code de l'environnement (Directive 2007/2/CE (INSPIRE), Article 13.1.b)"
-                ),
+    # TODO
```
streino, line 1773: TODO

streino (commented):
@maudetes Here's an annotated first draft. I'll need your feedback, esp. about https://github.com/opendatateam/udata/pull/3796/changes#r3311220902, to know how far to push this 🙏

udata/core/dataset/rdf.py
streino: Fixed in https://github.com/opendatateam/udata/pull/3832

udata/core/dataset/rdf.py
streino: I'll tackle this in a separate PR. Let's first upgrade to SEMIC main taking SEMIC output as-is.

streino:
Replaced by https://github.com/opendatateam/udata/pull/3832.

PR #3797, opened 2026-05-27
feat(harvest): harvest DCAT DatasetSeries as udata Dataset

### LINKED ISSUE #3767: Harvest DCAT `DatasetSeries`
The original GeoNetwork mapping converts ISO-19139 "series" to DCAT `Dataset` (DCAT 2), but the new mapping available in 4.4.6+ converts them to `DatasetSeries` (DCAT 3).

Depending on how csw-dcat is configured, we'll get `Dataset` (default config) or `DatasetSeries (GeoDCAT-AP option enabled). 

Currently, udata harvests `Dataset` but ignores `DatasetSeries` (even though the CSW request asks for "series"), leading to very different behavior depending on whether GeoDCAT-AP is enable or not. The worst case would be switching from default csw-dcat to GeoDCAT-AP on an existing harvest: this will cause all series records to disappear from the udata catalog.

Short of truly handling `DatasetSeries`, udata should probably at least treat `DatasetSeries` like `Dataset` to prevent the discrepancy between the two harvest modes. 

Example : https://sextant.ifremer.fr/geonetwork/DATAGOUVHVD/fre/catalog.search#/metadata/fc663284-e599-427e-9da6-9fe9519170f3

streino:
Fix https://github.com/opendatateam/udata/issues/3767

### DIFF
--- a/udata/commands/dcat.py
+++ b/udata/commands/dcat.py
@@ -72,7 +72,7 @@ def parse_url(url, csw, iso, quiet=False, rid=""):
         _subgraph = Graph(namespace_manager=namespace_manager)
         graph += _subgraph.parse(data=serialized, format=format)
 
-        for node in subgraph.subjects(RDF.type, DCAT.Dataset):
+        for node in subgraph.subjects(RDF.type, [DCAT.Dataset, DCAT.DatasetSeries]):
             identifier = subgraph.value(node, DCT.identifier)
             kwargs = {"nid": str(node), "page": page_number}
             kwargs["type"] = "uriref" if isinstance(node, URIRef) else "blank"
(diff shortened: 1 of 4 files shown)


ThibaudDauce (approved):
You could maybe also patch `udata dcat parse-url` to include `DatasetSeries` but not mandatory. Thanks!

streino:
> You could maybe also patch `udata dcat parse-url` to include `DatasetSeries` but not mandatory. Thanks!

Done. But that command would need some more love to match what's done in the harvester (from a quick glance, at least services are missing).

ThibaudDauce (approved):

PR #3804, opened 2026-06-01
feat: filter orgs with dataservices in suggest

Samuelfaure:
Linked with https://github.com/datagouv/cdata/pull/1106

fix https://linear.app/pole-api/issue/API2-314/retirer-les-organisations-par-defaut-du-filtre-organisation-lorsque

### DIFF
--- a/udata/core/organization/api.py
+++ b/udata/core/organization/api.py
@@ -1,6 +1,7 @@
 from datetime import UTC, datetime
 
 from flask import make_response, redirect, request, url_for
+from flask_restx.inputs import boolean
 from mongoengine.queryset.visitor import Q
 
 from udata.api import API, api, errors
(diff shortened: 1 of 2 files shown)


ThibaudDauce:
We want to look into facets to try to solve this issue with a generic answer and not add filters everywhere. Facets are not perfect right now but we'll try to fix them soon.

Samuelfaure:
@ThibaudDauce We had this discussion on https://github.com/datagouv/cdata/pull/1106

My position is to merge this small PR and then rewrite all as facets later, which would be on a bigger PR

Or do you want to start by the facets?

ThibaudDauce:
I still don't want to merge this before trying something better for the org facets (like other PRs like https://github.com/opendatateam/udata/pull/3738)

Issue #3793, opened 2026-05-27
Zombie resources in harvested datasets


streino:
By zombie I mean a `Resource` that was but isn't harvested from the source anymore, and is still attached to its `Dataset`, therefore showing up in its list of resources.

Zombies can occur in multiple situations:
- Resource is actually removed from the source record.
- Harvester configuration changes, and what was previously identified as a harvestable resource isn't anymore. For DCAT-related harvesters, this can happen:
  - When switching from the old GeoNetwork mapping (everything is a distribution) to the new one (only actual online resources are distributions, the rest is pages). This occurs when activating the "GeoDCAT-AP" option on `csw-dcat`.
  - When switching between `csw-iso-19139` (page unless proven a distribution) and `csw-dcat` (distribution unless proven a page), since they don't use the same heuristics.

***But***, datagouv lets users manually add resources to any dataset (via UI or API), including harvested datasets. So there isn't a guaranteed 1:1 mapping between source distributions and datagouv resources. We can't therefore just prune all distributions that aren't seen in the source anymore.

streino:
Discussed with @maudetes: First step is to add a timestamp in the `harvest` section of every (harvested) distribution, so we can track when they were last seen in the source.

We'll re-discuss the next steps, since we have to be careful about impacts on non-harvested resources on harvested datasets, impacts on hydra, whether we want to give options to harvester admins, ...

streino:
Resources stats for prod 28/07/2026 (resources of non-archived datasets):

| harvested | status | count | segment % | total % |
|---|---|---|---|---|
| true | current | 161690 | 67.25 | 41.53 |
| true | unknown | 68438 | 28.47 | 17.58 |
| true | outdated | 10299 | 4.28 | 2.65 |
| false | n/a | 148896 | 99.99 | 38.24 |
| false | outdated | 18 | 0.01 | 0 |

streino:
Updated stats for prod 04/08/2026, only considering datasets harvested during the last week (to tolerate some failed harvests runs, but still drop really stale harvests):

| harvested | status | count | segment % | total % |
|---|---|---|---|---|
| true | current | 160259 | 81.14 | 46.51 |
| true | unknown | 27614 | 13.98 | 8.01 |
| true | outdated | 9636 | 4.88 | 2.8 |
| false | n/a | 147029 | 99.99 | 42.67 |
| false | outdated | 18 | 0.01 | 0.01 |

[...]

PR #3805, opened 2026-06-01
feat(harvest): add resource harvest timestamp

streino:
Part of https://github.com/opendatateam/udata/issues/3793


ThibaudDauce (approved):
Not done for MAAF but not sure it's required?

streino:
> Not done for MAAF but not sure it's required?

MAAF doesn't set `Dataset.harvest` or `Resource.harvest` at all, so I'd say no?

PR #3810, opened 2026-06-02
fix(tags): move validation in harvest logic to fix tags unpickling

maudetes:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/293060/

```
DecodeError udata.tags in normalize
'<' not supported between instances of 'int' and 'LocalProxy'
```

We can't use `TAG_MIN_LENGTH`  LocalProxy in celery unpickling.
We shouldn't run validation in `clean`, since it is called outside of the app context.
Additionally, we don't want to silently drop invalid tags in most contexts.

Partially reverts https://github.com/opendatateam/udata/pull/3782 by moving tag validation in harvest logic.

### DIFF
--- a/udata/mongo/taglist_field.py
+++ b/udata/mongo/taglist_field.py
@@ -20,7 +20,7 @@ class TagListField(ListField):
             return []
 
     def clean(self, value):
-        return sorted(list(set([slugify_tag(v) for v in value])))
+        return sorted({slugify_tag(v) for v in value})
 
     def to_python(self, value):
         return super(TagListField, self).to_python(self.clean(value))
--- a/udata/mongo/taglist_field.py
+++ b/udata/mongo/taglist_field.py
@@ -1,8 +1,8 @@
 from mongoengine.fields import ListField, StringField
-from slugify import slugify
 
 from udata import tags
 from udata.i18n import lazy_gettext as _
+from udata.tags import slug as slugify_tag
 
 
 class TagListField(ListField):
@@ -20,7 +20,7 @@ class TagListField(ListField):
             return []
 
     def clean(self, value):
-        return sorted(list(set([slugify(v, to_lower=True) for v in value])))
+        return sorted(list(set([slugify_tag(v) for v in value])))
 
     def to_python(self, value):
         return super(TagListField, self).to_python(self.clean(value))
--- a/udata/mongo/taglist_field.py
+++ b/udata/mongo/taglist_field.py
@@ -1,4 +1,5 @@
 from mongoengine.fields import ListField, StringField
+from slugify import slugify
 
 from udata import tags
 from udata.i18n import lazy_gettext as _
@@ -19,7 +20,7 @@ class TagListField(ListField):
             return []
 
     def clean(self, value):
-        return sorted({n for n in (tags.normalize(v) for v in value) if n})
+        return sorted(list(set([slugify(v, to_lower=True) for v in value])))
 
     def to_python(self, value):
         return super(TagListField, self).to_python(self.clean(value))
@@ -28,8 +29,6 @@ class TagListField(ListField):
         return super(TagListField, self).to_mongo(self.clean(value))
 
     def validate(self, values):
-        if values is not None:
-            values[:] = self.clean(values)
         super(TagListField, self).validate(values)
 
         for tag in values:
--- a/udata/rdf.py
+++ b/udata/rdf.py
@@ -370,8 +370,7 @@ def themes_from_rdf(rdf):
             continue
         tags.append(tag.toPython())
     tags += theme_labels_from_rdf(rdf)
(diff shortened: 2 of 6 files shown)


udata/tests/test_tags.py
```diff
@@ -109,21 +110,47 @@ def test_normalize(self, app):
...
+    def test_taglist_field_deserialization_no_error(self):
+        """Regression test: TagListField deserialization should not raise errors.
+
+        This ensures the original bug (TypeError from LocalProxy during unpickling)
+        is fixed by verifying clean() doesn't call normalize() during to_python().
+        """
+        import pickle
+
+        from udata.mongo.taglist_field import TagListField
+
```
ThibaudDauce: Could be at the top?

udata/mongo/taglist_field.py
```diff
@@ -19,7 +20,7 @@ def from_input(input):
             return []
 
     def clean(self, value):
-        return sorted({n for n in (tags.normalize(v) for v in value) if n})
+        return sorted(list(set([slugify(v, to_lower=True) for v in value])))
```
ThibaudDauce: What's the difference between `tag.slug` and `slugify` should we use the same everywhere?

udata/rdf.py
```diff
@@ -29,6 +29,7 @@
 from udata import uris
 from udata.core.contact_point.models import ContactPoint
 from udata.frontend.markdown import parse_html
+from udata.harvest.filters import normalize_tag
```
ThibaudDauce, line 32: Should it be?
```suggestion
from  udata.tags import  normalize
```

udata/rdf.py
```diff
@@ -369,7 +370,8 @@ def themes_from_rdf(rdf):
             continue
         tags.append(tag.toPython())
     tags += theme_labels_from_rdf(rdf)
-    return list(set(tags))
+    normalized_tags = [normalize_tag(tag) for tag in tags if normalize_tag(tag)]
+    return list(set(normalized_tags))
```
ThibaudDauce:
```suggestion
    return sorted({normalize_tag(tag) for tag in tags if normalize_tag(tag)})
```

ThibaudDauce (commented):
The core fix here — moving keyword normalization into `themes_from_rdf` so a dirty `dcat:keyword` (empty, too short, too long) no longer crashes the harvest on `validate()` — isn't covered by a test. The existing keyword tests only use valid tags (`faker.tags`).
Could you add a regression test that builds a graph with a few invalid `DCAT.keyword` values (e.g. `""`, `"df"`, and an over-long string), then asserts `dataset_from_rdf(g)` + `dataset.validate()` succeeds and `dataset.tags` only contains the normalized, valid tags? That's exactly the scenario #3782 was about, so it's worth locking down.

maudetes:
> Could you add a regression test [...]

I kept [the regression test](https://github.com/opendatateam/udata/pull/3782/changes/ca8ce0a53220fcb06e842a599471490dc79821c4) added in the previous PR that build a graph with invalid keywords.

udata/rdf.py
maudetes: I kept with the existing pattern :shrug:

udata/mongo/taglist_field.py
maudetes: Seems like it's just a combination of slugify with lower. Used `tags.slug` in https://github.com/opendatateam/udata/pull/3810/commits/46fc53940e8e3eae993568742fbd9305729f1d4c.

udata/tests/test_tags.py
maudetes: Done in https://github.com/opendatateam/udata/pull/3810/changes/f4633f008a74b1f07407945d3ebfa34fd119f2c2

udata/mongo/taglist_field.py
```diff
@@ -19,7 +20,7 @@ def from_input(input):
             return []
 
     def clean(self, value):
-        return sorted({n for n in (tags.normalize(v) for v in value) if n})
+        return sorted(list(set([slugify_tag(v) for v in value])))
```
ThibaudDauce:
```suggestion
        return sorted({slugify_tag(v) for v in value})
```

ThibaudDauce (commented):
Not done for CKAN?

maudetes:
> Not done for CKAN?

CKAN was historically the only harvester filtering valid tags [in its schema](https://github.com/opendatateam/udata/blob/d79e8782cc79dbbb68f458173a7f0651b6e78cba/udata/harvest/backends/ckan/schemas/ckan.py#L37).
Added a test for CKAN in https://github.com/opendatateam/udata/pull/3810/commits/1a373caf22f3855ef23f31b5259adb3231a2dddf.

ThibaudDauce (approved):

PR #3811, opened 2026-06-02
feat(search): index Topic.elements

### LINKED ISSUE #1126: UnifiedSearch: indexation des facteurs

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/1126

Adds `Topic.elements[].(title|description|tags)` to the Topic ES index (like it's done in Mongo + tags).


udata_search_service/search_clients.py
```diff
@@ -483,7 +486,15 @@ def query_topics(
...
                         operator="and",
-                        fields=["id^5", "name^10", "description^4", "tags^3"],
+                        fields=[
+                            "id^8",
+                            "name^10",
+                            "description^6",
+                            "tags^5",
+                            "elements_titles^4",
+                            "elements_descriptions^3",
+                            "elements_tags^2",
```
abulte, line 496: Tried to make a kind of ladder Topic > Elements here, YMMV.

abulte: @streino if you have an opinion

ThibaudDauce (approved):
Maybe add a test on update/deletion to check the signals?

abulte:
> Maybe add a test on update/deletion to check the signals?

Done.

udata_search_service/search_clients.py
streino: A bit late to the party... No particular opinion. Logic makes sense but as you said YMMV, so let's see how it behaves.

PR #3812, opened 2026-06-03
fix: dataset deleted in accordion

ThibaudDauce:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/293899

And broken page in https://www.data.gouv.fr/posts/jeux-de-donnees-liees-aux-elections-municipales

- [x] fix missing accordion refs in purge
- [x] add a auto cleanup in marshalling to prevent 500 on page if purge fail in some way


maudetes (approved):
Thank you for the quickfix

PR #3816, opened 2026-06-05
feat(dataservices): compute reuses metric

Samuelfaure:
Add a 'reuses' metric to the Dataservice model, computed from the reuses that reference it via their 'dataservices' field, mirroring the existing Dataset.count_reuses pattern.

A signal handler recomputes the metric whenever a reuse is created, updated or deleted, so the count stays in sync. Without this the dataservice reuses tab always displayed 'Réutilisations (0)'.

related https://github.com/datagouv/cdata/pull/1120

### DIFF
--- a/udata/core/dataservices/models.py
+++ b/udata/core/dataservices/models.py
@@ -357,6 +357,7 @@ class Dataservice(
         "discussions_open",
         "followers",
         "followers_by_months",
+        "reuses",
         "views",
     ]
 
@@ -392,6 +393,16 @@ class Dataservice(
         self.metrics["followers"] = Follow.objects(until=None).followers(self).count()
         self.metrics["followers_by_months"] = get_stock_metrics(
             Follow.objects(following=self), date_label="since"
+        )
+        self.save(signal_kwargs={"ignores": ["post_save"]})
+
+    def count_reuses(self):
+        from udata.models import Reuse
+
+        # Not using visible() here because it excludes reuses without datasets,
+        # but a reuse can legitimately reference only a dataservice.
+        self.metrics["reuses"] = (
+            Reuse.objects(dataservices=self).filter(private__ne=True, deleted=None).count()
         )
         self.save(signal_kwargs={"ignores": ["post_save"]})
 
--- a/udata/core/dataservices/metrics.py
+++ b/udata/core/dataservices/metrics.py
@@ -1,0 +1,20 @@
+from udata.core.reuse.models import Reuse
+
+from .models import Dataservice
+
+
+@Reuse.on_create.connect
+@Reuse.on_update.connect
+@Reuse.on_delete.connect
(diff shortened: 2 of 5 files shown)


udata/core/dataservices/models.py
```diff
@@ -395,6 +396,14 @@ def count_followers(self):
         )
         self.save(signal_kwargs={"ignores": ["post_save"]})
 
+    def count_reuses(self):
```
maudetes, line 399: Could you add `dataservice.count_reuses` in [udata.core.metrics.commands.py](https://github.com/opendatateam/udata/blob/main/udata/core/metrics/commands.py) `update` command?

udata/core/dataservices/models.py
```diff
@@ -395,6 +396,14 @@ def count_followers(self):
...
+        from udata.models import Reuse
+
+        self.metrics["reuses"] = (
+            Reuse.objects(dataservices=self).filter(private__ne=True, deleted=None).count()
```
maudetes, line 405:
```suggestion
            Reuse.objects(dataservices=self).visible().count()
```

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Current (in progress)
+
+- feat(dataservices): compute the `reuses` metric so the dataservice reuses count is accurate [#XXXX](https://github.com/opendatateam/udata/pull/XXXX)
+
```
maudetes, line 15: Changelog is now updated automatically, you don't need to modify it

maudetes (approved):

ThibaudDauce:
@Samuelfaure I don't see you responses for the questions of @maudetes ?

Samuelfaure:
@ThibaudDauce I forgor 💀

Samuelfaure:
@ThibaudDauce Ok done (edit: need to fix a test)

udata/core/dataservices/models.py
```diff
@@ -395,6 +396,12 @@ def count_followers(self):
         )
         self.save(signal_kwargs={"ignores": ["post_save"]})
 
+    def count_reuses(self):
+        from udata.models import Reuse
+
+        self.metrics["reuses"] = Reuse.objects(dataservices=self).visible().count()
```
ThibaudDauce: `visible()` exclude reuses without datasets, it's not a problem for counting for datasets but for dataservices I think we want all the reuses even the ones without datasets?

ThibaudDauce (changes requested):

udata/core/dataservices/models.py
Samuelfaure: Oh yeah you're right, I shouldn't have applied the previous correction, cancelling

ThibaudDauce (approved):
Could you add a comment at least to explain why we don't use `.visible()` here (since it may not be obvious in a few months)

PR #3827, opened 2026-06-10
feat(geopf): add auth and sync

abulte:
Related https://github.com/datagouv/data.gouv.fr/issues/2043

This adds backend auth and sync to and from cartes.gouv.fr. See `docs/geopf.md` for a full description of the features and workflow.

Useful resources:
- geopf swagger https://data.geopf.fr/api/swagger-ui/index.html#/
- geopf docs https://geoplateforme.github.io


abulte:
@maudetes Current status is udata-only and triggered on resource upload automatically. Next, I'll work on cdata/oauth stuff and user triggers. But maybe you want to review the current process before this gets too big?

udata/geopf/tasks.py
```diff
@@ -0,0 +1,471 @@
...
+    return _DownloadToTempfile(resource.url)
+
+
+class _DownloadToTempfile:
+    """Download a remote URL to a temp file, yield it, clean up on exit.
+
+    FIXME: `self.url` is user-controlled (remote resource URL), so this fetch
+    is an SSRF vector from the worker (internal hosts, link-local metadata
+    endpoints). Switch to the SSRF-hardened http client once it lands
+    (see PRs #3877/#3878).
```
abulte: FIXME https://github.com/opendatateam/udata/pull/3877

abulte:
Putting this back in draft for a while, frontend review surfaced some lack on backend.

udata/geopf/tasks.py
```diff
@@ -0,0 +1,474 @@
...
+    dataset_id = dataset.id
+    resource_id = resource.id
+
+    upload_id = None
+    stored_data_id = None
+
+    try:
+        with _open_resource_file(resource) as f:
+            file_md5 = md5(f)
+            f.seek(0)
```
ThibaudDauce, line 151: This break with the new S3 storage for resources :-( Maybe use the same path as remote resources?

udata/geopf/auth.py
```diff
@@ -0,0 +1,104 @@
...
+            client_kwargs={"scope": app.config.get("GEOPF_OAUTH_SCOPE")},
+        )
+
+
+def store_token(user, token: dict) -> GeopfToken:
+    """Persist an authlib token dict (from the OAuth callback or a refresh) for `user`."""
+    geopf_token = GeopfToken.objects(user=user).first() or GeopfToken(user=user)
+    geopf_token.access_token = token["access_token"]
+    geopf_token.refresh_token = token["refresh_token"]
+    geopf_token.expires_at = datetime.now(UTC) + timedelta(seconds=token["expires_in"])
```
ThibaudDauce: `OAuth2Token` already derive the `expires_at` timestamp from the response (and it support the response containing `expires_in` or directly `expires_at`.

```suggestion
    geopf_token.expires_at = datetime.fromtimestamp(token["expires_at"], UTC)
```

udata/geopf/auth.py
```diff
@@ -0,0 +1,104 @@
...
+            server_metadata_url=app.config.get("GEOPF_OAUTH_OPENID_CONF_URL"),
+            client_kwargs={"scope": app.config.get("GEOPF_OAUTH_SCOPE")},
+        )
+
+
+def store_token(user, token: dict) -> GeopfToken:
+    """Persist an authlib token dict (from the OAuth callback or a refresh) for `user`."""
+    geopf_token = GeopfToken.objects(user=user).first() or GeopfToken(user=user)
+    geopf_token.access_token = token["access_token"]
+    geopf_token.refresh_token = token["refresh_token"]
```
ThibaudDauce: Should we keep the old refresh token if not present? It seems the OAuth spec doesn't require to send a new refresh_token…

```suggestion
    geopf_token.refresh_token = token.get("refresh_token", geopf_token.refresh_token)
```

udata/geopf/tasks.py
```diff
@@ -0,0 +1,474 @@
...
+def _offering_url(offering: dict) -> str:
+    urls = offering.get("urls") or []
+    return urls[0].get("url", "") if urls else ""
+
+
+def set_dataset_extras(dataset, extras: dict) -> None:
+    dataset.extras.update(extras)
+    Dataset.objects(id=dataset.id).update_one(
+        **{f"set__extras__{k}": v for k, v in extras.items()},
+    )
```
ThibaudDauce: This can set `None` for values and fail all the next `dataset.save()` on validation. `None` values should pop the key instead.

udata/geopf/tasks.py
```diff
@@ -0,0 +1,474 @@
...
+
+
+def set_resource_extras(dataset, resource, extras: dict) -> None:
+    """Update resource extras in-place and persist without reloading the full dataset."""
+    resource = get_by(dataset.resources, id=resource.id)
+    resource.extras.update(extras)
+    Dataset.objects(id=dataset.id, resources__id=resource.id).update_one(
+        **{f"set__resources__S__extras__{k}": v for k, v in extras.items()},
+    )
```
ThibaudDauce: This can set `None` for values and fail all the next `dataset.save()` on validation. `None` values should pop the key instead.

udata/geopf/client.py
```diff
@@ -0,0 +1,268 @@
...
+    def _url(self, path: str) -> str:
+        if not self.datastore:
+            raise GeopfError("GeopfClient: no datastore_id configured for this call")
+        return f"{self.base}/datastores/{self.datastore}/{path}"
+
+    def _raise(self, resp: requests.Response) -> None:
+        try:
+            resp.raise_for_status()
+        except requests.HTTPError as e:
+            raise GeopfError(f"{resp.status_code} {resp.url}: {resp.text}") from e
```
ThibaudDauce: This message is put into extras? Are we sure no private information can be returned by geopf in the body of the response?

udata/geopf/client.py
```diff
@@ -0,0 +1,268 @@
...
+        self._raise(resp)
+        memberships = resp.json().get("communities_member", [])
+        return [
+            {
+                "datastore_id": m["community"]["datastore"],
+                "name": m["community"].get("name"),
+                "rights": m.get("rights", []),
+            }
+            for m in memberships
+            if REQUIRED_PUBLISH_RIGHTS.issubset(m.get("rights", []))
```
ThibaudDauce, line 245: Is it possible to return all the datastores and annotate them with the missing permissions? It will allow the user to see the datastore it expects but notice him that he's missing one permission to use it? Is it better than having just an empty list and don't understand why it's not showing?

udata/geopf/tasks.py
```diff
@@ -0,0 +1,474 @@
...
+        raise
+
+    # Pin only after success, so a failed first push can't lock the dataset
+    # onto a bad datastore.
+    set_dataset_extras(dataset, {"geopf:push:datastore-id": datastore_id})
+
+
+def _run_pipeline(dataset, resource, datastore_id: str, client) -> None:
+    datasheet_name = str(dataset.id)
+    stored_data_name = f"_{resource.id}"
```
ThibaudDauce, line 140: Why `resource.id` is prefixed and not `dataset.id`?

udata/geopf/tasks.py
```diff
@@ -0,0 +1,474 @@
...
+        log.info(
+            "geopf: launched processing execution=%s dataset=%s resource=%s",
+            exec_id,
+            dataset_id,
+            resource_id,
+        )
+
+        exec_status, stored_data_id = client.poll_execution(exec_id)
+
+        # Delete upload after processing; API returns 409 if attempted while processing runs
```
ThibaudDauce, line 194: Should we check `exec_status != "SUCCESS"` before deleting the upload? To retry the processing or is it impossible and it's better to delete and retry from start?

udata/geopf/client.py
```diff
@@ -0,0 +1,268 @@
...
+                f"409 on metadata upload, could not locate existing record: {resp.text}"
+            )
+        self._raise(resp)
+        return resp.json()["_id"]
+
+    def _find_metadata_id(self, file_identifier: str) -> str | None:
+        resp = self.session.get(self._url("metadata"))
+        self._raise(resp)
+        for item in resp.json():
+            if item.get("file_identifier") == file_identifier:
```
ThibaudDauce, line 297: Is it possible to do this server side? I saw `"stored_data": stored_data_id` somewhere else. And what about pagination?  Is there a limit like in `_find_vector_processing_id` `"limit": 100`?

If it's not possible, at least have a comment explaining that the server doesn't support `file_identifier` as a filter?

udata/geopf/tasks.py
```diff
@@ -0,0 +1,474 @@
...
+    GeopfReauthRequired,
+    GeopfTimeoutError,
+)
+from .metadata import dataset_to_iso19115
+from .srs import DEFAULT_SRS, detect_srs
+
+log = logging.getLogger(__name__)
+
+
+@task(name="geopf.push_resource", bind=True, ignore_result=False)
```
ThibaudDauce, line 44: Maybe should be low priority if it's slow?

```suggestion
@task(name="geopf.push_resource", bind=True, ignore_result=False, route="low.geopf")
```

udata/geopf/tasks.py
```diff
@@ -0,0 +1,474 @@
...
+        log.info("geopf: updated metadata=%s dataset=%s", metadata_id, dataset.id)
+    else:
+        metadata_id = client.upload_metadata(xml)
+        log.info("geopf: uploaded metadata=%s dataset=%s", metadata_id, dataset.id)
+        client.tag_entity("metadata", metadata_id, datasheet_name)
+        set_dataset_extras(dataset, {"geopf:push:metadata-id": metadata_id})
+    return metadata_id
+
+
+@task(name="geopf.pull_offerings", bind=True, ignore_result=False)
```
ThibaudDauce, line 336: Maybe should be low priority if it's slow?

```suggestion
@task(name="geopf.pull_offerings", bind=True, ignore_result=False, route="low.geopf")
```

pyproject.toml
```diff
@@ -112,6 +114,9 @@ maaf = "udata.harvest.backends.maaf:MaafBackend"
 ckan = "udata.harvest.backends.ckan.harvesters:CkanBackend"
 dkan = "udata.harvest.backends.ckan.harvesters:DkanBackend"
 
+[project.entry-points."udata.plugins"]
+geopf = "udata.geopf"
```
ThibaudDauce: Is it really useful to have a plugin?  We can maybe check for the `GEOPF_API_BASE` setting with a simple `if` and avoid plugin complexity? It doesn't seems the plugin choice is bringing much to the feature?

udata/geopf/client.py
```diff
@@ -0,0 +1,268 @@
...
+        (push, reverse sync alike) runs as whichever user is currently
+        authenticated, there is no anonymous or service-account credential.
+        `datastore_id` scopes datastore-bound calls (uploads, processing,
+        tagging, metadata, offerings); it isn't needed for instance-level
+        calls like `list_datastores`.
+        """
+        self.base = current_app.config["GEOPF_API_BASE"]
+        self.datastore = datastore_id
+        self.poll_timeout = current_app.config.get("GEOPF_POLL_TIMEOUT", POLL_TIMEOUT)
+        self.session = requests.Session()
```
ThibaudDauce: Maybe add a default timeout? But it seems to need an override of `requests.Session`… Or set it manualy inside the requests in `while time.time() < deadline:` to avoid stalling?

udata/geopf/client.py
```diff
@@ -0,0 +1,268 @@
...
+        `token` is the acting user's bearer access token: every call here
+        (push, reverse sync alike) runs as whichever user is currently
+        authenticated, there is no anonymous or service-account credential.
+        `datastore_id` scopes datastore-bound calls (uploads, processing,
+        tagging, metadata, offerings); it isn't needed for instance-level
+        calls like `list_datastores`.
+        """
+        self.base = current_app.config["GEOPF_API_BASE"]
+        self.datastore = datastore_id
+        self.poll_timeout = current_app.config.get("GEOPF_POLL_TIMEOUT", POLL_TIMEOUT)
```
ThibaudDauce: `GEOPF_POLL_TIMEOUT` is never set in `settings.py` for the documentation?

udata/geopf/tasks.py
```diff
@@ -0,0 +1,474 @@
...
+
+    sync_metadata(dataset, client)
+
+    url = fiche_url(datastore_id, datasheet_name)
+    set_resource_extras(
+        dataset,
+        resource,
+        {
+            "geopf:push:status": "done",
+            "geopf:push:stored-data-id": stored_data_id,
```
ThibaudDauce: Maybe we could set this sooner, between `exec_status, stored_data_id = client.poll_execution(exec_id)` and this there is a lot a failure possible, having this sooner inside the extras could allow to cleanup the orphan or retry?

udata/geopf/tasks.py
```diff
@@ -0,0 +1,474 @@
...
+
+
+def set_dataset_extras(dataset, extras: dict) -> None:
+    dataset.extras.update(extras)
+    Dataset.objects(id=dataset.id).update_one(
+        **{f"set__extras__{k}": v for k, v in extras.items()},
+    )
+
+
+def set_resource_extras(dataset, resource, extras: dict) -> None:
```
ThibaudDauce: Is this the same as `Dataset.update_resource_extras`?

udata/geopf/tasks.py
```diff
@@ -0,0 +1,474 @@
...
+            dataset.update_resource(existing)
+        set_resource_extras(dataset, existing, extras_update)
+
+
+def _offering_url(offering: dict) -> str:
+    urls = offering.get("urls") or []
+    return urls[0].get("url", "") if urls else ""
+
+
+def set_dataset_extras(dataset, extras: dict) -> None:
```
ThibaudDauce: Maybe look into `Dataset.update_resource_extras` to implement the full save (quality, last update, indexing?) I don't think any of this is useful right now, but could catch us later?

udata/geopf/commands.py
```diff
@@ -0,0 +1,103 @@
...
+@click.argument("resource_id")
+@user_id_option
+@token_option
+@datastore_id_option
+def push_resource(dataset_id, resource_id, user_id, token, datastore_id):
+    """Push a GPKG resource to Géoplateforme (runs synchronously)."""
+    if bool(user_id) == bool(token):
+        raise click.ClickException("Provide exactly one of --user-id or --token")
+    if user_id and not User.objects(id=user_id).first():
+        raise click.ClickException(f"User {user_id} not found")
```
ThibaudDauce: Already in `_resolve_token_option`, maybe extract a `_check_credentials`?

udata/geopf/commands.py
```diff
@@ -0,0 +1,103 @@
...
+
+    datastore_id = _require_datastore_id(datastore_id)
```
ThibaudDauce, line 63: The doc says: "`--datastore-id` is required unless the dataset already has one pinned" but it's not what is done here?

udata/geopf/commands.py
```diff
@@ -0,0 +1,103 @@
...
+@datastore_id_option
+def push_metadata(dataset_id, user_id, token, datastore_id):
+    """Sync metadata for a dataset to Géoplateforme."""
+    access_token = _resolve_token_option(user_id, token)
+    datastore_id = _require_datastore_id(datastore_id)
+
+    try:
+        dataset = Dataset.objects.get(id=dataset_id)
+    except Dataset.DoesNotExist:
+        raise click.ClickException(f"Dataset {dataset_id} not found")
```
ThibaudDauce: We use `exit_with_error` in other commands but not really important…

udata/geopf/api.py
```diff
@@ -0,0 +1,263 @@
...
+    push_resource_to_geopf,
+    set_dataset_extras,
+    set_resource_extras,
+)
+
+ns = api.namespace("geopf", "Géoplateforme related operations")
+
+DATASET_SESSION_KEY = "geopf_oauth_dataset_id"
+
+common_doc = {"params": {"dataset": "The dataset ID or slug", "rid": "The resource ID"}}
```
ThibaudDauce: Not all endpoints have a `rid`?

udata/geopf/api.py
```diff
@@ -0,0 +1,263 @@
...
+    @api.response(400, "Unsupported resource format or missing datastore_id")
+    @api.response(404, "Resource not found")
+    @api.response(424, "Not connected to Géoplateforme")
+    def post(self, dataset, rid):
+        """Push a resource to Géoplateforme, as the current user."""
+        dataset.permissions["edit_resources"].test()
+
+        resource = get_by(dataset.resources, id=rid)
+        if resource is None:
+            api.abort(404, "Resource not found")
```
ThibaudDauce: Is this the same as `ResourceMixin.get_resource_or_404`?

udata/geopf/api.py
```diff
@@ -0,0 +1,263 @@
...
+
+        Resources are split into those eligible for a push (format in
+        `GEOPF_PUSHABLE_FORMATS`) and those that came back from a pull as
+        offerings; the two are disjoint.
+
+        A `status` of `null` means "never run"; otherwise it is one of
+        `pending`, `done`, `error` or `timeout`.
+        """
+        # endpoint only returns public data
+        dataset.permissions["read"].test()
```
ThibaudDauce: We abort(404) instead of 403 to avoid leaking private datasets.


```suggestion
        if not dataset.permissions["read"].can():
              if not dataset.private and dataset.deleted:
                  api.abort(410)
              api.abort(404)
```

(a little long, could be refactor one day…)

udata/geopf/api.py
```diff
@@ -0,0 +1,263 @@
...
+                f"Only {', '.join(sorted(pushable_formats))} resources can be pushed to Géoplateforme",
+            )
+
+        user = current_user._get_current_object()
+        try:
+            resolve_access_token(user=user)
+        except GeopfReauthRequired:
+            api.abort(424, "Not connected to Géoplateforme")
+
+        datastore_id = (request.get_json(silent=True) or {}).get("datastore_id")
```
ThibaudDauce, line 216: Maybe validate that `datastore_id` is a UUID?

ThibaudDauce (changes requested):
This PR adds a lot of stuff inside extras, I'm wondering if it's better to have a custom embed for this feature, it will allow to have better typing (real dates, no problem with the `None`…), better security (read only field on the API) and less magic? We have a lot of problem with old extras for harvesting (before DatasetHarvestMetadata and ResourceHarvestMetadata) and partial migrations leaving weird states in the database…

Tests mock the client with `MagicMock` while `test_client.py` already runs the real one against `rmock`. We never assert what we send (the srs for example) and the `user_id` path is never run?

udata/geopf/tasks.py
abulte: Fixed by downloading the "local" file if it can't be seeked. Bit of a shame to redownload but 🤷

udata/geopf/auth.py
abulte: Fixed

udata/geopf/auth.py
abulte: Fixed

udata/geopf/tasks.py
abulte: Moot with refactor to embedded docs.

udata/geopf/tasks.py
abulte: Moot with refactor to embedded docs.

udata/geopf/client.py
abulte: Not sure, no :-/ I truncated the content of the error body, it might protect us from sensitive traceback. In the end, it's more a problem for geopf than us.

udata/geopf/client.py
abulte: I don't think it's worth it at this point. FYI there'll always be at least one writable datastore (espace découverte), so the list should never be empty.

udata/geopf/tasks.py
abulte: Because geopf doesn't accept stored data name starting with a number 🤷

udata/geopf/tasks.py
abulte: We could but I kept it simple for a first version. Better to restart from scratch.

udata/geopf/client.py
abulte: Yep, fixed.

udata/geopf/tasks.py
abulte: It's tied to UI with a loader, so I'd rather not.

udata/geopf/tasks.py
abulte: Tied to UI, and this one is actually fast.

pyproject.toml
abulte: Removed the plugin.

udata/geopf/client.py
abulte: Added a configurable default timeout of 120 seconds.

udata/geopf/client.py
abulte: Fixed.

udata/geopf/tasks.py
abulte: Not worth it IMO, at least while we're not handling retries on failures. Restarting the upload will create a new stored data id anyway.

udata/geopf/tasks.py
abulte: Moot with refactor to embedded docs.

udata/geopf/tasks.py
abulte: Moot with refactor to embedded docs.

udata/geopf/commands.py
abulte: Yep, extracted.

udata/geopf/commands.py
abulte: True, but it's only a debug CLI and the failure paths are not catastrophic, while fixing it would not be trivial. I'd leave it like this.

udata/geopf/commands.py
abulte: Fixed.

udata/geopf/api.py
abulte: Fixed.

udata/geopf/api.py
abulte: Yep, factorized.

udata/geopf/api.py
abulte: Fixed.

udata/geopf/api.py
abulte: Not worth it IMO, id should come from geopf anyway IRL, and failure path is clean.

abulte:
> This PR adds a lot of stuff inside extras, I'm wondering if it's better to have a custom embed for this feature, it will allow to have better typing (real dates, no problem with the `None`…), better security (read only field on the API) and less magic? We have a lot of problem with old extras for harvesting (before DatasetHarvestMetadata and ResourceHarvestMetadata) and partial migrations leaving weird states in the database…

Refactored into embedded documents in latest version.

> Tests mock the client with `MagicMock` while `test_client.py` already runs the real one against `rmock`. We never assert what we send (the srs for example) and the `user_id` path is never run?

Added asserts on call args, but maybe you meant a more systematic call assertion system? And added tests on push and pull that exercise user -> token path.

abulte:
Frontend in sync https://github.com/datagouv/cdata/pull/1178/commits/b2c2757cd1252c556c217c87f7c31144bdfb53b2

udata/mongo/encrypted_field.py
```diff
@@ -0,0 +1,56 @@
...
+    (e.g. ``udata.mongo.datetime_fields.DateField``), in-memory access
+    before a save/reload round-trip sees the plaintext as assigned.
+
+    Stored values carry a ``fernet:`` prefix so decryption only ever runs
+    on actual ciphertext, and a decryption failure (wrong or rotated key)
+    raises loudly instead of silently passing ciphertext through as the
+    value.
+    """
+
+    def __init__(self, key_config="GEOPF_TOKEN_ENCRYPTION_KEY", **kwargs):
```
ThibaudDauce: Specific geopf key in a generic `udata/mongo/encrypted_field`?

udata/geopf/auth.py
```diff
@@ -0,0 +1,106 @@
...
+from .client import GeopfReauthRequired
+from .models import GeopfToken
+
+log = logging.getLogger(__name__)
+
+oauth = OAuth()
+
+
+def init_app(app):
+    if app.config.get("GEOPF_OAUTH_OPENID_CONF_URL"):
```
ThibaudDauce: `GEOPF_OAUTH_OPENID_CONF_URL` have a default value, maybe check `GEOPF_OAUTH_CLIENT_ID`?

udata/geopf/auth.py
```diff
@@ -0,0 +1,106 @@
...
+                "client_secret": current_app.config.get("GEOPF_OAUTH_CLIENT_SECRET"),
+            },
+            timeout=10,
+        )
+        resp.raise_for_status()
+    except Exception:
+        log.warning(f"geopf: failed to revoke token for user={geopf_token.user.id}", exc_info=True)
+
+
+def resolve_access_token(user=None, raw_token: str | None = None, min_validity: int = 0) -> str:
```
ThibaudDauce: Is this ever called with `raw_token` outside tests?

udata/geopf/tasks.py
```diff
@@ -0,0 +1,558 @@
...
+        self.url = url
+
+    def _chunks(self):
+        with ssrf_session().get(self.url, stream=True, timeout=60) as resp:
+            resp.raise_for_status()
+            yield from resp.iter_content(65536)
+
+    @property
+    def _subject(self) -> str:
+        return f"Remote file at {self.url}"
```
ThibaudDauce: It could be simpler with just functions, no?

```suggestion
@contextlib.contextmanager
def _copy_to_tempfile(chunks, subject: str):
    """Write a chunk stream to a local temp file, yield it, clean up on exit.

    `subject` describes what's being copied, for the size-limit error message."""
    max_size = current_app.config["GEOPF_MAX_FILE_SIZE"]
    with tempfile.NamedTemporaryFile(suffix=".gpkg") as tmp:
        size = 0
        for chunk in chunks:
            size += len(chunk)
            if size > max_size:
                raise GeopfError(f"{subject} exceeds GEOPF_MAX_FILE_SIZE ({max_size} bytes)")
            tmp.write(chunk)
        tmp.seek(0)
        yield tmp


def _storage_chunks(fs_filename: str):
    """Stream a stored resource, chunk by chunk."""
    with storages.resources.open(fs_filename, "rb") as src:
        yield from iter(lambda: src.read(65536), b"")


def _download_chunks(url: str):
    """Stream a remote URL, chunk by chunk.

    `url` is user-controlled (remote resource URL) and this fetch runs on the
    worker, so it goes through the SSRF-hardened session.
    """
    with ssrf_session().get(url, stream=True, timeout=60) as resp:
        resp.raise_for_status()
        yield from resp.iter_content(65536)
```

`_open_ressource` do not change:

```
def _open_resource_file(resource):
    """Return a context manager yielding an open, seekable binary file for the resource."""
    if resource.filetype == "file" and resource.fs_filename:
        try:
            return open(storages.resources.path(resource.fs_filename), "rb")
        except OperationNotSupported:
            # No direct filesystem access (e.g. the S3 backend): download the file
            return _copy_to_tempfile(
                _storage_chunks(resource.fs_filename),
                f"Stored resource {resource.fs_filename}",
            )
    return _copy_to_tempfile(_download_chunks(resource.url), f"Remote file at {resource.url}")
```

udata/geopf/client.py
```diff
@@ -0,0 +1,287 @@
...
+        super().__init__()
+        self.default_timeout = timeout
+
+    def request(self, *args, **kwargs):
+        kwargs.setdefault("timeout", self.default_timeout)
+        return super().request(*args, **kwargs)
+
+
+class GeopfClient:
+    def __init__(self, token: str | None = None, datastore_id: str | None = None):
```
ThibaudDauce: I don't see any call without a token except in tests?

udata/geopf/models.py
```diff
@@ -0,0 +1,146 @@
...
+
+class GeopfToken(db.Document):
+    """Per-user OAuth2 tokens used to call the Géoplateforme entrepôt API on their behalf.
+
+    One token per data.gouv.fr user, obtained via the authorization_code flow
+    against geopf's Keycloak (see udata/geopf/api.py) and refreshed as needed
+    before each call (see udata/geopf/auth.py).
+    """
+
+    user = db.ReferenceField(User, required=True, unique=True, reverse_delete_rule=db.CASCADE)
```
ThibaudDauce, line 133: `reverse_delete_rule=db.CASCADE` never trigger, we don't delete users, we mark them as deleted. You need to update `mark_as_deleted` for `User`

udata/geopf/api.py
```diff
@@ -0,0 +1,246 @@
...
+        `GEOPF_PUSHABLE_FORMATS`) and those that came back from a pull as
+        offerings; the two are disjoint.
+
+        A `status` of `null` means "never run"; otherwise it is one of
+        `pending`, `done`, `error` or `timeout`.
+        """
+        if not dataset.permissions["read"].can():
+            if not dataset.private and dataset.deleted:
+                api.abort(410, "Dataset has been deleted")
+            api.abort(404)
```
ThibaudDauce: Do we need to check for `edit_resources` or not here? Everyone can read a public dataset in data.gouv

udata/geopf/api.py
```diff
@@ -0,0 +1,246 @@
...
+
+        # Mark pending before enqueueing, so the status route doesn't read as
+        # "never pushed" until a worker picks the task up. The id only exists
+        # after `.delay()`, hence the second write.
+        set_resource_push_metadata(dataset, resource, status="pending", error=None)
+        task = push_resource_to_geopf.delay(
+            str(dataset.id), str(resource.id), str(user.id), datastore_id
+        )
+        set_resource_push_metadata(dataset, resource, task_id=task.id)
+        return {"task_id": task.id}, 202
```
ThibaudDauce, line 228: Should we check `resource_push_metadata(resource).status == "pending"` before queuing another push?

udata/geopf/client.py
```diff
@@ -0,0 +1,287 @@
...
+        ]
+
+    # --- offerings ---
+
+    def list_offerings(self, stored_data_id: str) -> list:
+        """Return offerings for a stored_data, including urls and type."""
+        resp = self.session.get(
+            self._url("offerings"),
+            params={"stored_data": stored_data_id, "fields": "urls,type,layer_name,status,open"},
+        )
```
ThibaudDauce: Should we have a limit here? I don't know what's the default `limit` in geopf but since we use this function to sync (and delete the local orphans), are you sure 1 call without pagination and limit return all the data?

udata/geopf/client.py
```diff
@@ -0,0 +1,287 @@
...
+            data={"type": "ISOAP", "open_data": "true"},
+            files={"file": ("metadata.xml", io.BytesIO(xml_bytes), "application/xml")},
+        )
+        if resp.status_code == 409:
+            fid = _extract_file_identifier(xml_bytes)
+            existing_id = self._find_metadata_id(fid)
+            if existing_id:
+                return self.update_metadata(existing_id, xml_bytes)
+            raise GeopfError(
+                f"409 on metadata upload, could not locate existing record: {resp.text}"
```
ThibaudDauce: This one is not truncated?

udata/geopf/client.py
```diff
@@ -0,0 +1,287 @@
...
+class _TimeoutSession(requests.Session):
+    """A requests.Session with a default per-request timeout."""
+
+    def __init__(self, timeout: float):
+        super().__init__()
+        self.default_timeout = timeout
+
+    def request(self, *args, **kwargs):
+        kwargs.setdefault("timeout", self.default_timeout)
+        return super().request(*args, **kwargs)
```
ThibaudDauce: Maybe convert `RequestException` to GeopfError here? `_raise` only catch response exceptions but not connection…

And we sometime do:

```
except GeopfError as e:
      log.warning("geopf: could not delete upload=%s …")
```

(a 120s timeout will not trigger this `except`)

The processing already succeeded at this point, so we go to the `except Exception`, we never reach `set_resource_push_metadata(…, stored_data_id=…)`, and `pull_offerings_for_dataset` will never find this stored data again…

udata/geopf/metadata.py
```diff
@@ -0,0 +1,251 @@
...
+        date_stamp=dataset.last_modified.strftime("%Y-%m-%d"),
+        title=escape(dataset.title),
+        created_at=dataset.created_at.strftime("%Y-%m-%d"),
+        abstract=escape(abstract),
+        point_of_contact_block=point_of_contact_block,
+        keywords_block=keywords_block,
+        topic_category_block=topic_category_block,
+        extent_block=extent_block,
+    )
+    return xml.encode("utf-8")
```
ThibaudDauce: Maybe use `escape_xml_illegal_chars` from the RDF code here too?

udata/geopf/models.py
```diff
@@ -0,0 +1,146 @@
...
+    )
+    last_synced_at = field(
+        DateTimeField(), readonly=True, allow_null=True, description="Last successful pull date"
+    )
+    error = field(
+        StringField(), readonly=True, allow_null=True, description="The last pull error, if any"
+    )
+    task_id = field(
+        StringField(), readonly=True, allow_null=True, description="The last pull Celery task's id"
+    )
```
ThibaudDauce, line 55: `GET /api/1/workers/tasks/<id>/` is `admin_permission` only (`udata/core/jobs/api.py:146`), so an editor can't do anything with this id? The doc says it's the bridge to the Celery result (`docs/geopf.md:110`) but they'll get a 403. Not sure if it's a real problem or not, but maybe worth noting if you plan to use it in the frontend

udata/geopf/tasks.py
ThibaudDauce: maybe add a comment in the code to explain this for future us :-)

udata/geopf/models.py
abulte: The frontend only uses the documents' metadata for now. I clarified the doc. It will come handy if we need an admin level management of the jobs.

udata/geopf/client.py
abulte: Good idea, done in a new `_GeopfSession`.

udata/geopf/client.py
abulte: Good catch, I implemented the geopf pagination.

udata/geopf/api.py
abulte: Good catch, done (not for pull, it doesn't matter there).

udata/geopf/api.py
abulte: Yes, fixed. The frontend also gates on edit, so it's fine.

udata/geopf/models.py
abulte: Fixed in `mark_as_deleted`. I left the reverse rule because it makes sense re the model, even if never triggered.

udata/geopf/client.py
abulte: Good catch, removed.

udata/geopf/tasks.py
abulte: Converted.

udata/geopf/auth.py
abulte: Good catch, removed.

udata/geopf/auth.py
abulte: Yep, changed to id.

udata/mongo/encrypted_field.py
abulte: Moved into caller.

abulte:
Added minimal public info for geopf in 558356a1bb1e8432f3a4a2e871f3775ba8260044. The pattern is now:
- full geopf status available through an edit-only route
- minimal selection on public route. need eg for disabling the edit button on frontend.

PR #3833, opened 2026-06-16
feat: add dataset check-files command to clean up orphan resource files

ThibaudDauce:
The FS is telling us 6.5To used but the MongoDB database give:

Dataset resources : 147,921 | 1.6 TiB (1,770,172,415,342.0 o) | 2,290 sans filesize
Community resources : 1,107 | 24.4 GiB (26,176,521,412.0 o) | 0 sans filesize
TOTAL hébergé : 1.6 TiB (1,796,348,936,754.0 o)

We could scan faster without using `os.stat()` I think, by just listing files and comparing but we'll loose the reclaimable space info and the security of the mtime (to not delete new files, but we could do a second pass on MongoDB to confirm the new files)

### DIFF
--- a/udata/core/dataset/commands.py
+++ b/udata/core/dataset/commands.py
@@ -262,13 +262,23 @@ def check_files(delete, days, output, yes):
     header("Deleting orphan files")
     deleted = 0
     errors = 0
-    for fs_filename in orphans:
-        try:
-            storage.delete(fs_filename)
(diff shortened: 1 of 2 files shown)


ThibaudDauce:
Will do with a one shot script or better maybe just copy the existing from files from the MongoDB listing to S3 and forget the others in NFS…

PR #3834, opened 2026-06-16
fix: protect hosted resource checksum

ThibaudDauce:


udata/core/dataset/forms.py
```diff
@@ -31,6 +33,14 @@
 
 __all__ = ("DatasetForm", "ResourceForm", "CommunityResourceForm")
 
+# Fields computed by the server at upload time for resources hosted on our
+# file storage. They must not be overridden by API clients: a client sending
+# stale metadata (e.g. fetched before a new file upload) would otherwise
+# overwrite the values describing the currently hosted file.
+# Same reasoning as the `url` protection from
+# https://github.com/opendatateam/udata/issues/2544
+HOSTED_RESOURCE_PROTECTED_FIELDS = ("filetype", "url", "checksum", "filesize")
```
Pierlou: I'm never sure: is `mime` one of these too?

ThibaudDauce: Done in cb8f2949b

Pierlou (approved):
Thanks for the fix!

PR #3836, opened 2026-06-16
perf: optimize harvest clean command

streino:
:warning: Stacked PR over https://github.com/opendatateam/udata/pull/3835, will need to be rebased once parent is merged.

Optimizes `udata harvest clean` by only recomputing metrics at the end of the cleanup.

Before/after on a local 700-dataset source: 1432s -> 15s.


ThibaudDauce (approved):
Just added a test (to see if I can run the CI with a commit on your branch but no :-( )

streino:
Thanks for the test :) GH is slow to catch up, but hopefully the branch merge will trigger the tests. Ran it locally anyway so we should be good.

PR #3837, opened 2026-06-17
chore(harvest): improve CSW request XML compliance

maudetes:
Fix CSW request to [PyCSW](https://pycsw.org/) instances, ie `udata dcat parse-url --iso https://meta.atmosud.org`

### DIFF
--- a/udata/harvest/backends/dcat.py
+++ b/udata/harvest/backends/dcat.py
@@ -301,6 +301,7 @@ class BaseCswDcatBackend(DcatBackend, ABC):
     <csw:GetRecords xmlns:apiso="http://www.opengis.net/cat/csw/apiso/1.0"
                     xmlns:csw="http://www.opengis.net/cat/csw/2.0.2"
                     xmlns:ogc="http://www.opengis.net/ogc"
+                    xmlns:gmd="http://www.isotc211.org/2005/gmd"
                     service="CSW" version="2.0.2" outputFormat="application/xml"
                     resultType="results" startPosition="{start}" maxRecords="25"
                     outputSchema="{output_schema}">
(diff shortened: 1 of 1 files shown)


ThibaudDauce (approved):

streino:
@maudetes @ThibaudDauce 

This change breaks GeoIDE harvests :( 
```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?><ExceptionReport version="1.2.0" xmlns="http://www.opengis.net/ows" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/ows ../../../ows/1.0.0/owsExceptionReport.xsd"><Exception exceptionCode="NoApplicableCode" locator=""><ExceptionText>Erreur dans l'analyse de la requête</ExceptionText></Exception></ExceptionReport>
```

And looking at OGC reference examples, `apiso:type` is lowercase: https://schemas.opengis.net/csw/2.0.2/profiles/apiso/1.0.0/examples-Catalogue/GetRecordsMultiTypes.xml.

ThibaudDauce:
> @maudetes @ThibaudDauce
> 
> This change breaks GeoIDE harvests :(
> 
> ```
> <?xml version="1.0" encoding="UTF-8" standalone="yes"?><ExceptionReport version="1.2.0" xmlns="http://www.opengis.net/ows" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/ows ../../../ows/1.0.0/owsExceptionReport.xsd"><Exception exceptionCode="NoApplicableCode" locator=""><ExceptionText>Erreur dans l'analyse de la requête</ExceptionText></Exception></ExceptionReport>
> ```
> 
> And looking at OGC reference examples, `apiso:type` is lowercase: https://schemas.opengis.net/csw/2.0.2/profiles/apiso/1.0.0/examples-Catalogue/GetRecordsMultiTypes.xml.

Trying something in https://github.com/opendatateam/udata/pull/3839 (ping @maudetes )

ThibaudDauce:
@streino Do you have an idea to support both servers? Something magical we could try or do we need to add a config to the source to specify the spec followed by the server?

ThibaudDauce:
@streino there is no GeoIDE container we could run in CI to test it too? Because GeoNetwork accept both case.

ThibaudDauce:
@streino The other solution (from a config key on the source) is to try first with uppercase and if it fails retry the request with the lowercase syntax… What do you think of it?

maudetes:
Oh god, I tested GeoNetwork but not GeoIDE :sweat_smile:
No more custom implementation please :crossed_fingers:

If the ground truth is `apiso:type`, why is pycsw Certified [OGC Compliant](https://www.ogc.org/resources/product-details/?pid=1661) and is an [OGC Reference Implementation](https://demo.pycsw.org/)?
Should we open an issue there? How come no one else stumble on this? Is there something we're missing?

In the meantime, I don't have much preference for the retry or config key approach, as I hope they can be temporary :crossed_fingers:

streino:
Aah, the wonderful world of flaky specs ✨😅 Kidding aside, I have no idea where pycsw took its spec and whether one version is more correct than the other...
An alternative to what @ThibaudDauce is suggesting is detecting geoide url and picking lowercase only in this situation. There's only one geoide url... and it might not be around too long if they migrate to the GPF. Might be the simplest solution.
I'm not aware of a geoide docker but i can ask if you need it.

ThibaudDauce:
@streino I'll defer to your expertise on the solution to adopt. If you say that geoide is rare, in the process of being migrated, and that there's only one URL to check, then yes, that's simpler than building a retry system.
In your opinion, which spec is better ucfirst or lowercase?

streino:
GeoIDE is centralized hosting by the MTE, so only one base url for sure. I'm not in front of the computer today but you can look it up on a DDT harvester config in prod, or I'll post it here tomorrow.

If pycsw doesn't like lowercase and geonetwork doesn't care, I'd say let's keep the change from this PR + exception for geoide. Should cover all cases 🤞

maudetes:
Yes, the base URL for GeoIDE is https://ogc.geo-ide.developpement-durable.gouv.fr/.

For your information, I've posted an issue on pycsw: https://github.com/geopython/pycsw/issues/1235

streino:
Thanks @maudetes for the pycsw ticket.
Alos, I got the answer from GeoIDE: no docker, they vms to openstack so don't have a need for it.

PR #3839, opened 2026-06-18
fix(harvest): allow sending lowercase properties to GeoIDE

ThibaudDauce:
Need to add `HARVEST_CSW_LOWERCASE_APISO_HOSTS = ["ogc.geo-ide.developpement-durable.gouv.fr"]` into production

Related to https://github.com/opendatateam/udata/pull/3837

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,6 +108,74 @@ jobs:
           path: reports/
           destination: reports
 
+  # Experimental: spin up real CSW servers (PyCSW + GeoNetwork 3.x) as service
+  # containers and send them our actual GetRecords request, to measure how long
+  # booting them takes in CI and which `apiso:` property casing each accepts.
+  # GeoNetwork (Java + H2 + Lucene) is the slow one — its "Wait for GeoNetwork"
+  # step duration is the timing signal we care about.
+  csw-integration:
+    resource_class: large
+    docker:
+      - image: ghcr.io/astral-sh/uv:python3.11-trixie
+      - image: mongo:7.0.28
+      - image: redis:alpine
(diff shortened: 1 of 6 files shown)


ThibaudDauce:
@maudetes what do you think of adding real pycsw and geonetwork to the CI to do E2E tests? The timings are not so bad…

maudetes:
> @maudetes what do you think of adding real pycsw and geonetwork to the CI to do E2E tests? The timings are not so bad…

Good idea I think. Since it runs in parallel, seems ok! I think it's quite overkill for this specific case but we can then build on it as we go

.circleci/config.yml
```diff
@@ -108,6 +108,74 @@ jobs:
...
+  # GeoNetwork (Java + H2 + Lucene) is the slow one — its "Wait for GeoNetwork"
+  # step duration is the timing signal we care about.
+  csw-integration:
+    resource_class: large
+    docker:
+      - image: ghcr.io/astral-sh/uv:python3.11-trixie
+      - image: mongo:7.0.28
+      - image: redis:alpine
+      - image: geopython/pycsw:2.6.1
+      - image: geonetwork:3.12.12
```
streino, line 123: 3.12 is pretty old, the majority of catalogs being on some version of 4.x now. But 4.x requires a more complex setup (elastic at least) and I don't know enough about the 3->4 changes to evaluate if it's worth the effort for what you want to be testing.

streino (approved):

ThibaudDauce:
The bug was reverted in https://github.com/opendatateam/udata/pull/3843, let's wait the response of the issue in https://github.com/geopython/pycsw/issues/1235

PR #3843, opened 2026-06-23
chore(harvest): partially revert csw xml update

maudetes:
See discussions in https://github.com/opendatateam/udata/pull/3837 for context.

Waiting for an answer from pycsw, we keep lowercase `apiso:title` and `apiso:identifier` that match the standard.
We keep the `gmd` namespace addition that is compliant.

If we don't have an answer or we need to add a workaround, see https://github.com/opendatateam/udata/pull/3839

### DIFF
--- a/udata/harvest/backends/dcat.py
+++ b/udata/harvest/backends/dcat.py
@@ -301,6 +301,7 @@ class BaseCswDcatBackend(DcatBackend, ABC):
     <csw:GetRecords xmlns:apiso="http://www.opengis.net/cat/csw/apiso/1.0"
                     xmlns:csw="http://www.opengis.net/cat/csw/2.0.2"
                     xmlns:ogc="http://www.opengis.net/ogc"
+                    xmlns:gmd="http://www.isotc211.org/2005/gmd"
                     service="CSW" version="2.0.2" outputFormat="application/xml"
                     resultType="results" startPosition="{start}" maxRecords="25"
                     outputSchema="{output_schema}">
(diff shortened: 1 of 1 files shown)


ThibaudDauce (approved):

udata/harvest/backends/dcat.py
```diff
@@ -301,6 +301,7 @@ class BaseCswDcatBackend(DcatBackend, ABC):
     <csw:GetRecords xmlns:apiso="http://www.opengis.net/cat/csw/apiso/1.0"
                     xmlns:csw="http://www.opengis.net/cat/csw/2.0.2"
                     xmlns:ogc="http://www.opengis.net/ogc"
+                    xmlns:gmd="http://www.isotc211.org/2005/gmd"
```
streino, line 304: Interesting that you had to keep that. The `gmd` namespace is not used anywhere in the XML, so I don't see why it would be needed. The only occurrence of `gmd` is in `csw:Query@typeName`, but as a string, so should require an `xmlns` declaration.

maudetes: It was required for pycsw, due to `<csw:Query typeNames="gmd:MD_Metadata">`.
Minimal reproduction:
```
curl -X POST https://demo.pycsw.org/gisdata/csw -d '
<csw:GetRecords xmlns:apiso="http://www.opengis.net/cat/csw/apiso/1.0"
                xmlns:csw="http://www.opengis.net/cat/csw/2.0.2"
                xmlns:ogc="http://www.opengis.net/ogc"
                service="CSW" version="2.0.2" outputFormat="application/xml"
                resultType="results" startPosition="1" maxRecords="10"
                outputSchema="http://www.isotc211.org/2005/gmd">
  <csw:Query typeNames="gmd:MD_Metadata">
    <csw:ElementSetName>full</csw:ElementSetName>
  </csw:Query>
</csw:GetRecords>'
```
It did not bother GeoNetwork or GeoIDE :muscle:

streino: Thanks for the command to reproduce! 
I wouldn't expect XML namespaces to apply to content, and indeed `xmllint` is perfectly happy to validate the above XML (but raises an error if you remove `xmlns:csw`). So I'd be tempted to think pycsw is mixing concerns here and is wrong to return an error in this case.
Not sure it's worth bothering, but it gives an idea how much we should trust pycsw compliance.

PR #3846, opened 2026-06-25
fix: wiping metrics on metric-api failures

### LINKED ISSUE #2044: Manque de certains indicateurs sur la page dashboard du site
## Description du bug
Auparavant, d'autres indicateurs comme le nombre de downloads des ressources ou le nombre de visites sur les pages datasets au global était disponible sur la page https://www.data.gouv.fr/dashboard
ils n'apparaissent plus, il faudrait les remettre

ThibaudDauce:
- fix https://github.com/datagouv/data.gouv.fr/issues/2044

### DIFF
--- a/udata/core/site/models.py
+++ b/udata/core/site/models.py
@@ -102,9 +102,9 @@ class Site(WithMetrics, Document):
         from udata.models import Dataset
 
         self.metrics["datasets"] = Dataset.objects.visible().count()
-        self.metrics["datasets_visits_by_months"] = get_metrics_for_model(
-            "site", None, ["visit_dataset"]
-        )[0]
+        visits = get_metrics_for_model("site", None, ["visit_dataset"])
+        if visits is not None:
+            self.metrics["datasets_visits_by_months"] = visits[0]
         self.save()
 
     def count_resources(self):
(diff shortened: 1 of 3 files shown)


udata/core/site/models.py
```diff
@@ -102,9 +102,9 @@ def count_datasets(self):
         from udata.models import Dataset
 
         self.metrics["datasets"] = Dataset.objects.visible().count()
-        self.metrics["datasets_visits_by_months"] = get_metrics_for_model(
-            "site", None, ["visit_dataset"]
-        )[0]
+        visits = get_metrics_for_model("site", None, ["visit_dataset"])
+        if visits is not None:
```
maudetes, line 106: Could we use a walrus operator?
```suggestion
        if visits := get_metrics_for_model("site", None, ["visit_dataset"]):
```

maudetes (approved):
Clean!

udata/core/site/models.py
ThibaudDauce: It'll match an empty array in the response, not sure it's what we want since we explicitly return `None` to say "keep the old value", other falsy value may not have the same meaning (even if it's not possible right now)

PR #3848, opened 2026-06-25
fix: notification dereferences issues

nicolaskempf57:


ThibaudDauce (commented):
Isn't it the same code on both listeners? Why do we need both?

nicolaskempf57:
> Isn't it the same code on both listeners? Why do we need both?

I'll remove one, and add a comment to the other one that it work for both of them

udata/core/dataservices/notifications.py
```diff
@@ -60,3 +60,16 @@ def cleanup_dataservice_notifications(dataservice, **kwargs):
...
+@Dataset.on_delete.connect
+def cleanup_dataservice_and_reuse_dataset_notifications(dataset, **kwargs):
+    """Clean up dataservice and reuse notifications when a referenced dataset is deleted"""
+    from udata.features.notifications.models import Notification
+
+    try:
+        Notification.objects(details__dataset=dataset).delete()
+    except Exception as e:
+        log.error(
+            f"Error cleaning up dataservice notifications for deleted dataset {dataset.id}: {e}"
```
ThibaudDauce: Talking only about dataservice and not reuse in this log

ThibaudDauce (approved):
Maybe you could improve the tests by adding other notification to check the count is 1 instead of checking 0 (here the filter could be absent and the test will still pass if all notifications are removed)

nicolaskempf57:
Moved to https://github.com/opendatateam/udata/pull/3867

PR #3849, opened 2026-06-26
fix: activity derefences when user doesnt exist

nicolaskempf57:
Try to fix https://errors.data.gouv.fr/organizations/sentry/issues/294820/?project=12

### DIFF
--- a/udata/core/activity/api.py
+++ b/udata/core/activity/api.py
@@ -107,14 +107,18 @@ class SiteActivityAPI(API):
         safe_items = []
         for item in qs.queryset.items:
             try:
+                item.actor
+                if item.organization:
+                    item.organization
                 item.related_to
             except DoesNotExist as e:
                 log.error(e, exc_info=True)
-            else:
-                if isinstance(item.related_to, Owned):
-                    if not OwnableReadPermission(item.related_to).can():
-                        continue
-                safe_items.append(item)
+                continue
+
+            if isinstance(item.related_to, Owned):
+                if not OwnableReadPermission(item.related_to).can():
+                    continue
+            safe_items.append(item)
         qs.queryset.items = safe_items
 
         return qs
--- a/udata/core/activity/api.py
+++ b/udata/core/activity/api.py
@@ -108,8 +108,7 @@ class SiteActivityAPI(API):
         for item in qs.queryset.items:
             try:
                 item.actor
-                if item.organization:
-                    item.organization
+                item.organization
                 item.related_to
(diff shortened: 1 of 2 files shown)


udata/core/activity/api.py
```diff
@@ -107,14 +107,18 @@ def get(self):
         safe_items = []
         for item in qs.queryset.items:
             try:
+                item.actor
+                if item.organization:
+                    item.organization
```
ThibaudDauce: I think we don't need both, the `if` already dereference?

udata/tests/api/test_activities_api.py
```diff
@@ -147,3 +150,51 @@ def test_activity_api_list_with_private_visible_to_org_member(self) -> None:
...
+        response: TestResponse = self.get(url_for("api.activity"))
+        assert200(response)
+
+    def test_activity_api_list_with_dangling_related_to(self) -> None:
+        """It should not crash when an activity references a deleted object."""
+        dataset = DatasetFactory()
+        FakeDatasetActivity.objects.create(actor=UserFactory(), related_to=dataset)
+
+        # Simulate a manual/document purge that leaves the activity with a DBRef.
+        get_db().dataset.delete_one({"_id": dataset.id})
```
ThibaudDauce: Maybe use `Dataset._get_collection().delete_one` to not hardcode the collection name?

udata/tests/api/test_activities_api.py
```diff
@@ -147,3 +150,51 @@ def test_activity_api_list_with_private_visible_to_org_member(self) -> None:
...
+        assert200(response)
+        assert len(response.json["data"]) == 0
+
+    def test_activity_api_list_with_dangling_actor(self) -> None:
+        """It should not crash when an activity actor has been hard-deleted."""
+        actor = UserFactory()
+        FakeDatasetActivity.objects.create(actor=actor, related_to=DatasetFactory())
+
+        # Simulate a manual hard-delete of the user leaving a dangling DBRef in the activity.
+        get_db().user.delete_one({"_id": actor.id})
```
ThibaudDauce: Same here `User._get_collection().delete_one`?

ThibaudDauce (approved):

udata/core/activity/api.py
nicolaskempf57: actually, according to the tests we don't need the if and the `.organization` is enough

nicolaskempf57:
Moved to https://github.com/opendatateam/udata/pull/3868

PR #3851, opened 2026-06-29
perf: targeted update_one for resource extras instead of full save

ThibaudDauce:
` PUT /api/2/datasets/:id/resources/:id/extras ` is the top endpoint (called by hydra twice for each checks)

From haproxy logs:

| Count     | Mean   | P50    | P75    | P95    | P99    | Max      | 5XX      |
| --------: | -----: | -----: | -----: | -----: | -----: | -------: | -------: |
|2 659 701 | 786 ms | 267 ms | 958 ms | 2,71 s | 6,83 s | 300,00 s | 0,0374 % |

Bench before/after with this PR

| Number of resources | BEFORE (`resource.save()`) | AFTER (`update_one`) | Gain |
| -------------------: | -------------------------: | -------------------: | ---: |
| 1 000                | 338,2 ms                   | 46,6 ms              | ×7   |
| 4 000                | 1 357,0 ms                 | 288,2 ms             | ×5   |
| 8 000                | 2 745,6 ms                 | 670,4 ms             | ×4   |
| 15 000               | 5 041,3 ms                 | 1 261,8 ms           | ×4   |


udata/core/dataset/models.py
```diff
@@ -1037,6 +1047,30 @@ def update_resource(self, resource):
...
+        last_update and quality_cached are recomputed because they both depend on
+        resource extras: a remote resource's last_modified can come from its
+        `analysis:last-modified-at` extra, and check:available feeds
+        `all_resources_available`. last_update is set first so the recomputed
+        quality_cached (whose next_update_for_update_fulfilled_in_time derives from
+        last_update) stays consistent, exactly as Dataset.clean() would do on save.
+        """
+        self.last_update = self.compute_last_update()
+        Dataset.objects(id=self.id, resources__id=resource.id).update_one(
+            set__resources__S__extras=resource.extras,
```
nicolaskempf57, line 1100: Do we want to validate these extras before updating ?

nicolaskempf57 (approved):
We might want to reindex the dataset after the resource update to have the dataset correct last_update in ES too. This issue seems to already exist in add_resource / update_resource / remove_resource.

PR #3853, opened 2026-07-01
feat(org): derive organization zone from SIRET via Recherche d'entreprise

ThibaudDauce:
Enrich an organization's GeoZone (commune, département, région, EPCI) from its SIRET using the public "Recherche d'entreprises" API, replacing the manual attach-zone/detach-zone CLI commands.

- Look up the zone on org create and on SIRET change, preserving existing enrichment on inconclusive lookups (outage, malformed payload, missing GeoZone) so a transient failure never wipes a correct zone.
- Carry parents/ancestors hierarchy on GeoZone from the upstream export.


udata/core/organization/api_entreprise.py
```diff
@@ -0,0 +1,151 @@
...
+NATURE_JURIDIQUE_TO_LEVEL: dict[str, str] = {
+    "7210": "commune",  # Commune et commune nouvelle
+    "7220": "departement",  # Département
+    "7230": "region",  # Région
+    "7340": "epci",  # Pôle métropolitain
+    "7343": "epci",  # Communauté urbaine
+    "7344": "epci",  # Métropole
+    "7346": "epci",  # Communauté d'agglomération
+    "7347": "epci",  # Communauté de communes
+    "7348": "epci",  # Communauté ou syndicat d'agglomération nouvelle
```
abulte, line 28: I'm not sure we have all those "types" in the current geozones database (to be verified), but I guess it doesn't really matter since it's forward compatible. Might be confusing for someone that enters the code path from here though: why isn't my syndicate not mapped to a geozone?

udata/core/spatial/models.py
```diff
@@ -50,11 +50,19 @@ class GeoZone(WithMetrics, Document[GeoZoneQuerySet]):
     code = StringField(required=True)
     level = StringField(required=True)
     uri = StringField()
+    # Hierarchy rebuilt upstream from INSEE subdivisions (see the geozones DAG).
+    # `parents`: closest containing zones (e.g. department + EPCI of a commune).
```
abulte, line 54: Not part of this PR, but why is `parents` plural? Shouldn't there be only one parent, and if not what is the logic for "stopping" at a given level? Might be explained briefly here.

PR #3857, opened 2026-07-03
feat: detect geozone from spatial.geom

abulte:
This feed `dataset.extras['analysis:spatial:zones']` with one or more detected geozones from `dataset.spatial.geom`. 

We're not using `dataset.spatial.zones` because:
- we don't want to override a producer's available attribute
- there currently can't be both a geom and zones in `dataset.spatial`

I believe this belongs in `udata` since it's working on dataset's metadata. The `analysis:` namespace might be wrong though.

The matching is done by comparing the geozone bbox to the spatial geometry bbox. This makes it fast and doesn't require any geo lib. See https://github.com/abulte/dataset-geom-to-zone for more info on matching method. It's triggered through Celery on dataset creation and dataset geometry update.

This requires associating a bbox to geozones, which this PR introduces a CLI command for. See attached file for a working draft of the expected input file. This should probably be industrialised through a DAG and published on data.gouv.fr. See https://github.com/abulte/dataset-geom-to-zone for more info on building the required input file.

Performance insights:

```
# no cache
In [3]: %timeit zones = get_zone_bboxes()
557 ms ± 3.69 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# with cache
In [3]: %timeit zones = get_zone_bboxes()
19.6 ms ± 153 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [8]: %timeit bbox = geom_to_bbox(poissy)
27.3 μs ± 66.3 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [10]: %timeit detect_zone(bbox, zones)
19.3 ms ± 790 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

I'm leaving this in draft for now since I want to take more time to review the code and do more tests on the matching threshold, but don't hesitate to comment on the principle and implementation choices.

[zones_bboxes.json](https://github.com/user-attachments/files/29630035/zones_bboxes.json)

### DIFF
--- a/docs/administrative-tasks.md
+++ b/docs/administrative-tasks.md
@@ -187,6 +187,30 @@ It's possible to index or reindex only last modified documents.
 time udata search index -f 2022-02-20-20-02
 ```
 
+## Geospatial zones
+
+To load zone bounding boxes (used for automatic zone detection from a dataset's spatial
+geometry) onto existing `GeoZone` documents:
+
+```shell
+$ udata spatial load-geozones-bboxes <geozones-bboxes-file>
+```
+
+`<geozones-bboxes-file>` can be either a local path or a remote URL. It's a JSON file: a flat
+object mapping each zone id to its bounding box `[minx, miny, maxx, maxy]` (WGS84
+longitude/latitude), e.g.:
+
+```json
+{
+  "fr:departement:32": [-0.2821, 43.3108, 1.2032, 44.08],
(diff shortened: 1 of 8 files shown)


abulte:
Related https://github.com/ecolabdata/ecospheres/issues/1191

abulte:
@maudetes re our conversation yesterday, this is probably a good starting point. Do you have feedback on the big picture before I refresh this PR?

PR #3861, opened 2026-07-07
chore(docker): disable elasticsearch disk threshold

streino:
Prevents Elasticsearch from going read-only when disk space is low.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,6 +27,7 @@ services:
           - node.name=es01
           - cluster.name=es-docker-cluster
           - cluster.initial_master_nodes=es01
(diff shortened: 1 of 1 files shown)


bolinocroustibat (approved):
LGTM but isn't that a bit dangerous?

streino:
> LGTM but isn't that a bit dangerous?

In prod yes, but AFAIK this compose file is used only in tests and local environments (??). In those types of environments I don't see how this particular setting is useful, and can cause silent search misses if you don't think of looking in the logs for proper indexing.

ThibaudDauce (approved):

PR #3862, opened 2026-07-07
fix(harvest): handle DCAT contact points more robustly

streino:
Required for https://github.com/ecolabdata/ecospheres/issues/1087 as workaround to https://github.com/geonetwork/core-geonetwork/issues/9342.

Context:
- GeoNetwork uses the deprecated `vcard:org` metadata structure to expose contact point org name in its new GeoDCAT-AP mapping. Supporting `vcard:org` is therefore needed if we want to move to the new (clearly improved) mapping. I don't have information whether this deprecated VCARD mapping could be used in other DCAT expositions.
- Under `vcard:org`, GeoNetwork uses misspelled `vcard:organisation-name` (UK spelling) instead of the spec `vcard:organization-name` (US spelling). Handling the misspelling is a tiny addition in `unwrap` that we can remove once GeoNetwork fixes the bug and all catalogs are up-to-date.


maudetes (approved):
Is there a risk that it's going to create new contacts and not match the existing ones?

streino:
> Is there a risk that it's going to create new contacts and not match the existing ones?

The patch in itself presents no such risk because it affects only the new GeoNetwork GeoDCAT-AP mapping and we have no csw-dcat harvesters with GeoDCAT-AP enabled in prod. But, if we enable GeoDCAT-AP *before* merging this patch, then it will indeed create new contacts for individuals with an organization, since contact "signature" will go from "individual" to "individual (organization)". So it would be better to deploy this patch before enabling GeoDCAT-AP on prod harvesters.

~~I also forgot to mention that this only affects the "contact point" role, as it's the only one [using vCard](https://semiceu.github.io/GeoDCAT-AP/releases/3.1.0/#Kind). Other roles [use FOAF](https://semiceu.github.io/GeoDCAT-AP/releases/3.1.0/#Agent).~~ *Update:* GeoNetwork new mapping uses vCard for all contact types (another bug, I'll add a fix).

The other related question is whether the switch from old GeoNetwork DCAT mapping to GeoDCAT-AP will create new contacts. I'll look into this and comment back here.

streino:
So yes, switching from old DCAT to GeoDCAT-AP can indeed create new contacts, see below.

The results below rely on the latest commit, which changes the criteria for parsing vCard/FOAF, from using the contact role (per spec) to its rdf:type (more lenient). This fixes the update from my previous comment (i.e. GeoNetwork uses vCard for all contact roles). There was a comment in the code about this, so I guess it's not the first time we think about improving this logic.

https://www.data.gouv.fr/datasets/carte-des-habitats-intertidaux-et-subtidaux-du-site-natura-2000-fr5300032-belle-ile-en-mer-2006-partie-1-2-1

- Old DCAT mapping:  
  

- New GeoDCAT-AP mapping without patch:
  

- New GeoDCAT-AP mapping with patch:
  

- Original: 
  

https://www.data.gouv.fr/datasets/dcsmm2018-d9c1-cartographie-de-la-contamination-en-yessotoxines-chez-les-mollusques-bivalves-collectes-entre-2010-et-2015

- Old DCAT mapping:
  

- New GeoDCAT-AP mapping (with or without patch):
  *Patch has no impact, because source is ISO-19115-3 and mapping outputs FOAF instead of vCard.*
  

- Original:

streino:
https://www.data.gouv.fr/datasets/les-quartiers-de-la-ville-damiens (and other datasets from this org) need this PR with the switch to GeoDCAT-AP:

**before**


**now**


TOOD: make sure it reverts back to **before** onces PR is deployed.

streino:
@ThibaudDauce @maudetes What's the plan for this PR? [There will be new contacts created](https://github.com/opendatateam/udata/pull/3862#issuecomment-4926158232), do we need to do something about it? Thanks.

udata/rdf.py
```diff
@@ -392,31 +392,47 @@ def contact_points_from_rdf(rdf, prop, role, dataset, dryrun=False):
...
+                email = (
+                    rdf_value(contact_point, FOAF.mbox)
+                    or (contact_point_org and rdf_value(contact_point_org, FOAF.mbox))
+                    or None
+                )
+                email = email.replace("mailto:", "").strip() if email else None
+                contact_form = None
+            else:
+                log.warning(f"Unsupported contact point with rdf:type={rdf_type}")
+                continue
```
ThibaudDauce: Same for types outside both namespaces: a `dct:publisher` typed `org:Organization` (or `dct:Agent`) with `foaf:name`/`foaf:mbox` is read on `main`, skipped here. Note the checks aren't symmetric — `rdf_type in VCARD` is a namespace prefix match, `rdf_type in FOAF` only matches terms declared in rdflib's `DefinedNamespace`, so an unusual foaf type lands in this `else`. Do you see such cases in the catalogs you harvest?

udata/rdf.py
```diff
@@ -392,31 +392,47 @@ def contact_points_from_rdf(rdf, prop, role, dataset, dryrun=False):
...
-            )
-            email = (
-                rdf_value(contact_point, FOAF.mbox)
-                or (contact_point_org and rdf_value(contact_point_org, FOAF.mbox))
-                or None
-            )
-            email = email.replace("mailto:", "").strip() if email else None
-            contact_form = None
+            log.warning(f"Unsupported contact point with role={role}")
+            continue
```
ThibaudDauce, line 430: What about contact nodes with vcard/foaf properties but no `rdf:type`? Locally, a `dcat:contactPoint` bnode with only `vcard:fn` + `vcard:hasEmail`, and a `dct:publisher` with only `foaf:name` + `foaf:mbox`, are parsed on `main` but dropped here. No fixture covers it, so the tests wouldn't catch it. Common enough in the wild to keep the previous `prop`-based dispatch as a fallback instead of `continue`?

streino: Good catch. I'm surprised this isn't in the tests at all. I'll add them.

streino: That was a bit more involved than I expected... But now we should have much better coverage, and a fix on `contact_form` (used `hasUrl` instead of [`hasURL`](https://www.w3.org/TR/vcard-rdf/#d4e605), so it never matched).

udata/rdf.py
streino: Fixed, see https://github.com/opendatateam/udata/pull/3862/changes#r3667808765

udata/rdf.py
```diff
@@ -475,7 +527,7 @@ def contact_points_to_rdf(contacts, graph=None):
             if contact.email:
                 node.set(VCARD.hasEmail, URIRef(f"mailto:{contact.email}"))
             if contact.contact_form:
-                node.set(VCARD.hasUrl, URIRef(contact.contact_form))
+                node.set(VCARD.hasURL, URIRef(contact.contact_form))
```
streino, line 535: This is not directly related to the PR since it's `to_rdf`, but might as well fix it both ways.

maudetes:
> @ThibaudDauce @maudetes What's the plan for this PR? [There will be new contacts created](https://github.com/opendatateam/udata/pull/3862#issuecomment-4926158232), do we need to do something about it? Thanks.

It's okay I think, unused old contacts won't be a huge bother :+1:

udata/rdf.py
```diff
@@ -382,69 +385,118 @@ def contact_point_name(agent_name: str | None, org_name: str | None) -> str:
...
-            email = (
-                rdf_value(contact_point, FOAF.mbox)
-                or (contact_point_org and rdf_value(contact_point_org, FOAF.mbox))
-                or None
-            )
-            email = email.replace("mailto:", "").strip() if email else None
-            contact_form = None
+            rdf_type = None
+            if val := obj.value(RDF.type):
+                rdf_type = val.identifier
```
ThibaudDauce: `val` can be a `Literal` if the source writes `<rdf:type>vcard:Kind</rdf:type>` instead of `rdf:resource`, and `Literal` has no `identifier`? Everything else here logs and continues, but this one would raise and fail the whole item.

```suggestion
            rdf_type = None
            if isinstance(val := obj.value(RDF.type), RdfResource):
                rdf_type = val.identifier
```

udata/rdf.py
```diff
@@ -382,69 +385,118 @@ def contact_point_name(agent_name: str | None, org_name: str | None) -> str:
...
             continue
-        yield contact
+
+        yield cast(ContactPoint, contact)
+
+
+def guess_contact_point_type(obj: RdfResource, predicate: URIRef) -> URIRef | None:
+    predicates = set(r.identifier for r in obj.predicates())
+    has_vcard = any(p in VCARD for p in predicates)
+    has_foaf = any(p in FOAF for p in predicates)
```
ThibaudDauce: `contact_point_from_foaf` also reads `skos:prefLabel` and `org:memberOf`, but neither is in `FOAF`, so an untyped node with only those is dropped here while it was read on main (a `dct:creator` with just a `skos:prefLabel` gave a contact). Maybe check SKOS and ORG too?

The two tests on this call `contact_point_from_foaf` directly so they don't see it, could we add one going through `contact_points_from_rdf` with an untyped node under `dct:creator`?

udata/rdf.py
```diff
@@ -382,69 +385,118 @@ def contact_point_name(agent_name: str | None, org_name: str | None) -> str:
...
+        rdf_value(obj, VCARD.fn),
+        rdf_value(obj, VCARD["organization-name"])
+        or rdf_value(  # deprecated vcard:org spec
+            obj,
+            VCARD.org,
+            unwrap=[
+                VCARD["organization-name"],
+                VCARD["organisation-name"],  # GeoNetwork incorrectly uses UK spelling
+            ],
+        ),
```
ThibaudDauce: If the `vcard:org` node has neither spelling, `serialize_value` ends on `value.identifier.toPython()`, so we store `Géo2France (N4f2c8b1a…)` (or the URI with `<vcard:org rdf:resource="…"/>`)? The fallback makes sense for `dct:format` or `dct:license` but not really for a display name.

Maybe read the child explicitly instead of going through `unwrap`?

udata/rdf.py
ThibaudDauce: > Fixed, see #r3667808765

Hum… `guess_contact_point_type` is only called under `if rdf_type is None`, no? So a `dct:publisher` typed `org:Organization` (or `dct:Agent`, `prov:Agent`) with `foaf:name`/`foaf:mbox` still falls in the `else: continue` while it was read as FOAF on main.

And the two checks are still not symmetric: `VCARD` is a plain `Namespace` (prefix match) but `FOAF` is a `DefinedNamespace`, so it only matches terms rdflib declares.

udata/rdf.py
streino: Indeed I missed `SKOS` (will add), but didn't check `ORG` on purprose because it made no sense to me to have a `foaf:Person` with nothing but a `org:memberOf`: there should either be something about the person, or it should be a `foaf:Organization`. I figured it's better to ignore such degenerate case and warn. You think we should instead cover it?

udata/rdf.py
streino: IMO it doesn't make sense in any case to return anonymous ids. I can either (in my order of preference):
1. Never return anonymous ids in `rdf_value`, logging a warning instead, if we're confident enough with such a change (we don't handle anonymous ids well anyway: it leaks to the user).
2. Add an option to `rdf_value` to fallback (or not) on anonymous ids.
3. Read children explicitly, but that's a bit more code for a deprecated VCARD spec...

Any opinion?

ThibaudDauce: I don't have a opinion on the subject, if you think it's not worth it, let's not do it. Logging will give the producer an useful information?

udata/rdf.py
streino: RDF in XML really is a minefield... I'll see if I can actually handle that case instead of ignore it.

udata/rdf.py
ThibaudDauce: No if you thinks it's not worth it…

udata/rdf.py
streino: I'll check, and go with either 1 or 2 depending on how helpful the warning can be.

udata/rdf.py
streino: > So a `dct:publisher` typed `org:Organization` (or `dct:Agent`, `prov:Agent`) with `foaf:name`/`foaf:mbox` still falls in the `else: continue` while it was read as FOAF on main.

Do we really want to go that far? DCAT does specify domain for its properties. There's a confusion (observed live in some DCAT implementations) between FOAF and VCARD because roles use both (VCARD for contactPoint, FOAF for other roles), but you'd really have to ignore the spec to go with other ontologies (ORG or PROV; `dct:Agent` doesn't even exist).

I mean we *can* go even more lenient, but do we want to? In that case reverting to the previous role-based logic (vs proposed type-based logic, which I embarked in this PR because of [the comment suggesting it](https://github.com/opendatateam/udata/pull/3862/changes#diff-4f35f2bdfbe2d1c4030a075566c489864c6f196e2e7db29eed3c662d1e7a2c8fL395)) is probably the better solution.

> And the two checks are still not symmetric: `VCARD` is a plain `Namespace` (prefix match) but `FOAF` is a `DefinedNamespace`, so it only matches terms rdflib declares.

You're talking about the `has_X` checks in `guess_contact_point_type` here, right? If so, I'm not sure it matters in practice? 
- `has_vcard` will be more lenient and be more likely to call `contact_point_from_vcard`, but if properties aren't defined in the `Namespace` there's no chance we'll be looking for them, so we'll get an "Empty contact" log. Not ideal but I'd say acceptable. Best solution would be to use a `DefinedNamespace` for DCAT if we really want be schema-safe; that'd be a big shift in how we parse RDF and required a dedicated PR (we'd have to parse the various plain-`Namespace` ontologies like DCAT into the graph for a start). 
- `has_foaf` will be more strict and be less likely to call `contact_point_from_foaf`, so if properties we're guessing on aren't defined in the `DefinedNamespace` we'll more appropriately get an "Unsupported contact". 
- There is a corner case where we have a record with an undefined VCARD property in an otherwise FOAF record for a `predicate=DCAT.contactPoint`: we'll guess wrong and end up calling `contact_point_from_vcard` when it would be better to call `contact_point_from_foaf`. We'll end up with an "Empty contact" log for the same reason as `has_vcard` above. Not ideal, but this is a pretty evil case I'd be really surprised to see in practice.
- I don't see a case where the lack of symmetry would result in missing an actually valid contact. Did you have something in mind?

If we do want symmetry for whatever reason, I'd say same as above: let's revert to role-based instead of complexifying type-based even more.

streino: Thinking some more about this, I think it's just simpler to go back to role-based. DCAT roles define an expected type, so it's actually cleaner to go with the role as the main driver. What I will do though is add a fallback to the other type (vcard falls back to foaf, and reverse) to support degenerate implementations.

I'll push the proposal to another branch and we can decide if we like it better.

streino: Alternative here: https://github.com/ecolabdata/udata/pull/12. IMO better :) WDYT?

udata/rdf.py
streino: Obsolete if we adopt https://github.com/ecolabdata/udata/pull/12.

udata/rdf.py
streino: Obsolete if we adopt https://github.com/ecolabdata/udata/pull/12.

udata/rdf.py
streino: Going with ecolabdata#12.

udata/rdf.py
streino: I finally went with option 3. I want to refactor rdf_value at some point but this isn't the PR to do it, and going with option 1 or 2 was going to be too involved.

udata/rdf.py
```diff
@@ -388,69 +391,104 @@ def contact_point_name(agent_name: str | None, org_name: str | None) -> str:
...
         except mongoengine.errors.ValidationError as validation_error:
             log.warning(f"Unable to validate contact point: {validation_error}", exc_info=True)
             continue
-        yield contact
+
+        yield cast(ContactPoint, contact)
+
+
+def contact_point_from_vcard(obj: RdfResource) -> tuple[str | None, str | None, str | None]:
+    contact_point_org = obj.value(VCARD.org)  # deprecated vcard:org spec
```
ThibaudDauce: if a catalog writes `<vcard:org>Foo</vcard:org>` instead of a node, `value()` gives back a `Literal` (`_cast` only wraps `BNode`/`URIRef`) and `rdf_value` calls `.objects()` on it. It raises, and the `except` below only catches `ValidationError`, so the whole item fails. Skip when it's not a `RdfResource`? Same for `obj.value(FOAF.member)` below.

udata/rdf.py
```diff
@@ -388,69 +391,104 @@ def contact_point_name(agent_name: str | None, org_name: str | None) -> str:
...
+    return name, email, contact_form
+
+
+def contact_point_from_foaf(obj: RdfResource) -> tuple[str | None, str | None, str | None]:
+    contact_point_org = obj.value(ORG.memberOf)
+    contact_point_member = obj.value(FOAF.member)
+    name = contact_point_name(
+        rdf_value(obj, FOAF.name)
+        or rdf_value(obj, SKOS.prefLabel)
+        or (rdf_value(contact_point_member, FOAF.name) if contact_point_member else None),
```
ThibaudDauce: This is never reached when the org has a `foaf:name`, and it always does in our fixtures. And when it is reached (anonymous org), the member name goes in the agent slot with no org name, so we get `Jean` and never `Jean (Géo2France)`, which is what the `org:memberOf` direction gives. Was it meant to be the other way around?

udata/rdf.py
```diff
@@ -388,69 +391,104 @@ def contact_point_name(agent_name: str | None, org_name: str | None) -> str:
...
+        or rdf_value(obj, VCARD.url)
+        or rdf_value(obj, VCARD.hasUrl)  # udata previously used incorrect case
+        or None
+    )
+    return name, email, contact_form
+
+
+def contact_point_from_foaf(obj: RdfResource) -> tuple[str | None, str | None, str | None]:
+    contact_point_org = obj.value(ORG.memberOf)
+    contact_point_member = obj.value(FOAF.member)
```
ThibaudDauce: `geonetwork-dcat-page-3.xml` has two `foaf:member` on the same org and `value()` keeps only one, so we pick one email over the other. Not a real problem but just wanted to note.

udata/rdf.py
streino: Good catch! 

`rdf_value` really gives a false sense of safety, and we already have a few places in the code that will cause the same sort of failure ([resource checksum](https://github.com/opendatateam/udata/blob/main/udata/core/dataset/rdf.py#L907), [dataset themes](https://github.com/opendatateam/udata/blob/main/udata/rdf.py#L357)). 

I'll fix it here but that's one more reason to refactor `rdf_value` (will get to that soon hopefully).

udata/rdf.py
streino: If we want to deal with this we can, but there's already a lot of code that behaves like this, so I'd say we'd need to deal with multi-values in a more generic way instead of just here.

udata/rdf.py
streino: I'm not entirely sure I understood your example, but I improved handling of contacts with `member` or `memberOf`. Does it cover your case?

ThibaudDauce (approved):

PR #3864, opened 2026-07-09
feat(harvest,catalog): sync dataset/dataservice harvest info

streino:
Needed for https://github.com/ecolabdata/ecospheres/issues/1130.

Changes:
- Expose missing `harvest` information in dataset and dataservices catalog exports.
- Add `harvest.modified_at` to dataservices, for symmetry with dataset harvest info.
- Reorganize fields to facilitate comparison between the datasets/dataservices harvest structures.

### DIFF
--- a/udata/core/dataservices/rdf.py
+++ b/udata/core/dataservices/rdf.py
@@ -89,7 +89,9 @@ def dataservice_from_rdf(
     )
     dataservice.harvest.created_at = rdf_value(d, DCT.created)
     dataservice.harvest.issued_at = rdf_value(d, DCT.issued)
-    dataservice.metadata_modified_at = rdf_value(d, DCT.modified)
+    dataservice.harvest.modified_at = rdf_value(d, DCT.modified)
+
+    dataservice.metadata_modified_at = dataservice.harvest.modified_at
 
     dataservice.tags = themes_from_rdf(d)
 
--- a/udata/core/dataset/models.py
+++ b/udata/core/dataset/models.py
@@ -120,18 +120,23 @@ def get_json_ld_extra(key, value):
 @generate_fields()
 class HarvestDatasetMetadata(EmbeddedDocument):
     backend = StringField()
+    domain = StringField()
+
+    source_id = StringField()
+
+    remote_id = StringField()
+    remote_url = URLField()
+
+    uri = StringField()
+
     created_at = DateTimeField()
     issued_at = DateTimeField()
     modified_at = DateTimeField()
-    source_id = StringField()
-    remote_id = StringField()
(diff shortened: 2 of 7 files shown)


streino:
I'm not sure how to easily test this locally because the "export-csv" job requires an existing catalog dataset. Is there a way to initialize it using `udata` commands, or do I have to create an appropriate dataset manually?

Edit: Tried creating a dataset manually. "export-csv" then creates the resources but I get 404 when trying to retrieve them. If you have instructions or a setup where you can test I'd appreciate :pray: 

Edit2: Hacked the csv tmp files to test and it looks OK (for datasets, because for dataservices none has harvest info, I'm guessing something's wrong with harvesting services -- I'll have a look later). I'm still interested in instructions to be able to test locally without the hack.

udata/core/dataservices/rdf.py
```diff
@@ -89,7 +89,9 @@ def dataservice_from_rdf(
     )
     dataservice.harvest.created_at = rdf_value(d, DCT.created)
     dataservice.harvest.issued_at = rdf_value(d, DCT.issued)
-    dataservice.metadata_modified_at = rdf_value(d, DCT.modified)
+    dataservice.harvest.modified_at = rdf_value(d, DCT.modified)
+
+    dataservice.metadata_modified_at = dataservice.harvest.modified_at
```
streino, line 94: Redundant but I suppose it's not a good idea to drop an existing field without a bit more planning...

udata/core/dataset/models.py
streino, line 1: Reorganized for easier comparison with dataservices `HarvestMetadata` => we're missing `source_url` in dataset and `dct_identifier` in dataservice harvest info. 

I didn't want to extend the scope of this PR but FYI. IMO it'd be nice to do a bit of refactoring around dataset/dataservice processing. Both paths are very similar but mostly copy/pasted, so they easily get out of sync at the moment.

maudetes (approved):
Thank you! As you said, I think we could have a bit of mutualization between dataset and dataservices, but this can be done in a different PR

maudetes:
> I'm not sure how to easily test this locally because the "export-csv" job requires an existing catalog dataset. Is there a way to initialize it using `udata` commands, or do I have to create an appropriate dataset manually?

I'll add https://www.data.gouv.fr/datasets/catalogue-des-donnees-de-data-gouv-fr in fixtures, so that it is more easily testable.
For the filesystem, I think you need to configure your udata.cfg with `FS_ROOT` I guess (see [doc](https://udata.readthedocs.io/en/stable/getting-started/#configure-udata)).

PR #3872, opened 2026-07-20
refactor(harvest): consolidate harvester core

streino:
Consolidates harvester core, and especially dataset/dataservice paths to avoid code duplication, which already caused several divergences.

This is a preliminary step for https://github.com/ecolabdata/ecospheres/issues/42, which is going to require reworking the DCAT `inner_havest` loop. Isolating item-type-specific code should hopefully make it less risky.

The refactor uses a single generic harvest path + a type-specific callback for the *harvested-record → datagouv-item* mapping. The callback pattern is used to support type-checking (existing `inner_process_...` doesn't) while avoiding type-specific signatures in `BaseBackend` (which would require all harvesters to implement all types even though they may not support them, e.g. CKAN/MAAF doesn't support services).

`Dataset`/`Dataservice` API was kept unchanged (\*), hence the need for aliases such as `Dataset.archived` → `archived_at`. (\*) `Harvest*Metadata` API did change (class name and `archived` → `archived_reason`) but it was [considered acceptable](https://github.com/opendatateam/udata/pull/3872#discussion_r3813507542).

Some changes/fixes result from the single generic path:
- Dataservices `HarvestItem` now record logs, like datasets.
- Dataservices have their `pk` set in dry-run, like datasets.

Some other misc changes/fixes are bundled in simply because I caught them while refactoring (discussed with @maudetes):
- Dataset/dataservice harvest metadata now have a shared `HarvestMetadata` superclass.
- Check on `max_items` is abstracted in base so backend implementations don't have to worry about it.
- Dataservices don't trigger org stats on each item, like datasets (computed once at the end of harvest). This is now abstracted in base as well.
- Archival is refactored to handle `Harvestable` items generically.
- `log_catcher` is connected at the very start of process_item, vs after the missing item check in current code. This is debatable, so let me know if you prefer I revert that one.
- `DcatBackend.inner_harvest` is refactored around sub-functions, so the main function reads like the main plan.
- Various misc `DcatBackend` cleanups (var names, arg order, types, ...).

Doc was updated a minima: I only updated parts of the doc that were up-to-date before the refactor.


udata/harvest/backends/base.py
```diff
@@ -246,185 +249,130 @@ def harvest(self):
...
-        harvest.remote_id = remote_id
-        harvest.domain = self.source.domain
-        harvest.last_update = datetime.now(UTC)
-        harvest.archived_at = None
-        harvest.archived = None
+    def update_harvest_metadata(self, metadata: HarvestMetadata, remote_id: str) -> HarvestMetadata:
+        metadata.backend = self.display_name or "unknown"
+        metadata.source_id = str(self.source.id)
+        if hasattr(metadata, "source_url"):
+            # FIXME: consolidate `source_url` to all HarvestMetadata (or remove)?
```
streino: Do we want to do this in this PR?

udata/harvest/backends/base.py
```diff
@@ -246,185 +249,130 @@ def harvest(self):
...
-        # TODO add `type` to `HarvestItem` to differentiate `Dataset` from `Dataservice`
-        item = self.add_item(
+    def process_item(
+        self,
+        remote_id: str,
+        item_processor: Callable[Concatenate[HarvestItem, ItemProcessorParams], Harvestable],
+        *args: ItemProcessorParams.args,
+        **kwargs: ItemProcessorParams.kwargs,
+    ) -> Never:
+        # FIXME: use typing.get_type_hints()["return"] or inspect.signature().return_annotation to get item type?
```
streino, line 275: Do we care about having the item type in the log instead of "item"? If yes I can implement the FIXME, otherwise I'll drop it.

udata/harvest/backends/base.py
```diff
@@ -246,185 +249,130 @@ def harvest(self):
...
-                    dataset.id = ObjectId()
+                if item.pk is None:
+                    item.id = ObjectId()
             else:
-                dataset.save()
-            item.dataset = dataset
-            item.status = "done"
+                item.save()
+
+            # FIXME: consolidate to a single field?
```
streino, line 321: ❓ What's the reason(s) for having separate per-type fields instead of a generic `item` field? I don't know enough about how those fields are used to go with a blind refactor...

streino:
PR passes test suite but I'll need to more thoroughly test manually and there are some FIXME left. 

Asking for reviews now @ThibaudDauce @maudetes to get feedback on the approach and if you'd be ok to merge something like that.

udata/harvest/backends/base.py
```diff
@@ -459,23 +407,23 @@ def autoarchive(self):
...
                 HarvestItem(
                     remote_id=str(dataservice.harvest.remote_id),
                     dataservice=dataservice,
                     status="archived",
                 )
             )
 
-    def get_dataset(self, remote_id):
-        """Get or create a dataset given its remote ID (and its source)
+    def get_item(self, remote_id: str, item_class: type[Harvestable]) -> Harvestable:
```
streino: TODO: add some doc about how this *must* be called from `process_item`'s `item_processor` callback

streino:
Manually tested on Ifremer (csw-dcat), IGN (csw-iso-19139), DREAL ARA (csw-iso-19139), Enedis (dcat), Parcs naturels PACA (ckan) and MinAgr (maaf). Same number of harvested items vs prod, and harvest metadata is properly set.

udata/harvest/backends/base.py
streino: My guess is that it's used to "type" the fields in the API. You know you always get a Dataset when getting `item["dataset"]`, and a Dataservice for `item["dataservice"]`. If those are merged into a single field, we'd need an explicit type and more complex logic client-side.
Still, I'd argue for a generic (properly typed) field because the current approach becomes cumbersome when adding other `Harvestable` types. Something similar to what we do with [`TopicElement`](https://github.com/opendatateam/udata/blob/main/udata/core/topic/models.py#L70).

udata/harvest/backends/base.py
streino: done

udata/core/dataservices/apiv2.py
```diff
@@ -10,7 +10,7 @@
 apiv2.inherit("DataservicePage", Dataservice.__page_fields__)
 apiv2.inherit("Dataservice (read)", Dataservice.__read_fields__)
 apiv2.inherit("DataserviceReference", Dataservice.__ref_fields__)
-apiv2.inherit("HarvestMetadata (read)", HarvestMetadata.__read_fields__)
+apiv2.inherit("HarvestDataserviceMetadata (read)", HarvestDataserviceMetadata.__read_fields__)
```
streino, line 13: ❓ Is this OK or should we preserve `HarvestMetadata`? This change will be visible in the swagger doc, but does it affect the actual json API?

udata/core/dataset/api_fields.py
```diff
@@ -38,10 +38,18 @@
     },
 )
 
+# FIXME: do we need this with @generate_fields?
```
streino, line 41: Unrelated to PR, but I can fix it here if needed.

udata/core/dataset/api_fields.py
streino, line 1: Reorganized fields so order is the same in all code locations (model, api, ...). Not sure how long this will survive but it makes comparing easier.

udata/core/dataset/api_fields.py
```diff
@@ -51,29 +59,24 @@
...
-        "remote_id": fields.String(
-            description="The dataset remote id on the source portal", allow_null=True
-        ),
-        "domain": fields.String(description="The harvested domain", allow_null=True),
         "last_update": fields.ISODateTime(
             description="The dataset last harvest date", allow_null=True
         ),
-        "remote_url": fields.String(description="The dataset remote url", allow_null=True),
+        "archived_at": fields.ISODateTime(description="The archive date", allow_null=True),
+        "archived_reason": fields.String(
```
streino, line 65: :question: ⚠️ `archived` is renamed to `archived_reason` => changes API.

I figured it's ok for harvest metadata because it may not be used that much and breaking change is OK? No impact on cdata.

udata/core/dataset/models.py
```diff
@@ -119,34 +120,18 @@ def get_json_ld_extra(key, value):
...
-    backend = StringField()
-    domain = StringField()
-
-    source_id = StringField()
-
-    remote_id = StringField()
-    remote_url = URLField()
-
+class HarvestDatasetMetadata(HarvestMetadata):
     uri = StringField()
```
streino, line 124: ❓  `uri` field is string here, but url in dataservice. Is this an intended difference or a mistake I should fix?

udata/core/dataset/models.py
```diff
@@ -707,6 +692,14 @@ def __str__(self):
 
     missing_resources = False
 
+    @property
+    def archived_at(self):
```
streino, line 696: ❓  Didn't want to break the `Dataset` API renaming `archived` to `archived_at`. But should we do it?

udata/harvest/backends/base.py
```diff
@@ -445,41 +410,34 @@ def autoarchive(self):
...
             # even when archiving has already been done (useful for debug)
-            self.add_item(
+            self.add_harvest_item(
                 HarvestItem(
-                    remote_id=str(dataservice.harvest.remote_id),
-                    dataservice=dataservice,
-                    status="archived",
+                    **{
+                        "remote_id": str(obj.harvest.remote_id),
+                        key: obj,  # FIXME: consolidate?
```
streino, line 442: See https://github.com/opendatateam/udata/pull/3872/changes#r3616718539

udata/harvest/tasks.py
```diff
@@ -20,18 +20,19 @@ def harvest(self, ident):
     backend.harvest()
 
 
-@task(ignore_result=False, route="low.harvest")
-def harvest_job_item(job_id, item_id):
-    log.info('Harvesting item %s for job "%s"', item_id, job_id)
+# FIXME: broken by https://github.com/opendatateam/udata/pull/3030
```
streino, line 23: This has been broken for a long time, I'd rather not investigate it in this PR.

udata/migrations/2026-08-18-rename-harvest-metadata-fields.py
```diff
@@ -0,0 +1,23 @@
+"""
+TODO
```
streino: TODO

udata/core/harvest.py
streino, line 1: We can't put `HarvestMetadata` in `harvest.models` or we get circular references.

I didn't dive into the whole `core` vs `harvest` split but I think if would benefit from some cleaner separation, i.e. `harvest` knows about `core`, but not the other way around.

Example of current ambiguity: `{dataset,dataservice}_from_rdf` manipulate harvest metadata so should either be:
- harvest-specific and go in `harvest`, or
- generic (e.g. push RDF through the API) but should not touch harvest metadata (that part should go in `harvest`).

Anyway, in the mean time `core/harvest.py` seems like the simpler option.

udata/migrations/2022-09-22-clean-inactive-harvest-datasets.py
streino, line 1: `test_migrations.py` loads all migrations so this rename of an old migration is needed to ensure test still passes.

udata/harvest/models.py
```diff
@@ -76,6 +74,9 @@
 DEFAULT_HARVEST_ITEM_STATUS = "pending"
 
 
+Harvestable = Dataset | Dataservice
```
streino, line 77: ❓ How about making `Harvestable` a core mixin, added to `Dataset` and `Dataservice`?

streino:
@ThibaudDauce @maudetes I plan on adding UT for the new harvester path, but I'd appreciate a review first (overall + comments flagged with ❓) to stabilize the proposed changes before I "freeze" them :pray:

streino:
> Manually tested on Ifremer (csw-dcat), IGN (csw-iso-19139), DREAL ARA (csw-iso-19139), Enedis (dcat), Parcs naturels PACA (ckan) and MinAgr (maaf). Same number of harvested items vs prod, and harvest metadata is properly set.

Re-tested all above harvesters. Looking good :+1:

udata/harvest/backends/dcat.py
streino, line 1: I think now would be a good time to reorganize the methods a bit into the three main groups: main driver, inner/implem-oriented and helpers. It's currently harder to follow than it should.

❓ I can do it in this PR if you don't mind combining changes+reorg, or I'll do a follow-up with just the reorg. Any preference?

udata/harvest/backends/dcat.py
```diff
@@ -72,62 +74,26 @@ class DcatBackend(BaseBackend):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.organizations_to_update = set()
+        self.graphs = list[tuple[Graph, int]]()
+        self.format = self.get_format()
```
streino: Moving `get_format` here means a possible `head` request in the constructor, as opposed to when calling `harvest`. I assume this is fine since backend instances are created right before calling harvest.

I do think this is more consistent with the rest of the code: `fmt` was the only parameter passed around.

If we don't like this, I can add a `begin_job` like we have a `end_job`, but I'd say it's not needed.

streino: This breaks tests... I'll have to take another look.

streino:
@ThibaudDauce @maudetes I'd really appreciate a review on this PR, starting with the comments flagged with :question:. 
It's slowly drifting from main (I'll sync ASAP, shouldn't affect the :question:), and I'm going to need to land series on top of it within the next month or so for OFB. :pray:

udata/harvest/backends/base.py
```diff
@@ -216,217 +237,177 @@ def harvest(self):
...
         self.save_job()
-        return item
+        return harvest_item
 
     def save_job(self):
         if not self.dryrun:
             self.job.save()
 
     def end_job(self):
+        self.inner_end_job()
```
ThibaudDauce: Called from the `finally` of `harvest()`: an S3 error in `store_graphs` escapes and the job stays `initialized`, and graphs are stored even when the harvest failed. Could it go at the end of the `try`?

udata/harvest/backends/base.py
```diff
@@ -216,217 +237,177 @@ def harvest(self):
...
-        if not harvest:
-            harvest = HarvestDataserviceMetadata()
+    def end_process_item(self, harvest_item: HarvestItem, logs: list[logging.LogRecord]):
+        harvest_item.ended = datetime.now(UTC)
+        harvest_item.logs = [
+            HarvestLog(level=log.levelname, message=log.getMessage()) for log in logs
+        ]
+        self.save_job()
+        if self.max_items and len(self.job.items) >= self.max_items:
+            raise StopHarvest()
```
ThibaudDauce: This escapes `inner_harvest` entirely, so in `DcatBackend` the second pass on dataservices never runs once the cap is reached (it used to `return` from the page loop). On a preview, capped at 20, any catalog with 20+ datasets now shows no dataservice at all. Should the cap be counted per item type?

udata/harvest/backends/base.py
```diff
@@ -216,217 +237,177 @@ def harvest(self):
...
-        return item
+        return harvest_item
 
     def save_job(self):
         if not self.dryrun:
             self.job.save()
 
     def end_job(self):
+        self.inner_end_job()
+        self.update_harvested_organizations()
```
ThibaudDauce, line 389: No dryrun guard here, so a preview writes the org metrics and runs the aggregations. Nothing was saved during a preview so there is nothing to recompute, could we early return on `self.dryrun`?

udata/harvest/backends/base.py
```diff
@@ -216,217 +237,177 @@ def harvest(self):
...
                 raise HarvestSkipException("missing identifier")
 
-            current_app.logger.addHandler(log_catcher)
-            dataset = self.inner_process_dataset(item, **kwargs)
-            if dataset.harvest:
-                item.remote_url = dataset.harvest.remote_url
+            item = item_processor(harvest_item, *args, **kwargs)
+
+            if item.organization:
+                item.organization.compute_aggregate_metrics = False
```
ThibaudDauce, line 293: `count_dataservices()` never reads `compute_aggregate_metrics`, so this has no effect for dataservices and the aggregation still runs on every item. Should it be guarded like `count_datasets()`?

udata/harvest/backends/base.py
```diff
@@ -205,7 +220,13 @@ def harvest(self):
...
 
         try:
-            self.inner_harvest()
+            try:
+                self.inner_harvest()
+            except StopHarvest:
+                error = HarvestError(
+                    message=f"{self.max_items} max items reached, not all datasets/dataservices were retrieved"
+                )
+                self.job.errors.append(error)
```
ThibaudDauce, line 231: On `main` this error was skipped on dryrun (`if not self.dryrun and self.has_reached_max_items()`). Previews are capped at 20, so any preview of a larger catalog now reports an error for doing what it was asked. Add back the dryrun guard?

udata/harvest/backends/base.py
ThibaudDauce: `end_process_item` both finalizes the item and decides to stop the harvest, from a `finally`. Could the cap check move to the end of `process_item`, after the try block?

udata/core/harvest.py
```diff
@@ -0,0 +1,37 @@
...
+    )
+    issued_at = field(
+        DateTimeField(), description="Date of release as provided by the harvested catalog"
+    )
+    modified_at = field(
+        DateTimeField(),
+        description="Date of last modification as provided by the harvested catalog",
+    )
+
+    last_update = field(DateTimeField(), description="The dataset last harvest date")
```
ThibaudDauce: This is the dataset wording, it now shows up in the dataservice OpenAPI doc:

```suggestion
    last_update = field(DateTimeField(), description="Date of the last harvesting")
```

udata/harvest/models.py
ThibaudDauce, line 77: `set_harvested()` is duplicated on both models and this stays an union alias, where `WithMetrics`/`BadgeMixin`/`Owned`/`Linkable` are all real mixins both models already inherit. A `Harvestable` mixin in `core/harvest.py` could hold `set_harvested()` once, resolving the class with `self._fields["harvest"].document_type()` like `BadgeMixin` does. Worth a follow-up?

udata/harvest/tasks.py
ThibaudDauce: No caller for `harvest_job_item` (the `api.harvest_job_items` matches are the HTTP endpoint), and the body calls a `process_item` signature that no longer exists. Git keeps it, could we just delete it? Same for `harvest_job_finalize` below, `backend.finalize()` doesn't exist on any backend either.

udata/migrations/2026-08-18-rename-dataset-harvest-archived.py
```diff
@@ -0,0 +1,23 @@
+"""
+Rename Dataset.harvest.archived to Dataset.harvest.archived_at.
```
ThibaudDauce: The rename targets `archived_reason`, and `archived_at` already exists as a separate date field, so this reads like a migration that would overwrite it:

```suggestion
Rename Dataset.harvest.archived to Dataset.harvest.archived_reason.
```

udata/harvest/backends/base.py
```diff
@@ -216,217 +237,177 @@ def harvest(self):
...
 
-    def update_dataservice_harvest_info(
-        self, harvest: HarvestDataserviceMetadata | None, remote_id: str
-    ):
-        if not harvest:
-            harvest = HarvestDataserviceMetadata()
+    def end_process_item(self, harvest_item: HarvestItem, logs: list[logging.LogRecord]):
+        harvest_item.ended = datetime.now(UTC)
+        harvest_item.logs = [
+            HarvestLog(level=log.levelname, message=log.getMessage()) for log in logs
```
ThibaudDauce: `log` is the module logger everywhere else in this file (`log.exception` 8 lines above), so `log.levelname` reads odd:

```suggestion
            HarvestLog(level=record.levelname, message=record.getMessage()) for record in logs
```

udata/harvest/backends/base.py
```diff
@@ -216,217 +237,177 @@ def harvest(self):
...
         self.job.ended = datetime.now(UTC)
         if not self.dryrun:
             self.job.save()
-
         after_harvest_job.send(self)
+        # Clean harvest_activity_user on global context
+        if hasattr(g, "harvest_activity_user"):
+            delattr(g, "harvest_activity_user")
+
+    def inner_end_job(self):
```
ThibaudDauce: One implementation, and it forces `pages` to become `self.graphs` on the instance. A `try/finally` inside `DcatBackend.inner_harvest` would keep it local, no?

udata/core/dataset/models.py
ThibaudDauce: Now that these fields go through `field()`, `HarvestDatasetMetadata.__write_fields__` holds 12 entries where it was empty. `Dataservice.harvest` has `readonly=True`, this one doesn't:

```suggestion
    harvest = field(
        EmbeddedDocumentField(HarvestDatasetMetadata),
        readonly=True,
        auditable=False,
    )
```

ThibaudDauce (changes requested):

udata/harvest/backends/base.py
streino: Before we would do the second pass but stop as soon as we checked `max_items`. So at most we harvested only a single service, and in the process went over `max_items`.

I don't know the origin of `max_items` (just saw [it's used for preview](https://github.com/opendatateam/udata/pull/3872#discussion_r4016072693)), so can't tell if per-type counts would make more sense. IMO having a limit (outside of preview) doesn't make much sense, except as a defensive measure in case something goes really wrong; in which case it probably doesn't matter whether it's overall or per-type counters. 

I'll let you decide with @maudetes and can change to per-type if you prefer.

ThibaudDauce: The `max_items` is used for the preview, no?

ThibaudDauce: Or for testing on big sources to avoid waiting 10 minutes just harvest a few hundred records

udata/harvest/backends/base.py
streino: I did miss that, but IMO the fix should be to add the `compute_aggregate_metrics` check in `count_dataservices()`?

udata/harvest/backends/base.py
streino: Note that per-type counters mean we'll have to scan at lot more of the remote catalog to ensure we process all types (worst case: all datasets first, then dataservices at last). So if `max_items` is more of a defensive move to avoid too much processing, it won't work as well.

And it could make preview take much longer.

udata/harvest/models.py
streino: Yeah, I initially wanted to and had that option flagged as :question: until this morning when I deleted it :p

I tried (and Claude too) to make it a mixin but failed, because if you start extracting everything that should go in `Harvestable` you end up with stuff that's also in other mixins and bump into MRO issues.

I can however make it a `Protocol`. I'll give it another shot and we can decide if it's worth the extra code.

udata/harvest/tasks.py
streino: Happy to delete it if it's OK on your side!

udata/harvest/backends/base.py
streino: > The max_items is used for the preview, no?

Yes just realized that reading your other comment.

> Or for testing on big sources to avoid waiting 10 minutes just harvest a few hundred records

Then we still may end up waiting those 10 minutes if we go with per-type limits. Idem for preview. See my previous comment where source returns all datasets before returning services => we'll still have to paginate the full list of datasets before we can start parsing dataservices.

Just raising the issue; it still might be an acceptable trade-off if we really want per-type limits.

udata/core/dataset/models.py
streino: You're talking about before/after introduction of `field()`, right, not about this change introduced in this PR? I see the same on `main`. 

I can do the change obviously, just checking to make sure I understand what you're saying about the observed change in behavior.

udata/migrations/2026-08-18-rename-dataset-harvest-archived.py
streino: And it's not what the migration actually does 🤦 
```
"$rename": {
  "harvest.archived": "harvest.archived_reason",
}
```

udata/core/dataset/models.py
ThibaudDauce: `HarvestDatasetMetadata` doesn't have any `field()` in `main`

streino: Ah ok, I see, the superclass `HarvestMetadata` has `field()` that `HarvestDatasetMetadata` now inherits. 

BTW, why didn't `HarvestDatasetMetadata` have `field()` in `main`, since it had the `generate_fields`? I didn't dig too much in the rather complex code in `udata/api_fields`, and assumed both should go together?

While I'm at it, I'm also confused whether we still need similar stuff in `dataset/api_fields`: https://github.com/opendatateam/udata/pull/3872/changes/BASE..94109c7dad9b891e7d54bc4ca8d8be258c699746#r3813488018

udata/core/dataset/api_fields.py
ThibaudDauce: Yes, @generate_fields are used with all models except datasets, see https://github.com/datagouv/data.gouv.fr/issues/1868

streino: Oh wow ok, good ticket to know about! I won't touch it here then.

udata/core/dataset/models.py
streino: Given https://github.com/datagouv/data.gouv.fr/issues/1868 will take care of Dataset and it's far from trivial, I'll do the minimum here, i.e. apply the suggestion above.

ThibaudDauce: Thanks! Just adding the readonly is enough to not be caught later when we activate on the dataset the api fields (and make the harvest data writable without noticing)

udata/harvest/backends/base.py
streino: Yes, I figured I'd generalize the callback, but I can go for try/finally and keep it local for now.

udata/migrations/2026-08-18-rename-dataset-harvest-archived.py
streino, line 1: FIXME: Change migration date in filename before merging in main.

udata/harvest/backends/base.py
streino: https://github.com/opendatateam/udata/pull/3872/commits/812089e73a27c80a2bc3a21074ceca47d5fa4641

udata/harvest/backends/base.py
streino: https://github.com/opendatateam/udata/pull/3872/commits/812089e73a27c80a2bc3a21074ceca47d5fa4641

udata/harvest/backends/base.py
streino: https://github.com/opendatateam/udata/pull/3872/commits/812089e73a27c80a2bc3a21074ceca47d5fa4641

udata/migrations/2026-08-18-rename-dataset-harvest-archived.py
streino: https://github.com/opendatateam/udata/pull/3872/commits/812089e73a27c80a2bc3a21074ceca47d5fa4641

udata/harvest/backends/base.py
streino: https://github.com/opendatateam/udata/pull/3872/commits/812089e73a27c80a2bc3a21074ceca47d5fa4641

udata/harvest/backends/base.py
streino: https://github.com/opendatateam/udata/pull/3872/commits/812089e73a27c80a2bc3a21074ceca47d5fa4641

udata/core/dataset/models.py
streino: https://github.com/opendatateam/udata/pull/3872/commits/812089e73a27c80a2bc3a21074ceca47d5fa4641

udata/harvest/tasks.py
streino: https://github.com/opendatateam/udata/pull/3872/commits/812089e73a27c80a2bc3a21074ceca47d5fa4641

udata/harvest/backends/base.py
streino: https://github.com/opendatateam/udata/pull/3872/commits/812089e73a27c80a2bc3a21074ceca47d5fa4641

udata/core/harvest.py
streino: https://github.com/opendatateam/udata/pull/3872/commits/812089e73a27c80a2bc3a21074ceca47d5fa4641

udata/core/dataservices/apiv2.py
ThibaudDauce: Yes it's ok

udata/core/dataset/api_fields.py
ThibaudDauce: Yes it's ok, need to do the change on the cdata side?

udata/core/dataset/models.py
ThibaudDauce: Not sure but don't care I think, we can unify later

udata/core/dataset/models.py
ThibaudDauce: No leave it `archived`

udata/harvest/backends/dcat.py
ThibaudDauce: Maybe a follow-up since this PR is large enough?

udata/harvest/backends/base.py
ThibaudDauce: I'm not sure what we want about dataservices, but leave it like you did, we can iterate on this later if needed. The only thing is the preview of 20 should work correctly because it's used in production

udata/harvest/backends/base.py
ThibaudDauce: Yes that's the change :-)

udata/harvest/models.py
ThibaudDauce: If it's too hard, leave it like this. We can change this later

udata/migrations/2026-08-18-rename-dataset-harvest-archived.py
ThibaudDauce: Why do you need to change date?

udata/harvest/backends/base.py
ThibaudDauce: Don't really care about one or the other. If you want to go the generic way (we have other generic `subject` for exemple and cdata can read this too)

udata/harvest/backends/base.py
ThibaudDauce: Let's drop it, the TODO just below (adding `type` to `HarvestItem`) would give us the info anyway :-)

udata/migrations/2026-08-18-rename-dataset-harvest-archived.py
streino: I assumed migrations were sorted by date and there have been several migrations added to `main` with later dates since I created this one.

If not, it still might be more understandable to have a date that is in sync (or at least closer) with the merge date?

udata/harvest/models.py
streino: I'll leave it like that then and prepare a follow-up when I have a bit more time.

udata/core/dataset/api_fields.py
streino: I think we don't need to do anything in cdata:

- I didn't find any mention of `archived(_reason)` in cdata that is related to harvest metadata.

- Only harvest-related type I found is [datagouv-components/src/types/harvest.ts](https://github.com/datagouv/cdata/blob/b371a9831eb147fc49ba9b3c2801da01e67852f9/datagouv-components/src/types/harvest.ts):
  ```
  export type Harvest = Record<string, unknown> & {
    backend?: string
  } | null
  ```

Did you have something specific in mind?

udata/harvest/backends/base.py
streino: Thanks. In the interest of time I'll leave it as is here and propose the change in another PR.

udata/harvest/backends/base.py
streino: [Will do later](https://github.com/opendatateam/udata/pull/3872#discussion_r4024539838).

udata/harvest/backends/base.py
streino: Added a test for that: https://github.com/opendatateam/udata/pull/3872/changes/812089e73a27c80a2bc3a21074ceca47d5fa4641#diff-acb450f1e8352540a99b3f88489e4ab7c6efd8d8730957d5166d90f66e558ce9R654

PR #3874, opened 2026-07-23
fix(topics): check element is part of Topic

abulte:


maudetes (approved):
Thank you for the clean fix 🙏

PR #3875, opened 2026-07-23
fix: stop leaking membership requests through the public organization API

ThibaudDauce:


udata/core/organization/models.py
```diff
@@ -255,7 +256,14 @@ class Organization(
 
     members = field(ListField(EmbeddedDocumentField(Member)), readonly=True)
     teams = field(ListField(EmbeddedDocumentField(Team)), readonly=True)
-    requests = field(ListField(EmbeddedDocumentField(MembershipRequest)), readonly=True)
+    requests = field(
```
maudetes: Do we actually need requests to be exposed by the API on the organization endpoint?

ThibaudDauce: Done in [2a3c14a](https://github.com/opendatateam/udata/pull/3875/commits/2a3c14a3e953b21e4ee7f07b765567a52bbfb584). Much simpler!

udata/api_fields.py
```diff
@@ -219,7 +219,14 @@ def constructor_read(**kwargs):
         #     2. `__additional_field_info__` of the inner field
         #     3. `__additional_field_info__` of the parent
         inner_info: dict = getattr(field.field, "__additional_field_info__", {})
-        nested_info = {**info, **inner_info, **info.get("inner_field_info", {})}
+        # `attribute` says where to read the list on the parent document. Propagating it
```
maudetes, line 222: I'm not sure I understand the need for this modifications

maudetes (approved):
Thank you for the fix 🙏

udata/api_fields.py
ThibaudDauce: It's not needed since we just removed the `field()` but it's a latent bug, see the test `test_list_attribute_is_resolved_on_the_document` maybe it's clearer? If we pass the `attribute` to the child, the child will try to marshall all its fields with the parent `attribute`

PR #3876, opened 2026-07-23
fix: never expose unpublished posts through the search and detail endpoints

ThibaudDauce:
Need a `udata search index post`

### DIFF
--- a/udata/core/post/api.py
+++ b/udata/core/post/api.py
@@ -101,6 +101,8 @@ class PostAPI(API):
     @api.marshal_with(Post.__read_fields__)
     def get(self, post):
         """Get a given post"""
+        if post.published is None and not AdminPermission().can():
+            api.abort(404)
         return post
 
     @api.doc("update_post")
@@ -129,7 +131,8 @@ class PublishPostAPI(API):
     @api.marshal_with(Post.__read_fields__)
     def post(self, post):
         """Publish an existing post"""
-        post.modify(published=datetime.now(UTC))
+        post.published = datetime.now(UTC)
+        post.save()
         return post
 
     @api.secure(admin_permission)
@@ -137,7 +140,8 @@ class PublishPostAPI(API):
     @api.marshal_with(Post.__read_fields__)
     def delete(self, post):
         """Unpublish an existing post"""
-        post.modify(published=None)
+        post.published = None
+        post.save()
         return post
 
 
--- a/udata/core/post/api.py
+++ b/udata/core/post/api.py
@@ -44,7 +44,7 @@ class PostsAPI(API):
         posts = Post.objects()
 
         if not (AdminPermission().can() and args["with_drafts"]):
-            posts = posts.published()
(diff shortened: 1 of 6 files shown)


udata/core/post/api.py
```diff
@@ -101,6 +101,8 @@ class PostAPI(API):
     @api.marshal_with(Post.__read_fields__)
     def get(self, post):
         """Get a given post"""
+        if post.published is None and not AdminPermission().can():
+            api.abort(404)
```
maudetes: We could add a permission logic on posts for coherence with other objects (`if not post.permissions["read"].can() and ...`):

udata/core/post/search.py
```diff
@@ -22,13 +22,27 @@ class PostSearch(ModelSearchAdapter):
     }
 
     @classmethod
-    def is_indexable(cls, post):
-        return True
+    def is_indexable(cls, post: Post) -> bool:
+        # Unpublished posts are drafts: they must never reach the search index,
+        # which is public and unauthenticated.
+        return post.published is not None
```
maudetes: Same, we could use a `is_visible` property to mutualize "visibility" usage all around?

udata/core/post/search.py
```diff
@@ -22,13 +22,27 @@ class PostSearch(ModelSearchAdapter):
...
 
     @classmethod
     def mongo_search(cls, args):
-        posts = Post.objects()
-        sort = cls.parse_sort(args["sort"]) or "-created_at"
+        posts = Post.objects().published()
```
maudetes: same here, using a query set named `visible`?

udata/tests/search/test_search_integration.py
```diff
@@ -397,6 +398,33 @@ def test_post_search(self):
         names = [p["name"] for p in response.json["data"]]
         assert "Actualités open data" in names
 
+    def test_post_search_follows_publication_state(self):
+        """A post should be searchable only while it is published."""
+        post = PostFactory(name="Brouillon confidentiel", published=None)
```
maudetes: I think it would be more relevant to name it
```suggestion
        post = PostFactory(name="Draft not yet published", published=None)
```

maudetes (approved):
Thank you 🙏 

We can merge this, but I have added comments for future iterations to standardize visibility patterns all around the codebase.

udata/core/post/search.py
ThibaudDauce: Done in 34f37b5cf

udata/core/post/api.py
ThibaudDauce: Done in 34f37b5cf

udata/tests/search/test_search_integration.py
ThibaudDauce: Done in 34f37b5cf

udata/core/post/search.py
ThibaudDauce: Done in 34f37b5cf

PR #3877, opened 2026-07-23
feat: SSRF hardened http client

ThibaudDauce:


maudetes (commented):
I'm not convinced by adding a new validation logic, completely independent from `uris.validate` used all around.

Could we imagine basing our http client on existing code? It may highlight needed updates for the existing code to harden it.

See [this POC](https://github.com/opendatateam/udata/pull/3878)

ThibaudDauce:
> I'm not convinced by adding a new validation logic, completely independent from `uris.validate` used all around.
> 
> Could we imagine basing our http client on existing code? It may highlight needed updates for the existing code to harden it.
> 
> See [this POC](https://github.com/opendatateam/udata/pull/3878)

I think your new PR doesn't close the SSRF, because you can still bind your domain to two IPs and use the randomness to pass the validation before the connect then connect to the forbidden host (Time Of Check Time Of Use problem)

ThibaudDauce:
> I'm not convinced by adding a new validation logic, completely independent from `uris.validate` used all around.
> 
> Could we imagine basing our http client on existing code? It may highlight needed updates for the existing code to harden it.
> 
> See [this POC](https://github.com/opendatateam/udata/pull/3878)

But yes we could reuse some of the existing validation (either getting the uris.validate() inside the ssrf check or the inverse). I didn't use uris.validate() because I think the SSRF check should be a dependency… I don't like having this kind of stuff inside our app code but didn't find a widely use library to do this in sync Python… async Python have some stuff about SSRF)

nicolaskempf57 (approved):

PR #3884, opened 2026-07-29
fix: restrict Discussion.subject to models supporting discussions

ThibaudDauce:
```
AttributeError: 'License' object has no attribute 'self_web_url'
(17 additional frame(s) were not displayed)
...
  File "flask_restx/marshalling.py", line 170, in __format_field
    value = field.output(key, data, ordered=ordered)
  File "flask_restx/fields.py", line 200, in output
    value = get_value(key if self.attribute is None else self.attribute, obj)
  File "flask_restx/fields.py", line 74, in get_value
    return key(obj)
  File "udata/core/discussions/api.py", line 80, in <lambda>
    attribute=lambda d: d.self_web_url(), description="The discussion web URL"
  File "udata/core/discussions/models.py", line 233, in self_web_url
    return self.subject.self_web_url(append="/discussions", discussion_id=self.id, **kwargs)
```

5 discussions exists in demo. Nothing in production.

### DIFF
--- a/udata/core/discussions/constants.py
+++ b/udata/core/discussions/constants.py
@@ -1,9 +1,10 @@
-from udata.core.dataservices.models import Dataservice
-from udata.core.dataset.models import Dataset
-from udata.core.post.models import Post
-from udata.core.reuse.models import Reuse
-from udata.core.topic.models import Topic
-
 COMMENT_SIZE_LIMIT = 50000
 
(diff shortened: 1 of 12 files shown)


nicolaskempf57 (approved):
Do we want to add a default subject to DiscussionFactory ?

ThibaudDauce:
> Do we want to add a default subject to DiscussionFactory ?

Done in a78657bed

PR #3889, opened 2026-07-30
refactor(test): extract parametrize_with_ids helper

streino:
Got tired of copy-pasting the same `parametrize` pattern over and over (more instances to come in https://github.com/opendatateam/udata/pull/3862).

Also, some instances generated test names that couldn't be called from the pytest cli:
```
$ pytest udata -k 'test_geonetwork[dcat-http://www.w3.org/ns/dcat#]'
ERROR: Wrong expression passed to '-k': test_geonetwork[dcat-http://www.w3.org/ns/dcat#]: at column 47: unexpected character "#"
```

### DIFF
--- a/udata/tests/helpers.py
+++ b/udata/tests/helpers.py
@@ -1,5 +1,5 @@
 import os
-from collections.abc import Callable
+from collections.abc import Callable, Iterable
 from contextlib import contextmanager
 from datetime import datetime, timedelta
 from io import BytesIO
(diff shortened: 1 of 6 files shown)


ThibaudDauce (commented):
Is it better to extract a constant and do the same thing as in `udata/tests/workers/test_tasks_routing.py:28`? It avoids replacing the standard `@pytest.mark.parametrize`? (or at least convert it to this new system?)

streino:
> Is it better to extract a constant and do the same thing as in `udata/tests/workers/test_tasks_routing.py:28`? It avoids replacing the standard `@pytest.mark.parametrize`? (or at least convert it to this new system?)

I'm not fond of extracting in constants when it's used for a single test, because it moves the test arguments further away from the test function itself.

I pushed another proposal with a more localized wrapper, around `argvalues` instead of `parametrize`. WDYT? If you think that works, I'll rename the PR and add some tests for `argvalues()` before you merge.

I also moved the default "id" parameter to the **last** tuple param instead of the first. It's more in line with how you'd declare a `pytest.param(..., id=...)` and might make associating `argvalues` tuple items with their corresponding `argnames` slightly easier.

ThibaudDauce (approved):

udata/tests/helpers.py
```diff
@@ -242,41 +243,23 @@ def security_gettext(string):
...
-      => @parameterize("a, b", [(1, 2), (3, 4)], ids=[3, 7])
+      @pytest.mark.parametrize("a, b", argvalues((1, 2), (3, 4), ids=lambda v: f"sum-{sum(v)}"))
+      => @pytest.mark.parametrize("a, b", [(1, 2), (3, 4)], ids=["sum-3", "sum-7"])
     """
-    if (
-        len(argvalues) == 1
-        and not isinstance(argvalues[0], tuple)
-        and isinstance(argvalues[0], Iterable)
-    ):
-        argvalues = list(argvalues[0])
```
streino: That part is needed to support generators. For instance `argvalues((foo(x), x) for x in [...])`. Without that bit, you'll need such a case to be called as `argvalues(*[(foo(x), x) for x in [...]])`.

ThibaudDauce: Added back in [d3113da](https://github.com/opendatateam/udata/pull/3889/commits/d3113dac91ad6c5ae11f020fd5453f2f32b06ebe) is it good for you?

streino: And thanks for the tests :)

PR #3891, opened 2026-08-05
fix: return 200 instead of 500 when a nested entry is submitted with an unknown id

ThibaudDauce:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/299180/ on `PUT` https://demo.data.gouv.fr/api/1/datasets/67f988bf9ea6979ac6079373/

```
AttributeError: 'NoneType' object has no attribute 'to_mongo'
(15 additional frame(s) were not displayed)
...
  File "udata/forms/__init__.py", line 20, in process
    super(CommonFormMixin, self).process(
  File "wtforms/form.py", line 128, in process
    field.process(formdata, data, extra_filters=field_extra_filters)
  File "udata/forms/fields.py", line 551, in process
    super(NestedModelList, self).process(formdata, data, **kwargs)
  File "wtforms/fields/list.py", line 91, in process
    self._add_entry(formdata, obj_data, index=index)
  File "udata/forms/fields.py", line 603, in _add_entry
    initial = flatten_json(self.nested_form, data.to_mongo(), prefix)
```

### DIFF
--- a/udata/core/dataset/apiv2.py
+++ b/udata/core/dataset/apiv2.py
@@ -33,7 +33,7 @@ from .api_fields import (
     temporal_coverage_fields,
 )
 from .constants import DEFAULT_LICENSE, FULL_OBJECTS_HEADER, UpdateFrequency
-from .models import CommunityResource, Dataset, Resource
+from .models import CommunityResource, Dataset, Resource, get_dataset_by_resource_id
 from .search import DatasetSearch
 
 DEFAULT_PAGE_SIZE = 50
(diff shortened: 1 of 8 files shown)


nicolaskempf57 (approved):
Do we want to update dataset model l.547 to use `get_dataset_by_resource_id` too ?

ThibaudDauce:
> Do we want to update dataset model l.547 to use `get_dataset_by_resource_id` too ?

Good catch, done in [3c0b9bc](https://github.com/opendatateam/udata/pull/3891/commits/3c0b9bc998690ee7d22f40f403b3f6608079fd5d)

PR #3897, opened 2026-08-10
fix(topics): update Topic.last_modified on element change

### LINKED ISSUE #1200: Date de mise à jour d'une collection non actualisée après modification d'un facteur
Il a été observé que la date de mise à jour n'est pas systématiquement actualisée lorsque la collection fait l'objet de modification au niveau des facteurs. La date de mise à jour permet de faire remonter les collections faisant l'objet de contributions. Exemple à partir du screenshot ci-dessous. Difficultés à reproduire : il semblerait que l'actualisation de la description d'un facteur soit bien prise en compte dans l'onglet activité mais pas pour l'actualisation de la date de mise à jour.

Image

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/1200

### DIFF
--- a/udata/core/topic/models.py
+++ b/udata/core/topic/models.py
@@ -1,3 +1,5 @@
+from datetime import UTC, datetime
+
 from blinker import Signal
 from flask import url_for
 from mongoengine.errors import DoesNotExist
@@ -94,6 +96,7 @@ class TopicElement(Auditable, Document):
             reindex.delay(*as_task_param(document.element))
         if document.topic:
             reindex.delay(*as_task_param(document.topic))
+            cls._bump_topic_last_modified(document.topic)
 
     @classmethod
     def post_delete(cls, sender, document, **kwargs):
@@ -103,10 +106,16 @@ class TopicElement(Auditable, Document):
                 reindex.delay(*as_task_param(document.element))
             if document.topic:
                 reindex.delay(*as_task_param(document.topic))
+                cls._bump_topic_last_modified(document.topic)
         except DoesNotExist:
             # Topic might have been deleted, causing dereferencing to fail
             pass
         cls.on_delete.send(document)
+
+    @staticmethod
+    def _bump_topic_last_modified(topic):
+        """Update the parent topic's last_modified without emitting a topic update activity"""
+        Topic.objects(id=topic.id).update(last_modified=datetime.now(UTC))
 
 
 @generate_fields()
--- a/udata/core/topic/models.py
+++ b/udata/core/topic/models.py
@@ -1,5 +1,3 @@
-from datetime import UTC, datetime
-
 from blinker import Signal
 from flask import url_for
 from mongoengine.errors import DoesNotExist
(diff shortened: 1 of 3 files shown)


udata/core/topic/models.py
```diff
@@ -103,11 +106,17 @@ def post_delete(cls, sender, document, **kwargs):
...
                 reindex.delay(*as_task_param(document.topic))
+                cls._bump_topic_last_modified(document.topic)
         except DoesNotExist:
             # Topic might have been deleted, causing dereferencing to fail
             pass
         cls.on_delete.send(document)
 
+    @staticmethod
+    def _bump_topic_last_modified(topic):
+        """Update the parent topic's last_modified without emitting a topic update activity"""
```
ThibaudDauce: Are you sure about that? `last_modified` is marked as `audited=False` so you can update it with `topic.save()` without creating a new activity I think

udata/core/topic/models.py
```diff
@@ -103,11 +106,17 @@ def post_delete(cls, sender, document, **kwargs):
                 reindex.delay(*as_task_param(document.element))
             if document.topic:
                 reindex.delay(*as_task_param(document.topic))
```
ThibaudDauce: `reindex.delay` will index a wrong `last_modified` date?

udata/core/topic/models.py
```diff
@@ -103,11 +106,17 @@ def post_delete(cls, sender, document, **kwargs):
...
+                cls._bump_topic_last_modified(document.topic)
         except DoesNotExist:
             # Topic might have been deleted, causing dereferencing to fail
             pass
         cls.on_delete.send(document)
 
+    @staticmethod
+    def _bump_topic_last_modified(topic):
+        """Update the parent topic's last_modified without emitting a topic update activity"""
+        Topic.objects(id=topic.id).update(last_modified=datetime.now(UTC))
```
ThibaudDauce: Do we want to set the `topic.last_modified` in memory to return the correct date to the caller?

ThibaudDauce (commented):
Two points outside the diff:
1. `purge-topics-elements` bump the `last_modified` now, is it what we want? maybe yes ^^
2. Do we need the `topic.save()` inside `TopicElementsAPI.post`?

udata/core/topic/models.py
abulte: True! Thanks, done.

udata/core/topic/models.py
abulte: True, changed the order.

udata/core/topic/models.py
abulte: Now using `.save()`.

abulte:
> 1. `purge-topics-elements` bump the `last_modified` now, is it what we want? maybe yes ^^

I guess both behaviour would be valid. I'll leave it like this, not strictly related to this PR anyway.

> 2. Do we need the `topic.save()` inside `TopicElementsAPI.post`?

Nope, removed.

udata/core/topic/models.py
```diff
@@ -93,6 +93,7 @@ def post_save(cls, sender, document, **kwargs):
         if document.topic and document.element and hasattr(document.element, "id"):
             reindex.delay(*as_task_param(document.element))
         if document.topic:
+            document.topic.save()
             reindex.delay(*as_task_param(document.topic))
```
ThibaudDauce: Is the reindex useful now that you use `topic.save()`?

ThibaudDauce (approved):

udata/core/topic/models.py
abulte: Good catch. Removed and tested.

ThibaudDauce (approved):

PR #3898, opened 2026-08-11
feat(topics): add lookup by element id/class

abulte:
Related https://github.com/ecolabdata/ecospheres/issues/1154

This lets us filter the Topics list by a nested element class/id couple, eg `/api/2/topics/?dataset=6a69ec676cd935c1caf8a184&tag=universe`.

Supports:
-  `dataset`
- `reuse`
- `dataservice`

For now, it replies with a standard list of Topics. In the **future**, we could augment the payload with a `matching_elements` structure as below:

```
  {
    "data": [
      {
        [...]
        "matching_elements": [
          {
            "id": "66aa11223344556677889900",
            "title": "Mesures PM10 2023",
            "description": "Placé dans la section « Particules fines »",
            "tags": ["particules-fines"],
            "extras": {},
            "element": { "id": "64f1...", "class": "Dataset" }
          },
          {
            "id": "66ab99887766554433221100",
            "title": null,
            "description": null,
            "tags": ["indicateurs-cles"],
            "extras": {},
            "element": { "id": "64f1...", "class": "Dataset" }
          }
        ]
      }
    ],
    [...]
  }
```

### DIFF
--- a/udata/core/topic/models.py
+++ b/udata/core/topic/models.py
@@ -52,6 +52,11 @@ class TopicQuerySet(OwnedQuerySet):
     def hidden(self):
         return self(private=True)
 
+    def for_element(self, element):
+        """Filter topics containing the given element (Dataset, Dataservice...)."""
+        topic_ids = TopicElement.objects(element=element).no_dereference().distinct("topic")
+        return self(id__in=topic_ids)
+
 
 @generate_fields()
 class TopicElement(Auditable, Document):
@@ -75,7 +80,12 @@ class TopicElement(Auditable, Document):
         "indexes": [
             {
                 "fields": ["$title", "$description"],
-            }
+            },
+            # Used to reverse-lookup the topics an element belongs to.
+            "element",
+            # Used by Topic.elements and the per-topic element count (incl. the
+            # API's HATEOAS "elements" link), so it stays indexed even on huge topics.
+            "topic",
         ],
         "auto_create_index_on_save": True,
     }
--- a/udata/core/topic/models.py
+++ b/udata/core/topic/models.py
@@ -68,6 +68,8 @@ class TopicElement(Auditable, Document):
     tags = field(ListField(StringField()))
     extras = field(ExtrasField())
     element = field(
+        # If modifying choices list below , also look at core/topic/parsers.py's
+        # ELEMENT_FILTER_MODELS for filtering support
         GenericReferenceField(choices=["Dataset", "Reuse", "Dataservice"]),
         nested_fields=api.model_reference,
         allow_null=True,
--- a/udata/core/topic/models.py
+++ b/udata/core/topic/models.py
@@ -69,7 +69,7 @@ class TopicElement(Auditable, Document):
     extras = field(ExtrasField())
     element = field(
         # If modifying choices list below , also look at core/topic/parsers.py's
-        # ELEMENT_FILTER_MODELS for filtering support
+        # parse_filters for filtering support
         GenericReferenceField(choices=["Dataset", "Reuse", "Dataservice"]),
         nested_fields=api.model_reference,
         allow_null=True,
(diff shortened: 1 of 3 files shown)


udata/core/topic/models.py
```diff
@@ -75,7 +80,12 @@ class TopicElement(Auditable, Document):
         "indexes": [
             {
                 "fields": ["$title", "$description"],
-            }
+            },
+            # Used to reverse-lookup the topics an element belongs to.
+            "element",
+            # Used by Topic.elements and the per-topic element count (incl. the
+            # API's HATEOAS "elements" link), so it stays indexed even on huge topics.
+            "topic",
```
abulte, line 90: Not strictly related to this PR. Should make the Topics list faster.

ThibaudDauce:
Have you tried using the same pattern `dataset=xxx`, `reuse=xxx` and `dataservice=xxx`? Not sure what is best but it may be simpler? (it avoids having two query params bound together and the `db.resolve_model()` on a user string). Also `element=xxx` targets `TopicElement.element.id`, which can be confusing with `TopicElement.id`?

abulte:
> Have you tried using the same pattern `dataset=xxx`, `reuse=xxx` and `dataservice=xxx`? Not sure what is best but it may be simpler? (it avoids having two query params bound together and the `db.resolve_model()` on a user string). Also `element=xxx` targets `TopicElement.element.id`, which can be confusing with `TopicElement.id`?

Yeah good idea, it's not as generic (need to wire in a new supported model manually), but the API looks better. Done.

udata/core/topic/parsers.py
```diff
@@ -6,9 +6,18 @@
...
+from udata.core.dataset.models import Dataset
+from udata.core.reuse.models import Reuse
 from udata.core.topic import DEFAULT_PAGE_SIZE
 from udata.core.topic.models import TopicElement
 
+ELEMENT_FILTER_MODELS = {
+    "dataset": Dataset,
+    "reuse": Reuse,
+    "dataservice": Dataservice,
+}
```
ThibaudDauce: As I understand, this is only to protect the filter with `element.permissions["read"].can()` but I don't think it's necessary, `TopicElementsAPI@get` doesn't have this protection anyway `/api/1/datasets/?topic=<id>` neither. And this constant add a lot of complexity to the feature. Maybe delay this protection and add something better later to protect all the `TopicElement` for the visibility of their child?

udata/core/topic/parsers.py
```diff
@@ -104,4 +125,15 @@ def parse_filters(topics, args):
...
+        for arg, model in ELEMENT_FILTER_MODELS.items():
+            element_id = args.get(arg)
+            if not element_id:
+                continue
+            if not ObjectId.is_valid(element_id):
+                api.abort(400, f"{arg} arg must be an identifier")
+            element = model.objects(id=element_id).first()
+            if element is None or not element.permissions["read"].can():
+                # return an empty queryset when nested element can't be read
+                return topics.none()
```
ThibaudDauce, line 137: Returning here prevent validating `ObjectId.is_valid(element_id)` for the other element types? But not sure `?dataset=<id privé non lisible>&reuse=garbage` is something we want to support anyway…

udata/core/topic/parsers.py
abulte: Agreed that the `read` check was a bit much given what we have elsewhere. The dict was not so complex I think, and we could use it for looping from a central place. I removed it, but not sure it's better?

udata/core/topic/parsers.py
abulte: I don't think there's a perfect solution for that with the API grammar we have today, but I now validate all args before applying the filters.

ThibaudDauce (approved):

PR #3910, opened 2026-08-21
fix: validate uploaded image content instead of the client-declared mimetype

ThibaudDauce:
Fix [UnidentifiedImageError](https://errors.data.gouv.fr/organizations/sentry/issues/300233/?referrer=alert_email&alert_type=email&alert_timestamp=1787307713153&alert_rule_id=10&notification_uuid=5c484318-4c9e-4f88-a8a3-3213e54a5952&environment=demo.data.gouv.fr) api.my_avatar
cannot identify image file <FileStorage: 'ch9.svg' ('image/png')>

### DIFF
--- a/udata/core/storages/api.py
+++ b/udata/core/storages/api.py
@@ -188,6 +188,9 @@ def parse_uploaded_image(field):
         image_format = Image.open(image).format
     except UnidentifiedImageError:
         image_format = None
+    # `Image.open` left the cursor right after the header it read. `field.save` may store
+    # the stream as-is (flask_storage only rewinds when it resizes or optimizes), which
+    # would silently truncate the stored file.
     image.seek(0)
     if image_format not in IMAGES_FORMATS:
         api.abort(400, "Unsupported image format")
--- a/udata/core/storages/api.py
+++ b/udata/core/storages/api.py
@@ -3,6 +3,7 @@ import os
 from datetime import UTC, datetime
 
 from flask import json
+from PIL import Image, UnidentifiedImageError
 from werkzeug.datastructures import FileStorage
 
 from udata.api import api, fields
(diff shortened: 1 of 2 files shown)


udata/core/storages/api.py
```diff
@@ -180,7 +181,15 @@ def parse_uploaded_image(field):
...
     image = args["file"]
-    if image.mimetype not in IMAGES_MIMETYPES:
+    # The mimetype is declared by the client and may not match the content at all
+    # (an SVG announced as `image/png` for instance): decode the file to know what
+    # it really is, otherwise Pillow raises further down and the request 500s.
+    try:
+        image_format = Image.open(image).format
+    except UnidentifiedImageError:
+        image_format = None
+    image.seek(0)
```
nicolaskempf57, line 199: what does is this line do ?

ThibaudDauce: Not in our configuration with image optimization but yes it could save a corrupted file (saving the file after the bytes used for detecting the type). Tested in 9f8b96e30

udata/core/storages/api.py
```diff
@@ -180,7 +181,15 @@ def parse_uploaded_image(field):
     args = image_parser.parse_args()
 
     image = args["file"]
-    if image.mimetype not in IMAGES_MIMETYPES:
+    # The mimetype is declared by the client and may not match the content at all
+    # (an SVG announced as `image/png` for instance): decode the file to know what
+    # it really is, otherwise Pillow raises further down and the request 500s.
+    try:
+        image_format = Image.open(image).format
+    except UnidentifiedImageError:
```
nicolaskempf57, line 189: Do we want to also do it for the Image.DecompressionBombError ?

ThibaudDauce: Good catch! `Image.open` raises it just after reading the header, so a 69 bytes PNG announcing 20000x20000 is enough to get a 500 (it was already the case on main, in the thumbnails). Done in c7a3b46c0

nicolaskempf57 (approved):

PR #3913, opened 2026-08-21
fix: make /organizations/<org>/datasets/ an alias of /datasets/?organization=<org>

ThibaudDauce:
`/organizations/<org>/datasets` was declaring `dataset_parser.parser` but didn't look into any of the filters (except sort and pagination…)

I think it should be the same as `/datasets?organization=<org>` but it removes the archived and deleted from the response, it's technically a breaking change…

### DIFF
--- a/udata/core/dataset/api.py
+++ b/udata/core/dataset/api.py
@@ -312,6 +312,21 @@ community_parser.add_argument(
 
 common_doc = {"params": {"dataset": "The dataset ID or slug"}}
 
+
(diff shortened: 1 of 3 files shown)


ThibaudDauce:
@maudetes don't want a breaking change on this endpoint (and I agree…) maybe we should just change the accepted filters? And by the way fix `archived` filter returning a 403 even if `archived` is not a visibility filter like `private` and `deleted`

PR #3914, opened 2026-08-21
fix: make the contact point name optional instead of an empty string

ThibaudDauce:
`ContactPoint.name` was `required=True`, but mongoengine only rejects `None`: an empty
string goes through (`mongoengine/base/document.py:415-431`). Harvesting created those,
`contact_point_name` falling back to `""` when neither the agent nor its organization
had a name.

In production 294 contact points have an empty name, and none has a null one. 190 of
them still carry an email or a contact form, so they can be reached and the front
already falls back to those (`ContactPoint.vue`, `ContactPointSelect.vue`): making the
name optional keeps them working as they are. The 104 others hold nothing but a role,
and the 563 datasets referencing them show an empty contact line, so the migration
detaches and deletes them.

A contact point now requires a name, an email or a contact form, whatever its role.
#3862 stopped harvesting from creating the empty ones on August 12th, this covers the
other write paths.

Needs `name: string | null` in the contact point type on the cdata side.

### DIFF
--- a/udata/migrations/2026-08-21-clean-nameless-contact-points.py
+++ b/udata/migrations/2026-08-21-clean-nameless-contact-points.py
@@ -1,0 +1,42 @@
+"""
+Harvesting used to name a contact point after its `foaf:Agent` or `vcard:Kind`, falling back to
+an empty string when neither carried a name. `ContactPoint.name` is now optional, so the empty
+string has to become an absent value: it is the same thing said in two ways, and only one of
+them is falsy everywhere.
+
+Those without a name also have no email and no contact form when they come from an agent that
+carried nothing at all, which harvesting stopped creating in #3862. Nobody can be reached
+through them and they render as an empty line under the datasets referencing them, so they are
+removed rather than normalized.
+
+Written against the collections rather than the models: this is a bulk rewrite of two fields,
+and loading every document through the ORM to save it back would only add ways to fail.
+"""
+
+import logging
+
+log = logging.getLogger(__name__)
+
+NO_INFORMATION = {"name": None, "email": None, "contact_form": None}
+
+
+def migrate(db):
+    normalized = db.contact_point.update_many({"name": ""}, {"$unset": {"name": ""}}).modified_count
+    log.info(f"{normalized} contact points had an empty name instead of no name.")
+
+    unreachable = [
+        contact_point["_id"] for contact_point in db.contact_point.find(NO_INFORMATION, {"_id": 1})
+    ]
+    if not unreachable:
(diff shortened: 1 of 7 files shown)


streino (approved):
Makes sense: email is in most cases a better contact info than a name, so requiring the latter only is counter-productive.

udata/migrations/2026-08-21-clean-nameless-contact-points.py
```diff
@@ -0,0 +1,42 @@
...
+
+import logging
+
+log = logging.getLogger(__name__)
+
+NO_INFORMATION = {"name": None, "email": None, "contact_form": None}
+
+
+def migrate(db):
+    normalized = db.contact_point.update_many({"name": ""}, {"$unset": {"name": ""}}).modified_count
```
maudetes: I'm always unsure if we should unset or set to None?
In any case, a future user could set a name as "", right?

udata/migrations/2026-08-21-clean-nameless-contact-points.py
```diff
@@ -0,0 +1,42 @@
...
+
+Written against the collections rather than the models: this is a bulk rewrite of two fields,
+and loading every document through the ORM to save it back would only add ways to fail.
+"""
+
+import logging
+
+log = logging.getLogger(__name__)
+
+NO_INFORMATION = {"name": None, "email": None, "contact_form": None}
```
maudetes, line 25: We could have cases where `email=""` right?

maudetes (commented):
Thank you! I have questions regarding the cleaning and expected empty values for `name`, `email` & `contact_form`.

udata/migrations/2026-08-21-clean-nameless-contact-points.py
ThibaudDauce: Yes, mongoengine unsets the field when it's `None` so both are the same thing.

And `patch()` already replaces blank strings by `None` on every `StringField` since https://github.com/opendatateam/udata/pull/3440, so it can't come back from the API. The empty ones came from harvesting, which doesn't go through `patch()`.

udata/migrations/2026-08-21-clean-nameless-contact-points.py
ThibaudDauce: Good catch! 26 in production. They all have a name so they're not in the 104 removed here, but a bare `mailto:` was read as an empty email. Fixed in ea5161aa9, with the `$unset` on `email` too.

maudetes (approved):

PR #3922, opened 2026-09-02
fix: fix dangling orgs refs

bolinocroustibat:
Fix [this Sentry error](https://errors.data.gouv.fr/organizations/sentry/issues/301053/?project=12):

If a discussion still points to an organization that was deleted, listing discussions crashed with a 500.
The API now returns null for that missing organization instead of failing the whole page.

[Problem discovered and discussed here](https://matrix.to/#/!FMXpGlfBgjvMumrnPQ:agent.dinum.tchap.gouv.fr/$1nKzs1nHV27p2dB7Ju6_4d9Qe1_01OPFn1kx-0PsclU?via=agent.dinum.tchap.gouv.fr).

PR #3923, opened 2026-09-03
fix(discussion): fix integrity on organization reference

maudetes:
Fix `Trying to dereference unknown document DBRef('organization', ObjectId('6a74a0726dfe7c1502ea806a'))` on a discussion with org reg integrity ([sentry error](https://errors.data.gouv.fr/organizations/sentry/issues/301115/))

1. We add organization ref cleaning in discussions purge task.
2. We apply a migration that remove existing inconsistent discussions or refs. Locally, I have:
```
  │ Processing Discussion references.
  │ Deleted 1 Discussion objects
  │ Unset 0 posted_by_organization in Discussion messages
  │ Unset 0 closed_by_organization in Discussion objects
```

`posted_by` and `closed_by` are always set on top of organization infos and used as fallback after organization reference removal.

Replace https://github.com/opendatateam/udata/pull/3922

### DIFF
--- a/udata/core/discussions/models.py
+++ b/udata/core/discussions/models.py
@@ -4,7 +4,7 @@ from datetime import UTC, datetime
 from urllib.parse import urlparse
 
 from bson import ObjectId
-from flask import current_app, url_for
+from flask import current_app, g, url_for
 from flask_login import current_user
 from flask_restx.inputs import boolean
 from mongoengine import EmbeddedDocument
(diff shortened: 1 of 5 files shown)


bolinocroustibat (approved):
Thank you, LGTM! Seems cleaner indeed to have a cleaning migration for dangling objects.

Do we know why it's happening though?

maudetes:
> Do we know why it's happening though?

An organization created a discussion and then the org was deleted, leaving the inconsistency.
I haven't checked into the whereabouts of this specific organization.

ThibaudDauce (approved):
I've just added some tests in 6f58d9 and fix one bug in the migration f0ab9a. I let you merge if your ok with my changes

PR #3927, opened 2026-09-07
fix!: remove the unpaginated org discussions endpoints in favor of /discussions/?org=

ThibaudDauce:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/300836

InvalidBSON  api.org_discussions (MemoryError hidden by pymongo)

Both are unpaginated and dereference `subject` once per discussion. On Météo-France, 488
discussions for 43 distinct datasets: 684 MB refetched per call, 500 or 502 after 12 to
17s. `/api/1/discussions/?org=<id>` returns the same list in 4s.

Over 118 days of HAProxy logs the org one got 43 requests (23 in 200, mostly scanners),
`/me/org_discussions/` got only 1 call.

`/me/org_discussions/` has no strict equivalent: it covered all my organizations plus my
own datasets and reuses in a single call, so it's now one `?org=` call per organization.

### DIFF
--- a/udata/core/organization/api.py
+++ b/udata/core/organization/api.py
@@ -760,19 +760,6 @@ class OrgReusesAPI(API):
         return list(qs)
 
 
-@ns.route("/<org:org>/discussions/", endpoint="org_discussions")
-class OrgDiscussionsAPI(API):
-    @api.doc("list_organization_discussions")
-    @api.marshal_list_with(Discussion.__read_fields__)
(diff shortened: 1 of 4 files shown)


nicolaskempf57 (approved):

ThibaudDauce:
I would like to have your opinion on this @maudetes, it's doesn't seem to be used but I'm not sure if it's the correct way to go?

PR #3928, opened 2026-09-07
fix: retry save when a concurrent writer takes the computed slug

ThibaudDauce:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/301000

NotUniqueError  api.datasets
Tried to save duplicate unique keys (E11000 duplicate key error collection: udata.dataset index: slug_1 dup key: { slug: "donnees-essentielles-des-marches-publics-region-auvergne-rhone-alpes-1936" }, full error: {'index': 0, 'cod

Useful for https://www.data.gouv.fr/organizations/region-auvergne-rhone-alpes/search?q=Donn%C3%A9es+essentielles&type=datasets&sort=-last_update which upload multiple datasets with the same title sometimes 21 times in a few minutes.

### DIFF
--- a/udata/mongo/document.py
+++ b/udata/mongo/document.py
@@ -2,6 +2,7 @@ import logging
 from collections.abc import Iterable
 from typing import TYPE_CHECKING, ClassVar, Generic
 
+from mongoengine.errors import NotUniqueError
 from typing_extensions import TypeVar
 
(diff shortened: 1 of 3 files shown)


maudetes (approved):

maudetes:
Nice! Thank you for looking up that bug and pushing this fix!

PR #3930, opened 2026-09-07
fix: transfer api reference injection

ThibaudDauce:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/301567/ and https://errors.data.gouv.fr/organizations/sentry/issues/301560

MultipleObjectsReturned api.transfers
2 or more items returned, instead of 1

AttributeError api.transfers
'License' object has no attribute 'organization'

Also fix https://errors.data.gouv.fr/organizations/sentry/issues/301890/
Error creating notification for admin user 6a9e59fb93ec631ffa38e226 and recipient 6a9e59fb93ec631.. api.transfers

### DIFF
--- a/udata/features/transfer/api.py
+++ b/udata/features/transfer/api.py
@@ -5,6 +5,7 @@ from udata.api import API, api, base_reference, fields
 from udata.core.dataservices.models import Dataservice
 from udata.core.dataset.api_fields import dataset_ref_fields
 from udata.core.organization.models import Organization
+from udata.core.topic.models import Topic
 from udata.features.transfer.permissions import (
     TransferPermission,
     TransferResponsePermission,
@@ -13,7 +14,7 @@ from udata.models import Dataset, Reuse, User, db
 from udata.utils import id_or_404
 
 from .actions import accept_transfer, refuse_transfer, request_transfer
-from .models import TRANSFER_STATUS, Transfer
+from .models import TRANSFER_PERSONS, TRANSFER_STATUS, TRANSFERABLE_SUBJECTS, Transfer
 
 RESPONSE_TYPES = ["accept", "refuse"]
 
@@ -53,6 +54,7 @@ subject_mapping = {
     Dataservice: Dataservice.__ref_fields__,
     Dataset: dataset_ref_fields,
     Reuse: Reuse.__ref_fields__,
+    Topic: Topic.__ref_fields__,
 }
 
 transfer_fields = api.model(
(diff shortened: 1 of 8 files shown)


udata/features/transfer/api.py
```diff
@@ -112,6 +114,37 @@
...
+    id: a dict reaches MongoDB as a set of operators (`{"$regex": …}`) selecting an
+    arbitrary document, which MongoEngine rejects on an `ObjectId` primary key but not on
+    a `StringField` one.
+    """
+    reference = data.get(field_name)
+    if not isinstance(reference, dict):
+        ns.abort(400, errors={field_name: "Expected an object with `class` and `id` keys"})
+
+    class_name = reference.get("class")
+    if class_name not in allowed_classes:
```
maudetes, line 133: It isn't tested (or not needed?), is it?

udata/migrations/2026-09-07-drop-transfers-with-unsupported-classes.py
nicolaskempf57, line 1: I think we should also delete the related notifications like `udata/migrations/2026-07-29-delete-discussions-on-unsupported-subjects.py`

udata/features/transfer/permissions.py
```diff
@@ -7,19 +7,29 @@ class TransferPermission(Permission):
...
-        super(TransferPermission, self).__init__(need)
+            needs.append(UserNeed(subject.owner.fs_uniquifier))
+        super().__init__(*needs)
 
 
 class TransferResponsePermission(Permission):
     """Permissions to transfer an object assets"""
 
     def __init__(self, transfer):
+        # Same as above: `Transfer.recipient` is a choice-less generic reference, so
```
nicolaskempf57, line 27: This new comment should be updated to match the one above about a deleted organization because recipient isn't choice-less anymore and the existing ones will be deleted by the migration

udata/migrations/2026-09-07-drop-transfers-with-unsupported-classes.py
ThibaudDauce: That's the inverse, no? The notifications weren't created because of this bug so nothing to remove?

nicolaskempf57: Right, I didn't saw the " Error creating notification for admin user 6a9e59fb93ec631ffa38e226 and recipient 6a9e59fb93ec631.. api.transfers"

PR #3932, opened 2026-09-08
feat(search): match dotted titles when querying with spaces

bolinocroustibat:
Searching the string "data gouv" on data.gouv.fr search bar currently misses titles like "data.gouv", because Elasticsearch treats the dot as part of the word. This adds a secondary analysis that turns `.` and `_` into spaces on both indexing and query, so the two forms match in both directions ("data gouv" <-> "data.gouv"), without changing the existing ranking.

After deploy, run a full reindex once (`data search index --reindex true`) on the server(s) where it is deployed, so existing documents get the new field.

cc @estellebertrand, @DorineLam as reported by them

### DIFF
--- a/udata/tests/search/test_search_integration.py
+++ b/udata/tests/search/test_search_integration.py
@@ -65,6 +65,25 @@ class SearchIntegrationTest(APITestCase):
         titles = [d["title"] for d in response.json["data"]]
         assert "Données spectaculaires sur les transports" in titles
 
+    def test_dataset_punctuation_search(self):
+        """Dots vs spaces in titles should match in both directions."""
+        DatasetFactory(title="Catalogue data.gouv.fr")
+        DatasetFactory(title="Portail data gouv")
+
+        self.refresh_index()
+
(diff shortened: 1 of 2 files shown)


nicolaskempf57 (approved):

estellebertrand:
Note for later, to verify before merging : 

I ran Claude to quickly verify if the assertion "without changing the existing ranking" is true and one feedback was that: 
"The fuzzy most_fields clause now double-counts titles. See search_clients.py:814-825, and the same change for reuses (~1216) and dataservices (~1507). most_fields adds up the scores of the matching fields. Most titles and organization names contain no . or _, so title and title.punct produce the same tokens and both match. That roughly doubles this clause's score for nearly every document. The clause has no function_score, and it sits in the outer bool.should next to the other clauses, which also add up. So its weight grows compared with the phrase, concat_title_org and cross_fields clauses. This contradicts the comment "the main french_dgv ranking is unchanged"."

We need to verify if that's a true issue or not !

PR #3938, opened 2026-09-09
feat: add or to chart filters

nicolaskempf57:


udata/api_fields.py
```diff
@@ -111,7 +115,7 @@ def fields_by_type(self):
     def format(self, value):
         # Value is one of the generic object
         data = marshal(value, self.fields_by_type[value.__class__.__name__])
-        data[DEFAULT_GENERIC_KEY] = value.__class__.__name__
+        data[self.generic_key] = value.__class__.__name__
```
ThibaudDauce, line 118: The output key for `series[].filters` goes from `class` to `_cls` here? Do `cdata` need to change something?

udata/core/visualizations/models.py
```diff
@@ -70,9 +71,36 @@ class Filter(EmbeddedDocument):
...
 
 
+class NestedFilterField(GenericEmbeddedDocumentField):
+    """Generic embedded field for filter groups, defaulting elements stored
+    without a `_cls` (before nested filter groups existed) to `Filter`."""
+
+    def to_python(self, value):
+        if isinstance(value, dict) and "_cls" not in value:
+            value = {**value, "_cls": "Filter"}
+        return super().to_python(value)
```
ThibaudDauce: Could we migrate the element instead of having to do this?

udata/tests/api/test_visualizations_api.py
```diff
@@ -149,6 +149,78 @@ def test_visualization_api_create_and_filter(self):
...
+        # GET should serialize the nested filter fields, not return empty dicts
+        response = self.get(url_for("api.visualization", visualization=visualization))
+        assert response.status_code == 200
+        filters_data = response.json["series"][0]["filters"]
+        assert filters_data["_cls"] == "OrFilters"
+        assert len(filters_data["filters"]) == 2
+        assert filters_data["filters"][0]["column"] == filters.filters[0].column
+
+    def test_visualization_api_create_nested_group_filters(self):
+        """AndFilters can contain OrFilters groups and vice versa"""
```
ThibaudDauce, line 177: "and vice versa" is not done in this test?

ThibaudDauce (approved):

udata/api_fields.py
nicolaskempf57: Yes, but no change is needed in cdata. It was a udata bug. We could change the generic key on the input, but not the output. 
Before this, cdata was sending `_cls` in POST / PUT and expecting `_cls` in response . But udata was accepting `_cls` and sending back `class` instead.

ThibaudDauce (approved):

PR #3939, opened 2026-09-10
feat(dataservices): add suggest endpoint for search autocomplete

bolinocroustibat:
Add external APIs (dataservices) search suggestion endpoint.

### DIFF
--- a/udata/core/dataservices/api.py
+++ b/udata/core/dataservices/api.py
@@ -5,6 +5,7 @@ from bson import ObjectId
 from feedgenerator.django.utils.feedgenerator import Atom1Feed
 from flask import make_response, redirect, request, url_for
 from flask_login import current_user
+from mongoengine import Q
 
 from udata.api import API, api, fields
 from udata.api_fields import patch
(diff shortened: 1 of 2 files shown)


udata/core/dataservices/api.py
```diff
@@ -89,6 +103,38 @@ def get(self):
...
+@ns.route("/suggest/", endpoint="suggest_dataservices")
+class DataserviceSuggestAPI(API):
+    @api.doc("suggest_dataservices")
+    @api.expect(suggest_parser)
+    @api.marshal_list_with(dataservice_suggestion_fields)
+    def get(self):
+        """Dataservices suggest endpoint using mongoDB contains"""
+        args = suggest_parser.parse_args()
+        dataservices = Dataservice.objects(
+            archived_at=None, deleted_at=None, private__ne=True
```
ThibaudDauce: Is it the same as `DataserviceQuerySet.visible`?

udata/tests/api/test_dataservices_api.py
```diff
@@ -705,6 +706,83 @@ def test_dataservice_api_update_org(self):
...
+            )
+        max_follower_dataservice = DataserviceFactory(
+            title="arealtestprefix-4", metrics={"followers": 10}
+        )
+
+        response = self.get(url_for("api.suggest_dataservices", q="arealtestpref", size=5))
+        assert200(response)
+
+        assert len(response.json) <= 5
+        assert len(response.json) > 1
```
ThibaudDauce: Why not create real datasets and have a real response here instead of ranges?

udata/core/dataservices/api.py
bolinocroustibat: Indeed, we can use it: https://github.com/opendatateam/udata/pull/3939/changes/9e7b9c945fcf712edd7238572163f55d62c9a530

udata/tests/api/test_dataservices_api.py
bolinocroustibat: https://github.com/opendatateam/udata/pull/3939/changes/9e7b9c945fcf712edd7238572163f55d62c9a530

